Issue #11, opened 2018-04-05
Bug in Bresenham's line algorithm (two of the octants)


byronwasti:
There seems to be a weird issue when drawing lines in two of the octants (the 4th and the 7th). 

Here is a plotting in Python from the output line values when trying to draw lines in a circle.
![figure_1](https://user-images.githubusercontent.com/5985936/38343239-b0637b6c-3850-11e8-846f-8c12fb75705f.png)

jamwaffles:
Thanks for investigating this! I haven't done much testing with the line drawing algo; just that it can draw a triangle. I'll fix as soon as I can. PRs welcome ;)

PR #12, opened 2018-04-05
Fix bugs in Bresenham algorithm implementation

jamwaffles:
See #11.

Not sure if this works as I can't test it on actual hardware, but there's now a simple test to ensure that a line is drawn within the bounding rectangle of the start and end coordinates. It doesn't check for line shape or gradient, but it's better than the nothing that there was before.

@byronwasti Can you try this branch and see if it fixes your problem?

### DIFF
--- a/src/primitives/line.rs
+++ b/src/primitives/line.rs
@@ -31,10 +31,16 @@ impl<'a> IntoIterator for &'a Line {
     fn into_iter(self) -> Self::IntoIter {
         let &Line { start, end, .. } = self;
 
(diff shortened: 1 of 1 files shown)


byronwasti:
The issue is slightly different now, with all negative-slope lines being broken:

figure_1

I've got a nice little testing setup here: https://github.com/byronwasti/embedded-graphics-line-test-suite

jamwaffles:
Oh no! A good demonstration of why writing code "blind" is terrible. Thank you for the tool! That's going to be super useful.

jamwaffles:
Closed in favour of #13

PR #13, opened 2018-04-05
Fix line algorithm implementation

### LINKED ISSUE #11: Bug in Bresenham's line algorithm (two of the octants)

byronwasti:
This implements the Bresenham's line algorithm from the wikipedia page with some slight modifications.

fixes #11

### DIFF
--- a/src/primitives/line.rs
+++ b/src/primitives/line.rs
@@ -168,10 +168,15 @@ impl Transform for Line {
 mod tests {
     use super::*;
 
-    fn test_expected_line(start: Coord, end: Coord, expected: &[Coord], error_message: &'static str) {
+    fn test_expected_line(
+        start: Coord,
+        end: Coord,
+        expected: &[Coord],
+        error_message: &'static str,
+    ) {
         let line = Line::new(start, end, 1);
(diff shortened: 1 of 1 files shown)


jamwaffles:
Thanks for taking a crack at this! A couple of things:

1. Can you `rustfmt` your code please?

2. It would be good to have at least a primitive unit test for this code, just to catch any glaring problems. I [wrote a crappy one for my fix](https://github.com/jamwaffles/embedded-graphics/blob/b6fdb1f6bd519c5268723b1bd85f95411ae5f5a2/src/primitives/line.rs#L172-L217) which passes when pasted into this PR. It's not the greatest test, but it should hopefully catch any regressions in the future.

byronwasti:
I've added tests for each octant. I'm not sure if its the best way to test things, but it should catch regressions.

jamwaffles:
Looks good! I like your approach to the testing, much better than mine as it actually checks the line. Thanks for fixing the lines!

PR #29, opened 2018-05-03
[WIP] Add basic simulator support

byronwasti:
This isn't a completed feature yet, but I think its a good baseline to get us started.

I'm wondering about the functionality of this feature: how are people going to use it and what form does it need to be in?

Currently it is implemented as a submodule in a workspace, but I'm not convinced that is the best form for it to take.


jamwaffles:
Sorry for letting this sit, didn't realise you'd pushed fixes. This should be a good base to build on even if there's a better place for it to live in the future. Thanks!

PR #41, opened 2018-08-02
Support multiple types of pixel

### LINKED ISSUE #33: Colour support
Hi all,

I've been working on a [SSD1351 driver crate](https://github.com/MabezDev/ssd1351) based on the work done with the ssd1306. I am at the point where I want to start integrating embedded_graphics traits for more advanced graphic work. The SSD1351 being a 16bit colour display driver the current u8 for colour is not going to work, and in the future this crate should probably be able to support even higher colour depths.

Just wanted to kick off some discussion on the matter :)

jamwaffles:
TODO:

* [x] Check/update docs

The `Drawing` trait has changed. It now accepts anything that implements `PixelColor: Clone + Copy + From<u8>` as seen [here](https://github.com/jamwaffles/embedded-graphics/blob/c89a54406820dc63ead528c83d4984b79a93c224/embedded-graphics/src/pixelcolor.rs#L4) as a pixel value. For example, the `simulator` looks like this:

```rust
/// Simple pixel that just wraps a `bool` (i.e. on or off only)
#[derive(Clone, Copy, PartialEq)]
pub struct SimPixelColor(pub bool);

/// Must implement `PixelColor`
impl PixelColor for SimPixelColor {}

/// Must also implement `From<u8>` to make migrating existing code easier. This will probably
/// expand in the future to include more types.
impl From<u8> for SimPixelColor {
    fn from(other: u8) -> Self {
        SimPixelColor(other != 0)
    }
}

/// Implement `Drawing` that accepts `SimPixelColor`
impl Drawing<SimPixelColor> for Display {
    fn draw<T>(&mut self, item_pixels: T)
    where
        T: Iterator<Item = Pixel<SimPixelColor>>,
    {
        for Pixel(UnsignedCoord(x, y), color) in item_pixels {
            if x >= DISPLAY_SIZE as u32 || y >= DISPLAY_SIZE as u32 {
                continue;
            }

            self.pixels[y as usize][x as usize] = color;
        }
    }
}
```

The display driver implementation handles how to convert a particular pixel struct into values that it can send down the wire to the physical display. The simulator above ends up turning `SimPixelColor` into a black/white colour to draw on the canvas.

Because `SimPixelColor` implements `From<u8>` (because of its trait bounds), to migrate existing code, it should be enough to just call `.into()` on any `u8`s currently being passed to draw objects. Example:

```rust
// Before
let line = Line::new(Coord::new(64, 64), Coord::new(0, 64), 1u8)

// After
let line = Line::new(Coord::new(64, 64), Coord::new(0, 64), 1u8.into())
```

There is currently a `PixelColorU8` struct available in `embedded_graphics::pixelcolor::PixelColorU8`. For different pixel types, they'd either be implemented next to it (`PixelColorU16` to help fix #33 for example), or the display driver implementor would export their own type (`Imaginary4DVoxelColor` or whatever, as long as it implements `PixelColor: Clone + Copy + From<u8>`).

Closes #33.

### DIFF
--- a/embedded-graphics/src/coord.rs
+++ b/embedded-graphics/src/coord.rs
@@ -1,4 +1,7 @@
-//! 2D coordinate in screen space
+//! 2D signed coordinate in screen space
+//!
+//! This struct is what you should use when positioning graphics elements as it can handle negative
+//! coordinates. To output pixel coordinates, use [`UnsignedCoord`](../unsignedcoord/index.html).
 
 use super::unsignedcoord::UnsignedCoord;
 
--- a/embedded-graphics/src/drawable.rs
+++ b/embedded-graphics/src/drawable.rs
@@ -1,13 +1,11 @@
 //! `Drawable` trait and helpers
 
-use super::unsignedcoord::UnsignedCoord;
+use pixelcolor::PixelColor;
+use unsignedcoord::UnsignedCoord;
 
 /// A single pixel
 #[derive(Copy, Clone, Debug)]
-pub struct Pixel<C: PixelColor + Clone>(pub UnsignedCoord, pub C);
-
-/// Color trait
-pub trait PixelColor {}
+pub struct Pixel<C: PixelColor>(pub UnsignedCoord, pub C);
 
 /// Marks an object as "drawable". Must be implemented for all graphics objects
 pub trait Drawable {}
--- a/embedded-graphics/src/drawable.rs
+++ b/embedded-graphics/src/drawable.rs
@@ -4,7 +4,7 @@ use pixelcolor::PixelColor;
 use unsignedcoord::UnsignedCoord;
 
 /// A single pixel
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct Pixel<C: PixelColor>(pub UnsignedCoord, pub C);
 
 /// Marks an object as "drawable". Must be implemented for all graphics objects
--- a/embedded-graphics/src/drawable.rs
+++ b/embedded-graphics/src/drawable.rs
@@ -2,12 +2,12 @@
 
 use super::unsignedcoord::UnsignedCoord;
 
-// TODO: Refactor to use both with monochrome and multicolour displays
-/// Monochrome colour type
-pub type Color = u8;
-
 /// A single pixel
-pub type Pixel = (UnsignedCoord, Color);
+#[derive(Copy, Clone, Debug)]
+pub struct Pixel<C: PixelColor + Clone>(pub UnsignedCoord, pub C);
+
+/// Color trait
+pub trait PixelColor {}
 
 /// Marks an object as "drawable". Must be implemented for all graphics objects
 pub trait Drawable {}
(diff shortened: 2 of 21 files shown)


MabezDev:
Thanks for the implementation, I've been testing this PR with the ssd1351 crate and I've hit a road block.

```rust
#[cfg(feature = "graphics")]
use self::embedded_graphics::pixelcolor::PixelColor;
#[cfg(feature = "graphics")]
use self::embedded_graphics::unsignedcoord::UnsignedCoord;

#[cfg(feature = "graphics")]
#[derive(Copy, Clone, PartialEq)]
pub struct PixelColorU16(pub u16);

#[cfg(feature = "graphics")]
impl From<u16> for PixelColorU16 {
    fn from(other: u16) -> Self {
        PixelColorU16(other)
    }
}

#[cfg(feature = "graphics")]
impl From<u8> for PixelColorU16 {
    fn from(other: u8) -> Self {
        PixelColorU16(other as u16)
    }
}

#[cfg(feature = "graphics")]
impl PixelColor for PixelColorU16 {}

#[cfg(feature = "graphics")]
impl<DI> Drawing<PixelColorU16> for GraphicsMode<DI> 
    where
    DI: DisplayInterface,
{
    fn draw<T>(&mut self, item_pixels: T)
    where
        T: Iterator<Item = drawable::Pixel<PixelColorU16>>,
    {
        for drawable::Pixel(UnsignedCoord(x, y), color) in item_pixels {
            self.set_pixel(x, y, color.into());
        }
    }
}
```

when I try to compile I get

```rust
error[E0277]: the trait bound `u16: core::convert::From<mode::graphics::PixelColorU16>` is notsatisfied
   --> src/mode/graphics.rs:140:40
    |
140 |             self.set_pixel(x, y, color.into());
    |                                        ^^^^ the trait `core::convert::From<mode::graphics::PixelColorU16>` is not implemented for `u16`
    |
    = help: the following implementations were found:
              <u16 as core::convert::From<bool>>
              <u16 as core::convert::From<u8>>
    = note: required because of the requirements on the impl of `core::convert::Into<u16>` for`mode::graphics::PixelColorU16`
```

but from what I can see it is?

Also the `PixelColor` requiring `From<u8>` means that I need to implement it for the Display even if the display doesn't support that. For instance If a user `into()`'d a u8 would that compile or would it get blocked because `set_pixel()` takes a `u16`?

jamwaffles:
Thanks for giving this a try! `From<u8>` is required for the `Image*BPP` objects to work; internally they use `u8`s. It's noisy to have to implement `From<u8>` for a struct outside the library, but a necessary evil for now I think. I'm at a loss as to why you get that error. I'll have a dig and get back to you.

jamwaffles:
Ah, take a look at the error message; Rust can't figure out how to turn a `PixelColorU16` _into a raw `u16`_. I suppose you could impl `From<PixelColorU16>`, or do the following which worked for me:

```rust
impl ColorType {
    pub fn into_inner(self) -> u16 {
        self.0
    }
}
```

Up to you which way you go, but I think that should fix your problem.

jamwaffles:
FWIW, I've just added `PixelColorU16` and `PixelColoru32`. Can you try the 16 bit one out in `embedded_graphics::pixelcolor::PixelColoru16` and see if it does what you need?

MabezDev:
Thanks, this is working very well for me! [Here's a crappy phone camera video](https://streamable.com/t0uty). 
About the images, is there anything stopping those becoming generic i.e, `imagedata: &'a [C]` ?

jamwaffles:
Excellent, thanks!

PR #48, opened 2018-10-03
Font Refactor

TeXitoi:
I've seen annoying bugs on fonts: glinchs on the font8x16, panics when writing unsupported chars (typically `\n`) and non working extended chars as `°`.

This solve all of that, and factorize the code:
 - a font builder is added, allowing to easily implement new fonts
 - clean support of the characters of the font
 - change the image of `font8x16` to support latin1
 - add a mock of display to check the rendering of the fonts


jamwaffles:
Thanks for taking the time to improve the fonts! I don't have time to do a deeper review/test right now, but the thing that stands out to me is the display simulator; can you use [the simulator already present in this repo instead?](https://github.com/jamwaffles/embedded-graphics/tree/master/simulator) I'd prefer not to duplicate efforts. 

I'll do a deeper review when I next have a bit of time.

TeXitoi:
I don't understand the simulator: that's just for development purpose, no. Do you want I check on it by hand?

Or you want I remove the mock_display? Because the goal is to have unit tests on rendering.

TeXitoi:
As an helper to review this PR, I added font_builder that is almost exactly the code in the different fonts, with a trait to configure it.

TeXitoi:
Great, thanks!

No problem with the delay, less than one week is really acceptable :-)

TeXitoi:
I think that's not a breaking change as I declared types with the same name, but I'm not really sure because the kind of type change.

TeXitoi:
Can you please publish the update on crates.io?

jamwaffles:
Released as [0.4.1](https://crates.io/crates/embedded-graphics/0.4.1) :)

TeXitoi:
Thanks.

PR #51, opened 2018-10-16
Make other background colors than (0/"black") possible for fonts

Caemor:
When I tested the embedded graphics library today with my eink I realised that only black background colors for fonts was possible.

This PR make different background colors for fonts possible and adds a default stroke_color (a default for fill_color was already set before, so it's just the opposite) so the panic can be removed.

That means the following changes (only shows the differences):
- If no stroke and fill color is set: It doesn't panic anymore and uses 1u8 as stroke color
- If stroke color is set but fill color is not: No change
- If stroke and fill color are set: now fill color is used instead of the default 0u8 ("black")


embedded-graphics/src/fonts/mod.rs
```diff
@@ -45,6 +45,8 @@ where
     ///     disp.draw(text.into_iter());
     /// }
     /// ```
+    /// 
+    /// Defaults to 1u8 for stroke_color and 0u8 for fill_color
```
jamwaffles: I think this should be above the example; it's less easily missed up there.

embedded-graphics/src/style.rs
```diff
@@ -6,8 +6,6 @@ use pixelcolor::PixelColor;
 #[derive(Debug, Copy, Clone)]
 pub struct Style<P: PixelColor> {
     /// Fill colour of the object
-    ///
-    /// Has no effect for fonts
```
jamwaffles: I think this doc should mention that it sets the _background_ colour for a font. There's some semantic difference between text background and fill colour, so it's worth mentioning.

jamwaffles:
Can you also add or edit an example in the [simulator examples folder](https://github.com/jamwaffles/embedded-graphics/tree/master/simulator/examples) to demonstrate this behaviour?

Caemor:
Yes, I am gonna add an example and some tests.

In the font_builder we are currently using Style::default() which returns None for both of the colors. Should we change this so it's more visible that we use a default fill and stroke color for None values in the next iterator? (https://github.com/jamwaffles/embedded-graphics/blob/master/embedded-graphics/src/fonts/font_builder.rs#L61)

New:
```Rust
fn render_str(text: &'a str) -> Self {
    Self {
        pos: Coord::new(0, 0),
        text,
        style: Style::default()
            .with_fill(0u8.into())
            .with_stroke(1u8.into()),
        _conf: Default::default(),
    }
}
```

Caemor:
When I was trying to make a testcase for inverted I might have found a bug in the current implementation:
```
// produced result:
[0, 0, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 0, 1, 0, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 0, 1, 0, 1, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

// what it should look like (at least thats what i thought)
[1, 0, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 0, 1, 0, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 0, 1, 0, 1, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 
[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

jamwaffles:
That's a bug in master, thanks for surfacing it! Running the simulator with a font where the pixel colour is always set, I see this in master:

image

The black rectangle to the middle left has its first row shifted by one which obviously shouldn't happen. I'll look at getting a fix into master asap.

Caemor:
> where the pixel colour is always set

What do you mean by that? 

Can you maybe paste your code in here for that example? That looks like the old behaviour before this PR and shouldn't happen.

jamwaffles:
I've just merged #53. Can you updated from master? The off-by-one error should be gone now.

> What do you mean by that?

Sorry for the confusion, it was just a quick temporary change to help debugging by always setting the pixel black :slightly_smiling_face:

embedded-graphics/src/fonts/font6x8.rs
```diff
@@ -101,6 +137,40 @@ mod tests {
...
+            display,
+            Display([
+                [1, 0, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 0, 0, 1, 0, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 0, 1, 0, 1, 0, 1, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
jamwaffles: Base on the fixes in #53, I got this working by stripping the front column off and moving it after the last char:

```
[0, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 1, 0, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 1, 0, 1, 0, 1, 0, 0, 1, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```

Caemor: Okay, so the "empty" first column got moved to the end with #52 if I saw that correctly?

jamwaffles: Correct :)

Caemor:
Travis is also happy now :-)

jamwaffles:
[0.4.2](https://crates.io/crates/embedded-graphics/0.4.2) released with these changes in it!

PR #61, opened 2018-10-25
Add Dimensions trait

jamwaffles:
The `Dimensions` trait adds the methods `top_left()`, `bottom_right()` and `size()` to all embedded graphics objects. This might be useful to mark the update area of a display that supports partial refreshes, for example.


jamwaffles:
Glad to hear it! FWIW, this is to allow partial updates on displays that support it, however I'm struggling to get the dimension info into `Drawable::draw()`. Please let me know if you run into the same problem! I'm a bit stumped on a solution right now.

PR #82, opened 2019-03-24
Fix off-by-one error for off-screen fonts

### LINKED ISSUE #81: Infinite font iterator when negative translation happens to be the font height
I've run into an issue where the fontbuilder iterator will loop indefinitely under certain translation conditions.

Here's a minimal example:
```        
display.draw(Font6x12::render_str("Testing string")
            .with_stroke(Some(0xF1FA_u16.into()))
            .translate(Coord::new(0, -12))
            .into_iter());
display.flush();
```

if a negative `y` translation happens to be equal to the height of the font, it will loop indefinitely in here: https://github.com/jamwaffles/embedded-graphics/blob/master/embedded-graphics/src/fonts/font_builder.rs#L193

jamwaffles:
Closes #81 

If the font is translated by exactly the negative of its font height, the guard checking whether any pixels should be returned was failing, meaning the iterator spins indefinitely. This PR fixes that off-by-one error.

@MabezDev can you give this a test in your application and let me know if it fixes #81 properly? I'll wait to merge based on your feedback.

### DIFF
--- a/embedded-graphics/src/fonts/font6x12.rs
+++ b/embedded-graphics/src/fonts/font6x12.rs
@@ -30,6 +30,7 @@ mod tests {
     use crate::drawable::Dimensions;
     use crate::fonts::Font;
     use crate::mock_display::Display;
+    use crate::pixelcolor::PixelColorU16;
     use crate::style::Style;
     use crate::style::WithStyle;
     use crate::transform::Transform;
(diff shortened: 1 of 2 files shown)


MabezDev:
Sorry I couldn't get to this yesterday. I've just tested with my application and it now works as expected. Thanks!

jamwaffles:
Spooctacular. Not sure when this'll get released so you'll have to use a Git URL for now. Sorry about that!

MabezDev:
That's okay, for now I'm just using font-height + 1, it's hard to notice the 1px difference.

PR #84, opened 2019-03-27
Reduce code duplication in image module

rfuest:
The different image types contain some duplicated code (e.g. the implementation of the `Dimension` and `Transform` traits). This PR changes the code to use a single `Image` struct with an additional type parameter for the different bit depth (like `FontBuilder`).


jamwaffles:
Thanks for the PR! I don't have time to review it right now, but I'll get to it when I can. There's quite a lot of duplication in Embedded Graphics so it's good to see some of it cleaned up :slightly_smiling_face:

embedded-graphics/src/image/mod.rs
```diff
@@ -1,16 +1,163 @@
...
+use core::marker::PhantomData;
 
 mod image16bpp;
 mod image1bpp;
 mod image8bpp;
 mod image_bmp;
 
-/// Raw image trait
-pub trait Image<'a>: Dimensions {
+/// Raw image type
```
jamwaffles:
```suggestion
/// Trait implemented by all concrete image types
```

embedded-graphics/src/image/mod.rs
```diff
@@ -1,16 +1,163 @@
...
+pub trait ImageType {}
+
+/// Raw image
```
jamwaffles:
```suggestion
/// An image constructed from a slice
```

embedded-graphics/src/image/mod.rs
```diff
@@ -1,16 +1,163 @@
...
+    C: PixelColor,
+    T: ImageType,
+{
+    /// Image width in pixels
+    width: u32,
+
+    /// Image height in pixels
+    height: u32,
+
+    /// Image data, 1 bit per byte, 1 byte per 8 horizontal pixels
```
jamwaffles: This isn't correct any more as the data in the slice will be interpreted differently by the different *BPP image types. Perhaps:

```suggestion
    /// Image data, packed as dictated by image type `T` (`Image1BPP` uses 8 bits per pixel, etc)
```

embedded-graphics/src/image/mod.rs
```diff
@@ -29,7 +176,7 @@ pub trait ImageFile<'a>: Dimensions + Sized {
     fn height(&self) -> u32;
 }
 
-pub use self::image16bpp::Image16BPP;
-pub use self::image1bpp::Image1BPP;
-pub use self::image8bpp::Image8BPP;
+pub use self::image16bpp::{Image16BPP, ImageType16BPP};
+pub use self::image1bpp::{Image1BPP, ImageType1BPP};
+pub use self::image8bpp::{Image8BPP, ImageType8BPP};
```
jamwaffles: Do these `*Type*` types need to be publicly exported? I'd rather not expand the API surface if possible.

embedded-graphics/src/image/image16bpp.rs
```diff
@@ -21,87 +18,27 @@ use core::marker::PhantomData;
...
-pub struct Image16BPP<'a, C: PixelColor> {
-    /// Image width
-    width: u32,
+pub type Image16BPP<'a, C> = Image<'a, C, ImageType16BPP>;
 
-    /// Image height
-    height: u32,
+/// 16 bits per pixel image type
+#[derive(Debug, Copy, Clone)]
+pub enum ImageType16BPP {}
```
jamwaffles: Can this be a `struct` instead? What's the reasoning behind using an `enum`?

embedded-graphics/src/image/image1bpp.rs
```diff
@@ -17,84 +14,29 @@ use core::marker::PhantomData;
...
-pub struct Image1BPP<'a, C> {
-    /// Image width in pixels
-    width: u32,
+pub type Image1BPP<'a, C> = Image<'a, C, ImageType1BPP>;
 
-    /// Image height in pixels
-    height: u32,
+/// 1 bit per pixel image type
+#[derive(Debug, Copy, Clone)]
+pub enum ImageType1BPP {}
```
jamwaffles: See my other `enum`/`struct` comment.

embedded-graphics/src/image/image8bpp.rs
```diff
@@ -17,88 +14,28 @@ use core::marker::PhantomData;
...
-pub struct Image8BPP<'a, C: PixelColor> {
-    /// Image width
-    width: u32,
+pub type Image8BPP<'a, C> = Image<'a, C, ImageType8BPP>;
 
-    /// Image height
-    height: u32,
+/// 8 bits per pixel image type
+#[derive(Debug, Copy, Clone)]
+pub enum ImageType8BPP {}
```
jamwaffles: See my other `enum`/`struct` comment.

embedded-graphics/src/image/mod.rs
```diff
@@ -1,16 +1,163 @@
...
 
 mod image16bpp;
 mod image1bpp;
 mod image8bpp;
 mod image_bmp;
 
-/// Raw image trait
-pub trait Image<'a>: Dimensions {
+/// Raw image type
+pub trait ImageType {}
```
jamwaffles: Nit: I think this could use a better name, but I'm not sure what to go for. Maybe `ImageExt`? I've seen that used in a couple of other crates but it's not very descriptive. `ImageObject` perhaps?

embedded-graphics/src/image/image16bpp.rs
rfuest: The main reason is  that I modeled it after the `Font...Conf` types, which are also using enums. If I understand it correctly the main advantage of an empty enum is that it can never be instantiated and is only used at type level.

embedded-graphics/src/image/mod.rs
rfuest: I didn't like the name myself but didn't have a better idea at the time.
Aren't traits like `ImageExt` used extend other traits with additional functionality, like the `ReadBytesExt` trait in the ` byteorder` crate? Which isn't really the case here.
IMHO `ImageObject` is as generic as `ImageType` and doesn't reflect that the trait is used to differentiate between different kinds of image.
How about `ImageFormat` or more verbose `ImageDataFormat`?

jamwaffles: Ah yes, that seems to be the pattern.

Those names don't sound great either, so let's just go with `ImageType` :). AFAICS it's internal anyway so changing the name later shouldn't be too difficult, right?

embedded-graphics/src/image/image16bpp.rs
jamwaffles: Woops, I missed that. Ok yeah keep these as enums then :slightly_smiling_face: makes sense for consistency.

embedded-graphics/src/image/mod.rs
rfuest: Currently `ImageType` is public. I wasn't sure if this would be useful for users of e-g.

I would suggest to make it private now and expose it late if necessary.

embedded-graphics/src/image/mod.rs
rfuest: If `ImageType` is change to be internal these souldn't be exposed either.

embedded-graphics/src/image/mod.rs
jamwaffles: Sounds good to me!

embedded-graphics/src/image/mod.rs
jamwaffles: Got it, let's make it internal then :slightly_smiling_face:

embedded-graphics/src/image/mod.rs
rfuest: Simply changing `ImageType` to be only crate public doesn't seem to work because the `Image` would leak the private type. I'll need to look into it some more in the next few day to find a better solution.

jamwaffles:
Hi, sorry for the delay. Is there anything left outstanding in this PR that needs to be fixed still?

PR #86, opened 2019-05-04
Triangles

weichweich:
I added an iterator which can draw triangles. It uses the line iterator, to draw the edges. To fill the triangle I start to draw the edges at the top left corner and fill the space between the points on the edge.

Because the current way of drawing lines not always starts drawing at the start point, I had to reimplement the line drawing algorithm.

There are some issues with my triangle implementation. First issue is that triangles have no stroke width. The second issue is that the triangle iterator returns some points multiple times. The corner point e.g. are return at least twice because they are part of two edges which are drawn separately.

### DIFF
--- a/embedded-graphics/src/primitives/line.rs
+++ b/embedded-graphics/src/primitives/line.rs
@@ -198,7 +198,7 @@ where
         Self {
             start: self.start + by,
             end: self.end + by,
-            ..self.clone()
+            ..*self
         }
     }
 
--- a/embedded-graphics/src/primitives/line.rs
+++ b/embedded-graphics/src/primitives/line.rs
@@ -101,8 +101,7 @@ impl<'a, C: PixelColor> IntoIterator for &'a Line<C> {
             d = Coord::new(d[0], -d[1]);
         }
 
-        let s = match (self.start[0] >= self.end[0], 
-                       self.start[1] >= self.end[1]) {
+        let s = match (self.start[0] >= self.end[0], self.start[1] >= self.end[1]) {
             (false, false) => Coord::new(1, 1),
             (false, true) => Coord::new(1, -1),
             (true, false) => Coord::new(-1, 1),
@@ -111,7 +110,7 @@ impl<'a, C: PixelColor> IntoIterator for &'a Line<C> {
 
         LineIterator {
             style: self.style,
-            
+
             start: self.start,
             end: self.end,
             d,
@@ -163,8 +162,10 @@ impl<C: PixelColor> Iterator for LineIterator<C> {
                 self.start += Coord::new(0, self.s[1]);
             }
             if p_coord[0] >= 0 && p_coord[1] >= 0 {
-                return Some(Pixel(p_coord.to_unsigned(), 
-                                  self.style.stroke_color.unwrap()));
+                return Some(Pixel(
+                    p_coord.to_unsigned(),
+                    self.style.stroke_color.unwrap(),
+                ));
             }
         }
         None
@@ -335,7 +336,7 @@ mod tests {
     fn draws_octant_8_correctly() {
         let start = Coord::new(10, 10);
         let end = Coord::new(15, 7);
-        let expected = [(10, 10), (11, 9), (12, 9), (13, 8), (14, 8), (15, 7),];
+        let expected = [(10, 10), (11, 9), (12, 9), (13, 8), (14, 8), (15, 7)];
         test_expected_line(start, end, &expected);
     }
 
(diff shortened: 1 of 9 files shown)


embedded-graphics/src/primitives/line.rs
```diff
@@ -90,121 +90,84 @@ where
...
-            slow,
-            end,
+            style: self.style,
+            
+            start: self.start,
+            end: self.end,
+            d,
+            s,
+            err: d[0] + d[1],
+            e2: 0,
```
jamwaffles: I think these single letters are from Bresenham's algorithm, no? Can you give `d`, `s` and `e2` more descriptive names please?

embedded-graphics/src/primitives/line.rs
```diff
@@ -90,121 +90,84 @@ where
...
-            increment,
-            error,
-
-            quick,
-            slow,
-            end,
+            style: self.style,
+            
+            start: self.start,
+            end: self.end,
```
jamwaffles: These three lines can be replaced with `..self` I think.

embedded-graphics/src/primitives/line.rs
```diff
@@ -90,121 +90,84 @@ where
...
 // [Bresenham's line algorithm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm)
-impl<'a, C: PixelColor> Iterator for LineIterator<'a, C> {
+impl<C: PixelColor> Iterator for LineIterator<C> {
     type Item = Pixel<C>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.quick > self.end {
-            return None;
-        }
+        self.style.stroke_color?;
```
jamwaffles: Can you add a quick comment explaining that this bails early if there's no stroke? It took me a moment to figure out so might confuse others too. Quite an elegant solution though, nice!

simulator/examples/fill.rs
```diff
@@ -6,14 +6,14 @@ use std::time::Duration;
...
-use embedded_graphics::primitives::{Circle, Rect};
+use embedded_graphics::primitives::{Circle, Rect, Triangle};
 
 use simulator::DisplayBuilder;
 
 static CIRCLE_SIZE: i32 = 32;
 
 fn main() {
-    let mut display = DisplayBuilder::new().size(256, 128).scale(2).build();
+    let mut display = DisplayBuilder::new().size(304, 128).scale(2).build();
```
jamwaffles: Why did this need to change?

simulator/src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub struct DisplayBuilder {
 impl DisplayBuilder {
     pub fn new() -> Self {
         Self {
-            width: 256,
+            width: 320,
```
jamwaffles: Why does this need to change?

embedded-graphics/src/primitives/line.rs
weichweich: I tried that, but this results in an error. 

```
expected struct `primitives::line::LineIterator`, found struct `primitives::line::Line`
```

simulator/examples/fill.rs
weichweich: I added some triangles so I had to increase the window size in order for it to fit all primitives.

simulator/src/lib.rs
weichweich: I added triangles to the stroke example and wanted to make them fit into the window. I didn't realize that this should be done in the example itself. Should be fixed with my last commit.

weichweich:
> Thanks for the PR! I've left a preliminary review, and I'll test the simulator examples when I'm next at a Linux machine. Can you add a `Triangle` bullet to the `Primitives` list on the [docs homepage](https://docs.rs/embedded-graphics/0.4.7/embedded_graphics/) please?

Done

> If a line has zero length, surely it shouldn't be drawn at all? What's the current behaviour in Embedded Graphics for this? Is there an established/expected behaviour from the wider graphics community?

I didn't put much thought in that. I think the behaviour should not be changed. So zero length lines should no longer be drawn.

> Don't worry about stroke width - that's something that needs to be fixed for the line primitive as well!

Lines with a stroke width are rectangles. If the triangle solution is stable enough, it could be reused for lines and polygons.

embedded-graphics/src/primitives/line.rs
weichweich: should be fixed now

embedded-graphics/src/primitives/line.rs
jamwaffles: Oops, you're right

simulator/examples/fill.rs
jamwaffles: Ah I see, no problem

jamwaffles:
Released in [0.4.8](https://crates.io/crates/embedded-graphics/0.4.8).

PR #126, opened 2019-06-17
Change simulator to be publishable on crates.io

jamwaffles:
This will set up the simulator to be publishable on crates.io under the `embeded-graphics-simulator` name. A bit of a mouthful, but will avoid collisions and ambiguity.

@rfuest Can you take a quick look at this and let me know if it's what you had in mind?

Closes #122

PR #127, opened 2019-06-17
Add commonly used color types

rfuest:
Here is a first work in progress version of my work on the new color types. It is still missing some features and lacks polish, but should be enough to get a first impression and discuss further details.

## Notes about the implementation
- I replaced the `From<u...>` implementations for the `PixelColor` types with a `FromSlice` trait. This allows the creation of instances of `PixelColor` implementations in a more generic way which isn't constrained to a fixed number of bytes.
- `FromSlice` supports little and big endian data
- In the old `Rgb565` implementation the `r()`, `g()` and `b()` functions returned the 5- or 6-bit values padded with zeroes. But `Rgb565::new(0x1F, 0, 0)` should be the same color as `Rgb888::new(0xFF, 0, 0)`, which doesn't work with simple padding.
Because of that I changed the behavior to always return the raw channel data in the `r()`... functions and require explicit conversion to e.g. `Rgb888` if the user needs values in the range 0-255.
- The simulator now uses the new color types instead of `SimPixelColor`.

## Open questions
- I'd like to know your opinion on the used names for the different color types:
  - `Y8` is used for grayscale images. Other possible names would be `Luma8` or `Gray8`. I chose `Y8` because it is similar to the `Rgb...` types and could be extended to `Ay17` or `Ay88` to describe grayscale image with alpha channel in the future.
  - `BinaryColor` is used to describe colors on monochrome display. I didn't want to call it `Y1` because an active pixel on an OLED display is bright and an active pixel on an B/W LCD is dark. Because of that I used an enum with `Off` and `On` variants for inactive and active pixels. Typing `BinaryColor::On` can get annoying really fast, but this can be abbreviated by a use statement like: `use BinaryColor::On as C1;`.

## Todos

- [x] ~~Support big and little endian in `Image16BBP`~~  (postponed: isn't necessary for these changes)
- [x] Add color conversions with correct rounding
- [x] Rebase after #126 is merged
- [x] Documentation


rfuest:
There is still some room for future improvements, but I think the code is ready for review.

embedded-graphics/src/image/image_bmp.rs
```diff
@@ -129,15 +130,16 @@ where
...
             } else if self.im.bmp.bpp() == 16 {
-                let offset = offset * 2; // * 2 as two bytes per pixel
-
-                (self.image_data[offset as usize] as u16)
-                    | ((self.image_data[(offset + 1) as usize] as u16) << 8)
+                &self.image_data[offset * 2..] // * 2 as two bytes per pixel
+            } else if self.im.bmp.bpp() == 24 {
+                &self.image_data[offset * 3..] // * 3 as two bytes per pixel
+            } else if self.im.bmp.bpp() == 32 {
+                &self.image_data[offset * 4..] // * 4 as two bytes per pixel
```
jamwaffles: Nit: Can you make this if/else chain a `match` instead? I think that'd look cleaner.

embedded-graphics/src/lib.rs
```diff
@@ -182,7 +185,7 @@ use crate::pixelcolor::PixelColor;
 /// Here's an example for an imaginary display that has a 64x64px framebuffer of 8 bit values that
 /// communicates over a (simplified) SPI interface:
 ///
-/// ```rust
+/// ```ignore
```
jamwaffles: Why does this test fail? Is it in-scope for this PR to fix it?

embedded-graphics/src/lib.rs
```diff
@@ -259,7 +262,7 @@ where
 /// framebuffer. It sends pixels one by one to over the SPI bus which isn't very efficient, but that
 /// could be fixed by using a fixed length chunked buffering scheme.
 ///
-/// ```rust
+/// ```ignore
```
jamwaffles: Why does this test fail? Is it in-scope for this PR to fix it?

embedded-graphics/src/pixelcolor/mod.rs
```diff
@@ -1,57 +1,443 @@
...
+            BinaryColor::Off => BinaryColor::On,
+        }
+    }
+
+    /// Maps active and inactive colors to a different `PixelColor` type.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use embedded_graphics::pixelcolor::{BinaryColor, RgbColor, Rgb565};
```
jamwaffles: Nit: I think it's useful for examples to show where the items used are imported from

```suggestion
    /// use embedded_graphics::pixelcolor::{BinaryColor, RgbColor, Rgb565};
```

simulator/Cargo.toml
```diff
@@ -24,5 +24,6 @@ circle-ci = { repository = "jamwaffles/embedded-graphics", branch = "master" }
 sdl2 = "0.32.2"
 
 [dependencies.embedded-graphics]
-version = "0.5.0"
+version = "0.5.2-alpha.0"
+path = "../embedded-graphics"
```
jamwaffles: Can you remove this from the PR? Testing major changes like this needs a better story, but I don't want to release the simulator referencing a local path :slightly_smiling_face:

simulator/examples/chaining.rs
```diff
@@ -2,23 +2,27 @@
 
 use embedded_graphics::fonts::Font6x8;
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles: Can this be aliased to `On` instead? I think it's clearer than `C1` for the examples' sake.

simulator/examples/fill-macros.rs
```diff
@@ -1,6 +1,8 @@
 //! Demonstrate usage of primitives like `fill.rs` but use macros instead for shorter code
 
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::Off as C0;
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::On as On;
```

simulator/examples/fill-macros.rs
```diff
@@ -1,6 +1,8 @@
 //! Demonstrate usage of primitives like `fill.rs` but use macros instead for shorter code
 
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::Off as C0;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::Off as Off;
```

simulator/examples/fill.rs
```diff
@@ -1,4 +1,6 @@
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::Off as C0;
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::On as On;
```

simulator/examples/fill.rs
```diff
@@ -1,4 +1,6 @@
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::Off as C0;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::Off as Off;
```

simulator/examples/fonts.rs
```diff
@@ -1,13 +1,18 @@
 use embedded_graphics::fonts::{Font6x12, Font6x8, Font8x16};
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::Off as C0;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::Off as Off;
```

simulator/examples/fonts.rs
```diff
@@ -1,13 +1,18 @@
...
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::On as On;
```

simulator/examples/hello.rs
```diff
@@ -1,24 +1,28 @@
 use embedded_graphics::fonts::Font6x8;
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::On as On;
```

simulator/examples/offscreen.rs
```diff
@@ -1,17 +1,22 @@
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::On as On;
```

simulator/examples/stroke.rs
```diff
@@ -1,30 +1,34 @@
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::On as On;
```

simulator/examples/triangles.rs
```diff
@@ -1,4 +1,6 @@
 use embedded_graphics::icoord;
+use embedded_graphics::pixelcolor::BinaryColor::Off as C0;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::Off as Off;
```

simulator/examples/triangles.rs
```diff
@@ -1,4 +1,6 @@
...
+use embedded_graphics::pixelcolor::BinaryColor::On as C1;
```
jamwaffles:
```suggestion
use embedded_graphics::pixelcolor::BinaryColor::On as On;
```

jamwaffles:
Aside: I've turned on CircleCI PR builds, so if you push any new changes CircleCI should hopefully pick them up now.

rfuest:
> * What could we change `Y8` to? I don't think it's a very descriptive name

I agree with you. Did you see the reasoning and alternatives for this name in the original message?
 
> * Using set colour types loses portability to different display types (I think). The same object should be drawable on a monochrome OLED and in a 32BPP simulator with the same code. In the `transaprent-fonts` example, I tried drawing items using `Rgb565` to a `BinaryDisplay` and got a bunch of compile errors. This is a case I think e-g should support for things like UI libraries that work in both colour and monochrome/binary. How could we modify this PR to support this case?

How would you implement drawing a colored pixels on an binary display in a generic way? I think that this should be possible but require an explicit conversion.
In most cases libraries should still be able to implement routines for types implementing the `PixelColor` trait.

> * `.map_color()` is interesting. Could it be replaced with a `From<>` impl that maps to `DEFAULT_FG` and `DEFAULT_BG`?

The purpose of map_color is different. You can for example use it to draw a Image1BPP with red and green pixels on a color display.
I originally didn't want to implement `From<BinaryColor>` for the RGB types because `BinaryColor` wasn't supposed to imply anything about what an active or inactive pixel should look like. But maybe a `From<BinaryColor>` implementation with off -> black and on -> white for the RGB types might make some things clearer.

> * To continue my above point, can we lose `DEFAULT_FG` and `DEFAULT_BG` and use other consts instead? Are they only used to convert to `BinaryColor`?

`DEFAULT_FG` is used as a fallback for font rendering if no stroke is specified, which was implemented this way to retain the current behavior. I didn't want to change the behavior of text rendering in this PR, but maybe it should work the same as primitives and don't use a default stroke, which would make `DEFAULT_FG` unnecessary. `DEFAULT_BG` is mainly used for the simulator and should be replaceable.

> * That's an awful lot of macro code! I get that hand-writing it all would be a horrible pain, but it's a lot of macro to maintain. Can it at least be rustfmt-ed? I'd like to at least see some doc comments explaining what each does. The file is quite long as well, so is there a sensible place to split it up?

Yes this could certainly be cleaned up more, but I didn't want to spend more time on it before I was certain this PR had a chance to be merged. I'll look into this some more and try to find a balance between readability and ease of implementations. Maintaining lots of copy and pasted code isn't fun either.

embedded-graphics/src/lib.rs
rfuest: This was disabled because there was no 8 bit color type when I started development, but it is fixed now.

simulator/examples/chaining.rs
rfuest: I didn't want to use `On` and `Off` because they have a higher chance of already being used and I wanted to show that you can easily use any name if someone doesn't like typing `BinaryColor::On` and `BinaryColor::Off` everywhere. Maybe `BcOn` and `BcOff` instead?

simulator/Cargo.toml
rfuest: How can I remove this without breaking CI?

rfuest:
The latest changes should make the code more readable. The code is now split in multiple files and the use of macros reduced.

simulator/Cargo.toml
jamwaffles: Hm, not sure to be honest. Leave it for now and I'll fix it forward after merge/before release.

jamwaffles:
Thanks for addressing those changes. I'll do another review pass on the new commits soon. Thanks for bearing with me on this. As a reward, here's another wall of text :grin::

> Did you see the reasoning and alternatives for this name in the original message?

Ah, I did! I was living in the `Files changed` tab and missed it. `Gray8` seems like a better idea as it makes clear that the pixel is in 8BPP greyscale.

While we're bikeshedding names (:grin:), `BinaryColor` is on the edge of cumbersome. Here are some alternative suggestions I've come up with:

* `Gray1` - fits with `Gray8` but doesn't make much sense because it's not greyscale
* `Binary` - removes having to type `BinaryColor` everywhere, but is quite ambiguous
* `OnOff` - works quite nicely when using variants like `OnOff::On` which is nice and short
* `Bin` - makes more sense when imported as `pixelcolor::Bin`, but is quite weird when used like `Bin::Off`

I'm leaning towards `OnOff` as it's short and works nicely with the `On` and `Off` variants. There's a reasonably low chance of it conflicting with other imports as well. The `C1` suggestions I left could be replaced by just using `OnOff::On` instead, without aliasing the import. Thoughts? I'm open to better suggestions :slightly_smiling_face:

> How would you implement drawing a colored pixels on an binary display in a generic way? I think that this should be possible but require an explicit conversion.

I think the conversion should be implicit by default even if it is a bit more magic. The new colours in this PR implement a lot of `From<OtherColor>` (which is rad by the way), so what about changing the `Drawing` trait to accept `Into<C>` instead of just `C`? This would allow passing of any colour that has the right conversion traits implemented. e-g can provide sane conversion defaults like converting `Rgb565` into `BinaryColor` using a threshold function or whatever looks good for most cases. If different behaviour than the default is desired, the programmer can use `.into_iter().map(...)` to add custom conversion behaviour at the call site. It's a bit more boilerplatey, but gives a good compromise between sane defaults and customisability I think.

> The purpose of `map_color` is different. You can for example use it to draw a Image1BPP with red and green pixels on a color display.

If you agree with my above point, `map_color` would be unnecessary as it would be replaced with a `From<...>` impl, right? Again, if a different mapping is desired, the programmer can add a `.map()`. It's a bit longer to write for the programmer, but affords them much more flexibility in what colours to map to.

> I originally didn't want to implement From<BinaryColor> for the RGB types because BinaryColor wasn't supposed to imply anything about what an active or inactive pixel should look like. But maybe a From<BinaryColor> implementation with off -> black and on -> white for the RGB types might make some things clearer.

e-g should be somewhat batteries included, so implementing `From<BinaryColor` is a good idea to me. I think this might require the retention of `DEFAULT_FG`/`DEFAULT_BG` though?

> I'll look into this some more and try to find a balance between readability and ease of implementations. Maintaining lots of copy and pasted code isn't fun either.

Thanks :slightly_smiling_face:. The macros do reduce a pile of boilerplate so we might just have to swallow that pill.

> DEFAULT_FG is used as a fallback for font rendering if no stroke is specified, which was implemented this way to retain the current behavior.

I see. I'm starting to think these two consts need to stay in, particularly if the above changes are made around `From<>` impls. Let's see where the rest of this PR goes.

rfuest:
> `Gray8` seems like a better idea as it makes clear that the pixel is in 8BPP greyscale.

OK, I'll change that.

> I'm leaning towards `OnOff` as it's short and works nicely with the `On` and `Off` variants. There's a reasonably low chance of it conflicting with other imports as well. The `C1` suggestions I left could be replaced by just using `OnOff::On` instead, without aliasing the import. Thoughts? I'm open to better suggestions slightly_smiling_face

I don't like `OnOff` as a type name. In my opinion `OnOff::On` and `OnOff::Off` are visually hard to distinguish, which gets worse if the code contains many of these expressions. I'm leaning towards `Binary` at the moment.

> ..., so what about changing the `Drawing` trait to accept `Into<C>` instead of just `C`?

The simulated `RgbDisplay` already does this by using the binding `C: PixelColor + Into<Rgb888>`. But maybe having this feature by default is a good idea.

> e-g can provide sane conversion defaults like converting `Rgb565` into `BinaryColor` using a threshold function or whatever looks good for most cases. If different behaviour than the default is desired, the programmer can use `.into_iter().map(...)` to add custom conversion behaviour at the call site. It's a bit more boilerplatey, but gives a good compromise between sane defaults and customisability I think.

OK. I'll implement `RgbColor`s to `Gray8` and `BinaryColor` by calculating the luminance and using a thresholding operation for `BinaryColor`.

> If you agree with my above point, `map_color` would be unnecessary as it would be replaced with a `From<...>` impl, right? Again, if a different mapping is desired, the programmer can add a `.map()`. It's a bit longer to write for the programmer, but affords them much more flexibility in what colours to map to.

I still think that `map_color` is useful and would like to keep it, but change it to a trait and also implement it for iterators. This would allow code like this:
```rust
// Draw red icon on black background.
display.draw(icon.into_iter().map_color(Rgb565::BLACK, Rgb565::RED));
```
I was also thinking about adding more variants of this function, like:
```rust
// Draw green icon on transparent background.
// The BinaryColor:Off pixels will be skipped by the MapOn iterator.
display.draw(icon.into_iter().map_on(Rgb565::GREEN));
```
Do you think this would be useful? But I can also remove `map_color` for now and open another PR after this one is merged.

> e-g should be somewhat batteries included, so implementing `From<BinaryColor` is a good idea to me. I think this might require the retention of `DEFAULT_FG`/`DEFAULT_BG` though?

OK, I'll implement conversions from `BinaryColor` to `Gray8` and the `RgbColor`s assuming `Off` is black and `On` is white. `DEFAULT_FG` and `DEFAULT_BG` can than be replaced with `BinaryColor::Off.into()` and `BinaryColor::On.into()`.

embedded-graphics/src/image/image_bmp.rs
```diff
@@ -129,17 +131,14 @@ where
                 return None;
             }
 
-            let offset = ((h - 1 - y) * w) + x;
+            let offset = (((h - 1 - y) * w) + x) as usize;
```
evq: hello! I was testing out this branch and noticed this calculation (and the former revision) don't take stride into account, so BMP images that have a width that is not a multiple of 4 don't display properly

jamwaffles: Hi, thanks for spotting this! As the bug is present in the currently released version, can you open a separate issue describing it please? This PR is already pretty long :slightly_smiling_face:. It would be great if you could also include a failing test case and/or a sample image to test with.

jamwaffles:
Hey @rfuest, thanks for making those changes and replying to my comments. This PR is definitely on my radar but I don't have much time to sit down and properly review it right now. Thanks for your patience!

jamwaffles:
Let's just stick with `BinaryColor` at least for now. It's easy enough to alias if some feel it's too long to type.

> Do you think this would be useful? But I can also remove map_color for now and open another PR after this one is merged.

A suite of `map_*()` functions would be great! I think it makes sense to explore that feature in a separate PR, so can you split it out from this one?

> `DEFAULT_FG` and `DEFAULT_BG` can than be replaced with `BinaryColor::Off.into()` and `BinaryColor::On.into()`.

Makes sense to me 🙂. I think that would completely remove them then, right?

I'm going to do another review pass now so I'll add another comment with anything else I find.

jamwaffles:
Ah, I missed this:

> The simulated RgbDisplay already does this by using the binding C: PixelColor + Into<Rgb888>. But maybe having this feature by default is a good idea.

I'd expect this to work:

```rust
/// From `simulator/examples/tga.rs`

use embedded_graphics::image::ImageTga;
use embedded_graphics::pixelcolor::BinaryColor;
use embedded_graphics::prelude::*;
use embedded_graphics_simulator::DisplayBuilder;
use std::thread;
use std::time::Duration;

fn main() {
    let image: ImageTga<BinaryColor> = ImageTga::new(include_bytes!("./rust-pride.tga")).unwrap();

    let mut display = DisplayBuilder::new()
        .title("TGA image")
        .size(304, 128)
        .scale(2)
        .build_binary();

    display.draw(&image);
}
```

I currently get an error because `Rgb888` can't be converted to `BinaryColor`. I also tried `Gray8` which gave me similar errors. It works in `master` so I'd like to keep that functionality. Maybe there are just some missing trait impls?

rfuest:
> I'd expect this to work:
> 
> ```rust
> /// From `simulator/examples/tga.rs`
> 
> use embedded_graphics::image::ImageTga;
> use embedded_graphics::pixelcolor::BinaryColor;
> use embedded_graphics::prelude::*;
> use embedded_graphics_simulator::DisplayBuilder;
> use std::thread;
> use std::time::Duration;
> 
> fn main() {
>     let image: ImageTga<BinaryColor> = ImageTga::new(include_bytes!("./rust-pride.tga")).unwrap();
> 
>     let mut display = DisplayBuilder::new()
>         .title("TGA image")
>         .size(304, 128)
>         .scale(2)
>         .build_binary();
> 
>     display.draw(&image);
> }
> ```
> 
> I currently get an error because `Rgb888` can't be converted to `BinaryColor`. I also tried `Gray8` which gave me similar errors. It works in `master` so I'd like to keep that functionality. Maybe there are just some missing trait impls?

'Rgb888' can be converted to `BinaryColor` but the simulator currently doesn't use `Into<BinaryColor>` for binary display. If I implement this your TGA example would work fine (except that it should be `let image: ImageTga<Rgb888> = ImageTga::new(include_bytes!("./rust-pride.tga")).unwrap();` instead). But this change breaks the font example because the compiler can't infer the type for texts which use default stoke and fill anymore. Maybe we could change the fonts to behave in the same way as all other primitives and default to no stoke and no fill. This would make it easier for the compiler to infer types and be more consistent.

> Edit: I'd also expect to be able to do this:
> 
> ```rust
> let _r: Rectangle<Rgb565> = egrectangle!(
>     (10, 20),
>     (30, 40),
>     stroke = Some(1u8.into()),
>     fill = Some(Rgb565::GREEN)
> );
> ```
> 
> Note the `1u8.into()`. It's currently possible in `master` and I want to keep this feature available.

Hmm, to be honest I don't like this at all. What color is `1u8` supposed to be? The same as `Gray8(1)` or `Rgb565(0, 0, 1)` or `BinaryColor::On`?

embedded-graphics/src/pixelcolor/gray_color.rs
```diff
@@ -0,0 +1,72 @@
+use crate::pixelcolor::{BinaryColor, FromRawData, PixelColor};
+
+/// 8 bit grayscale color stored in a `u8`
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Gray8(u8);
+
+impl Gray8 {
+    /// Creates a new color.
+    pub const fn new(y: u8) -> Self {
```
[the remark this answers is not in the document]
rfuest: Ok, but I used `luma` instead. I think it is more accurate and also shorter.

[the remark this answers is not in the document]
jamwaffles: Works for me, thanks

jamwaffles:
> 'Rgb888' can be converted to BinaryColor but the simulator currently doesn't use Into<BinaryColor> for binary display. If I implement this your TGA example would work fine...

I just tried changing the sim's `Drawing` impl to this:

```rust
impl<C> Drawing<C> for BinaryDisplay
where
    C: PixelColor + Into<BinaryColor>,
{
    fn draw<T>(&mut self, item_pixels: T)
    where
        T: IntoIterator<Item = Pixel<C>>,
    {
        for Pixel(coord, color) in item_pixels {
            let x = coord[0] as usize;
            let y = coord[1] as usize;

            self.pixels.set(x, y, color.into());
        }
    }
}
```

This works, but gives an odd result:

image

This is to be expected with knowledge of how the conversion routines work, but will look really odd to end users I think. This could easily be fixed by customising the mapping with the `.map_color()`, etc methods in a future PR. For now, let's go with requiring a manual `.into_iter().map(|p| ...)`. What do you think?

> What color is 1u8 supposed to be? The same as Gray8(1) or Rgb565(0, 0, 1) or BinaryColor::On?

Right, sure 😅 that doesn't make any sense does it. I think it made sense before because the simulator was mostly monochrome anyway? This works though, which I'm happy with:

```rust
let _r: Rectangle<Rgb565> = egrectangle!(
    (10, 20),
    (30, 40),
    stroke = Some(Rgb888::RED.into()),
    fill = Some(Rgb565::GREEN)
);
```

(as does `BinaryColor::On.into()`)

At this point, it's probably worth doing an alpha release with this PR in it so it can be tested in real world scenarios. I still don't have a great idea of how it'll play out, but chatting over a PR isn't the best way to prove something 😁. Can you rebase this branch please? Then I'll do a final review and release `0.6.0-alpha`.

rfuest:
> I just tried changing the sim's `Drawing` impl to this: [...]
> 
> This works, but gives an odd result:
>
> This is to be expected with knowledge of how the conversion routines work, but will look really odd to end users I think. This could easily be fixed by customising the mapping with the `.map_color()`, etc methods in a future PR. For now, let's go with requiring a manual `.into_iter().map(|p| ...)`. What do you think?

That's the reason I didn't want to implement the `RgbXXX` to `BinaryColor` default conversion at first, because there is no single "right" way to do this conversion. But I think most users will prefer storing smaller 1 BPP data anyway if they don't need color, so this conversion might not be used often.
 
> At this point, it's probably worth doing an alpha release with this PR in it so it can be tested in real world scenarios. I still don't have a great idea of how it'll play out, but chatting over a PR isn't the best way to prove something grin. Can you rebase this branch please? Then I'll do a final review and release `0.6.0-alpha`.

Done. I would like to do some more (breaking) changes before a final 0.6.0 release but they all depend on this PR, so a alpha release is a good idea. And sorry again for the size of this PR, but there wasn't really a good place to split it into smaller chunks.

jamwaffles:
> That's the reason I didn't want to implement the RgbXXX to BinaryColor default conversion at first, because there is no single "right" way to do this conversion.

Yeah... I'm starting to regret asking you to add that in now 😬. Do you think that kind of conversion would be better served by a `.map_<something>()` function? If so, maybe removing it would be better - it's clear to me now that big conversions like these are unlikely to work the way users expect. Calling `.into_iter().map(...)` isn't so bad in the interim while some syntactic sugar gets worked on. Thoughts?

> And sorry again for the size of this PR, but there wasn't really a good place to split it into smaller chunks.

Absolutely no problem. Thank you for bearing with me on this - I haven't had much spare time or brain capacity to really understand it so the dumb questions have been plentiful.

rfuest:
> Yeah... I'm starting to regret asking you to add that in now grimacing. Do you think that kind of conversion would be better served by a `.map_<something>()` function? If so, maybe removing it would be better - it's clear to me now that big conversions like these are unlikely to work the way users expect. Calling `.into_iter().map(...)` isn't so bad in the interim while some syntactic sugar gets worked on. Thoughts?

Here is a list of currently implemented conversions and how I would like to handle them for now:
- `RgbXXX` -> `RgbYYY`
  Keep as is.
- `Gray8` -> `RgbXXX`
  Keep as is.
- `RgbXXX` -> `Gray8`
  This currently uses HSI intensity, but there are other methods available.
  Replace by a `RgbXXX::intensity() -> u8` function.
- `BinaryColor` -> `Gray8` & `BinaryColor` -> `RgbXXX`
  Keep as is.
- `Gray8` -> `BinaryColor`
  Replace by `Gray8::threshold(threshold: u8) -> BinaryColor` function.
- `RgbXXX` -> `BinaryColor`
  Can be replaced by `RgbXXX` -> `Gray8` -> `BinaryColor` conversion for now.

jamwaffles:
Sounds good to me. Thanks for listing those out. Do you want to make those changes in this PR? I'm fine with that if you do. Let me know when you're done and I'll do a final review pass.

rfuest:
I did remove the ambiguous color conversions, but didn't add any of the mentioned additional functions to replace them. These aren't essential and can be added later.

jamwaffles:
Ok, this is looking pretty good. I'm still not sure how it'll work in practice, but let's get this merged so we can answer that question. One thing I didn't consider is `Gray16` for displays like the [SSD1351](https://crates.io/crates/ssd1351), but that can be added later.

jamwaffles:
Released as [e-g 0.6.0-alpha.1](https://crates.io/crates/embedded-graphics/0.6.0-alpha.1) and [e-g-sim 0.2.0-alpha.1](https://crates.io/crates/embedded-graphics-simulator/0.2.0-alpha.1). Thanks for all your hard work on this!

PR #128, opened 2019-06-18
Add list of drivers with e-g support to README.md

rfuest:
I'm not sure if this list is complete and some drivers might not support the latest versions of e-g. But I think an overview of available drivers is useful anyway.


rfuest:
Oops, I completely missed that such a list already existed in the documentation. I will sync the lists when I'm back next week.

jamwaffles:
I just cherry-picked this into the next release. Turns out the lists are basically identical already, I just missed a couple. I synced the README with `lib.rs` and it's all good now. Thanks for adding the couple of missing drivers! I think it's really useful to put the list in the README for stuff like crates.io to pick up.

PR #130, opened 2019-06-25
Rename primitives macros

### LINKED ISSUE #129: Macro-using this crate breaks panic
Hi !

The `line` macro in this crate conflicts with the [`std::line`](https://doc.rust-lang.org/1.2.0/std/macro.line!.html) macro, which is invoked by `panic` for instance. This means that adding a 
```rust
#[macro_use]
extern crate embedded_graphics;
```
in a program will prevent usage of `panic`.

jamwaffles:
This primarily fixes #129 so that the `line!()` macro provided by std doesn't conflict with the `line!()` macro from Embedded Graphics. The macros are now renamed as such:

* `line` -> `egline`
* `triangle` -> `egtriangle`
* `rect` -> `egrectangle`
* `circle` -> `egcircle`

Note that `rect` is also now called `rectangle` for consistency with the other macros that use the shape's full name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ Some smaller breaking changes, mostly polish around public APIs
 
 * The simulator is now [available on crates.io](https://crates.io/crates/embedded-graphics-simulator) as a standalone crate. You can now create simulated displays for testing out embedded_graphics code or showing off cool examples.
 
+### Changed
+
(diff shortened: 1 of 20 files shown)


althonos:
Note that you added this as a `0.5.1` project but for semver consistency this should be released as `0.6` !

jamwaffles:
Correct. I don't want to bump a minor version for what I think is a feature that isn't widely used yet, so I'm ok dealing with the fallout releasing a breaking change in a patch release. The [semver spec](https://semver.org/#spec-item-4) state that this _is_ allowed, if we want to get technical :grin:.

PR #148, opened 2019-07-31
Use character patterns in MockDisplay

rfuest:
The PR simplifies testing of drawing routines by using character patterns to specify the expected output.

Apart from being easier to read and write, the previous tests couldn't distinguish if a pixel wasn't drawn at all or was drawn with `BinaryColor::Off`. This is now possible by using `Option<C>` for storing pixels. See `font6x8::tests::correct_inverse_coloured_m` for an example which expects `BinaryColor::Off` and all other font tests which expect a transparent background.

`fmt::Debug` for `MockDisplay` now outputs the same ASCII art representation of the display which is useful to debug failing tests.

Currently char patterns are only supported for `BinaryColor`, but this could easily be extended to other color types for more complex tests which might for example use different colors for stroke and fill.

### DIFF
--- a/embedded-graphics/src/mock_display.rs
+++ b/embedded-graphics/src/mock_display.rs
@@ -1,3 +1,24 @@
+//! Mock display for use in tests.
+//!
+//! [`MockDisplay`] can be used to replace a real display in tests. The internal
+//! framebuffer wraps the color values in `Option` to be able to test which
+//! pixels were modified by drawing operations.
+//!
+//! The [`from_pattern`] method provides a convenient way of creating expected
+//! test results. The same patterns are used by the implementation of `Debug`
+//! and will be shown in failing tests.
+//!
+//! [`MockDisplay`]: struct.MockDisplay.html
+//! [`from_pattern`]: struct.MockDisplay.html#method.from_pattern
+//!
+//! # Characters used in `BinaryColor` patterns
+//!
+//! | Character | Color                    | Description                             |
+//! |-----------|--------------------------|-----------------------------------------|
+//! | `' '`     | `None`                   | No drawing operation changed the pixel  |
+//! | `'.'`     | `Some(BinaryColor::Off)` | Pixel was changed to `BinaryColor::Off` |
+//! | `'#'`     | `Some(BinaryColor::On)`  | Pixel was changed to `BinaryColor::On`  |
+
 use crate::drawable::Pixel;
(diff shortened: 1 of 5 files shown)


embedded-graphics/src/mock_display.rs
```diff
@@ -1,24 +1,125 @@
...
+where
+    C: PixelColor,
+{
+    /// Creates a new empty mock display.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns a iterator over all pixels on the display.
+    pub fn pixels<'a>(&'a self) -> impl Iterator<Item = Option<C>> + 'a {
```
jamwaffles: Nit: This method can return `impl Iterator<Item = &'a Option<C>>`. The copy here doesn't really matter for tests of course, but it feels cleaner to me to return a reference.

This change introduces an error in `font6x8.rs` which can be fixed by adding `.cloned()` there. Like this:

```rust
assert!(inverted_display_normal.eq(display_inverse.pixels().copied()));
```

embedded-graphics/src/mock_display.rs
```diff
@@ -1,24 +1,125 @@
...
+    C: PixelColor + ColorMapping<C>,
+{
+    /// Creates a new mock display from a character pattern.
+    ///
+    /// The color pattern is specified by a slice of string slices. Each string slice represents
+    /// a row of pixels and every character a single pixel.
+    ///
+    /// A space character in the pattern represents a pixel which wasn't modified by any
+    /// drawing routine and is left in the default state. All other characters are converted
+    /// by implementations of the `ColorMapping` trait.
```
jamwaffles: Please turn `ColorMapping` into a doc link

```suggestion
    /// by implementations of the [`ColorMapping`](./trait.ColorMapping.html) trait.
```

To make the char conversion easier for other contributors to use, I'd like to see a quick bullet point list in either this doc comment or another describing which characters map to what colour. I'm fine with just supporting `BinaryColor` for tests so that mapping should be easy to document. Maybe in the `//!` module docs?

embedded-graphics/src/mock_display.rs
rfuest: To be more useful `pixels` should have returned `Pixel<...>` anyway, but for now I removed it completely and replaced it by `get_pixel` and `set_pixel`. If we need this iterator later it's easy to add it again.

Issue #139, opened 2019-07-23
Option in WithStyle


rfuest:
`WithStyle::stroke` and `WithStyle::fill` currently take an `Option<C>` as an argument. For this builder style interface the `Some(X)` variant is used in almost all instances. I only found one use of `stoke(None)` in the e-g source and it wasn't even necessary in this case.

I would suggest to change `stroke` and `fill` to use `C` as an argument directly. For the rare case that someone wants to remove the stroke or fill additional methods could be added.

If you don't like this change maybe this can at least be implemented for the primitive macros, where the `Some`s look a little out of place IMHO.

jamwaffles:
Yes, losing the `Option`s would clean up both the macros and method call chains. That said, what would unsetting a style look like? `no_*()` methods would end up being pretty cumbersome I think, particularly with more style options in the future. Do you have any suggestions in mind for how the API might look?

rfuest:
I probably would use `no_*()` methods.

Implementing the functions for `T: Into<Option<C>>` (to accept `C` and `Option<C>`) would also work but make the function declarations in `WithStyle` more complex. And I don't think that this is a common pattern in Rust APIs, or have you seen this used before in any commonly used crate?

> `no_*()` methods would end up being pretty cumbersome I think, particularly with more style options in the future.

Cumbersome to use or cumbersome to implement?

jamwaffles:
> ... but make the function declarations in WithStyle more complex. 

Not that much more complex, right? The function signatures would change a bit and there would be a `.into()` call in the body, but there wouldn't be any other changes.

> And I don't think that this is a common pattern in Rust APIs, or have you seen this used before in any commonly used crate?

I haven't seen any other (ab)uses of the builder pattern like setting styles in e-g, so I don't have much to go on. 

> Cumbersome to use or cumbersome to implement?

Both; it's twice the number of methods in the trait to have to implement, but it's a pain to use as well. If someone is animating styles between `Some(...)` or `None`, it's more ergonomic to be able to just pass a variable in instead of swapping between e.g. `stroke()` and `no_stroke()` I think. Here's a contrived example:

```rust
let mut border = Some(C1);
let mut circle = Circle::new(icoord!(64, 64), 64).stroke(border);

loop {
    display.draw(circle);

    border = match border {
        Some(_) => None,
        None => Some(C1),
    };

    circle = circle.stroke(border);

    thread::sleep(Duration::from_millis(200));
}
```

[...]

PR #149, opened 2019-07-31
[WIP] Add experimental drawing context

rfuest:
While I was thinking about ways to simplify drawing in e-g, I decided to try to implement a more conventional drawing API on top of the current API. I'm not sure if you'll like this, because it might not fit your vision of using iterators for everything, but I wanted to share it anyway.

The implementation is only a quick proof of concept, but should be enough to discuss if this might be useful.

Related to #163

### DIFF
--- a/embedded-graphics/src/context.rs
+++ b/embedded-graphics/src/context.rs
@@ -1,0 +1,145 @@
+//! Drawing context
+
+use crate::coord::Coord;
+use crate::drawable::Pixel;
+use crate::fonts::Font;
+use crate::pixelcolor::{BinaryColor, PixelColor};
+use crate::primitives::{Circle, Line, Rectangle, Triangle};
+use crate::style::Style;
+use crate::transform::Transform;
+use crate::Drawing;
+use core::marker::PhantomData;
+
+/// Drawing context.
+#[derive(Debug)]
+pub struct Context<'a, C, D>
+where
+    D: Drawing<C>,
+    C: PixelColor,
+{
+    drawing: &'a mut D,
+
+    style: Style<C>,
+    offset: Coord,
+
+    color_type: PhantomData<C>,
+}
+
+impl<'a, C, D> Context<'a, C, D>
+where
+    D: Drawing<C>,
+    C: PixelColor + From<BinaryColor>,
+{
+    /// Draws a circle.
+    pub fn circle(&mut self, center: Coord, radius: u32) {
+        let circle = Circle {
+            center: center,
+            radius,
+            style: self.style,
+        };
+
+        self.drawing.draw(circle.translate(self.offset));
+    }
+
+    /// Draws a line.
+    pub fn line(&mut self, start: Coord, end: Coord) {
+        let line = Line {
+            start,
+            end,
+            style: self.style,
+        };
+
+        self.drawing.draw(line.translate(self.offset));
+    }
+
+    /// Draws a rectangle.
+    pub fn rectangle(&mut self, top_left: Coord, bottom_right: Coord) {
+        let rectangle = Rectangle {
+            top_left,
+            bottom_right,
+            style: self.style,
+        };
+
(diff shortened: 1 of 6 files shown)


jamwaffles:
I really like the idea of reducing repeated boilerplate when drawing complex graphics built up with loads of e-g primitives. This PR definitely surfaces a usecase for that.

The `Iterator`-based approach in e-g is a fundamental requirement (to me at least), so if you can find a way to merge the two aspects that would be great.

rfuest:
> The `Iterator`-based approach in e-g is a fundamental requirement (to me at least), so if you can find a way to merge the two aspects that would be great.

This PR wasn't intended to replace the `Iterator`-based approach at all, but provide an additional way of building primitives. Mixing the "old" way of drawing with the `Context` based on is always possible:
https://github.com/jamwaffles/embedded-graphics/blob/ad9af8956e9b28208cb99216691359c9f6b8b896/simulator/examples/fonts-context.rs#L18-L41

Some changes might be required to make the `Context` based approach more efficient, like not having to clone the `Style` for every drawing operation. https://github.com/jamwaffles/embedded-graphics/issues/139#issuecomment-515121637, you mentioned, might be a solution for that.

Offering additional ways of doing things is nice, but I think there should be one recommended way of drawing, otherwise new users might get confused. Having more examples (see #150) and evaluating different ways of drawing with them should make it easier to decide which way to recommend.

jamwaffles:
We'll have to figure out the `Style` refactor, but the other question I have is around maintenance. There's quite a lot of API surface on `ctx` which is only a thin wrapper over the primitives, fonts, etc present in e-g. Instead of methods for circle, rect, text, etc, what do you think about a `ctx.draw()` method that has e-g items passed to it? Contexts could then support iterator chaining like `Drawing` does, reducing cognitive overhead.

rfuest:
The API is designed to resemble canvas APIs like in HTML5 or cairo. This API behaves a little differently because it can't store an path of unknown length and draws primitives immediately instead, but should still feel familiar to users of the other APIs.

One change I would make to this implementation is to replace the `line` method by the `move_to` and `line_to` methods used in other APIs, which can't easily be replicated by using a `draw` method with primitives as parameters.

> Contexts could then support iterator chaining like `Drawing` does, reducing cognitive overhead.

I don't quite understand what you mean by that. Could you provide an example how this would work?

jamwaffles:
> I don't quite understand what you mean by that. Could you provide an example how this would work?

Sure, something like this contrived thing:

```rust
let display = AmazingDisplay::new();

let line = Line::new(Coord::new(0, 1), Coord::new(2, 3)).into_iter();
let circle = Circle::new(Coord::new(5, 5), 3).into_iter();
let chained = line.chain(circle);

let mut ctx = display.context();

ctx.set_stroke_color(BinaryColor::On);
ctx.draw(text_6x8!("Hello World! - default style 6x8", icoord!(15, 15)));

ctx.set_stroke_color(BinaryColor::Off);
ctx.set_fill_color(BinaryColor::On);
ctx.draw(chained);
```

There would be one way of getting e-g items into a context - `.draw()` - which can accept iterators like `.draw()` from `Drawing`. I don't have an answer for how styling would work with iterator support, but if it's possible this is the API I'd like to see. It's different from stuff like canvas and Cairo, but still provides the benefits of setting styles once for multiple calls to `ctx.draw()`, which is I think the main problem a drawing context solves.

I'd prefer not to have two e-g object creation APIs; one on objects (e.g. `Circle::new(...)`) and one on a `Context` (e.g. `ctx.circle(...)`), which may differ in subtle ways. What problems would be caused by leveraging the existing e-g objects?

rfuest:
> I don't have an answer for how styling would work with iterator support, but if it's possible this is the API I'd like to see.

Maybe we should postpone this discussion for now and first handle this: https://github.com/jamwaffles/embedded-graphics/issues/139#issuecomment-515121637. Decoupling the shapes from the styling should make this easier.

Issue #142, opened 2019-07-24
Dimension::size() off by one error


rfuest:
The size returned by `Dimension::size` for `Rectangle` and `Circle` seems to be one pixel to small. There might be other `Drawable`s which also return the wrong size.

## Code
```rust
lett rectangle = egrectangle!((5, 5), (7, 7), stroke = Some(BinaryColor::On));
display.draw(&rectangle);
println!("rectangle: {:?}", rectangle.size());

let circle = egcircle!((15, 6), 1, stroke = Some(BinaryColor::On));
display.draw(&circle);
println!("circle: {:?}", circle.size());
```

## Stdout
```
rectangle: UnsignedCoord(2, 2)
circle: UnsignedCoord(2, 2)
```

## Display
![dimensions](https://user-images.githubusercontent.com/226123/61818339-e854d080-ae50-11e9-9eeb-488dfab4fb03.png)

caemor:
It seems like all of the primitives have this "error"/undocumented behaviour. (see added simulation example in my branch but with already "fixed" behaviour - you can change this in the cargo.toml of the simulator )
But I found multiple confusing parts:
- Lines: A line of size 1 is not drawn at all and there is even a test for it:
 https://github.com/jamwaffles/embedded-graphics/blob/2ddbaad976a67dcb55d5023ed81384128e795726/embedded-graphics/src/primitives/line.rs#L309-L315
 I would have expected that a line of size 1 is the same as a dot, since for a variable sized line of lengths from zero to let's say ten you would need a special case just for this line of size 1 to make it a dot instead of a line
- Circles: You can't make circles with an even size (you always get size 3,5,7...), the opposite is for filled circles without a stroke color where you only have even sized fillings
- Triangles: Triangles of Size One are not shown - is it same no dot policy as for lines?
- Rectangles: Here you can make dots with 1 sized rectangles :-D

My branch with the tests and maybe fixes can be found here: https://github.com/caemor/embedded-graphics/tree/fix-dimension-size
I also added a new simulation example called size to show the different sizes.

jamwaffles:
Thanks for adding some extra detail. Can you open a PR with your fixes so it's easier to review?

The dot/no dot thing has cropped up before with I think a PR contributed to change the behaviour to _not_ draw a dot. Is there a common convention/expectation for elements with no size that you know of? Things start to get a bit weird with dimensions as well - what size should a line from `(10,10)` to `(10,10)` have? `0`? `1`?

[...]

PR #158, opened 2019-08-21
Rename (Unsigned)Coord to Point and Size

### LINKED ISSUE #141: Coordinate types
e-g currently uses two coordinate types `Coord` and `UnsignedCoord`. Their name is purely based on the data type they use internally and not on their intended use. This makes it harder to know which one a function expects.

I think it would be better to use an approach similar to servo's euclid crate or flutter's Offset and Size classes. This could be accomplished by renaming `Coord` to `Point` or `Position` and `UnsignedCoord` to `Size`.

The one place where this semantic meaning isn't already followed is the `Pixel` type which uses `UnsignedCoord` to make sure that only positive coordinates can be drawn. But changing this to use signed coordinates could be beneficial. Currently every drawing iterator has to manually check if all coordinates are >= 0, which can be quite error prone. This check could be easily implemented in the display drivers, which need to check for too large coordinates anyway. Not clipping the drawing inside the iterators would also allow to translate the pixels using `map()` without loosing the normally hidden parts. And using one type for all coordinates will get rid of many `to_unsigned()` and `to_signed()` calls.

Both examples mentioned above also implement a `Rect` type which can be constructed by combining a `Point` and a `Size`, which could be useful in many places.

### LINKED ISSUE #155: `.map()`ping iterators doesn't retain clipped pixels
Crap title. Let me explain. From #141:

> Currently every drawing iterator has to manually check if all coordinates are >= 0, which can be quite error prone. This check could be easily implemented in the display drivers, which need to check for too large coordinates anyway. Not clipping the drawing inside the iterators would also allow to translate the pixels using map() without loosing the normally hidden parts.

e-g iterators will check if a pixel coordinate is less than `(0,0)`. This is usually fine, however if `.map()` is called on the iterator when drawing to the display, those off-screen pixels won't show up. Example:

image

This is obviously incorrect; those discarded pixels should be retained.

The less-than-`(0,0)` checking should be moved to driver implementations so that the unexpected behaviour described above doesn't occur.

**This is a breaking change.**

rfuest:
This PR implements the name changes discussed in #141. The comments aren't updated yet and some tests are ignored at the moment (mostly in the image module, which will be changed anyway in #152).

Some notes:
* I removed the `Neg` implementation for `UnsignedCoord`, which didn't seem to make sense for `Size`
* I added support for `point + size` and `point - size` which return a `Point`. `point += size` and `point -= size` are also implement.
* I removed the `To(Un)Signed` traits.
  `UnsignedCoord` to `Coord` can be replaced by `Point::orign() + size`, when necessary.
  `Coord` to `UnsignedCoord` is replaced by the `Size::from_bounding_box` function. (Not a great name, but maybe this could be replaced by a function on the `Rect(angle)` struct soon)

Questions:
* Names! Any suggestions for a name for the module, which contains `Point` and `Size`? I called it `geometry` for now.
* The `icoord` and `ucoord` macros are replaced by `point` and `size` functions. This works and I like to not use a macro if not necessary, but `point` and `size` are pretty common name and `point!` and `size!` macros would use another namespace. Do you think we should change this, or keep it?
* Do we need the `Point::clamp_positive` function? It's not used anywhere in e-g and I can't immediately see where this would be useful.

Closes #155 
Closes #141


Cargo.toml
```diff
@@ -6,3 +6,9 @@ members = [
     "tinybmp",
     "tinytga",
 ]
+
+[patch.crates-io]
```
jamwaffles: I asked the same question in #152 - can you explain the reasoning behind this change? Does it allow crates.io releases to work as normal, whilst enabling development against unreleased versions or something?

embedded-graphics/src/geometry/mod.rs
```diff
@@ -0,0 +1,17 @@
+//! Geometry module.
+
+mod point;
+mod size;
+
+pub use point::Point;
+pub use size::Size;
+
+/// Creates a new `Point`.
+pub const fn point(x: i32, y: i32) -> Point {
```
jamwaffles: Hm, doesn't look like these `point()` and `size()` fns are really necessary now. Sure they're a bit shorter than calling `Point::new()`, but it confuses the API surface when there are two ways to create a `Point`. It made a bit more sense when they were macros and the name `UnsignedCoord` is a lot longer, but now, probably not. It also confuses between `Point(10, 20)` (where `Point` is a tuple struct) and `point(10, 20)` (this function). Thoughts?

embedded-graphics/src/geometry/point.rs
```diff
@@ -0,0 +1,368 @@
...
+    pub y: i32,
+}
+
+impl Point {
+    /// Create a new coordinate with X and Y values
+    pub const fn new(x: i32, y: i32) -> Self {
+        Point { x, y }
+    }
+
+    /// Creates a point with x and y equal to zero.
```
jamwaffles: Capitalisation consistent with above doc comment

```suggestion
    /// Creates a point with X and Y equal to zero.
```

embedded-graphics/src/geometry/point.rs
```diff
@@ -0,0 +1,368 @@
...
+    ///
+    /// assert_eq!(point.clamp_positive(), Point::new(0, 10));
+    /// ```
+    pub fn clamp_positive(&self) -> Self {
+        Point::new(self.x.max(0), self.y.max(0))
+    }
+
+    /// Remove the sign from a coordinate
+    ///
+    ///
```
jamwaffles: Dumb nit: extra line

```suggestion
```

embedded-graphics/src/geometry/point.rs
```diff
@@ -0,0 +1,368 @@
...
+    fn add(self, other: Point) -> Point {
+        Point::new(self.x + other.x, self.y + other.y)
+    }
+}
+
+impl Add<Size> for Point {
+    type Output = Point;
+
+    fn add(self, other: Size) -> Point {
+        Point::new(self.x + other.width as i32, self.y + other.height as i32)
```
jamwaffles: As mentioned in some of my previous PRs, what if `u32::max_value()` is used here? Would it be worth changing `Size` to use `u16`s for storage so this conversion is supported? Embedded targets _probably_ don't need more than 65k values but someone out there might disagree.

There are some other cases down this file that would also suffer from the same problem.

embedded-graphics/src/geometry/size.rs
```diff
@@ -0,0 +1,244 @@
...
+
+impl Size {
+    /// Create a new coordinate with X and Y values
+    pub const fn new(width: u32, height: u32) -> Self {
+        Size { width, height }
+    }
+
+    /// Creates a size from two corner points of a bounding box.
+    ///
+    /// TODO: issue #142
```
jamwaffles: This probably shouldn't be a doc comment

embedded-graphics/src/geometry/size.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub fn from_bounding_box(corner_1: Point, corner_2: Point) -> Self {
```
jamwaffles: Naming suggestion: `from_points` or is that too abstract? `from_corners`? 😬 

```suggestion
    pub fn from_points(corner_1: Point, corner_2: Point) -> Self {
```

embedded-graphics/src/geometry/point.rs
```diff
@@ -0,0 +1,368 @@
...
+}
+
+impl Point {
+    /// Create a new coordinate with X and Y values
+    pub const fn new(x: i32, y: i32) -> Self {
+        Point { x, y }
+    }
+
+    /// Creates a point with x and y equal to zero.
+    pub const fn origin() -> Self {
```
jamwaffles: Do you think `Point::zero()` or even `Point::default()` (with a `Default` impl) might be better than `origin()`? `origin()` implies (to me at least) that the origin can be other values than `(0, 0)`.

embedded-graphics/src/geometry/point.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// Clamp coordinate components to positive integer range
+    ///
+    /// ```
+    /// # use embedded_graphics::geometry::Point;
+    /// #
+    /// let point = Point::new(-5, 10);
+    ///
+    /// assert_eq!(point.clamp_positive(), Point::new(0, 10));
+    /// ```
+    pub fn clamp_positive(&self) -> Self {
```
jamwaffles: If it's unused, then yeah let's remove it. If there's a desperate need for it in the future, this method can be added back in.

Can you add a note to `CHANGELOG.md` for this breaking change? I'll pick it up and write a more comprehensive bullet point later.

embedded-graphics/src/image/image16bpp.rs
```diff
@@ -94,9 +96,7 @@ where
                 self.y += 1;
             }
 
-            if current_pixel[0] >= 0 && current_pixel[1] >= 0 {
-                break Pixel(current_pixel.to_unsigned(), C::from_raw_data(data));
-            }
+            break Pixel(current_pixel, C::from_raw_data(data));
```
jamwaffles: I think this is also a breaking change - images no longer self-check if pixels are greater than zero, so it's up to the display driver, right? A changelog note would be appreciated so I don't forget when I'm writing it up.

embedded-graphics/src/image/image16bpp.rs
```diff
@@ -140,14 +139,15 @@ mod tests {
...
-        assert_eq!(image.top_left(), Coord::new(100, 200));
-        assert_eq!(image.bottom_right(), Coord::new(103, 203));
-        assert_eq!(image.size(), UnsignedCoord::new(3, 3));
+        assert_eq!(image.top_left(), Point::new(100, 200));
+        assert_eq!(image.bottom_right(), Point::new(103, 203));
+        assert_eq!(image.size(), Size::new(3, 3));
     }
 
     #[test]
+    #[ignore]
```
jamwaffles: 🙁 ?

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -139,8 +138,7 @@ where
             top_left: self.top_left,
             bottom_right: self.bottom_right,
             style: self.style,
-            x: self.top_left[0],
-            y: self.top_left[1],
+            p: self.top_left,
```
jamwaffles: Could this be given a more descriptive name, or at least a doc comment explaining what it's for? It's got the same name in the other primitives also.

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -61,44 +60,22 @@ impl<C> Dimensions for Triangle<C>
...
-            self.p1[1]
-        } else if self.p2[1] <= self.p1[1] && self.p2[1] <= self.p3[1] {
-            self.p2[1]
-        } else {
-            self.p3[1]
-        };
-        Coord::new(x, y)
+    fn top_left(&self) -> Point {
+        let &x = [self.p1.x, self.p2.x, self.p3.x].iter().min().unwrap();
+        let &y = [self.p1.y, self.p2.y, self.p3.y].iter().min().unwrap();
```
jamwaffles: Thanks for making this code much nicer!

image

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -146,15 +123,15 @@ where
     }
 }
 
-fn sort_two_yx(p1: Coord, p2: Coord) -> (Coord, Coord) {
-    if p1[1] < p2[1] || (p1[1] == p2[1] && p1[0] < p2[0]) {
+fn sort_two_yx(p1: Point, p2: Point) -> (Point, Point) {
```
jamwaffles: Does/should `Point` and `Size` impl `Ord` and/or `PartialOrd`, etc?

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -411,130 +384,82 @@ where
...
-        assert_eq!(tri.p1, Coord::new(5, 10));
-        assert_eq!(tri.p2, Coord::new(15, 20));
-        assert_eq!(tri.p3, Coord::new(5, 20));
-        assert_eq!(tri.size(), UnsignedCoord::new(10, 10));
-
-        assert_eq!(moved.p1, Coord::new(-5, 0));
-        assert_eq!(moved.p2, Coord::new(5, 10));
-        assert_eq!(moved.p3, Coord::new(-5, 10));
-        assert_eq!(moved.size(), UnsignedCoord::new(10, 10));
+            Triangle::new(Point::new(5, 10), Point::new(15, 25), Point::new(5, 25));
```
jamwaffles: Did these values need to change? It adds noise to the diff.

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -411,130 +384,82 @@ where
...
-        let moved = tri.translate(Coord::new(10, 10));
+            Triangle::new(Point::new(5, 10), Point::new(15, 20), Point::new(10, 15));
+        let moved = tri.translate(Point::new(5, 10));
 
-        assert_eq!(moved.p1, Coord::new(15, 20));
-        assert_eq!(moved.p2, Coord::new(25, 30));
-        assert_eq!(moved.p3, Coord::new(20, 25));
+        assert_eq!(moved.p1, Point::new(10, 20));
+        assert_eq!(moved.p2, Point::new(20, 30));
+        assert_eq!(moved.p3, Point::new(15, 25));
```
jamwaffles: Did these values need to change? It adds noise to the diff.

jamwaffles:
Ah, missed this in my review:

> Names! Any suggestions for a name for the module, which contains Point and Size? I called it geometry for now.

`geometry` seems fine to me. Perhaps the `Dimensions` trait could be moved into it?

Cargo.toml
rfuest: It is used to use local versions for development, but I'm not sure how it's handled when uploading to crates.io.

embedded-graphics/src/geometry/mod.rs
rfuest: The main reason I added these was to replicate the old macros, but I don't really like them myself. I'll remove them.

embedded-graphics/src/geometry/point.rs
rfuest: I was expecting this comment because we did talk about this before.

We could change the conversion to use `try_into` and panic if the conversion from `u32` to `i32` fails.

embedded-graphics/src/primitives/triangle.rs
rfuest: They don't and probably shouldn't because there are multiple ways to order 2D points.

embedded-graphics/src/primitives/triangle.rs
rfuest: I changed some of the tests because they were resulting in equal x and y coordinates. Having different results for x and y makes it easier to spot if x and y where mixed up somewhere.

embedded-graphics/src/image/image16bpp.rs
rfuest: As I mentioned in the original comment for this PR, I didn't care to much about the tests in the image module because it will probably be replaced by the code in #152 anyway.

embedded-graphics/src/primitives/rectangle.rs
rfuest: Suggestions for a better name?

rfuest:
> `geometry` seems fine to me. Perhaps the `Dimensions` trait could be moved into it?

Yes, it could. Should I include this into this PR or add another one later?

Cargo.toml
jamwaffles: Ok, I'll merge #152 and release another alpha. Hopefully it doesn't have any effect.

embedded-graphics/src/geometry/point.rs
jamwaffles: We could, although I don't like the idea of panicking over a simple `+`. Do you know of any other `impl`s of `Add` out in the wild that panic in cases like this?

embedded-graphics/src/primitives/rectangle.rs
jamwaffles: It's the current point within the iterator, right? I don't mind keeping it at `p` for conciseness in the `next()` function, as long as there's a doc comment describing what it is. If you want to rename it, one of these perhaps?

* `current_point`
* `pos`
* `current_pos`
* `current_position`

jamwaffles:
Ouch, lots of conflicts from #152 :grimacing:

embedded-graphics/src/primitives/triangle.rs
jamwaffles: Ah I see. Good catch, thank you!

embedded-graphics/src/geometry/point.rs
rfuest: The normal primitive integer types panic on overflow in debug builds. Should we change this to be checked in debug builds and panic but just overflow in release builds?

rfuest:
> Ouch, lots of conflicts from #152 grimacing

That was expected, I'll update the PR later.

embedded-graphics/src/geometry/point.rs
jamwaffles: Sounds good to me. If we can stick as close to Rust's own behaviour as possible that would be best. Can you add some doc comments for this and other places that panic/overflow?

rfuest: I've added some comments and debug assertions to check if width or height are too large.

embedded-graphics/src/image/image16bpp.rs
rfuest: Yes, i've added a changelog note, but #155 also describes this change.

embedded-graphics/src/geometry/size.rs
rfuest: I would really like to remove this completely when we've found solutions for #142 and `Rect(angle)`. Should I leave the temporary name for now or change it?

jamwaffles: Ok, in that case leave the name 🙂. Could be worth adding a `#[deprecated]` attribute so it's harder to forget that this is left in.

embedded-graphics/src/image/image16bpp.rs
jamwaffles: Thanks, good point on #155 but all these changes are tricky to keep track of (for me at least) so I'd prefer to have the redundancy of a note in the changelog.

jamwaffles:
I've edited the description to close both #155 and #141 when this PR is merged. I don't think there are any outstanding items in those issues that aren't solved in this PR, but please let me know if I'm wrong.

There's also another merge conflict which is fun 👌. Once that's fixed I'll give this a final review and merge it - we resolved all the points that came up I think.

rfuest:
One outstanding item was the "from_bounding_box" function. I've changed it to be "pub(crate)" for now, so that we can decide on how to change the name or replace it later.

You also mentioned the name of the current point variables in the iterators. I left them unchanged, because I want to take another look at the iterators after #142 is closed and the name changes and/or additional documentation can be added then, if that's OK with you.

jamwaffles:
> One outstanding item was the "from_bounding_box" function. I've changed it to be "pub(crate)" for now

Good idea, thanks

> if that's OK with you

Cool and good, it's not a public facing API change - just something I wanted to make the code easier to read for me/you in 6 weeks/future maintainers.

PR #167, opened 2019-09-02
Clippy fixes

rfuest:
This PR fixes some issues clippy reported. (Two additional warnings are fixed in #166)

One warning remains:
```
warning: module has the same name as its containing module
 --> embedded-graphics/src/image/mod.rs:6:1
  |
6 | mod image;
  | ^^^^^^^^^^
  |
  = note: #[warn(clippy::module_inception)] on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#module_inception
```
Should we ignore this or move the contents of image/image.rs to image/mod.rs?


rfuest:
> I saw that warning too. It's a lot of code to move into `mod.rs` but it would reduce a little bit of confusion. Do what you think is best slightly_smiling_face.

OK, I'll move it.

> I also get these two warnings which seem like false-positives to me:

I don't think they are false positives. `BinaryColor` implements copy and can be treated the same way as an `u8`, for which you wouldn't normally use a reference. But you've already merged the fixes for these warnings anyway in #166 :wink:.

jamwaffles:
Mint. Can you give this branch a rebase and I'll give it a final check and merge it.

rfuest:
I've renamed image.rs to image_raw.rs. This fixes the warning and doesn't require to move all the code into `mod.rs`.

jamwaffles:
Oops, didn't see the build failures. Looks like an issue with `rustfmt`. Get that fixed and I'll merge :slightly_smiling_face:

PR #170, opened 2019-09-03
24x32 font support

TDHolmes:
I know the end game is to change how fonts are handled in general, but in the mean time, I needed a large font for a project I'm working on so I implemented a 24x32 font in the existing format. This is just a dumbly doubled version of the 12x16 font, but still looks pretty good. 

Here's a video of "hello Rust", with hello being in 12x16 and Rust being the new 24x32 font:
hello_rust


TDHolmes:
Wow, sorry for the wave of commits... Luckily you squash merge! At least I finally have `cargo test` compiling on my machine so this shouldn't happen again 😅

embedded-graphics/src/fonts/font24x32.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        assert_eq!(text.into_iter().count(), 24 * 32 * "Hello World!".len());
+    }
+
+    #[test]
+    fn text_dimensions() {
+        let hello: Font24x32<BinaryColor> = Font24x32::render_str("Hello World!");
+        let empty: Font24x32<BinaryColor> = Font24x32::render_str("");
+
+        assert_eq!(hello.size(), Size::new(288, 32));
```
BroderickCarlin: nit: 288 seems to be pretty arbitrary. Recommend creating a var out of `"Hello World!"` so that this number can be derived from the str length (Same for `text_corners` test)

jamwaffles: FWIW there's an example [here](https://github.com/jamwaffles/embedded-graphics/blob/master/embedded-graphics/src/fonts/font6x8.rs#L111).

TDHolmes: Should I fix all of the fonts then? All of the other fonts have tests with magic numbers in the same way mine does:
```rust
    #[test]
    fn text_dimensions() {
        let hello: Font6x8<BinaryColor> = Font6x8::render_str("Hello World!");
        let empty: Font6x8<BinaryColor> = Font6x8::render_str("");

        assert_eq!(hello.size(), Size::new(72, 8));
        assert_eq!(empty.size(), Size::new(0, 0));
    }
```

TDHolmes: I went ahead and refactored all of the font tests to not have significant magic numbers. Let me know what you think!

embedded-graphics/src/fonts/font24x32.rs
```diff
@@ -0,0 +1,243 @@
...
+        }
+        fallback
+    }
+}
+
+/// 24x32 pixel monospace font
+///
+/// There is also the [`text_24x32`] macro to provide an easier to use interface.
+///
+/// The 24x32 font is just a scaled up version of the 12x16 font, so [![here is the 12x16 font spritemap screenshot](https://raw.githubusercontent.com/jamwaffles/embedded-graphics/master/embedded-graphics/data/font12x16.png)](https://raw.githubusercontent.com/jamwaffles/embedded-graphics/master/embedded-graphics/data/font12x16.png)
```
jamwaffles: This looks a bit odd when rendered out as the link text is actually the image caption:

image

I'd suggest this instead:

```suggestion
/// The 24x32 font is a 2x scaling of the [12x16 font].
///
/// The 12x16 font sprite is shown here for reference:
///
/// [![12x16 font spritemap screenshot](https://raw.githubusercontent.com/jamwaffles/embedded-graphics/master/embedded-graphics/data/font12x16.png)](https://raw.githubusercontent.com/jamwaffles/embedded-graphics/master/embedded-graphics/data/font12x16.png)
```

With the link to the `12x16 font` added at the bottom of the doc comment:

```diff
+ /// [12x16 font]: ./type.Font12x16.html
```

The rendered doc then looks like this, which I think is much better:

image

Feel free to come up with something different if you think this can be improved further.

embedded-graphics/src/fonts/font24x32.rs
```diff
@@ -0,0 +1,243 @@
...
+    use crate::style::WithStyle;
+    use crate::transform::Transform;
+    use crate::Drawing;
+
+    const WIDTH: usize = Font24x32Conf::CHAR_WIDTH as usize;
+    const HEIGHT: usize = Font24x32Conf::CHAR_HEIGHT as usize;
+    const HELLO_WORLD: &'static str = "Hello World!";
+
+    #[test]
+    fn off_screen_text_does_not_infinite_loop() {
```
jamwaffles: I _think_ this test was made redundant by #158 - negative coordinate checks were moved out of e-g into the drivers' domain of responsibility. Seeing as you're refactoring the font tests anyway, can you remove all these `off_screen_text_does_not_infinite_loop` tests?

If you think it should be left in, the Y value in the `.translate()` call should be larger than the font height to shift the entire font into negative coordinates.

embedded-graphics/src/fonts/font6x12.rs
```diff
@@ -43,33 +43,46 @@ mod tests {
     use crate::transform::Transform;
     use crate::Drawing;
 
+    const WIDTH: usize = Font6x12Conf::CHAR_WIDTH as usize;
+    const HEIGHT: usize = Font6x12Conf::CHAR_HEIGHT as usize;
+    const HELLO_WORLD: &'static str = "Hello World!";
+
     #[test]
     fn off_screen_text_does_not_infinite_loop() {
```
jamwaffles: Same as other `off_screen_text_does_not_infinite_loop` comment.

embedded-graphics/src/fonts/font8x16.rs
```diff
@@ -46,33 +46,46 @@ mod tests {
     use crate::transform::Transform;
     use crate::Drawing;
 
+    const WIDTH: usize = Font8x16Conf::CHAR_WIDTH as usize;
+    const HEIGHT: usize = Font8x16Conf::CHAR_HEIGHT as usize;
+    const HELLO_WORLD: &'static str = "Hello World!";
+
     #[test]
     fn off_screen_text_does_not_infinite_loop() {
```
jamwaffles: Same as other `off_screen_text_does_not_infinite_loop` comment.

embedded-graphics/src/fonts/font6x8.rs
```diff
@@ -47,33 +47,46 @@ mod tests {
...
     use crate::Drawing;
 
+    const WIDTH: usize = Font6x8Conf::CHAR_WIDTH as usize;
+    const HEIGHT: usize = Font6x8Conf::CHAR_HEIGHT as usize;
+    const HELLO_WORLD: &'static str = "Hello World!";
+
     #[test]
-    fn text_with_negative_y_does_not_infinite_loop() {
-        let text: Font6x8<BinaryColor> = Font6x8::render_str("Hello World!")
+    fn off_screen_text_does_not_infinite_loop() {
```
jamwaffles: Same as other `off_screen_text_does_not_infinite_loop` comment.

embedded-graphics/src/fonts/font12x16.rs
```diff
@@ -43,33 +43,46 @@ mod tests {
     use crate::transform::Transform;
     use crate::Drawing;
 
+    const WIDTH: usize = Font12x16Conf::CHAR_WIDTH as usize;
+    const HEIGHT: usize = Font12x16Conf::CHAR_HEIGHT as usize;
+    const HELLO_WORLD: &'static str = "Hello World!";
+
     #[test]
     fn off_screen_text_does_not_infinite_loop() {
```
jamwaffles: Same as other `off_screen_text_does_not_infinite_loop` comment.

rfuest:
> @rfuest do you have anything to add?

The 12x16 already is a pixel doubled version of the 6x8 font, so doubling again gets really blocky. But as a temporary solution I'm OK with adding this, until something better is available.

A longer term solution for this would be to add a scaling option to font rendering. This way you could easily display large texts if font quality isn't important and/or you have limited storage space available for font data.

jamwaffles:
> The 12x16 already is a pixel doubled version of the 6x8 font

Didn't realise that!

> But as a temporary solution I'm OK with adding this...

Yes, I'm overlooking the fact it's just a 2x scaling for now until we have a better font/scaling story. I had a go with a `.map()` on the pixel iterator but didn't get very far. The scaling logic would have to be internal to the font/its iterator.

> ...until something better is available

e-g doesn't make any commitments to a particular typeface for its fonts, so it'll be easy to change this simple typeface the future.

rfuest:
> Yes, I'm overlooking the fact it's just a 2x scaling for now until we have a better font/scaling story. I had a go with a `.map()` on the pixel iterator but didn't get very far. The scaling logic would have to be internal to the font/its iterator.

A simple version of this, which works for all `Drawable`s, isn't hard to implement: https://gist.github.com/rfuest/2e83462c8e3733bf21d439202e530d8b
But this simple implementation would break `SizedDrawing`, which could be easily supported with scaling implemented in the font rendering itself. On the other hand having a simple way of scaling any `Drawable` when performance isn't critical could be a useful feature anyway.

Issue #150, opened 2019-08-03
Advanced example ideas


rfuest:
The recent discussions about how different functions of e-g could be implemented or improved made me think that e-g should have some more advanced examples. The current examples are pretty basic and only show how to use single aspect of e-g. Having more "real world" examples would make it much easier to evaluate where the API could be improved.

I currently don't have any really good ideas for such examples, expect from different styles of clocks (the hello world of graphics libraries :grin:). Please comment if you have any good ideas for examples, which might look visually interesting or could help to improve e-g for your use case.

jamwaffles:
Good point! I guess these should all go in the simulator so people can see what they look like. This is what I've got so far:

* Clocks sounds good - it would show some usage of loops or reusable functions (drawing the ticks) and some basic state management for updating the hands
* A super simple up/down menu system that shows some layout and style changes (like highlighting the current item)
* A "DVD player screensaver" type bouncing logo demo that shows animating an object around the screen
* Might be a bit too large for an example, but some kind of super simple game like Snake, controlled with the arrow keys.

The simulator doesn't currently support input or animation, but adding that shouldn't be too difficult, and would allow some much richer examples.

jamwaffles:
Another idea to add to the list:

A Raspberry Pi (for WiFi support) to show sunrise/sunset like DuckDuckGo does:

![image](https://user-images.githubusercontent.com/800791/64053240-635c8580-cb79-11e9-9610-80e50cdbc278.png)

Would demonstrate drawing text and more complex lines/shapes.

jamwaffles:
Another one:

A basic compass using an MPU9250 or similar

rfuest:
I've merged all example ideas into the top post.

@jamwaffles I did hide your comments that only contained new ideas, which are now added to the top post. I hope that's OK.
Are there any examples you are already working on or would like to work on soon? If not I would like to start working on the menu example next.

jamwaffles:
Thanks, that's much cleaner. I'm not working on anything at the moment so feel free to take the menu example. Perhaps open a draft PR and edit its number into this thread so we can keep track of who's doing what?

rfuest:
I don't have any WIP code yet to create a draft PR, but I'll open one and add it to the list when I do. Until then I will add a comment to the list.

I don't want to split this issues in separate issues for each example, but for other issues we should start to assign ourself to issues that we are working on to keep track of who's doing what.

JerelJr:
[...]

Issue #143, opened 2019-07-24
Circles without stroke are drawn incorrectly


rfuest:
The stroke on circles is applied from the outside edge of the circle inwards. This means that a circle with only fill set should render the same way as one with stroke and fill set to the same color.

## Code
```rust
display.draw(egcircle!(
    (10, 16),
    3,
    fill = Some(Rgb888::RED)
));

display.draw(egcircle!(
    (22, 16),
    3,
    fill = Some(Rgb888::GREEN),
    stroke = Some(Rgb888::GREEN)
));
```

## Display
![circles](https://user-images.githubusercontent.com/226123/61818999-62398980-ae52-11e9-898b-098d8bf05918.png)

jamwaffles:
Good catch, and thanks for the test case.

rfuest:
There still seems to be a problem with this for some radii. The test `issue_143_stroke_and_fill` fails if the radius is changed to 4 pixel. I noticed this in the `fill-macros` example, which uses a 32 pixel radius, so maybe all even radii cause an issue.

PR #171, opened 2019-09-05
Add more complicated clock example to simulator

jamwaffles:
~There are some findings listed at the top of the file which some issues should come out of~ - see #172, #173, #174.

This adds the clock example as suggested in #150.

image

TODO

* [x] Create issues for findings/at least discuss them
* [x] Remove findings from top of demo file
* [x] Massively clean up demo file - lots of dedupe/etc work can be done


simulator/examples/clock.rs
```diff
@@ -0,0 +1,182 @@
...
+//! # Development findings
+//!
+//! Some of these should probably be turned into issues and discussed before this example is
+//! released.
+//!
+//! 1. The primitives macros should accept both `()` tuples and variables/expressions for their
+//! coordinate positions. In this code, I often find myself replacing a macro with a call to
+//! `[Primitive]::new()` so I can pass in a `Point`
+//! 2. I often want a relative offset from a `Point`. Maybe we implement `Transform` for `Point`? Or
+//! a new trait/set of methods to do relative moves.
```
rfuest: How would this be different than adding 2 points?

simulator/examples/clock.rs
```diff
@@ -0,0 +1,182 @@
...
+    // Basic line hand
+    Line::new((CENTER, CENTER).into(), end).stroke(Some(BinaryColor::On))
+}
+
+fn main() {
+    let mut display = DisplayBuilder::new()
+        .title("Clock")
+        .size(DISP_SIZE as usize, DISP_SIZE as usize)
+        .scale(2)
+        .build_rgb();
```
rfuest: Why do you use an RGB display?

simulator/examples/clock.rs
jamwaffles: ... because I should've gone to bed long before I wrote this :wink:. I've removed this comment and just started adding/subbing `Point`s and `Size`s.

simulator/examples/clock.rs
jamwaffles: Oops, I wanted to see how colour conversions were handled. I've changed this to `build_binary()`.

jamwaffles:
Looks like this now:

image

embedded-graphics/src/primitives/mod.rs
```diff
@@ -47,6 +66,13 @@ macro_rules! egcircle {
         $crate::primitives::Circle::new($crate::geometry::Point::new($cx, $cy), $r)
             $( .$style_key($style_value) )*
     }};
+
+    ($center:expr, $r:expr $(, $style_key:ident = $style_value:expr )* $(,)?) => {{
```
rfuest: Wouldn't it be possible to use `Point::from($center)` in the macros? This should work with all types that implement `From<...> for Point` (including tuples and `Point` expressions) and only requires one macro with `$center::tt` as first parameter.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -83,18 +121,32 @@ macro_rules! egline {
         $crate::primitives::Line::new($crate::geometry::Point::new($x1, $y1), $crate::geometry::Point::new($x2, $y2))
             $( .$style_key($style_value) )*
     }};
+
+    ($start:expr, $end:expr $(, $style_key:ident = $style_value:expr )* $(,)?) => {{
```
rfuest: See `egcircle`

embedded-graphics/src/primitives/mod.rs
```diff
@@ -118,18 +181,34 @@ macro_rules! egrectangle {
         $crate::primitives::Rectangle::new($crate::geometry::Point::new($x1, $y1), $crate::geometry::Point::new($x2, $y2))
             $( .$style_key($style_value) )*
     }};
+
+    ($top_left:expr, $bottom_right:expr $(, $style_key:ident = $style_value:expr )* $(,)?) => {{
```
rfuest: See `egcircle`

embedded-graphics/src/primitives/mod.rs
```diff
@@ -153,16 +239,25 @@ macro_rules! egtriangle {
         $crate::primitives::Triangle::new($crate::geometry::Point::new($x1, $y1), $crate::geometry::Point::new($x2, $y2), $crate::geometry::Point::new($x3, $y3))
             $( .$style_key($style_value) )*
     }};
+
+    ($p1:expr, $p2:expr, $p3:expr $(, $style_key:ident = $style_value:expr )* $(,)?) => {{
```
rfuest: See `egcircle`

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+    let decoration_offset = Point::new(
+        (seconds_radians.cos() * (20.0)) as i32,
+        (seconds_radians.sin() * (20.0)) as i32,
+    );
+
+    // Add a fancy circle near the end of the hand
+    let decoration = Circle::new(end - decoration_offset, 5)
+        .fill(Some(BinaryColor::Off))
+        .stroke(Some(BinaryColor::On))
+        .stroke_width(1);
```
rfuest: This is unnecessary because `stroke_width` defaults to 1.

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+    // Basic line hand
+    Line::new(CENTER, end).stroke(Some(BinaryColor::On))
+}
+
+/// Draw digital clock just above center with black text on a white background
+///
+/// NOTE: The formatted time str must be passed in as references to temporary values in a
+/// function can't be returned.
+fn draw_digital_clock<'a>(time_str: &'a str) -> impl Iterator<Item = Pixel<BinaryColor>> + 'a {
+    let text = Font12x16::<BinaryColor>::render_str(&time_str)
```
rfuest: The turbofish shouldn't be required here.

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+            time.second()
+        );
+
+        // Draw digital clock just above center
+        display.draw(draw_digital_clock(&clock_text));
+
+        // Draw a small circle over the hands in the center of the clock face. This has to happen
+        // after the hands are drawn so they're covered up
+        display.draw(
+            Circle::new(CENTER, 6)
```
rfuest: IMO a smaller circle looks better, maybe `r = 3`

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+/// function can't be returned.
+fn draw_digital_clock<'a>(time_str: &'a str) -> impl Iterator<Item = Pixel<BinaryColor>> + 'a {
+    let text = Font12x16::<BinaryColor>::render_str(&time_str)
+        .stroke(Some(BinaryColor::Off))
+        .translate(CENTER - Size::new(48, 48));
+
+    // Add some padding on the background around the time digits, here it's 2px in each direction
+    let offset = Size::new(2, 2);
+
+    let background = Rectangle::new(text.top_left() - offset, text.bottom_right() + offset)
```
rfuest: Because the font rendering already adds some padding to the bottom right this looks better if the offset is only applied to the top left.

CHANGELOG.md
```diff
@@ -2,9 +2,15 @@
 
 Embedded Graphics is a `no_std` library for adding graphics features to display drivers. It aims to use the minimum amount of memory for builtin graphics objects by leveraging Rust's iterators to avoid large allocations. It targets embedded environments, but can run anywhere like a Raspberry Pi up to full desktop machines.
 
+## Unreleased
+
+### Added
+
+- Added a more complex `clock` example to the simulator - [check it out](https://github.com/jamwaffles/embedded-graphics/tree/embedded-graphics-v0.6.0-alpha.3/simulator/examples/clock.rs) for some more in-depth usage of Embedded Graphics.
```
rfuest: Could you rename the example to `analog_clock` or something similar? I have an idea for a way to draw seven segment displays and would probably demonstrate that with another clock example.

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+    );
+
+    // Create 12 `Line`s starting from the outer edge and drawing inwards by `tic_len` pixels
+    let tics = (0..12).into_iter().map(move |index| {
+        // Start angle around the circle, in radians
+        let angle = START + (PI * 2.0 / 12.0) * index as f32;
+
+        // Start point on circumference
+        let start = CENTER
+            + Point::new(
```
rfuest: Factoring out this code into a function like:
```rust
fn polar(angle: f32, radius: f32) -> Point {
    CENTER + Point::new(
        (angle.cos() * radius) as i32,
        (angle.sin() * radius) as i32,
    )
}
```
could remove some repeated code:
```rust
let start = polar(angle, SIZE as f32);
let end = polar(angle, SIZE as f32 - tic_len);
```

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+    face.into_iter().chain(tics.flatten())
+}
+
+/// Draw the seconds hand given a seconds value (0 - 59)
+fn draw_seconds_hand(seconds: u32) -> impl Iterator<Item = Pixel<BinaryColor>> {
+    // Convert seconds into a position around the circle in radians
+    let seconds_radians = ((seconds as f32 / 60.0) * 2.0 * PI) + START;
+
+    let end = CENTER
+        + Point::new(
```
rfuest: Use `polar` function.

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+            (seconds_radians.cos() * (SIZE as f32)) as i32,
+            (seconds_radians.sin() * (SIZE as f32)) as i32,
+        );
+
+    // Basic line hand
+    let hand = Line::new(CENTER, end).stroke(Some(BinaryColor::On));
+
+    // Offset from end of hand
+    let decoration_offset = Point::new(
```
rfuest: Use `polar` function.

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/// Draw the hour hand (0-11)
+fn draw_hour_hand(hour: u32) -> Line<BinaryColor> {
+    // Convert hour into a position around the circle in radians
+    let hour_radians = ((hour as f32 / 12.0) * 2.0 * PI) + START;
+
+    let hand_len = SIZE as f32 - 60.0;
+
+    let end = CENTER
```
rfuest: Use `polar` function.

simulator/examples/clock.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/// Draw the minute hand (0-59)
+fn draw_minute_hand(minute: u32) -> Line<BinaryColor> {
+    // Convert minute into a position around the circle in radians
+    let minute_radians = ((minute as f32 / 60.0) * 2.0 * PI) + START;
+
+    let hand_len = SIZE as f32 - 30.0;
+
+    let end = CENTER
```
rfuest: Use `polar` function.

simulator/examples/clock.rs
jamwaffles: `r = 4` and a fill looks good too 🙂. I'm gonna go with that and call this bikeshed resolved haha

jamwaffles:
Changes addressed. Thanks for the review! I renamed the demo to `analog-clock` to reduce conflicts.

simulator/examples/analog-clock.rs
```diff
@@ -155,33 +140,32 @@ fn main() {
...
+        display.draw(draw_digital_clock(&digital_clock_text));
 
         // Draw a small circle over the hands in the center of the clock face. This has to happen
         // after the hands are drawn so they're covered up
         display.draw(
-            Circle::new(CENTER, 6)
-                .fill(Some(BinaryColor::Off))
+            Circle::new(CENTER, 4)
+                .fill(Some(BinaryColor::On))
                 .stroke(Some(BinaryColor::On)),
```
rfuest: The stroke isn't normally necessary if the stroke and fill colors are equal, but #143 currently breaks circles without stroke. Should we remove the stroke here and accept the small visual error until #143 is fixed?

simulator/examples/analog-clock.rs
```diff
@@ -70,26 +72,18 @@ fn draw_seconds_hand(seconds: u32) -> impl Iterator<Item = Pixel<BinaryColor>> {
...
 
     // Basic line hand
     let hand = Line::new(CENTER, end).stroke(Some(BinaryColor::On));
 
     // Offset from end of hand
-    let decoration_offset = Point::new(
-        (seconds_radians.cos() * (20.0)) as i32,
-        (seconds_radians.sin() * (20.0)) as i32,
-    );
+    let decoration_offset = polar(seconds_radians, SIZE as f32 - 20.0);
```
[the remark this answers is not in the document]
jamwaffles: Renamed :)

simulator/examples/analog-clock.rs
jamwaffles: > Should we remove the stroke here and accept the small visual error until #143 is fixed?

Ah true, yeah makes sense. I've pushed a change that just removes the `.stroke()` call.

jamwaffles:
All good. The nightly build is failing but CircleCI doesn't support allowing failed build steps like Travis does, so feel free to ignore that.

PR #176, opened 2019-09-08
Font doc screenshots and write macro examples

### LINKED ISSUE #172: Add font screenshots to doc pages
A finding from #171.

A nice QoL improvement for those looking to see what the font looks like. I think for now this can 
just be the font bitmap images themselves as a lowest effort solution.

### LINKED ISSUE #173: Add `write!()` font examples
A finding from #171.

The font docs would be much more useful if there's an example of using a fixed length buffer (probably with [`arrayvec`](https://crates.io/crates/arrayvec) with the `write!()` macro to render formatted text in `no-std` contexts.

### LINKED ISSUE #174: Allow `Point * Scalar` and `Point / Scalar`
A question from #171

As the title. Questions:

1. Should this be supported/allowed (`Mul`, `MulAssign`, `Div`, `DivAssign`):

```rust
let point = Point::new(10, 20);
let larj_point = point * 10;

let size = Size::new(5, 5);
let larj_point = point * size;

let larj_point = point * Point::new(7, 7);
```

And naturally with `/` also

2. Should the above only support integers, or floats as well?

jamwaffles:
Closes #172 
Closes #173


rfuest:
I've edited your post to reference #172 and #173 instead of #173 and #174 to match the title.

rfuest:
Should we change the organization of the `font` module documentation?

Currently each font type contains the same examples (except for the used font) which seems a bit redundant. By moving the examples to the `font` module we could share the examples between all font types and they would be immediately be visible in the module-level documentation, which should also make it clearer that these examples work for all fonts.

jamwaffles:
> Should we change the organization of the font module documentation?

I had this in the front of my head whilst adding those examples. We probably should, yes. The homepage `Text with multiple fonts` link goes to the module which would immediately show examples to people browsing from the landing page. 

The only (tiny) negative I can see is the mild frustration with having to click a "Please see the module docs for examples" link, but it makes maintenance an awful lot easier. I'll dedupe the docs in a new commit to this PR.

embedded-graphics/src/fonts/font12x16.rs
```diff
@@ -27,98 +27,8 @@ impl FontBuilderConf for Font12x16Conf {
 ///
 /// # Examples
 ///
-/// ## Write some text to the screen at the default `(0, 0)` position
+/// See the [module docs](./index.html) for examples.
```
rfuest: The official rust docs always use `[module-level documentation]` for this link, which I also used in the image module docs.

rfuest:
> The only (tiny) negative I can see is the mild frustration with having to click a "Please see the module docs for examples" link, but it makes maintenance an awful lot easier. I'll dedupe the docs in a new commit to this PR.

I don't think that's a problem, because most users will discover the module-level documentation first and might not even look at the docs for the individual font structs.

embedded-graphics/src/fonts/font12x16.rs
jamwaffles: Thanks for this - I've updated the PR to be consistent with rust docs and the `image` module.

PR #177, opened 2019-09-08
Fix circle stroke fill

### LINKED ISSUE #143: Circles without stroke are drawn incorrectly

jamwaffles:
TODO

* [x] Add failing unit test
* [x] Actually fix the bug
* [x] Remove debugging demo from simulator before merge

Closes #143


rfuest:
I wanted to wait with fixing this until we've decided how coordinates should work. But if you can easily fix it now go ahead and with the additional test it hopefully won't break again later

> Is there a nicer way of comparing iterators?

There is, look at the `it_handles_negative_coordinates` test in `circle.rs` for an example.

jamwaffles:
> I wanted to wait with fixing this until we've decided how coordinates should work.

Sure, but I think this issue is somewhat separate. Once fixed, it can be tweaked to fit the new coordinates behaviour I think.

> There is, look at the it_handles_negative_coordinates test in circle.rs for an example.

Ah, I wasn't sure what to grep for and it's at the bottom of the same file I'm editing lol. Thanks. Shoulda gone to bed earlier...

jamwaffles:
All good. The actual change is this:

```diff
- let is_fill = len <= outer_radius_sq;
+ let is_fill = len <= outer_radius_sq + 1;
```

Pretty hacky and I don't fully understand the reasoning behind the off-by-one but it fixes the example given in #143.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -272,6 +271,23 @@ mod tests {
...
+    #[test]
+    fn issue_143_stroke_and_fill() {
+        let circle_no_stroke: Circle<BinaryColor> =
+            Circle::new(Point::new(10, 16), 3).fill(Some(BinaryColor::On));
+        let circle_stroke: Circle<BinaryColor> = Circle::new(Point::new(10, 16), 3)
+            .fill(Some(BinaryColor::On))
+            .stroke(Some(BinaryColor::On));
+
+        assert_eq!(circle_stroke.size(), circle_no_stroke.size());
+        assert_eq!(
```
rfuest: This assert is redundant, because the iters are checked for equality below.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -272,6 +271,23 @@ mod tests {
     use super::*;
     use crate::pixelcolor::BinaryColor;
 
+    #[test]
+    fn issue_143_stroke_and_fill() {
```
rfuest: We should define a default scheme to refer to GitHub issues in tests. I don't care what it is but having a defined way would make it easier to grep for in the future. I think currently only one other test exists that refers to an issue:
https://github.com/jamwaffles/embedded-graphics/blob/9196c227f4acd8b1d52bd270fbe2cd9c6257ba98/embedded-graphics/src/image/image_bmp.rs#L340-L342
Could you change one or the other to match?

jamwaffles: Good shout. I've made the names and comments consistent in 51d3eb9.

embedded-graphics/src/primitives/circle.rs
jamwaffles: Ah yeah, it's just noise at this point. Removed in b875b8c.

PR #180, opened 2019-09-14
Remove SizedDrawing trait

### LINKED ISSUE #123: SizedDrawing trait
I'm not sure I understand the way `SizedDrawing` is supposed to work. Doesn't this require special `Drawable` implementations, which make sure the data is continuous and in the correct iteration order?

The example in the documentation for `SizedDrawing` does only output 56 pixels to the display, but the bounding box contains 400 pixels. This will leave the display in an illegal state after the circle is drawn.

rfuest:
This closes #123 by removing the `SizedDrawing` trait.

@jamwaffles: Could you open a new issue to discuss how this might be implemented in a better way?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
   );
   ```
 
+### Removed
+
+- **(breaking)** The `SizedDrawing` trait is removed.
+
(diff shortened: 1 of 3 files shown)


jamwaffles:
> Could you open a new issue to discuss how this might be implemented in a better way?

#181 :grin:

PR #183, opened 2019-09-15
Add input handling to simulator

ReeceStevens:
@jamwaffles Firstly, thanks for the great work you are doing on `embedded-graphics`! I am using this library to build the front-end for a low-cost patient monitor written in Rust, and it is a great tool. I recall you mentioning this library a while ago on the embedded-rust IRC channel-- I'm glad to see it's growing in usage. I will be writing an implementation of this API for the RA8875 display driver very soon.

I wanted to put up a small PR to add basic input event handling to the embedded-graphics simulator. Although input event handling isn't in the scope of the embedded-graphics API, it is a very nice tool to add to the simulator from a developer perspective:

- A simple touch interface is common in many embedded device displays
- Typically, the touch screen controller and display driver are tightly bound at the firmware level, sometimes even being reported by the same chip
- For the purposes of simulation, adding input controls makes it significantly easier to quickly iterate on responsive UI designs without having to hook up the display hardware and deal with finicky drivers

It also seemed necessary to add input control at the simulator level since it wraps the SLD2 window instance, which handles all of the input events on the canvas. An input event solution running external to the simulator did not seem like an easy thing to pull off.

Let me know what you think when you have time to look this over-- and no worries if this is outside the scope of what you'd like in the library. Just thought I'd put it up as a PR since it's working well for me, and it might be useful to others!


rfuest:
> @rfuest You mentioned you were picking up a menu demo which will require input handling. Might be an idea to base it off this PR/wait for merge?

Yes, this definitely looks like a great start. @ReeceStevens would you be interested in extending this PR to also support keyboard input? I would, for example, like to be able control my menu demo using the cursor keys.

rfuest:
The coordinates used in the mouse events currently use SDL window coordinates, which don't take the display scaling into account. I'd prefer the mouse events to use `Point`s which correspond with the coordinates used by the drawing routines.

ReeceStevens:
@jamwaffles thanks for the feedback! Unfortunately I don't have a public repo for my patient monitor project yet, but I hope to open source it in the future. In the meantime I am trying to open source as many of the drivers as I can (e.g. the [`f4` board support crate][f4-crate]) and utility crates [like this digital filter one][digital-filter]. I intend to open source the RA8875 driver as well.

[f4-crate]: https://github.com/ReeceStevens/f4
[digital-filter]: https://github.com/ReeceStevens/digital-filter

And definitely, I understand about heavy development :) I've got a pretty heavy emphasis on unit + integration testing for the software side of things; I assume risk for any dependencies I pull in, so I do a lot of source code diving and try to cover my bases as much as is reasonably possible.

I've got an example running for clicking to move a red circle around the screen; I'll push that up shortly.

@rfuest I don't mind taking a crack at adding keyboard support, but it might be a little bit till I can get to it since keyboard controls are out of the scope of what I am using the simulator for currently. Keyboard events also seem a lot more context-dependent; would it make sense to just queue the entire `Event::Keydown` and `Event::Keyup` event and let the user decide what to do with it? (that implementation would certainly be more straightforward)

Also, good catch on the SDL window coordinates and display scaling, I will address that.

simulator/src/window.rs
```diff
@@ -87,10 +90,19 @@ impl Window {
                 } => {
                     return true;
                 }
+                Event::MouseButtonUp { x, y, .. } => {
+                    self.input_events
+                        .push(Point::new(x / self.scale as i32, y / self.scale as i32));
```
rfuest: This doesn't take the pixel spacing into account: https://github.com/jamwaffles/embedded-graphics/blob/e3412829cb16f19ec3d5a80a007e4f0b61132a47/simulator/src/window.rs#L69-L72

rfuest:
> I don't mind taking a crack at adding keyboard support, but it might be a little bit till I can get to it since keyboard controls are out of the scope of what I am using the simulator for currently.

That's great, I don't mind waiting a bit.

> Keyboard events also seem a lot more context-dependent; would it make sense to just queue the entire `Event::Keydown` and `Event::Keyup` event and let the user decide what to do with it? (that implementation would certainly be more straightforward)

I would prefer to not have a direct dependency on SDL2 by reusing it's types, but this would be the easiest solution. @jamwaffles what's your opinion on this?

jamwaffles:
> That's great, I don't mind waiting a bit.

Should this be in a different PR or the same one do you think?

> what's your opinion on this?

I don't really have one, but we can go with the simple solution for now and abstract it later. The simulator is more of an internal project (and is sub-1.0 too) so it's up for plenty of breakage :)

simulator/examples/input-handling.rs
```diff
@@ -0,0 +1,38 @@
...
+    display.draw(Circle::new(position, 100).fill_color(foreground_color));
+
+    loop {
+        let end = display.run_once();
+
+        if end {
+            break;
+        }
+
+        if let Some(Point { x, y }) = display.get_input_event() {
```
rfuest: Nit: The `Point` doesn't need to be destructured.
```suggestion
        if let Some(new_position) = display.get_input_event() {
```

simulator/examples/input-handling.rs
```diff
@@ -0,0 +1,38 @@
...
+            // Clear old circle
+            display.draw(Circle::new(position, 100).fill_color(background_color));
+            position = Point::new(x, y);
```
rfuest:
```suggestion
            position = new_position;
```

simulator/examples/input-handling.rs
```diff
@@ -0,0 +1,38 @@
...
+extern crate embedded_graphics;
+extern crate embedded_graphics_simulator;
+
+use embedded_graphics::pixelcolor::Rgb888;
+use embedded_graphics::prelude::*;
+use embedded_graphics::primitives::Circle;
+use embedded_graphics_simulator::DisplayBuilder;
+
+fn main() {
+    let mut display = DisplayBuilder::new().size(800, 480).build_rgb();
```
rfuest: Could you add some window title like: `Click to move the circle`. This would help if someone starts the example without reading it first.

simulator/examples/input-handling.rs
```diff
@@ -0,0 +1,38 @@
...
+    let background_color = Some(Rgb888::new(0, 0, 0));
```
rfuest: You could use the color constants here:
```suggestion
    let background_color = Some(Rgb888::BLACK);
```

simulator/examples/input-handling.rs
```diff
@@ -0,0 +1,38 @@
...
+    let foreground_color = Some(Rgb888::new(255, 0, 0));
```
rfuest:
```suggestion
    let foreground_color = Some(Rgb888::RED);
```

simulator/src/window.rs
```diff
@@ -87,10 +90,20 @@ impl Window {
                 } => {
                     return true;
                 }
+                Event::MouseButtonUp { x, y, .. } => {
+                    let pitch = (self.scale + self.pixel_spacing) as i32;
+                    self.input_events
+                        .push(Point::new(x / pitch, y / pitch as i32));
```
rfuest: The cast to `i32` isn't necessary any more.

rfuest:
> Should this be in a different PR or the same one do you think?

I don't mind, whatever @ReeceStevens prefers.

> I don't really have one, but we can go with the simple solution for now and abstract it later. The simulator is more of an internal project (and is sub-1.0 too) so it's up for plenty of breakage :)

OK, I agree.

ReeceStevens:
@rfuest I had a few more minutes today to work than I thought, so I went ahead and did the straightforward implementation of keyboard events. I also updated the example to move the circle around based on the arrow keys as well as mouse clicks. Let me know what you think-- it may be better just to forward the sdl2 events directly, but I made a separate enum so we could do the coordinate system conversion internally before it hits the user. Open to changing that if need be.

rfuest:
> Let me know what you think-- it may be better just to forward the sdl2 events directly, but I made a separate enum so we could do the coordinate system conversion internally before it hits the user.

Let's stick with this solution for now and see how it works out in practice.

I didn't have time to look at your recent changes in detail, but here are some suggestions for additional features that might be useful:
- Support for both `Up` and `Down` variants of the `Key` and `MouseButton` events
- Adding the `repeat` flag to the keyboard events
- Adding support for `MouseWheel` events, which can be used to emulate rotary encoders. Something simple like this should work:
    ```
    Event::MouseWheel {
        y, ..
    } => {
        if y > 0 {
            self.input_events.push(SimulatorEvent::MouseWheelUp);
        } else if y < 0 {
            self.input_events.push(SimulatorEvent::MouseWheelDown);
        }
    }
    ```

ReeceStevens:
Hey @rfuest thanks for the review-- I haven't forgotten about your comments, I am hoping to get to them either tonight or next Wednesday.

ReeceStevens:
@rfuest changes made

ReeceStevens:
@rfuest thanks for the review! Apologies for missing the details on the wheel event-- I should have read the SDL docs more carefully there.

I believe I have addressed all of your comments-- let me know if there is anything else that stands out to you.

simulator/src/window.rs
```diff
@@ -87,10 +135,83 @@ impl Window {
...
             }
         }
 
         false
     }
+
+    /// Return all captured input events
+    pub fn get_input_events(&mut self) -> Vec<SimulatorEvent> {
+        let input_events = self.input_events.clone();
+        self.clear_event_queue();
```
rfuest: Is the additional method `clear_event_queue` necessary? It's only called once and
```suggestion
        self.input_events.clear();
```
isn't much harder to understand (maybe with an additional comment).

ReeceStevens: D'oh, you're right. That was an oversight on my part.

rfuest:
LGTM. Thanks for all the work!

jamwaffles:
Mint! Thank you everyone!

PR #184, opened 2019-09-15
Add basic issue and PR templates

jamwaffles:
Just to save some back and forth, mainly for PRs where a changelog entry is missing, or the build is failing due to `rustfmt` throwing a fit. These are certainly not my finest work but should get the job done.

Anything else to add to the issue or PR template?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,8 @@
-* Version of embedded-graphics in use (if applicable): [version here]
+- Version of embedded-graphics in use (if applicable): [version here]
+- Target: [simulator] / [real hardware]
+- If using real hardware: [display driver used]
+
+Please link to the driver on <https://crates.io> if applicable.
 
 ## Description of the problem/feature request/other
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,11 @@
+* Version of embedded-graphics in use (if applicable): [version here]
+
+## Description of the problem/feature request/other
+
+[description here]
+
+## Test case (if applicable)
+
+```rust
+// Failing test case demonstrating the issue here
+```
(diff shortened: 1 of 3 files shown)


rfuest:
Did you check out the newer issue template: https://help.github.com/en/articles/creating-issue-templates-for-your-repository
I hadn't realized these existed before, but I've activated them in my e-g fork (with default template) if you want to try them from the user perspective.

jamwaffles:
Ugh why do they keep changing things :joy:. It looks like those new templates don't support PRs, so I'd say we go with the old style for now until PRs are supported.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+* Version of embedded-graphics in use (if applicable): [version here]
```
rfuest: Knowing the target platform would also be useful in some cases: Simulator or real hardware and which display/driver is used.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+Hi! Thank you for helping out with Embedded Graphics development! Please:
+
+* [ ] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading
+* [ ] Run `rustfmt` on the project - please check CI for failures
```
rfuest: > please check CI for failures

I think it could be clearer that this is only possible after the PR is already opened. All other checks are supposed to be done before opening the PR.

embedded-graphics/README.md
```diff
@@ -126,6 +126,23 @@ fn main() {
 
 There may be other drivers out there we don't know about yet. If you know of a driver to add to this list, please open [an issue](https://github.com/jamwaffles/embedded-graphics/issues/new)!
 
+## Development setup
+
+Ensure you have a reasonably recent version of Rust installed, preferably through <https://rustup.rs>.
```
rfuest:
```suggestion
Ensure you have the latest stable version of Rust installed, preferably through <https://rustup.rs>.
```

I would like to target the latest stable version for now, because e-g and the whole embedded ecosystem is relatively new and constantly changing. When e-g is more mature this should change to support the latest n stable releases.

embedded-graphics/README.md
```diff
@@ -126,6 +126,23 @@ fn main() {
...
+
+### Ubuntu/Linux Mint
+
+```bash
+# Ensure rustfmt is installed
```
rfuest: If you agree with my previous comment this should also include `rustup update`.

jamwaffles:
Comments addressed!

Issue #175, opened 2019-09-08
Extended display API


rfuest:
PR #171 adds clear functions to the simulator. It would be useful to have this function available for all display driver crates with e-g support, without having to use lower level APIs. 

This could be implemented by changing the `Drawing` trait to a `Display` trait that can provide additional functions:
```rust
pub trait Display<C>
where
    C: PixelColor,
{
    /// Draw an object from an iterator over its pixels
    fn draw<T>(&mut self, item: T)
    where
        T: IntoIterator<Item = drawable::Pixel<C>>;

    /// Returns the display size.
    fn size(&self) -> Size;

    /// Fills the display with a solid color.
    fn clear(&mut self, color: C);
}
```
The `size` function can be used to query the size of the display and layout elements independent of display size. This could also be used to provide a default implementation for `clear`:
```rust
fn clear(&mut self, color: C) {
   let r = Rectangle::new(Point::zero(), self.size()).fill(color);
   self.draw(r);
}
```
Which could be replaced by the driver crate if a more efficient way of clearing is available.

jamwaffles:
Sounds good to me, but a couple of things:

1. I think it should have a different name to `std::fmt::Display` (even though it's in a `no-std` context most of the time) to reduce conflicts. Maybe `DrawTarget` or `DrawingTarget`?
2. Is it worth either:
    * splitting `clear()` into `fn clear(&mut self)` and `fn fill_color(&mut self, color: C)`, or 
    * just having `fn clear(&mut self)`? 

    `clear`ing a display with a fill colour seems counterintuitive to me.

rfuest:
1. Good point. I think `DrawTarget` is OK and more generic than `Display`, for example for drawing to a image file.
2. IMO it's useful to have a way to fill the display with any color. If, for example, a UI uses a gray background a `clear` method that only supports black would be useless.

jamwaffles:
1. Cool and good
2. Absolutely, my comment was about whether to split clearing the display (filling with transparency) and filling with a colour into to separate methods, or leaving it as a single `clear()`. Calling `clear()` and having the display _fill_ with colour instead of _clearing_ seems counterintuitive.

rfuest:
I don't think there are many benefits of having a separate `clear` and `fill` method. Explicitly requiring the user to specify the color has the advantage that it works the same way as other fills and doesn't require an additional convention which color is used if none is specified.

[...]

PR #187, opened 2019-10-02
Extended display api

ReeceStevens:
## Note
There are a lot of changes in this PR, but mostly all of them are updating the usage of `draw` to be on the `Drawable` instead of the `DrawTarget`. The key changes are the modifications to those two traits (renamed `Display` -> `DrawTarget`).

I will also circle back around to this checklist and fill out more detail in this PR description :) 

## Checklist

## PR description

Extend the drawing API to allow hardware accelerated drawing functionality to be used transparently. This is done in two steps:

1. Change `Drawing` to `DrawTarget`, and provide primitive drawing methods with default `draw_iter()` implementations. This means that a display driver implementing `DrawTarget` does not have to specify anything other than the basic pixel drawing behavior, but can optionally supply hardware-accelerated methods to speed up the drawing of graphics primitives.

2. Add the `draw()` required method to the `Drawable` trait. This defines how an object is drawn by a `DrawTarget`, and it is the primary interface for the consumer of the library.


embedded-graphics/src/fonts/font_builder.rs
```diff
@@ -165,7 +166,7 @@ where
     }
 }
 
-impl<'a, C, Conf> IntoIterator for &'a FontBuilder<'a, C, Conf>
+impl<'a, 'b, C, Conf> IntoIterator for &'b FontBuilder<'a, C, Conf>
```
ReeceStevens: @jamwaffles this was also an important change to get the &Self trait bounds working. It looks like the reference in this trait implementation had its lifetime tied to the lifetime of the string reference inside FontBuilder; however, in order to implement the `Drawable` trait, those lifetimes needed to be made independent.

ReeceStevens:
Unfortunately, I've hit a bit of a roadblock with the tests. In particular:

With a trait definition of this:

```rust
/// Marks an object as "drawable". Must be implemented for all graphics objects
pub trait Drawable<C>
where
    C: PixelColor,
    for<'a> &'a Self: IntoIterator<Item = Pixel<C>>,
{
    /// Draw the graphics object. Override this method with primitive drawing methods as
    /// applicable.
    fn draw<T: DrawTarget<C>>(&self, display: &mut T) {
        display.draw_iter(self);
    }
}
```

We assert that in order to be `Drawable`, a type must be able to convert a reference to itself into an iterator. Unfortunately it looks like that is not true for core::iter types. You can see this most clearly in the analog-clock example, when specifying the return type of the `draw_face()` function:

```
error[E0277]: `&'a std::iter::Chain<embedded_graphics::primitives::circle::CircleIterator<embedded_graphics::pixelcolor::binary_color::BinaryColor>, st
d::iter::Flatten<std::iter::Map<std::ops::Range<{integer}>, [closure@simulator/examples/analog-clock.rs:50:40: 63:6 tic_len:_]>>>` is not an iterator
 --> simulator/examples/analog-clock.rs:38:19
  |
38| fn draw_face() -> impl Drawable<BinaryColor> {
  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ `&'a std::iter::Chain<embedded_graphics::primitives::circle::CircleIterator<embedded_graphics::pixelco
lor::binary_color::BinaryColor>, std::iter::Flatten<std::iter::Map<std::ops::Range<{integer}>, [closure@simulator/examples/analog-clock.rs:50:40: 63:6
tic_len:_]>>>` is not an iterator
  |
  = help: the trait `std::iter::Iterator` is not implemented for `&'a std::iter::Chain<embedded_graphics::primitives::circle::CircleIterator<embedded_g
raphics::pixelcolor::binary_color::BinaryColor>, std::iter::Flatten<std::iter::Map<std::ops::Range<{integer}>, [closure@simulator/examples/analog-clock
.rs:50:40: 63:6 tic_len:_]>>>`
  = note: required because of the requirements on the impl of `embedded_graphics::drawable::Drawable<embedded_graphics::pixelcolor::binary_color::Binar
yColor>` for `std::iter::Chain<embedded_graphics::primitives::circle::CircleIterator<embedded_graphics::pixelcolor::binary_color::BinaryColor>, std::it
er::Flatten<std::iter::Map<std::ops::Range<{integer}>, [closure@simulator/examples/analog-clock.rs:50:40: 63:6 tic_len:_]>>>`
  = note: the return type of a function must have a statically known size
```

Do y'all have any ideas on how to proceed? It seems like the core problem here is that `core::iter::Chain` does not necessarily imply that `&core::iter::Chain` implements `IntoIterator`.

Alternatively we could change `Drawable` to take ownership of `self` instead of handling a reference; however, that's clearly not desirable since then you could never use an object after drawing it. We could also return `Self` from draw, thus returning the reference to self after use? but that would produce some awkward use semantics, like

```rust
    let mut button = SomeDrawableButtonType();
    button = button.draw(&mut display)
```

ReeceStevens:
@jamwaffles @rfuest Aha! I realized that a reference to an iterator must be mutable in order to change the internal iterator state. Sure enough, changing all those `&self`s to `&mut self`s fixed up the compiler errors and the analog clock example is now working! When I get some more time either this week or next I'll patch up the remaining test cases.

jamwaffles:
Awesome! Well done for figuring that out! Please let me know when this PR is no longer WIP and I'll give it a review.

jamwaffles:
Something I noticed in my quick look at the code is the `draw_<primitive>()` methods. I can't remember if we decided, but doesn't this PR remove the need for those methods, whilst still supporting hardware acceleration? I like the simplicity of just having a `.draw()` method, but I might be missing something here. I think it's also a bit odd that there are two ways of drawing something; one with the builder pattern, the other with `.draw_<thing>()`. What if a crate implements a `Drawable` `Button` trait - it can be drawn with `Button::new().....` but `display.draw_button()` doesn't exist.

ReeceStevens:
@jamwaffles Good question RE: `draw_<primitive>()` methods. I believe that, given the current approach,  we _do_ still need to have the individual primitive handles. The reason for this is that primitive-specific (or in other words, hardware-specific) drawing methods are defined per-display rather than defined by the primitive.

The benefit of this approach, though, is that `draw_<primitive>` methods should never have to be used by a consumer of the library. A consumer can transparently call `Rectangle().draw(&mut display)`; the primitive is already defined to access the primitive-specific draw method, which falls back to the basic iterator approach unless the display explicitly provided a different implementation. If it were possible to make private trait methods (I'm not sure if it is, but I think not?) we could mark those `draw_<primitive>` methods as private. In lieu of that, we could put a statement in the docstring for those functions strongly dissuading using them or mangling the name (`fn __internal__draw_rectangle()` or something).

Regarding the two ways to draw something, I agree that's a downside to this approach. Really, we never want a user to use any interface other than `<Drawable>.draw(&mut <DrawTarget>)` to draw things; the "draw" methods on `DrawTarget` are strictly for use in an `impl Drawable for <sometype>` block. I'm not sure of a good way to "hide" this, but we may be able to make some explicit statement in the documentation explaining this. Or perhaps there is a better naming convention we can use to avoid saying "draw" so much :)

I have finally fixed up all the tests and examples, so everything is now working with the new approach. I'll go ahead and say this is no longer a WIP and is instead a proposal, although the documentation changes still need to be made.

ReeceStevens:
(seems there's a problem downloading rustfmt on the most recent nightly, hence the precheck-nightly test failing on circleci)

ReeceStevens:
@rfuest I implemented your suggestion in #175 of easing the `IntoIterator` bound on `Drawable`, and I think it greatly increased the ergonomics of the Drawable/DrawTarget system (see bfbb5bd, in particular the docstring example in `embedded_graphics/src/drawable.rs`).

embedded-graphics/src/image/image_bmp.rs
```diff
@@ -26,11 +27,11 @@ use tinybmp::Bmp;
...
 ///
 /// // Load `patch_16bpp.bmp`, a 16BPP 4x4px image
-/// let image = ImageBmp::new(include_bytes!("../../../assets/patch_16bpp.bmp")).unwrap();
+/// let mut image = ImageBmp::new(include_bytes!("../../../assets/patch_16bpp.bmp")).unwrap();
 ///
 /// // Equivalent behavior
-/// display.draw(&image);
-/// display.draw(image.into_iter());
+/// image.into_iter().draw(&mut display);
+/// image.draw(&mut display);
```
jamwaffles: This doc comment needs refactoring - the `Note` part doesn't make sense any more, and I think the last two lines could be collapsed to the last `image.draw(&mut display);`. Probably worth checking other doc comments as well. I'll point out any I spot in this review but I might miss some.

embedded-graphics/src/image/image_bmp.rs
```diff
@@ -121,10 +122,33 @@ where
...
-impl<'a, C> Drawable for ImageBmp<'a, C> where C: PixelColor + From<<C as PixelColor>::Raw> {}
+impl<'a, 'b, C> IntoIterator for &'b ImageBmp<'a, C>
+where
+    'b: 'a,
+    C: PixelColor + From<<C as PixelColor>::Raw>,
+{
+    type Item = Pixel<C>;
+    type IntoIter = ImageBmpIterator<'a, C>;
+
+    // NOTE: `self` is a reference already, no copies here!
```
jamwaffles: Can you lose this copypasta? I'm sick of seeing my comment everywhere :joy:

embedded-graphics/src/image/image_bmp.rs
```diff
@@ -231,6 +263,11 @@ mod tests {
         ))
         .unwrap()
         .translate(Point::new(-1, -1));
+
+        {
+            assert_eq!(image.into_iter().count(), 9);
+        }
+
```
jamwaffles: Hm?

embedded-graphics/src/lib.rs
```diff
@@ -177,17 +179,25 @@ pub mod transform;
...
 use crate::pixelcolor::PixelColor;
 
-/// To use this crate in a driver, `Drawing` must be implemented. This allows display drivers to
-/// support all embedded_graphics objects through the `draw()` method.
+/// To use this crate in a driver, `DrawTarget` must be implemented. This trait defines how a
+/// display draws pixels, and optionally provides a way to define accelerated drawing methods for
+/// graphical primitives such as lines, rectangles, triangles, and circles.
+///
+/// Once a `DrawTarget` is defined, it can be used to render [`Drawable` objects][drawable]. Note
+/// that any object that can produce an iterator of `Pixels` has a default implementation for the
```
jamwaffles: Nit

```suggestion
/// that any object that can produce an iterator of `Pixel`s has a default implementation for the
```

embedded-graphics/src/lib.rs
```diff
@@ -177,17 +179,25 @@ pub mod transform;
...
+/// `Drawable` trait-- see the trait documentation for more details.
```
jamwaffles: Nit

```suggestion
/// `Drawable` trait. See [the `Drawable` trait documentation](drawable) for more details.
```

embedded-graphics/src/lib.rs
```diff
@@ -232,18 +237,130 @@ use crate::pixelcolor::PixelColor;
...
+///     // Update the display
+///     display.flush().expect("Failed to send data to display");
+/// }
+///
+/// ```
+pub trait DrawTarget<C>
 where
     C: PixelColor,
 {
+    /// Draw a pixel on the display
```
jamwaffles: Should there be an extra note here explaining that some displays will require a `flush()` of some sort to actually write the framebuffer? Others will work in immediate mode so won't require the `flush()` call. Should this be commented on here, or left to display drivers to describe?

embedded-graphics/src/lib.rs
```diff
@@ -232,18 +237,130 @@ use crate::pixelcolor::PixelColor;
...
+    fn draw_pixel(&mut self, item: drawable::Pixel<C>);
+
     /// Draw an object from an iterator over its pixels
```
jamwaffles: Same as above flush comment

embedded-graphics/src/lib.rs
```diff
@@ -232,18 +237,130 @@ use crate::pixelcolor::PixelColor;
...
     where
-        T: IntoIterator<Item = drawable::Pixel<C>>;
+        T: IntoIterator<Item = drawable::Pixel<C>>,
+    {
+        for pixel in item {
+            self.draw_pixel(pixel);
+        }
+    }
+
+    /// Draw a line primitive.
```
jamwaffles: Nit

```suggestion
    /// Draw a line primitive
```

@rfuest likes full stops but the other doc comments on this trait don't have them so let's keep this consistent

embedded-graphics/src/mock_display.rs
```diff
@@ -166,22 +166,18 @@ where
...
-        for Pixel(Point { x, y }, color) in item_pixels {
-            if x < 0 || y < 0 || x >= SIZE as i32 || y >= SIZE as i32 {
-                continue;
-            }
-
-            let i = x + y * SIZE as i32;
-            self.0[i as usize] = Some(color);
+    fn draw_pixel(&mut self, pixel: Pixel<C>) {
+        let Pixel(Point { x, y }, color) = pixel;
+        if x < 0 || y < 0 || x >= SIZE as i32 || y >= SIZE as i32 {
```
jamwaffles: Just having some fun with Rust syntax:

```suggestion
        if !(0..SIZE).contains(&x) || !(0..SIZE).contains(&y) {
```

This also [optimises out to better assembly](https://godbolt.org/z/5LGkBo). Completely irrelevant for an example so feel free to ignore this comment, I just thought the range check is cool.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -24,23 +25,23 @@ use crate::style::WithStyle;
...
 ///     .stroke_color(None)
 ///     .fill_color(Some(Rgb565::BLUE))
 ///     .translate(Point::new(65, 35));
 ///
-/// display.draw(c1);
-/// display.draw(c2);
-/// display.draw(c3);
+/// c1.draw(&mut display);
+/// c2.draw(&mut display);
+/// c3.draw(&mut display);
```
jamwaffles: Could these be moved onto the ends of the `Circle::new()` chains above? The `let mut`s could be removed then as well.

embedded-graphics/src/primitives/line.rs
```diff
@@ -24,19 +25,19 @@ use crate::style::WithStyle;
...
-/// let l3 = Line::new(Point::new(50, 20), Point::new(60, 35))
+/// let mut l3 = Line::new(Point::new(50, 20), Point::new(60, 35))
 ///     .translate(Point::new(65, 35));
 ///
-/// display.draw(l1);
-/// display.draw(l2);
-/// display.draw(l3);
+/// l1.draw(&mut display);
+/// l2.draw(&mut display);
+/// l3.draw(&mut display);
```
jamwaffles: Same as circle example comment

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -24,21 +25,21 @@ use crate::style::WithStyle;
...
-/// let r3 = Rectangle::new(Point::new(50, 20), Point::new(60, 35))
+/// let mut r3 = Rectangle::new(Point::new(50, 20), Point::new(60, 35))
 ///     .translate(Point::new(65, 35));
 ///
-/// display.draw(r1);
-/// display.draw(r2);
-/// display.draw(r3);
+/// r1.draw(&mut display);
+/// r2.draw(&mut display);
+/// r3.draw(&mut display);
```
jamwaffles: Same as circle example comment

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -25,19 +26,19 @@ use crate::style::WithStyle;
...
-/// let t3 = Triangle::new(Point::new(50, 20), Point::new(60, 35), Point::new(70, 80))
+/// let mut t3 = Triangle::new(Point::new(50, 20), Point::new(60, 35), Point::new(70, 80))
 ///     .translate(Point::new(65, 35));
 ///
-/// display.draw(t1);
-/// display.draw(t2);
-/// display.draw(t3);
+/// t1.draw(&mut display);
+/// t2.draw(&mut display);
+/// t3.draw(&mut display);
```
jamwaffles: Same as circle example comment

embedded-graphics/src/drawable.rs
```diff
@@ -1,11 +1,73 @@
...
+///     text: &'a str
+/// }
+///
+/// impl<'a, C: 'a> Drawable<'a, C> for Button<'a, C>
+/// where
+///     C: PixelColor + From<BinaryColor>,
+/// {
+///     fn draw<D: DrawTarget<C>>(&mut self, display: &mut D) {
+///         egrectangle!(self.p1, self.p2, stroke_color=Some(self.bg_color)).draw(display);
+///         text_6x8!(self.text, fill_color = Some(self.fg_color)).translate(Point::new(20, 16)).draw(display);
```
jamwaffles: Can you `rustfmt` this code block? This line in particular causes unnecessary horizontal scrolling.

embedded-graphics/src/drawable.rs
```diff
@@ -1,11 +1,73 @@
...
+///
+/// ```
+///
+/// [draw_iter]: ../trait.DrawTarget.html#method.draw_iter
+pub trait Drawable<'a, C>
+where
+    C: PixelColor,
+{
+    /// Draw the graphics object. Override this method with primitive drawing methods as
+    /// applicable.
```
jamwaffles: I'm not entirely sure what "Override this method with primitive drawing methods as applicable." means here. Can you clarify the comment a bit?

jamwaffles:
> Good question RE: draw_<primitive>() methods. I believe that, given the current approach, we do still need to have the individual primitive handles. The reason for this is that primitive-specific (or in other words, hardware-specific) drawing methods are defined per-display rather than defined by the primitive.

> The benefit of this approach, though, is that draw_<primitive> methods should never have to be used by a consumer of the library. A consumer can transparently call Rectangle().draw(&mut display); the primitive is already defined to access the primitive-specific draw method, which falls back to the basic iterator approach unless the display explicitly provided a different implementation.

Thanks for explaining this again - it's tricky keeping track of all these changes going on. Looking at the code makes this clear as well :)

> In lieu of that, we could put a statement in the docstring for those functions strongly dissuading using them or mangling the name (fn __internal__draw_rectangle() or something).

I think a doc comment discouraging their use by end-users is fine. The hardware acceleration example in the trait doc is excellent and should help driver writers implement things correctly.

> (seems there's a problem downloading rustfmt on the most recent nightly, hence the precheck-nightly test failing on circleci)

Don't worry about it :grin: it's not marked as a required build step anyway. 

Also, seeing as this PR aims to close #175, should the `clear()` and `size()` methods be added?

embedded-graphics/src/lib.rs
rfuest: @rfuest likes to follow: https://doc.rust-lang.org/1.0.0/style/style/comments.html#sentence-structure

jamwaffles: Ah, makes sense! We should do a docs pass before 0.6.0 to make everything consistent

jamwaffles: #189 :)

ReeceStevens:
@jamwaffles thanks for the review! I'm looking forward to sitting down and going through it line-by-line-- just wanted to give you a heads up that I'll be w/o internet access for the next week so it may be a bit before I can give a proper response.

jamwaffles:
No problem, thanks for letting me know!

embedded-graphics/src/drawable.rs
ReeceStevens: Ah, my `rustfmt` did not pick up on docstring code blocks. Fixed, thanks!

embedded-graphics/src/drawable.rs
ReeceStevens: That second sentence is unnecessary, I have revised this docstring. All that I intended to convey was that this method is where you draw the object via the exposed methods on the supplied `DrawTarget`.

embedded-graphics/src/fonts/mod.rs
```diff
@@ -107,11 +104,11 @@ mod font6x8;
...
-pub use self::font12x16::Font12x16;
-pub use self::font24x32::Font24x32;
-pub use self::font6x12::Font6x12;
-pub use self::font6x8::Font6x8;
-pub use self::font8x16::Font8x16;
+pub use self::font12x16::{Font12x16, Font12x16Conf};
+pub use self::font24x32::{Font24x32, Font24x32Conf};
+pub use self::font6x12::{Font6x12, Font6x12Conf};
+pub use self::font6x8::{Font6x8, Font6x8Conf};
+pub use self::font8x16::{Font8x16, Font8x16Conf};
```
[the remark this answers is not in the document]
ReeceStevens: This was a vestigial change from the old approach where any drawable object had to implement `IntoIterator` for a mutable reference to itself. This resulted in a user of the library having to specify the full concrete `IntoIterator` type for any new Drawable object. This type ended up getting really gnarly and required references to the `Font*Conf` in the case of text. Thankfully this is no longer required after removing the IntoIterator constraint. I'll remove these exports

embedded-graphics/src/image/image_bmp.rs
ReeceStevens: lol, that is embarrassing! Totally missed that. Turns out the copy-pasta isn't even necessary now so I'm just going to remove that whole impl for &mut.

embedded-graphics/src/image/image_bmp.rs
ReeceStevens: Leftover from an old attempt to fix some mutable reference issues, fixing

embedded-graphics/src/lib.rs
ReeceStevens: Hmm, that's a good point. That seems like a reasonable thing to caution about in the documentation-- it's good for consumers of an eg-compatible library to know about as well.

In that vein, should there be a `flush()` method on `DrawTarget`? It seems like it should be required, and immediate mode drivers can just implement it as a noop function.

embedded-graphics/src/mock_display.rs
ReeceStevens: Woah, nice! I had no idea that was an option. I also have never seen that Compiler Explorer website before-- very cool, I will definitely use that again in the future.

It doesn't hurt to show off some Rust power in examples-- I'll add it in :)

embedded-graphics/src/lib.rs
jamwaffles: > In that vein, should there be a flush() method on DrawTarget? It seems like it should be required, and immediate mode drivers can just implement it as a noop function.

I don't think so, that seems pretty driver dependent to me.

rfuest:
I didn't have time to review the code in detail, but I noticed a lot of added `mut` modifiers.

Did you try the code I suggested in #175? Sorry that I didn't explain the code better in that comment. By using `self` instead of a reference in the `Drawable` trait you can either implement `Drawable` for `&T` or `&mut T`. This way primitives can implement `Drawable` for `&T` and don't need to be mutable. And `Drawable` can still be implemented for `Iterator`s which need to be mutable.

embedded-graphics/src/lib.rs
ReeceStevens: @jamwaffles I just pushed a commit that added a default flush method to a `DrawTarget` that is just a noop. I can revert this if you would like; however, I thought it might make sense to include flushing as part of the interface since it would not be possible to write portable code for flushing displays if that functionality were not exposed. Furthermore, this seems to shadow what is done in the stdlib `io` crate, which requires a flushing method even though not all `Write` objects necessarily need it.

ReeceStevens:
@rfuest aha! I had missed the finer details of that code snippet you posted in #175 but now I see that `Drawable` was implemented on `&Rectangle`-- that seems like a good solution to this problem.

ReeceStevens:
@jamwaffles @rfuest I believe I have addressed all review comments-- thanks again for the reviews :) Please let me know if anything else stands out to you.

embedded-graphics/src/lib.rs
```diff
@@ -232,18 +243,183 @@ use crate::pixelcolor::PixelColor;
...
+
+    /// Draws a line primitive.
+    ///
+    /// This default trait method should be overridden if a display provides hardware-accelerated
+    /// methods for drawing lines.
+    ///
+    /// ## Caution
+    ///
+    /// This method should not be called directly from application code. It is used to define the
+    /// internals of the `draw()` method used for the `Line` primitive. To draw a line, call
```
rfuest: The types like `Line` in this and the methods below should be links.

embedded-graphics/src/pixelcolor/mod.rs
```diff
@@ -13,6 +13,7 @@
 //! ```
 //! use embedded_graphics::prelude::*;
 //! use embedded_graphics::egrectangle;
+//! use embedded_graphics::geometry::Size;
```
rfuest: Contained in the prelude.

embedded-graphics/src/pixelcolor/raw/mod.rs
```diff
@@ -66,7 +66,7 @@
 //!
 //! fn main() {
 //!     // Create new image with RGBI colors.
-//!     let image: Image<RGBI> = Image::new(IMAGE_DATA, 2, 2);
+//!     let mut image: Image<RGBI> = Image::new(IMAGE_DATA, 2, 2);
```
rfuest: Doesn't need to be mutable.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -147,6 +146,23 @@ where
...
+    fn into_iter(self) -> Self::IntoIter {
+        CircleIterator {
+            center: self.center,
+            radius: self.radius,
+            style: self.style,
+            p: Point::new(-(self.radius as i32), -(self.radius as i32)),
+        }
+    }
+}
+
```
rfuest: Not required anymore.

embedded-graphics/src/primitives/line.rs
```diff
@@ -157,6 +156,39 @@ impl<'a, C: PixelColor> IntoIterator for &'a Line<C> {
...
+            start: self.start,
+            end: self.end,
+            delta,
+            direction,
+            err: delta.x + delta.y,
+            stop: self.start == self.end, // if line length is zero, draw nothing
+        }
+    }
+}
+
```
rfuest: Not required anymore.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -143,6 +142,23 @@ where
...
+    fn into_iter(self) -> Self::IntoIter {
+        RectangleIterator {
+            top_left: self.top_left,
+            bottom_right: self.bottom_right,
+            style: self.style,
+            p: self.top_left,
+        }
+    }
+}
+
```
rfuest: Not required anymore.

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -189,6 +189,44 @@ where
...
+            next_ac,
+            next_b,
+            x: 0,
+            min_y: v1.y,
+            max_y: v3.y,
+            style: self.style,
+        }
+    }
+}
+
```
rfuest: Not required anymore.

embedded-graphics/tests/chaining.rs
```diff
@@ -1,9 +1,9 @@
 extern crate embedded_graphics;
 
-use embedded_graphics::geometry::Point;
+use embedded_graphics::geometry::{Point, Size};
```
rfuest: Include in the prelude.

embedded-graphics/tests/chaining.rs
```diff
@@ -1,9 +1,9 @@
...
 use embedded_graphics::prelude::*;
 use embedded_graphics::primitives::{Circle, Line, Rectangle};
-use embedded_graphics::Drawing;
+use embedded_graphics::DrawTarget;
```
rfuest: Include in the prelude.

simulator/examples/bmp.rs
```diff
@@ -17,15 +17,15 @@ use std::thread;
 use std::time::Duration;
 
 fn main() {
-    let image: ImageBmp<Rgb565> = ImageBmp::new(include_bytes!("./rust-pride.bmp")).unwrap();
+    let mut image: ImageBmp<Rgb565> = ImageBmp::new(include_bytes!("./rust-pride.bmp")).unwrap();
```
rfuest: Doesn't need to be mutable.

simulator/examples/tga.rs
```diff
@@ -17,15 +17,15 @@ use std::thread;
 use std::time::Duration;
 
 fn main() {
-    let image: ImageTga<Rgb888> = ImageTga::new(include_bytes!("./rust-pride.tga")).unwrap();
+    let mut image: ImageTga<Rgb888> = ImageTga::new(include_bytes!("./rust-pride.tga")).unwrap();
```
rfuest: Doesn't need to be mutable.

ReeceStevens:
@rfuest thanks for the feedback! I believe I've addressed all your comments except for the import/prelude related ones. I'll wait on changing that until @jamwaffles and you decide on whether to avoid prelude use internally or not.

embedded-graphics/src/fonts/font12x16.rs
```diff
@@ -39,9 +40,9 @@ mod tests {
     use crate::geometry::{Dimensions, Point, Size};
     use crate::mock_display::MockDisplay;
     use crate::pixelcolor::BinaryColor;
+    use crate::prelude::*;
     use crate::style::WithStyle;
     use crate::transform::Transform;
-    use crate::Drawing;
```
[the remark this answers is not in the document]
jamwaffles: I'd prefer not to use the prelude internally so it's clear which bits of the crate a module depends on. The prelude exists for end users that want a convenient way to import commonly used stuff.

jamwaffles:
This thread is getting super long so I'll repost my comment here in case you missed it:

> I'd prefer not to use the prelude internally so it's clear which bits of the crate a module depends on. The prelude exists for end users that want a convenient way to import commonly used stuff.

embedded-graphics/src/lib.rs
jamwaffles: Ok, that's fair. I didn't know about `flush` in `io` but that does make sense, particularly with a default impl.

embedded-graphics/src/lib.rs
```diff
@@ -174,20 +176,30 @@ pub mod primitives;
...
 pub mod transform;
 
-use crate::geometry::Dimensions;
+use crate::geometry::{Dimensions, Point, Size};
 use crate::pixelcolor::PixelColor;
+use crate::prelude::*;
 
-/// To use this crate in a driver, `Drawing` must be implemented. This allows display drivers to
-/// support all embedded_graphics objects through the `draw()` method.
+/// To use this crate in a driver, `DrawTarget` must be implemented. This trait defines how a
```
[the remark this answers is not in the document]
jamwaffles: Might be misunderstanding what you mean by "summary line", but isn't that line 183:

```rust
/// Defines a display that can be used to render [`Drawable'] objects.
```

?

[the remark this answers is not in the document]
rfuest: Yes it is, but this was added by @ReeceStevens in response to my comment.

The term "summary line" is used in https://doc.rust-lang.org/1.0.0/style/style/comments.html

embedded-graphics/src/drawable.rs
```diff
@@ -1,11 +1,75 @@
...
+/// [`draw_iter`]: ../trait.DrawTarget.html#method.draw_iter
+pub trait Drawable<C>
+where
+    C: PixelColor,
+{
+    /// Draw the graphics object using the supplied DrawTarget.
+    fn draw<T: DrawTarget<C>>(self, display: &mut T);
+}
+
+impl<C, T> Drawable<C> for T
```
rfuest: Could the `Drawable` for iterators use `&mut T` instead? That might be a little more flexible than taking ownership of the iterator. Although it won't matter in most cases, because iterators aren't reused normally.

embedded-graphics/src/lib.rs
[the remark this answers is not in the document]
jamwaffles: Ah right, I guess Github's diff view got me on that one :ok_hand:

embedded-graphics/src/fonts/mod.rs
```diff
@@ -134,10 +131,10 @@ where
     /// fn main() {
     ///     let mut disp = Display::default();
     ///     // Render a string with a red stroke
-    ///     let text = Font6x8::render_str("Hello world")
+    ///     let mut text = Font6x8::render_str("Hello world")
```
rfuest: Haven't checked this, but I think this `mut` could also be removed.
```suggestion
    ///     let text = Font6x8::render_str("Hello world")
```

embedded-graphics/src/lib.rs
```diff
@@ -174,20 +176,33 @@ pub mod primitives;
...
 pub mod transform;
 
-use crate::geometry::Dimensions;
+use crate::geometry::{Dimensions, Point, Size};
 use crate::pixelcolor::PixelColor;
+use crate::prelude::*;
 
-/// To use this crate in a driver, `Drawing` must be implemented. This allows display drivers to
-/// support all embedded_graphics objects through the `draw()` method.
+/// Defines a display that can be used to render [`Drawable'] objects.
```
rfuest:
```suggestion
/// Defines a display that can be used to render [`Drawable`] objects.
```

embedded-graphics/src/lib.rs
```diff
@@ -174,20 +176,33 @@ pub mod primitives;
...
+///
+/// To use this crate in a driver, `DrawTarget` must be implemented. This trait defines how a
+/// display draws pixels, and optionally provides a way to define accelerated drawing methods for
+/// graphical primitives such as lines, rectangles, triangles, and circles.
+///
+/// Once a `DrawTarget` is defined, it can be used to render [`Drawable`]. Note that any iterator
```
rfuest: Typo: You removed the last word of the sentence when you changed the link.
```suggestion
/// Once a `DrawTarget` is defined, it can be used to render [`Drawable`] objects. Note that any iterator
```
or maybe
```suggestion
/// Once a `DrawTarget` is defined, it can be used to render [`Drawable`]s. Note that any iterator
```

ReeceStevens:
Great, nice catches @rfuest and @jamwaffles! I've removed all internal use of prelude use statements except for docstring examples.

Thanks for being so thorough in the reviews-- this is my first non-trivial contribution to an open-source Rust project, so I'm still getting used to the workflow, tools, and conventions. Nice work with all the CI, documentation validation, and link checking. It makes it very easy to jump in and make sure I didn't break anything :)

jamwaffles:
Awesoommeee! Many thanks to you, @rfuest for being a thorough reviewer and to you @ReeceStevens for putting in the work implementing this feature. Thank you both for your considerable time spent on this PR! 🎉

PR #190, opened 2019-10-21
simulator speedup

jacobrosenthal:
Ive been meaning to dig into why the simulator is really slow, and in the process realized presumably this is probably what that extended graphics pr was partially addressing. Hope Im not stepping on anyones toes.

'works' and fast! But doesn't properly clean the screen--After a few moves there a multiple circles hanging around.
Lots of todos, too

Any pointers appreciated.

### DIFF
--- a/simulator/src/lib.rs
+++ b/simulator/src/lib.rs
@@ -143,18 +143,8 @@ where
 
         let Point { x, y } = center;
 
-        let fill_color: Option<Rgb888> = if let Some(color) = style.fill_color {
-            let color: Rgb888 = self.theme.convert(color.into());
-            Some(color)
-        } else {
-            None
-        };
-        let stroke_color: Option<Rgb888> = if let Some(color) = style.stroke_color {
-            let color: Rgb888 = self.theme.convert(color.into());
-            Some(color)
-        } else {
-            None
-        };
+        let fill_color = style.fill_color.map(|c| self.theme.convert(c.into()));
+        let stroke_color = style.stroke_color.map(|c| self.theme.convert(c.into()));
 
         self.window
             .draw_circle(*x, *y, *radius, fill_color, stroke_color);
@@ -166,18 +156,8 @@ where
         let Point { x: x1, y: y1 } = start;
         let Point { x: x2, y: y2 } = end;
 
-        let fill_color: Option<Rgb888> = if let Some(color) = style.fill_color {
-            let color: Rgb888 = self.theme.convert(color.into());
-            Some(color)
-        } else {
-            None
-        };
-        let stroke_color: Option<Rgb888> = if let Some(color) = style.stroke_color {
-            let color: Rgb888 = self.theme.convert(color.into());
-            Some(color)
-        } else {
-            None
-        };
+        let fill_color = style.fill_color.map(|c| self.theme.convert(c.into()));
+        let stroke_color = style.stroke_color.map(|c| self.theme.convert(c.into()));
 
         self.window
             .draw_line(*x1, *y1, *x2, *y2, fill_color, stroke_color);
@@ -197,18 +177,8 @@ where
         let Point { x: x1, y: y1 } = top_left;
         let Point { x: x2, y: y2 } = bottom_right;
 
-        let fill_color: Option<Rgb888> = if let Some(color) = style.fill_color {
-            let color: Rgb888 = self.theme.convert(color.into());
-            Some(color)
-        } else {
-            None
-        };
-        let stroke_color: Option<Rgb888> = if let Some(color) = style.stroke_color {
-            let color: Rgb888 = self.theme.convert(color.into());
-            Some(color)
-        } else {
-            None
-        };
+        let fill_color = style.fill_color.map(|c| self.theme.convert(c.into()));
+        let stroke_color = style.stroke_color.map(|c| self.theme.convert(c.into()));
 
         self.window
             .draw_rectangle(*x1, *y1, *x2, *y2, fill_color, stroke_color);
@@ -266,17 +236,8 @@ where
 
         let Point { x, y } = center;
 
-        //really?
-        let fill_color: Option<Rgb888> = if let Some(color) = style.fill_color {
-            Some(color.into())
-        } else {
-            None
-        };
-        let stroke_color: Option<Rgb888> = if let Some(color) = style.stroke_color {
-            Some(color.into())
-        } else {
-            None
-        };
+        let fill_color = style.fill_color.map(|c| c.into());
+        let stroke_color = style.stroke_color.map(|c| c.into());
 
         self.window
             .draw_circle(*x, *y, *radius, fill_color, stroke_color);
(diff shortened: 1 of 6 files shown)


rfuest:
Accelerating the simulator is certainly appreciated but I'm not sure if the HW acceleration API is the right way to do it. Because the simulator is often used to test e-g's drawing routines it shouldn't use an external drawing API by default IMO.

But as an example and test of the HW acceleation API an optional feature that enables the acceleration using sdl-gfx would be an useful feature.

I did look into the slow drawing myself a while ago and think that the SDL `fill_rect` method is really slow. If you are interested in fixing the root cause of the problem you might want to investigate changing the way `RgbDisplay::run_once` and/or `Window::draw_pixel` are implemented. Maybe building a list of rects and drawing them at once using `Canvas::fill_rects` is faster.

jacobrosenthal:
I think I agree, I traced it to fill_rect and decided to try to use
fill_rects to batch transactions I found that youd have to some how sort by
color because you send color separate from fill_rect,  I realized by the
time you've atomized your work into 384000 pixels and 384000 fil_rect  with
individual colors, you cant do any speedups anymore.

Then as you presumably know, nothing other than rect and fill_rects is
available outside of the gfx library. Now that we have the extended
primitives api it seems like you might be able to chop up all other
primitives into a rectangle plus a much smaller amount of pixels, which
only occasionally might draw miserably. I dont have any history in
graphics, perhaps this is a solved problem? For my uses Im not sure how
interesting that is, I think Id rather have the feature gated gfx.

I can look to feature gating and leave space if anyone else can help add
those pixelizers.

jacobrosenthal:
Naive googling netted the keywords "minimum dissection into rectangles of a rectilinear polygon" and some links:
* https://stackoverflow.com/questions/5919298/algorithm-for-finding-the-fewest-rectangles-to-cover-a-set-of-rectangles-without
* https://arxiv.org/pdf/0908.3916.pdf
* https://link.springer.com/content/pdf/10.1007%2FBF02189307.pdf
* https://github.com/mikolalysenko/rectangle-decomposition

rfuest:
You are right using `fill_rects` would be hard for color displays, I only thought about `BinaryDisplay`. But there has to be a method to accelerate this without complex calculations.

If I disable hardware acceleration in window.rs by changing line 85 to: `let canvas = window.into_canvas().software().build().unwrap();` I get a ~15% performance improvement on my system. So maybe HW accelerated drawing isn't working as expected in this use case. Perhaps we could draw to a texture directly and then only copy the result to the window canvas.

jacobrosenthal:
Yeah seems like I can get rid of the opengl() call i just cargo culted it
thinking I might need it for circles, etc
Oh interesting, .software() also fixes my problem of ghosting of previous
images too. Wonder why.

simulator/src/lib.rs
```diff
@@ -178,50 +225,113 @@ impl DrawTarget<BinaryColor> for BinaryDisplay {
...
+            style,
+        } = item;
+
+        let Point { x, y } = center;
+
+        //really?
+        let fill_color: Option<Rgb888> = if let Some(color) = style.fill_color {
+            Some(color.into())
+        } else {
+            None
```
rfuest:
```rust
let fill_color = color.map(|c| c.into());
```
should work

jacobrosenthal:
Id love any more comments, is there still an inclination this cant use the gfx library?

@ReeceStevens what do you think, didnt you notice the slowness of the previous input handling? or is it just my mac laptops where it was second+ between input and circle movement.

My remaining todos that I would take any help or feedback on are
* adding the gfx feature to the example doesnt seem to work? Can anyone spot why?
* I like the cleanup of returning events iterator instead of a vec, but then theres lifetime problems from display.get_input_events and move_circle taking display
* draw_triangle not implemented, not sure how to map that to sdl

ReeceStevens:
Hi @jacobrosenthal -- I definitely did notice slowness in the input handling example, especially with quick repetition of keystrokes. My particular use case for the simulator was to emulate a touch screen, so I haven't needed anything more responsive than basic click handling.

To be clear, clicking at a new location moved the circle at sub-second speeds, there was an observable delay but it was miniscule at best. I noticed significant lag when I held down arrow keys.

RE not having an implementation for `draw_triangle`: I'm not sure I fully understand the context here, but I just wanted to throw out there that adding HW accelerated methods isn't an all-or-nothing proposition-- you can provide accelerated rectangles and lines but fall back to iterator-based drawing for circles and triangles by not providing an alternative implementation for `draw_circle` and `draw_triangle`.

rfuest:
> Id love any more comments, is there still an inclination this cant use the gfx library?

Sorry for the delayed replay.

IMO the gfx library approach has some problems, as stated above. But that's only my opinion and I think @jamwaffles  should comment on this.

> @ReeceStevens what do you think, didnt you notice the slowness of the previous input handling? or is it just my mac laptops where it was second+ between input and circle movement.

More than a second sounds really bad and needs to be improved. My tests showed similar performance as @ReeceStevens describes.

I did some test with using a streaming texture to draw and got promising results: Drawing was ~5x faster than before. I'll upload my branch in the next few day so you can test it on your machine.

> * adding the gfx feature to the example doesnt seem to work? Can anyone spot why?

This is probably caused by the layout of the repository as a workspace. Did you try to run `cargo run --features gfx --example input-handling` from the `simulator` directory and not the root directory?
 
> * I like the cleanup of returning events iterator instead of a vec, but then theres lifetime problems from display.get_input_events and move_circle taking display

You could just set the new position variable inside the loop over the input events and call `move_circle` after the loop:
```rust
let mut new_positon = None;
for event in display.get_input_events() {
   ...
   new_postion = Some(Point::new(...));
   ...
}
if let Some(p) = new_position {
   move_circle(&mut display, position, new_position);
   position = new_position;
}
```
> * draw_triangle not implemented, not sure how to map that to sdl

I haven't used the gfx libary before, but shouldn't that map to the `trigon` functions?

> * is it ok to truncate x and y to u16

This won't be a problem in most cases but I would prefer to have some kind of check for this conversion.

jamwaffles:
Thanks for the PR! Apologies for the slow reply, other stuff has got in the way. I'll try and answer some bits of the thread above.

> Id love any more comments, is there still an inclination this cant use the gfx library?

I think the simulator should remain pixel based. For example, I'm working on thick lines support at the moment. If the simulator just uses an accelerated line function, I can't debug my code. Using `gfx` also feels like a bandaid for the underlying problem of a slow simulator. Is there something else that could be done to speed up the simulator itself? I use Linux and WSL2 (with XMing) and haven't had slowdowns that make the simulator unusable, but it's clear you and others have so we need to fix that. Is macOS the common factor here?

Could we use a different approach like drawing into a texture and scaling that appropriately? Or something based off @rfuest's suggestion above:

> I did look into the slow drawing myself a while ago and think that the SDL fill_rect method is really slow. If you are interested in fixing the root cause of the problem you might want to investigate changing the way RgbDisplay::run_once and/or Window::draw_pixel are implemented. Maybe building a list of rects and drawing them at once using Canvas::fill_rects is faster.

jacobrosenthal:
Look forward to your work @rfuest 
Beyond that I think the optimization to be made is to not flatten to pixels and store that, requiring an 800x400 draw each update, but to instead store visible and hidden primitives and only draw them.

jamwaffles:
There was also an interesting approach in jamwaffles/ssd1306#94 where the bounding box of any updated pixels is stored, so only those pixels are sent to the display. Could be adapted for the simulator? It doesn't help full-screen updates, but if only a few things are being drawn an efficient clear + iterator over only changed pixels could be a decent interim solution. 

That said, it'd be best to find a solution to the general case.

rfuest:
I've opened PR #192 to discuss an alternative approach by using a texture.

PR #193, opened 2019-10-30
Events as iterator instead of vec, move quit to an event instead of bool

jacobrosenthal:
Perhaps moving the quit bool into an event survives from #190?

## checklist
- [x] Add a simulator example(s) where applicable *Not applicable* 
- [x] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) *Not applicable* 

### DIFF
--- a/simulator/src/window.rs
+++ b/simulator/src/window.rs
@@ -60,6 +60,7 @@ pub struct Window {
     scale: usize,
     pixel_spacing: usize,
     theme: BinaryColorTheme,
+
     canvas: render::Canvas<sdl2::video::Window>,
     event_pump: sdl2::EventPump,
 }
--- a/simulator/src/window.rs
+++ b/simulator/src/window.rs
@@ -51,6 +51,8 @@ pub enum SimulatorEvent {
         /// The directionality of the scroll (normal or flipped)
         direction: MouseWheelDirection,
     },
+    /// An exit event
+    Quit,
 }
 
 /// Simulator window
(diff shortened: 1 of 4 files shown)


jacobrosenthal:
Simulator hasnt been released though, so theres no change to log

jacobrosenthal:
ping @jamwaffles or anyone else, any thoughts on this?

simulator/src/window.rs
```diff
@@ -51,17 +51,17 @@ pub enum SimulatorEvent {
...
+    /// An exit event
+    Quit,
 }
 
 /// Simulator window
 pub struct Window {
     scale: usize,
     pixel_spacing: usize,
     theme: BinaryColorTheme,
-
```
rfuest: This empty line was intentional, to separate the SDL fields from the other fields.

simulator/src/window.rs
```diff
@@ -154,39 +153,41 @@ impl Window {
...
     }
 
     /// Handle events
-    pub fn handle_events(&mut self) -> bool {
-        for event in self.event_pump.poll_iter() {
-            match event {
+    /// Return an iterator of all captured SimulatorEvent
+    pub fn events(&mut self) -> impl Iterator<Item = SimulatorEvent> + '_ {
+        let scale = self.scale;
+        let pixel_spacing = self.pixel_spacing;
```
rfuest: Could you move the `map_input_to_point` function into `impl Window` and remove the `scale` and `pixel_spacing` arguments instead?

jacobrosenthal: Can move map_input_to_point easily enough, but I need copies to get around accessing self.scale inside of self.event_pump

rfuest: Ok, I didn't noticed that.

jamwaffles:
Thanks!

Issue #162, opened 2019-09-01
Factor styling out of e-g objects


jamwaffles:
Carrying on with #139, this issue is to discuss refactoring how styling works to decouple styles from primitives. This ties in with #156 in allowing e-g objects to be useful for more than just drawing pixels.

So far, there's been a suggestion of something like the following (from https://github.com/jamwaffles/embedded-graphics/issues/139#issuecomment-515121637):

```rust
struct Styled<T, C> {
   primitive: T,
   style: Style<C>,
}
```

`IntoIterator` could then be implemented for e.g. `Styled<Rectangle>` instead of `Rectangle`. `Rectangle` itself would be unstyled by default.

The current issue with this approach as mentioned in https://github.com/jamwaffles/embedded-graphics/issues/139#issuecomment-525946121 is that it breaks chaining. I.e., this is no longer possible:

```rust
let objects = Circle::new(icoord!(64, 64), 64)
    .into_iter()
    .chain(Line::new(icoord!(64, 64), icoord!(0, 64))))
    .chain(Line::new(icoord!(64, 64), icoord!(80, 80))))
    .chain(
        Font6x8::render_str("Hello World!")
            .translate(icoord!(5, 50)),
    );

let styled = Styled::new(objects).stroke(Some(C1)).fill(Some(C0));
```

One possible solution to this is a "context" API as explored in #149.

jamwaffles:
How would style chaining work now? One option is `Styled::new(...)` but being able to chain a style right onto a primitive would be a great feature to retain.

I've [written up](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c278173b591e5486f262fe3bf5c9633c) a quick idea of how I think it could work. The crux of it is this:

Have a default impl for `WithStyle`:

```rust
trait WithStyle<C: PixelColor>: Sized {
    fn fill(self, fill: Option<C>) -> Styled<Self, C> {
        Styled {
            item: self,
            style: Style {
                fill,
                ..Style::default()
            }
        }
    }

    /* ... etc ... */
}
```

Then, implement `Styled`:

```rust
struct Styled<I, C> {
    item: I,
    style: Style<C>
}

impl<I: Drawable, C: PixelColor> Styled<I, C>  {
    fn fill(self, fill: Option<C>) -> Self {
        Self {
            item: self.item,
            style: Style {
                fill,
                ..self.style
            }
        }
    }

    /* ... etc ... */
}
```

[...]

Issue #156, opened 2019-08-21
Additional uses for primitive shapes


rfuest:
From #141:

> I'd like to leave `Rect` for a different PR. I'm envisaging a clipping/masking API that can accept various different clip shapes, not just rectangles. e-g primitives could be reused for this purpose. For example, to get a circular mask:
> 
> ```rust
> // Naming/import structure to be decided!
> 
> use embedded_graphics::{coord, mask::Mask, primitives::Circle};
> 
> let mask = Mask::new(Circle::new(coord!(10, 20), 10));
> 
> let point = Point::new(15, 25);
> 
> mask.contains(&point);
> ```
> 
> Let's open a new issue if you'd like to discuss this further.

Being able to use the primitive shapes for more uses would be a great feature. As mentioned in #139  and https://github.com/jamwaffles/embedded-graphics/issues/141#issuecomment-517929501 I would suggest to remove the styling from the primitives and have structs that only represent geometric shapes.

jamwaffles:
Cc #202 - there seems to be some real-world desire for things like hit testing which is good validation of this issue.

rfuest:
Related comment from #162:

Having a way to iterate over the coordinates of the points inside a shape would be useful. This would make it pretty easy to implement more advanced iterators:

```rust
// a filled red rectangle
let filled_rect = rect.points().map(|p| Pixel(p, Rgb565:::Red));
filled_rect.draw(&mut display);

// inset is a method that shrinks a Rectangle that by n pixels in each direction
let inside_area = rect.inset(1);

// a rectangle with a 1px green stroke
let stroked_rect = rect.points().filter(|p| !inside_area.contains(p)).map(|p| Pixel(p, Rgb565:::Green));
stroked_rect.draw(&mut display);
```

jacobrosenthal:
Oh II think related to #202 

I think Id like to be able to optionally iterate over the primitives, not the points of chained primitives.

Im thinking like a snake game or something. I want a fast way to see if a new point or shape is contained in the existing chain, and doing so on a point by point basis would be too onerous, but I think checking only bounding boxes would be possible.

rfuest:
> I think Id like to be able to optionally iterate over the primitives, not the points of chained primitives.

I'm not quite sure I understand your comment. Chaining is only used for drawing in e-g and there is no way to chain primitives before they are converted into an pixel iterator. Could you provide a small code example of what you would expect to be able to do?

jacobrosenthal:
[...]

PR #198, opened 2019-11-06
Factor styling out of e-g objects

rfuest:

## PR description

This PR implements some of the changes we discussed in #162. Iterators over non styled primitives and the changes to the `Transform` trait aren't implemented yet. But it should be enough to get a better feel for the API.


jamwaffles:
Looking at some of the examples, `Primitive::new().into_styled().translate()` feels a bit weird to me - translations aren't part of styling. Does this also mean `Primitive::new().translate()` wont't work? I think that should work instead of `.into_styled().translate()` as other uses for primitives _a la_ #156 would require extra work.

Should we allow this pattern?:

```rust
Primitive::new()
    .styled(
        Style::new().stroke_color(Some(1u8.into())
    )
```

`Style::new()` returns `Style::default()` which can be modified with a chain on the style itself. I'm not sure if this would actually make the API cleaner, but it would mean that style stuff is grouped nicely into the `Style` struct, and the same style can be copied/`clone()`d into various instances of primitives.

`.styled()` returns the same thing as `.into_styled()`, but allows passing of a pre-built style struct. Might be handy if a user is drawing lots of things with the same style (list menus come to mind as an example).

I'll do a more thorough review later when I have a bit of time.

rfuest:
> Looking at some of the examples, `Primitive::new().into_styled().translate()` feels a bit weird to me - translations aren't part of styling. Does this also mean `Primitive::new().translate()` wont't work? I think that should work instead of `.into_styled().translate()` as other uses for primitives _a la_ #156 would require extra work.

No both should work, as you suggested in: https://github.com/jamwaffles/embedded-graphics/issues/162#issuecomment-527360982. `Transform` is implemented for the primtives and `Styled<T, C> where T: Transform`.

> Should we allow this pattern?:
> 
> ```rust
> Primitive::new()
>     .styled(
>         Style::new().stroke_color(Some(1u8.into())
>     )
> ```

Yes, this was something I also considered adding.

Should we also use `Styled` for fonts? I would prefer to use another struct for styling text, because there was some confusion about what `fill_color` and `stroke_color` means for font rendering. `foreground_color` (or `font_color`, `text_color`) and `background_color` would be easier to understand IMO.

rfuest:
The latest version adds an style parameter to `into_styled` instead of adding a separate `styled` method. This works fine for styling single objects or sharing styles between multiple objects.

Code like this:
```rust
Circle::new(position, 100)
    .into_styled()
    .fill_color(FOREGROUND_COLOR)
    .draw(&mut display);
```
can now be written as:
```rust
Circle::new(position, 100)
    .into_styled(Style::fill(FOREGROUND_COLOR))
    .draw(&mut display);
``` 

Using styles that set a stroke and fill currently requires to build a style first and use it later:
https://github.com/rfuest/embedded-graphics/blob/2db8a5aef94f6018425a6a81fb1809648f440d5b/simulator/examples/analog-clock.rs#L83-L89
But this could be simplified by adding a `StyleBuilder` to keep a builder like API:
```rust
let decoration = Circle::new(decoration_position, 5)
    .into_styled(
        StyleBuilder::new()
        .fill(BinaryColor::Off)
        .stroke(BinaryColor::On, 1)
        .build()
    );
```

embedded-graphics/src/primitives/mod.rs
```diff
@@ -1,19 +1,32 @@
...
-
 pub use self::circle::Circle;
 pub use self::line::Line;
 pub use self::rectangle::Rectangle;
+pub use self::styled::Styled;
 pub use self::triangle::Triangle;
+use crate::style::Style;
+
+/// Primitive trait
+pub trait Primitive: Dimensions {
```
jamwaffles: Can we call this `Stylable` instead? Or at least something as generic. I'm imagining something like a struct in part of a UI library that draws a complex widget (i.e. it's not primitive) which breaks the `Primitive` name.

rfuest: > > Should we also use Styled for fonts? I would prefer to use another struct for styling text, because there was some confusion about what fill_color and stroke_color means for font rendering.
> 
> Yeah that makes sense. Would the fonts also have the same `.into_styled()` method, but which takes a `FontStyle` instead of a `Style`? `.into_styled()` would have to take a generic arg in that case, right?

I haven't really thought about how font styling could be implemented. But I was currently planning to use `Style` and `Styled` exclusively for primitives.
 
> > But this could be simplified by adding a `StyleBuilder` to keep a builder like API:
> Not sure why there's this limitation tbh. Can we remove `WithStyle` now that the only way of styling things is through the `.into_styled()` method on `Primitive`? If so, why not move all the methods on `WithStyle` into `Style`?

Keeping `WithStyle` was temporary to keep font rendering intact.
Having the methods from `WithStyle` implemented on `Style` seemed unnecessary because the fields are public anyway and there is no need for additional setter methods IMO.

> Or, if mutating is preferred, stick a `.clone()` in:
> 
> ```rust
> // ...
> 
> let other_thing = Circle::new(decoration_position, 5)
>     .into_styled(decoration_style.clone().stroke_width(10));
> ```

This only works if `stroke_width` (and the other setters) would take ownership of the `Style` object:
```rust
fn stroke_width(self, width: u32) -> Self { ... ; self }
```
But this wouldn't be as flexible as using a mutable reference for `self`. I think that was the reason I suggested to introduce the `StyleBuilder` instead.

> As a general API convention, I'm starting to think if e-g should mutate by default and require `.clone()` for non-mutations seeing as `.clone()` is going to have the same cost as a copy for all e-g objects (at time of writing). I think you've mentioned this in the past but looking at all the `translate`/`translate_mut` stuff in the diff seems a bit superfluous now.
> 
> Edit: ah yes, in the linked issue no less [#162 (comment)](https://github.com/jamwaffles/embedded-graphics/issues/162#issuecomment-533886275) roll_eyes

At first I also wanted to change the `Transform` implementation in this PR, but there were some problems with this change and I decided to postpone this changes to discuss this further in another issue or PR.

rfuest:
> > Should we also use Styled for fonts? I would prefer to use another struct for styling text, because there was some confusion about what fill_color and stroke_color means for font rendering.
> 
> Yeah that makes sense. Would the fonts also have the same `.into_styled()` method, but which takes a `FontStyle` instead of a `Style`? `.into_styled()` would have to take a generic arg in that case, right?

I haven't really thought about how font styling could be implemented. But I was currently planning to use `Style` and `Styled` exclusively for primitives.
 
> > But this could be simplified by adding a `StyleBuilder` to keep a builder like API:
> Not sure why there's this limitation tbh. Can we remove `WithStyle` now that the only way of styling things is through the `.into_styled()` method on `Primitive`? If so, why not move all the methods on `WithStyle` into `Style`?

Keeping `WithStyle` was temporary to keep font rendering intact.
Having the methods from `WithStyle` implemented on `Style` seemed unnecessary because the fields are public anyway and there is no need for additional setter methods IMO.

> Or, if mutating is preferred, stick a `.clone()` in:
> 
> ```rust
> // ...
> 
> let other_thing = Circle::new(decoration_position, 5)
>     .into_styled(decoration_style.clone().stroke_width(10));
> ```

This only works if `stroke_width` (and the other setters) would take ownership of the `Style` object:
```rust
fn stroke_width(self, width: u32) -> Self { ... ; self }
```
But this wouldn't be as flexible as using a mutable reference for `self`. I think that was the reason I suggested to introduce the `StyleBuilder` instead.

> As a general API convention, I'm starting to think if e-g should mutate by default and require `.clone()` for non-mutations seeing as `.clone()` is going to have the same cost as a copy for all e-g objects (at time of writing). I think you've mentioned this in the past but looking at all the `translate`/`translate_mut` stuff in the diff seems a bit superfluous now.
> 
> Edit: ah yes, in the linked issue no less [#162 (comment)](https://github.com/jamwaffles/embedded-graphics/issues/162#issuecomment-533886275) roll_eyes

At first I also wanted to change the `Transform` implementation in this PR, but there were some problems with this change and I decided to postpone this changes to discuss this further in another issue or PR.

embedded-graphics/src/primitives/mod.rs
rfuest: Do you think that an UI library is likely to reuse the `Style` struct for styling complex widgets? Most widgets need more styling properties then provided by `Style` and, similar to font styling, would benefit from more specific names.

I was also thinking about changing the names and came up with this:
- Change `primitives` to `shapes`: This might be a better name if the objects are also used for other purposes (#156).
- If `Styled` is only used to apply styling to shapes it could then be called `Primitve`.

jamwaffles: Yeah my bad, now that I'm a bit more awake the properties on the `Style` struct only really make sense for e-g primitives. 

`shapes`/`Shape` is easier to type but I can't think of a better reason to change the name unless you want to. We're still missing a couple of items like ellipses and polygons/polylines, and `Primitive` and `Shape` work equally well for those cases so up to you if you want to change the name :slightly_smiling_face:

jamwaffles:
> I was currently planning to use Style and Styled exclusively for primitives.

Surely `Styled` could be reused? `Style` is definitely specific to the primitives but `Styled` could be more general, no?

> Having the methods from WithStyle implemented on Style seemed unnecessary because the fields are public anyway and there is no need for additional setter methods IMO.

Ahh that's true, I don't think you explicitly mentioned going in that direction. Makes for a lighter API for sure. I don't think there's need for a `StyleBuilder` if we're just going with struct access.

> At first I also wanted to change the `Transform` implementation in this PR, but there were some problems with this change and I decided to postpone this changes to discuss this further in another issue or PR.

Sounds good, my original comment on this was just a spitball.

rfuest:
> > I was currently planning to use Style and Styled exclusively for primitives.
> 
> Surely `Styled` could be reused? `Style` is definitely specific to the primitives but `Styled` could be more general, no?

Should we rename `Style` to `PrimitiveStyle` to distinguish it from `FontStyle`?

I guess `Styled` can be reused. The main reason to keep it separate was to have a shorter type. `Styled<Circle, PrimitiveStyle<BinaryColor>>` is pretty long. But it will be inferred by the compiler most of the time and doesn't need to be specified, so this might not be a problem.

> > Having the methods from WithStyle implemented on Style seemed unnecessary because the fields are public anyway and there is no need for additional setter methods IMO.
> 
> Ahh that's true, I don't think you explicitly mentioned going in that direction. Makes for a lighter API for sure. I don't think there's need for a `StyleBuilder` if we're just going with struct access.

OK. `StyleBuilder` could also be added later if necessary.

embedded-graphics/src/primitives/mod.rs
rfuest: If I don't come up with a better reason for this change I'll keep the current names.

rfuest:
The primitive refactoring is ready for review. Text styling could use some more work, but this was outside the scope of this PR.

@jamwaffles Which of the new types should be included in the prelude? `Style` was included before, but I'm not sure if it should be.

CHANGELOG.md
```diff
@@ -51,6 +51,12 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 - **(breaking)** The type of `Style::stroke_width` changed from `u8` to `u32`.
 
+- **(breaking)** Primitives shapes need to be converted into `Styled`s to be drawn.
+
+- **(breaking)** The primitive macros like `egline` don't accept a `style` setting anymore. Use `object.style = new_style` instead.
```
jamwaffles: Just to clarify - they still accept individual style properties though, right?

jamwaffles:
One thing stands out to me in a real quick drive-by review:

1. I'm really not keen on patterns like

  ```rust
  let mut s = PrimitiveStyle::default();
  s.stroke_color = Some(BinaryColor::Off);
  s.fill_color = Some(BinaryColor::On);
  ```

  Could we bring back the `WithStyle`-like builder interface directly on `PrimitiveStyle` (and `TextStyle` too I think) and use this instead:

  ```rust
  let s = PrimitiveStyle::new()
    .stroke_color(Some(BinaryColor::Off))
    .fill_color(Some(BinaryColor::On));
  ```

  This gets rid of `mut` and feels a bit more Rust-like IMO. Also a couple fewer characters to type. The signatures for these fns could look like:

  ```rust
  fn stroke_color(&mut self, stroke_color: Option<C>) -> &mut Self {
      self.stroke_color = stroke_color;

      self
  }
  ```

  I don't think it's worth separating those methods into a `StyleBuilder`. IMO it adds unnecessary noise to the API - I think it's better to have the methods under the `PrimitiveStyle` or `TextStyle` structs.

---

> Text styling could use some more work, but this was outside the scope of this PR.

What kind of things were you thinking? Would be good to note those down somewhere. Is it worth opening an issue?

> Which of the new types should be included in the prelude?

Is there anything that the user wouldn't type out that's required? Any new traits and the like? If not then yeah let's keep the new stuff out of the prelude. Can always add it later!

Quick question: I noticed a lot of changes like this:

```patch
- let object: Triangle<Gray8> =
-     Triangle::new(Point::new(5, 10), Point::new(15, 20), Point::new(5, 20));
+ let object = &Triangle::new(Point::new(5, 10), Point::new(15, 20), Point::new(5, 20))
+     .into_styled(PrimitiveStyle::stroke(Gray8::new(10), 1));
```

Is the reference required, or is it just some cleanup?

CHANGELOG.md
rfuest: Yes, only replacing the complete style object doesn't work anymore.

rfuest:
> ```rust
> let s = PrimitiveStyle::new()
>  .stroke_color(Some(BinaryColor::Off))
>  .fill_color(Some(BinaryColor::On));
> ```

If the methods in `WithStyle` are implemented for mutable references this code would result in `s` being a reference. This was the reason to try the `StyleBuilder`.

> This gets rid of mut and feels a bit more Rust-like IMO.

Is chaining setter methods Rust-like? Do you have any other examples of this being used (except for builders)?

> > Text styling could use some more work, but this was outside the scope of this PR.
> 
> What kind of things were you thinking? Would be good to note those down somewhere. Is it worth opening an issue?

I'm not quite sure yet and need to look at the text rendering code in more detail first.
But adding a `Text` struct could help the distinguish between fonts and text drawables. This way `Stylable` could also be used for text rendering.
```rust
struct Text<'a>{
   text: &'a str,
   position: Point,
};
```

> Quick question: I noticed a lot of changes like this:
> 
> ```diff
> - let object: Triangle<Gray8> =
> -     Triangle::new(Point::new(5, 10), Point::new(15, 20), Point::new(5, 20));
> + let object = &Triangle::new(Point::new(5, 10), Point::new(15, 20), Point::new(5, 20))
> +     .into_styled(PrimitiveStyle::stroke(Gray8::new(10), 1));
> ```
> 
> Is the reference required, or is it just some cleanup?

Ah, that was something I forgot to discuss before.
`IntoIterator` is only implemented for `&Styled<...>` anymore, it was previously implemented for e.g. `&Rectangle` and `Rectangle`. Is this Rust-like? For example `IntoIterator` isn't implemented for arrays but only for slices.

rfuest:
> > Which of the new types should be included in the prelude?
> 
> Is there anything that the user wouldn't type out that's required? Any new traits and the like? If not then yeah let's keep the new stuff out of the prelude. Can always add it later!

The `Primitive` trait provides the `into_styled` method now, so it should be in the prelude IMO. But I've removed all `crate::style::*` types from the prelude.

jamwaffles:
> If the methods in WithStyle are implemented for mutable references this code would result in s being a reference. 

Sorry yeah not `&mut` - good point. Was thinking this instead (with just `mut`):

```rust
pub fn fill_color(mut self, fill_color: Option<C>) -> Self {
    self.fill_color = fill_color;

    self
}
```

I tried this quickly in `examples/analog-clock.rs` like this:

```rust
let decoration_style = PrimitiveStyle::default()
    .stroke_color(Some(BinaryColor::On))
    .fill_color(Some(BinaryColor::Off));
```

which seems to work ok. Am I missing other patterns?

> Is chaining setter methods Rust-like? Do you have any other examples of this being used (except for builders)?

Not setters, no. I feel like `PrimitiveStyle` is a builder, or at least presents a builder-like interface which is why I said it's Rust-like. I see how they could also look like setter methods as well, however I think in a lot of use cases it will be used as a builder, with multiple chained methods to set a style up. This is why I think having a builder-like interface on `PrimitiveStyle` is preferable.

> Ah, that was something I forgot to discuss before.
> IntoIterator is only implemented for `&Styled<...>` anymore, it was previously implemented for e.g. `&Rectangle` and `Rectangle`. Is this Rust-like? For example IntoIterator isn't implemented for arrays but only for slices.

Ah right I see. I'm not sure what the right Rust idiom is here tbh. Whichever is less difficult to use I guess? Let's leave it just for `&Styled` and fill in any gaps later.

> The Primitive trait provides the into_styled method now, so it should be in the prelude IMO.

I agree :)

rfuest:
> > If the methods in WithStyle are implemented for mutable references this code would result in s being a reference.
> 
> Sorry yeah not `&mut` - good point. Was thinking this instead (with just `mut`):
> 
> ```rust
> pub fn fill_color(mut self, fill_color: Option<C>) -> Self {
>     self.fill_color = fill_color;
> 
>     self
> }
> ```
> 
> I tried this quickly in `examples/analog-clock.rs` like this:
> 
> ```rust
> let decoration_style = PrimitiveStyle::default()
>     .stroke_color(Some(BinaryColor::On))
>     .fill_color(Some(BinaryColor::Off));
> ```
> 
> which seems to work ok. Am I missing other patterns?

This works, but you can't use these methods anymore if you only have access to a mutable reference. I'm not sure how often this would be a problem.
 
> > Is chaining setter methods Rust-like? Do you have any other examples of this being used (except for builders)?
> 
> Not setters, no. I feel like `PrimitiveStyle` is a builder, or at least presents a builder-like interface which is why I said it's Rust-like. I see how they could also look like setter methods as well, however I think in a lot of use cases it will be used as a builder, with multiple chained methods to set a style up. This is why I think having a builder-like interface on `PrimitiveStyle` is preferable.

I'm not a fan of the almost a builder pattern. I agree that a builder pattern is probably the nicest way of constructing a style. But having one struct that is used as a builder and as an object that represents a style leads to some slightly confusing naming IMO. For a builder I would expect to be able to set the fill color using `style.fill_color(color)`, but for any other object `style.fill_color()` would be a getter. (see https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-[rfc-344])

jamwaffles:
> This works, but you can't use these methods anymore if you only have access to a mutable reference. I'm not sure how often this would be a problem.

Just [refreshed my memory](https://www.ameyalokare.com/rust/2017/11/02/rust-builder-pattern.html) and I see your point there. Ok, let's leave `Style` as is for this PR and think about builders later. Good point on the getter/setter thing as well.

I'm wondering if it's a good idea to recommend this method of creating styles in the examples/docs:

```rust
let style = Style {
    stroke_color: Some(BinaryColor::On),
    ..Style::default()
}
```

That feels better than creating a bunch of `let mut`s. Seeing as this is code people will probably copypasta quite a bit, I think it's reasonably important to set a clean precedent. This can be replaced with a builder when we figure that out. The other option is to just leave the `let mut` pattern in for now. Which would you prefer?

rfuest:
Yes, that's a better way to build these styles, I'll change that.

`PrimitiveStyle` has two constructors `PrimtiveStyle::stroke` and `PrimitiveStyle::fill`. The names were chosen to be short, but they don't follow the naming convention. Should they be renamed to `PrimitiveStyle::with_stroke` and `PrimitiveStyle::with_style` like the API guidelines suggests? And if they are renamed, should `TextStyle::new` be renamed to `TextStyle::with_text_color` to match, even if there currently is only one constructor?

When a style is constructed using `PrimitiveStyle::stroke` the width must explicitly specified, but `PrimitiveStyle::default` sets a default stroke width of 1 px. I would prefer if both variants would behave similar by either assuming the 1px default width or require setting the stroke width explicitly:
```rust
// 1. current implementation

let _s1 = PrimitiveStyle::stroke(color, 1);
let _s1 = PrimitiveStyle {
    stroke_color: Some(color),
    //setting stroke_width is not necessary because default is 1
    ..PrimitiveStyle::default()
};
let _s10 = PrimitiveStyle::stroke(color, 10);
let _s10 = PrimitiveStyle {
    stroke_color: Some(color),
    stroke_width: 10,
    ..PrimitiveStyle::default()
};


// 2. width is specified explicitly

let _s1 = PrimitiveStyle::stroke(color, 1);
let _s1 = PrimitiveStyle {
    stroke_color: Some(color),
    stroke_width: 1,
    ..PrimitiveStyle::default()
};
let _s10 = PrimitiveStyle::stroke(color, 10);
let _s10 = PrimitiveStyle {
    stroke_color: Some(color),
    stroke_width: 10,
    ..PrimitiveStyle::default()
};


// 3. width has a 1px default value

let _s1 = PrimitiveStyle::stroke(color);
let _s1 = PrimitiveStyle {
    stroke_color: Some(color),
    ..PrimitiveStyle::default()
};
let mut _s10 = PrimitiveStyle::stroke(color);
_s10.stroke_width = 10;
let _s10 = PrimitiveStyle {
    stroke_color: Some(color),
    stroke_width: 10,
    ..PrimitiveStyle::default()
};
```
I would prefer variant 2.

jamwaffles:
> I would prefer variant 2.

Looks good. Does `PrimitiveStyle::default()` then give a stroke width of `0` as the default then?

> Should they be renamed to PrimitiveStyle::with_stroke and PrimitiveStyle::with_style like the API guidelines suggests?

Makes sense to me :slightly_smiling_face: 

> And if they are renamed, should TextStyle::new be renamed to TextStyle::with_text_color to match, even if there currently is only one constructor?

Yeah I think so. It keeps consistency with `PrimitiveStyle` and I imagine there'll be `TextStyle::with_background_color` added at some point, right?

rfuest:
> Looks good. Does `PrimitiveStyle::default()` then give a stroke width of `0` as the default then?

Yes.

> Yeah I think so. It keeps consistency with `PrimitiveStyle` and I imagine there'll be `TextStyle::with_background_color` added at some point, right?

The `with_*` constructors were intended for common use cases and I don't think that rendering invisible transparent text with opaque background is very common. Or did you have a special use case for that constructor in mind?

But I would like to add additional `PrimitiveStyle::new` constructors which behave the same way as `PrimitiveStyle::default`. They can be `const fn` and are expected to exist by most users.

jamwaffles:
> Yes.

Cool. More breaking changes for the changelog. Worth adding some doc comments about this behaviour too.

> Or did you have a special use case for that constructor in mind?

Good point, no I don't :grin: 

> But I would like to add additional PrimitiveStyle::new constructors which behave the same way as PrimitiveStyle::default. They can be const fn and are expected to exist by most users.

ConstructorS plural? Can you implement them in a commit and show me what you mean?

rfuest:
> Cool. More breaking changes for the changelog. Worth adding some doc comments about this behaviour too.

It is a breaking change, but I think the 1px default width behavior was never documented.

> ConstructorS plural? Can you implement them in a commit and show me what you mean?

I meant `PrimitiveStyle::new` and `TextStyle::new`.

jamwaffles:
> It is a breaking change, but I think the 1px default width behavior was never documented.

True, but it's worth mentioning that now a shape with a stroke colour but no set width won't render any pixels.

> I meant PrimitiveStyle::new and TextStyle::new.

Ah I see :ok_hand:

embedded-graphics/src/lib.rs
```diff
@@ -124,8 +125,8 @@
 //!
 //! fn build_thing(text: &'static str) -> impl Iterator<Item = Pixel<Rgb565>> {
 //!     egrectangle!((0, 0), (40, 40)).into_iter()
-//!         .chain(egcircle!((20, 20), 8, fill_color = Some(Rgb565::RED)))
-//!         .chain(text_6x8!(text, fill_color = Some(Rgb565::GREEN)).translate(Point::new(20, 16)))
+//!         .chain(&egcircle!((20, 20), 8, fill_color = Some(Rgb565::RED)))
+//!         .chain(text_6x8!(text, text_color = Some(Rgb565::GREEN)).translate(Point::new(20, 16)))
```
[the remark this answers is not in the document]
jamwaffles: Another option would be to implement the same iter for both `&Thing` and `Thing`. Could use a macro to reduce duplication. Is it even useful to have both?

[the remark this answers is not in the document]
rfuest: > `egcircle` being a ref and `text_6x8` not being one is odd (and might lead to end user confusion). Not necessarily for this PR, but should we change the text iterators to work with refs so they're consistent with the primitive iters?

Yes this is confusing and wasn't intended to be final, but I tried to keep the changes in the font module to a minimum for this PR. `IntoIterator` is also implemented for `&FontBuilder` so adding a `&` in front of `text_6x8` would already work. I decided not to change this now because I wanted to see the impact of removing `IntoIterator` for `FontBuilder` when working on the font module next.

> Another option would be to implement the same iter for both `&Thing` and `Thing`. Could use a macro to reduce duplication. Is it even useful to have both?

We discussed this earlier in this PR and had decided to only implement `IntoIterator` for `&Thing` for now. Implementing `IntoIterator` for `&Thing` instead of `Thing` has the advantage that the `Thing` object isn't consumed and can be reused multiple times.

rfuest:
> It's a shame we can't make `PrimitiveStyle::with_stroke()`, etc, `const` yet. Can you open an issue to track when these can be `const`-ified when `#![feature(const_fn)]` is stabilised?

I noticed that at least the constructors for primitive can be `const` now, because they don't have a type parameter anymore.

embedded-graphics/src/lib.rs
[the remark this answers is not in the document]
jamwaffles: Ok, that all sounds good!

PR #201, opened 2019-11-13
Build docs in project root

jamwaffles:

- [x] Check that you've added passing tests and documentation (N/A)
- [x] Add a simulator example(s) where applicable (N/A)
- [x] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) (N/A)
- [x] Run `rustfmt` on the project (N/A)

## PR description

A small change introduced in #187 meant that only docs in the e-g project were built, breaking the linkchecker. Not sure why we've only seen it now but it might be a caching thing. This PR builds docs in the workspace root, making `linkchecker` work again.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,10 +4,12 @@ set -e
 
 cargo fmt --all -- --check
 cargo test --release
+
 pushd embedded-graphics
 cargo test --release --all-features
-cargo doc --all-features
 popd
+
+cargo doc --all-features
 cargo bench --no-run
 
 linkchecker target/doc/embedded_graphics/index.html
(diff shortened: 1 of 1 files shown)


rfuest:
Looking at the changes in #187 this change was introduced because this build had failed:
https://circleci.com/gh/jamwaffles/embedded-graphics/3258
This build failure was caused by a bug in that nightly build of cargo that is fixed in newer versions.

We should revert the entire change, otherwise `cargo test --release --all-features` is also only called for the `embedded-graphics` crate.

jamwaffles:
Thanks for catching that. I've now reverted `build.sh` to before #187 was merged.

PR #203, opened 2019-11-21
Pretty pictures

### LINKED ISSUE #194: Pretty pictures
As per https://twitter.com/gcouprie/status/1191332050185412609:

> please add pictures in the readme :)
projects that work in graphical space have a free advantage to make presentation pages more attractive

I'd like to do two things:

1. Add the graphic in the lib docs to the readme
2. Add some photos of driver crates in action to both the readme and lib docs to add a bit of snaz

jamwaffles:

## PR description

Closes #194 by adding screenshots to README. Also updates them with better examples and a real-world photo of an SSD1331. The docs/readme will only update when this branch is merged to master, so here's the new screenshot image:

simulator-demo

It's sized to work at 1:1 scale both in the Github README section (888px wide) and rustdoc (a bit wider). Crates.io uses a width of 600px so it'll look a bit fuzzy there, but will still be readable.


rfuest:
Could we replace the stroke example with something else? Because half of the lines don't have the correct width yet.

jamwaffles:
That's the first one I found that fits in that space but good catch. I've updated the image to this:

simulator-demo

PR #206, opened 2019-11-23
Use Styled for font rendering

rfuest:

## PR description

This PR adds a `Text` struct that is used, together with the recently added `Styled` struct, to make the text rendering behave similar to primitive rendering.

A `font` property was added to the `TextStyle` struct. This way fonts can be defined the same way as other style properties. To make this possible the font types are now unit-like structs.

The `text_...` macros aren't necessary anymore and are replaced by a single `egtext` macro. This scales much better IMO and can also be used for external fonts.

## TODO
- [x] `TextStyle` constructors
     `TextStyle::with_text_color` doesn't really make sense anymore if the first argument is the font. Any suggestions for better names? Or should this be replaced by a builder?
- [x] Update documentation
- [x] Add custom font example


embedded-graphics/src/fonts/mod.rs
```diff
@@ -107,200 +111,78 @@
...
+pub use font24x32::Font24x32;
+pub use font6x12::Font6x12;
+pub use font6x8::Font6x8;
+pub use font8x16::Font8x16;
 
-/// Internal macro used to implement `text_*` on fonts. Do not use directly!
-#[doc(hidden)]
+/// TODO: docs
+/// TODO: don't require font as first parameter
+/// TODO: position parameter
```
jamwaffles: Should `Text::new()` take a position argument, or rely on `Transform` for positioning? The primitives all take coordinates in their constructors so I understand keeping the consistency with them.

embedded-graphics/src/fonts/mod.rs
```diff
@@ -107,200 +111,78 @@
...
+    use crate::style::{Styled, TextStyle};
 
     #[test]
     fn font_macros() {
-        let _text: Font6x8<BinaryColor> = text_6x8!("Hello!");
-        let _text: Font6x12<BinaryColor> = text_6x12!("Hello!");
-        let _text: Font8x16<BinaryColor> = text_8x16!("Hello!");
-        let _text: Font12x16<BinaryColor> = text_12x16!("Hello!");
-        let _text: Font24x32<BinaryColor> = text_24x32!("Hello!");
+        let _text: Styled<Text<'_>, TextStyle<BinaryColor, Font6x8>> =
```
jamwaffles: Ouch, I hope end users don't have to type this out very often :sweat_smile:

embedded-graphics/src/prelude.rs
```diff
@@ -2,7 +2,7 @@
 
 pub use super::drawable::Drawable;
 pub use super::drawable::Pixel;
-pub use super::fonts::Font;
+//pub use super::fonts::Font;
```
jamwaffles: Will you add this back before you're done? We export stuff like `Primitive` so this makes sense to leave in, no?

rfuest: Yes, this was a temporary change while I was refactoring this code.

embedded-graphics/src/fonts/mod.rs
rfuest: Yes, the `Styled<...>` types for texts and also primitives are pretty long. If this turns out to be a problem we could add type aliases to shorten these types later:
```rust
type StyledText<'a, C, F> = Styled<Text<'a>, TextStyle<C, F>>;
let _text: StyledText<'_, BinaryColor, Font6x8> = egtext!("Hello!", font = Font6x8);
```

embedded-graphics/src/fonts/mod.rs
rfuest: I think `Text::new()` and the `egtext` macro should take the a position parameter, because setting the position is necessary for all texts. Another benefit of setting the position in `Text::new()` is that it is possible to create constant text objects this way.

The comment on the `egtext` macro was added because I wanted to consider different variants of adding the position parameter:
```rust
egtext!(
    "Hello!",
    Point::new(10, 20),
    font = Font24x32
);
// or with assignments for all settings
egtext!(
    text = "Hello!",
    position = Point::new(10, 20),
    font = Font24x32
);
```
But this would need changes to the primitive macros to keep them consistent:
```rust
egcircle!(
    (CIRCLE_SIZE, CIRCLE_SIZE),
    CIRCLE_SIZE as u32,
)
// would become:
egcircle!(
    radius = CIRCLE_SIZE as u32,
)
```
This is a bit longer, but it's immediately clear what the values mean.

But this is totally outside the scope of this PR and I don't know if this would be easily implementable without using procedual macros. I'll simply add the position parameter without a label in the `egtext` macro for now.

jamwaffles: Ok cool :+1:

> This is a bit longer, but it's immediately clear what the values mean.

Very true - I like that idea. I don't think it adds too much bloat to the convenience macros. 

> But this is totally outside the scope of this PR and I don't know if this would be easily implementable without using procedual macros.

Agreed, but would be neat if it _was_ implementable.

embedded-graphics/src/fonts/mod.rs
jamwaffles: Ah yeah that's true. Might not ever be a problem so let's see what community feedback we get over time.

embedded-graphics/src/fonts/text.rs
```diff
@@ -0,0 +1,226 @@
...
+        self.top_left() + self.size()
+    }
+
+    /// Returns the size of the bounding box of a styled text.
+    ///
+    /// Currently does not handle newlines (but neither does the rasteriser).
+    /// It will return `Size::zero()` if the string to render is empty.
+    fn size(&self) -> Size {
+        // TODO: Handle height of text with newlines in it
+        let width = F::CHARACTER_SIZE.width * self.primitive.text.len() as u32;
```
jamwaffles: Shouldn't this take into account `F::CHARACTER_SPACING` too?

rfuest: Yes it should, good catch! I've fixed this and added a test.

jamwaffles: Mint!

embedded-graphics/src/fonts/text.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub position: Point,
+}
+
+impl<'a> Text<'a> {
+    /// Creates a text.
+    pub const fn new(text: &'a str, position: Point) -> Self {
+        Self { text, position }
+    }
+
+    /// Attaches a text style the the text object.
```
jamwaffles:
```suggestion
    /// Attaches a text style to the text object.
```

embedded-graphics/src/fonts/text.rs
```diff
@@ -0,0 +1,285 @@
...
+/// [`Styled`]: ../style/struct.Styled.html
+/// [module-level documentation]: index.html
+#[derive(Debug, PartialEq, Eq)]
+pub struct Text<'a> {
+    /// The string.
+    pub text: &'a str,
+
+    /// The position.
+    ///
+    /// The position of a text is defined as the top left pixel of the text.
```
jamwaffles:
```suggestion
    /// Defines the top-left starting pixel of the text object
```

embedded-graphics/src/fonts/text.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    fn bottom_right(&self) -> Point {
+        self.top_left() + self.size()
+    }
+
+    /// Returns the size of the bounding box of a styled text.
+    ///
+    /// Currently does not handle newlines (but neither does the rasteriser).
+    /// It will return `Size::zero()` if the string to render is empty.
```
jamwaffles:
```suggestion
    /// It will return [`Size::zero()`] if the string to render is empty.
    ///
    /// [`Size::zero()`]: struct.Size.html#method.zero
```

embedded-graphics/src/fonts/text.rs
```diff
@@ -0,0 +1,285 @@
...
+            F::CHARACTER_SIZE.height
+        } else {
+            0
+        };
+
+        Size::new(width, height)
+    }
+}
+
+/// Pixel iterator for styled texts.
```
jamwaffles:
```suggestion
/// Pixel iterator for styled text.
```

rfuest:
Thanks for the review. I want to add some documentation to the new simulator example before merging this and will fix your comments then.

jamwaffles:
No problem. Just re-request my

rfuest:
I don't have time to work on the `custom-font` docs right now and opened #214 instead. But the rest should be ready to merge.

jamwaffles:
That's fine, let's do a complete example docs pass in their own PR :slightly_smiling_face:

PR #209, opened 2019-11-27
Fix circle rendering (again :()

### LINKED ISSUE #143: Circles without stroke are drawn incorrectly

jamwaffles:

## PR description

Closes #143 again. Not sure if this is a regression or the test never caught it in the first place.

The test for #143 now loops from sizes `0..10` which does a slightly better job of ensuring filled/stroked circles are the same.

TODO:

* [x] Fix drawing behaviour to be in line with master (stroked and filled or just filled, pick one). Compare:

  Master:

  master

  This branch:
  
  fix

### DIFF
--- a/embedded-graphics/src/primitives/circle.rs
+++ b/embedded-graphics/src/primitives/circle.rs
@@ -137,7 +137,8 @@ where
     // https://stackoverflow.com/a/1237519/383609
     // https://stackoverflow.com/questions/1201200/fast-algorithm-for-drawing-filled-circles#comment80182898_1237519
     fn next(&mut self) -> Option<Self::Item> {
-        // If border or stroke colour is `None`, treat entire object as transparent and exit early
+        // If the fill and stroke colors are `None`, treat entire object as transparent and exit
+        // early.
         if self.style.stroke_color.is_none() && self.style.fill_color.is_none() {
             return None;
         }
@@ -151,15 +152,9 @@ where
             let is_fill = len <= self.outer_threshold;
 
             let item = if is_stroke && self.style.stroke_color.is_some() {
-                Some(Pixel(
-                    self.center + t,
-                    self.style.stroke_color.expect("Border color not defined"),
-                ))
+                Some(Pixel(self.center + t, self.style.stroke_color.unwrap()))
             } else if is_fill && self.style.fill_color.is_some() {
-                Some(Pixel(
-                    self.center + t,
-                    self.style.fill_color.expect("Fill color not defined"),
-                ))
+                Some(Pixel(self.center + t, self.style.fill_color.unwrap()))
             } else {
                 None
             };
--- a/embedded-graphics/src/primitives/circle.rs
+++ b/embedded-graphics/src/primitives/circle.rs
@@ -223,20 +223,31 @@ mod tests {
     /// Test for issue #143
     #[test]
     fn issue_143_stroke_and_fill() {
-        let circle_no_stroke: Styled<Circle, PrimitiveStyle<BinaryColor>> =
-            Circle::new(Point::new(10, 16), 3)
-                .into_styled(PrimitiveStyle::with_fill(BinaryColor::On));
+        for size in 0..10 {
+            let circle_no_stroke: Styled<Circle, PrimitiveStyle<BinaryColor>> =
+                Circle::new(Point::new(10, 16), size)
+                    .into_styled(PrimitiveStyle::with_fill(BinaryColor::On));
 
-        let style = PrimitiveStyle {
-            fill_color: Some(BinaryColor::On),
-            stroke_color: Some(BinaryColor::On),
-            stroke_width: 1,
-        };
-        let circle_stroke: Styled<Circle, PrimitiveStyle<BinaryColor>> =
-            Circle::new(Point::new(10, 16), 3).into_styled(style);
+            let style = PrimitiveStyle {
+                fill_color: Some(BinaryColor::On),
+                stroke_color: Some(BinaryColor::On),
+                stroke_width: 1,
+            };
+            let circle_stroke: Styled<Circle, PrimitiveStyle<BinaryColor>> =
+                Circle::new(Point::new(10, 16), size).into_styled(style);
 
-        assert_eq!(circle_stroke.size(), circle_no_stroke.size());
-        assert!(circle_no_stroke.into_iter().eq(circle_stroke.into_iter()));
+            assert_eq!(
+                circle_stroke.size(),
+                circle_no_stroke.size(),
+                "Filled and unfilled circle iters are unequal for radius {}",
+                size
+            );
+            assert!(
+                circle_no_stroke.into_iter().eq(circle_stroke.into_iter()),
+                "Filled and unfilled circle iters are unequal for radius {}",
+                size
+            );
+        }
     }
 
     #[test]
--- a/embedded-graphics/src/primitives/circle.rs
+++ b/embedded-graphics/src/primitives/circle.rs
@@ -146,9 +146,9 @@ where
             let t = self.p;
             let len = t.x.pow(2) + t.y.pow(2);
 
-            let is_stroke = len > self.inner_threshold && len < self.outer_threshold;
+            let is_stroke = len > self.inner_threshold && len <= self.outer_threshold;
 
-            let is_fill = len < self.outer_threshold;
+            let is_fill = len <= self.outer_threshold;
 
             let item = if is_stroke && self.style.stroke_color.is_some() {
                 Some(Pixel(
--- a/embedded-graphics/src/primitives/circle.rs
+++ b/embedded-graphics/src/primitives/circle.rs
@@ -147,9 +147,8 @@ where
             let t = self.p;
             let len = t.x.pow(2) + t.y.pow(2);
 
-            let is_stroke = len > self.inner_threshold && len <= self.outer_threshold;
-
-            let is_fill = len <= self.outer_threshold;
+            let is_fill = len <= self.inner_threshold;
+            let is_stroke = len <= self.outer_threshold && !is_fill;
 
             let item = if is_stroke && self.style.stroke_color.is_some() {
                 Some(Pixel(self.center + t, self.style.stroke_color.unwrap()))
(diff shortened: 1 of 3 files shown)


jamwaffles:
Looks like this now:

Screenshot from 2019-11-27 12-08-23
Screenshot from 2019-11-27 12-08-36

Circles with a 1px stroke particularly look a lot better now.

jamwaffles:
I noticed a strange issue when making the stroke width greater than the radius. It seems to walk back in on itself. This happens in master too though so I won't fix it for this PR.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -132,26 +134,21 @@ where
...
-        let radius_sq = radius * radius;
-        let outer_radius_sq = outer_radius * outer_radius;
-
         loop {
             let t = self.p;
-            let len = t.x * t.x + t.y * t.y;
+            let len = t.x.pow(2) + t.y.pow(2);
 
-            let is_border = len > radius_sq - radius && len < outer_radius_sq + radius;
+            let is_border = len > self.inner_threshold && len < self.outer_threshold;
```
rfuest: `is_border` should be called `is_stroke` IMO.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -132,26 +134,21 @@ where
...
 
-            let is_fill = len <= outer_radius_sq + 1;
+            let is_fill = len < self.outer_threshold;
```
rfuest: Should these use `<` or `<=` for the comparisons? The size 4 circle in the debug example looks better with `<=` IMO (and matches an equally sized circle drawn in GIMP). But this change breaks size 1 rendering.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -206,11 +203,21 @@ where
             -(self.primitive.radius as i32),
             -(self.primitive.radius as i32),
         );
+
+        // A stroke width of zero renders a 1px border, so add 1 to inner radius to compensate
```
```suggestion
        // A stroke width of zero renders a 1px stroke, so add 1 to inner radius to compensate
```

embedded-graphics/src/primitives/circle.rs
jamwaffles: Hm yeah it does look a lot better. I've made them `<=` in eab8abd.

I also added a kludge in d6b0b86 that fixes tiny circles. I can't really think of a better way of doing this at the moment so please let me know. It does _work_, but it doesn't feel clean.

jamwaffles:
> I assume you will remove the debug example before merging the code?

Yeah I will. It doesn't make much sense as a full demo. I'll remove it once this PR is approved so I can still debug stuff up until that point.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -233,6 +233,51 @@ mod tests {
     use super::*;
     use crate::pixelcolor::BinaryColor;
 
+    fn test_expected_circle(
```
rfuest: Couldn't you use ` MockDisplay::from_pattern` for these tests?

embedded-graphics/src/primitives/circle.rs
```diff
@@ -206,37 +203,105 @@ where
...
+    // Check that tiny filled circle render as a "+" shape with NO hole in the center
+    #[test]
+    fn tiny_circle_filled() {
+        let mut display = MockDisplay::new();
+
+        Circle::new(Point::new(1, 1), 1)
+            .into_styled(PrimitiveStyle::with_fill(BinaryColor::On))
+            .draw(&mut display);
+
+        #[rustfmt::skip]
```
jamwaffles: Added so `from_pattern(...)` stays on multiple lines. I would've liked to have put this on the `MockDisplay::from_pattern` line but this is the closest I could get without throwing warnings.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -206,37 +203,105 @@ where
...
+    // Check that tiny circles render as a "+" shape with a hole in the center
+    #[test]
+    fn tiny_circle() {
+        let mut display = MockDisplay::new();
+
+        Circle::new(Point::new(1, 1), 1)
+            .into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 1))
+            .draw(&mut display);
+
+        #[rustfmt::skip]
```
jamwaffles: Same as https://github.com/jamwaffles/embedded-graphics/pull/209/files#r351842083

embedded-graphics/src/primitives/circle.rs
jamwaffles: Oop yeah I can. I just saw the `Line` harness and copied that. Tests now ported to `MockDisplay`. I had to add `#[rustfmt::skip]` to get the expected values to break lines. See https://github.com/jamwaffles/embedded-graphics/pull/209/files#r351842083

jamwaffles: I couldn't find a rule in the [config options](https://rust-lang.github.io/rustfmt/) to make it more specific so I've just disabled all formatting for that block with `#[rustfmt::skip]`.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -206,37 +203,105 @@ where
             -(self.primitive.radius as i32),
             -(self.primitive.radius as i32),
         );
+
+        // A stroke width of zero renders a 1px stroke, so add 1 to inner radius to compensate
+        let inner_radius = self.primitive.radius as i32 - self.style.stroke_width_i32() + 1;
+        let outer_radius = self.primitive.radius as i32;
+
+        let inner_threshold = inner_radius.pow(2) - inner_radius;
+        let mut outer_threshold = outer_radius.pow(2) + outer_radius;
```
rfuest: Why is the `inner_radius` subtracted and the `outer_radius` added? IMO they should be treated equally and both be added. This way these examples should output the exact same results (with `r2 < r1`):
```rust
egcircle!(c, r1, fill_color = Some(Rgb888::RED));
egcircle!(c, r2, fill_color = Some(Rgb888::GREEN));
```
```rust
egcircle!(
    c, r1,
    fill_color = Some(Rgb888::GREEN),
    stroke_color = Some(Rgb888::RED),
    stroke_width = r1 - r2
);
```

embedded-graphics/src/primitives/circle.rs
```diff
@@ -206,37 +203,105 @@ where
...
+
+        // Special case for small circles. This kludge removes the top-left pixel and leaves the
```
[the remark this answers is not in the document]
jamwaffles: Yeah, tricky to get right without a bunch of special cases. This'll have to do for now I guess.

embedded-graphics/src/primitives/circle.rs
jamwaffles: Honestly, because [this comment](https://stackoverflow.com/questions/1201200/fast-algorithm-for-drawing-filled-circles#comment80182898_1237519) recommended it. I didn't look into why it works any further than implementing it.

If I do change both to an add, I get this incorrect result:

image

embedded-graphics/src/primitives/circle.rs
```diff
@@ -132,28 +134,23 @@ where
...
         loop {
             let t = self.p;
-            let len = t.x * t.x + t.y * t.y;
+            let len = t.x.pow(2) + t.y.pow(2);
 
-            let is_border = len > radius_sq - radius && len < outer_radius_sq + radius;
+            let is_stroke = len > self.inner_threshold && len <= self.outer_threshold;
 
-            let is_fill = len <= outer_radius_sq + 1;
+            let is_fill = len <= self.outer_threshold;
```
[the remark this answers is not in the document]
jamwaffles: I just tried this but the `tiny_circle_filled` test failed with:

```
###
###
##
```

I think the original is ok because the `if` below checks for stroke before fill, so no overdraw occurs AFAIK.

embedded-graphics/src/primitives/circle.rs
jamwaffles: With the original code, I think I get the correct result with `r1 = 20`, `r2 = 10`:

image

(ignore the on-screen text)

Am I missing something?

rfuest: I think this would be the mathematically correct version:
https://gist.github.com/rfuest/83bf0b06c82dec13449932adfc520863

This version uses 2 times larger length values to represent half pixel values.

The threshold calculation is derived from:
`(2(r + 0.5))^2 = (2r + 1)^2 = 4r^2 + 4r + 1`

jamwaffles: Ah that looks loads better. I hope you don't mind, I've shamelessly plugged that gist into this PR. Nice one figuring out the doubled solution.

rfuest: Oops, my code isn't perfect at all:
```rust
#[test]
fn stroke_width_doesnt_affect_fill() {
    let mut expected = MockDisplay::new();
    let mut style = PrimitiveStyle::with_fill(BinaryColor::On);
    Circle::new(Point::new(5, 5), 4)
        .into_styled(style)
        .draw(&mut expected);

    let mut with_stroke_width = MockDisplay::new();
    style.stroke_width = 1;
    Circle::new(Point::new(5, 5), 4)
        .into_styled(style)
        .draw(&mut with_stroke_width);

    assert_eq!(expected, with_stroke_width);
}
```

fixed version of `next`:
```rust
/ https://stackoverflow.com/a/1237519/383609
// https://stackoverflow.com/questions/1201200/fast-algorithm-for-drawing-filled-circles#comment80182898_1237519
    // If the fill and stroke colors are `None`, treat entire object as transparent and exit
    // early.
    if self.style.stroke_color.is_none() && self.style.fill_color.is_none() {
        return None;
    }

    loop {
        let len = (2 * self.p.x).pow(2) + (2 * self.p.y).pow(2);

        let color = if len < self.inner_threshold {
            self.style.fill_color
        } else if len < self.outer_threshold {
            // Use fill_color if no stroke_color was set
            self.style.stroke_color.or(self.style.fill_color)
        } else {
            None
        };
        let item = color.map(|c| Pixel(self.center + self.p, c));

        self.p.x += 1;

        if self.p.x > self.radius as i32 {
            self.p.x = -(self.radius as i32);
            self.p.y += 1;
        }

        if self.p.y > self.radius as i32 {
            break None;
        }

        if item.is_some() {
            break item;
        }
    }
}
```

jamwaffles: Fixed :slightly_smiling_face: test added in as well.

jamwaffles:
Thanks!

rfuest:
@jamwaffles You forgot to remove the debug example.

jamwaffles:
🙈 #213

Issue #134, opened 2019-07-21
embedded-graphics logo


rfuest:
I saw that you added a temporary logo to the repository, so I decided to open this issue to collect logo ideas. Having a nice logo would be really useful for a create which deals with graphics.

This is my contribution, which I created a while ago to test some changes. It is only programmer art but maybe a little less temporary. The logo is just an stylized e and g and is designed to also be usable at small pixel sizes.

Full color with text:
![e-g](https://user-images.githubusercontent.com/226123/61593174-69576200-abdc-11e9-9d47-1a6d5f9d694e.png)
B/W on OLED display:
![e-g oled](https://user-images.githubusercontent.com/226123/61593434-cbfe2d00-abdf-11e9-8113-e848b16f31ba.png)

jamwaffles:
😱 so much better than my attempt! I'll post this on Twitter and see if I can get some more input. Can you think of any other places to post this for more embedded Rustaceans to see?

jamwaffles:
I just had a play around with the OLED logo example and tried it out on docs.rs. Unfortunately the image is too large to fit in the space and scale properly (notice the fuzz):

![image](https://user-images.githubusercontent.com/800791/64157809-4cdf5400-ce2f-11e9-8cda-0adaa5ea3d34.png)

For docs.rs I quite like the transparent background/black pixels idea. Could you create a version like that with a max height of 100px to use on docs.rs? The "full colour with text" version could go at the top of the readme.

rfuest:
The 2 color version was just a very quick test and could still be tweaked a bit, but here are some versions of that logo, for testing:

Unscaled:
![logo1](https://user-images.githubusercontent.com/226123/64444780-3c500780-d0d5-11e9-90d3-1299d5609fd9.png)
Scaled 3x:
![logo2](https://user-images.githubusercontent.com/226123/64444781-3c500780-d0d5-11e9-84d2-6403befc4b8e.png)
Scaled 3x with pixel grid (the 2:1 ratio between gap and pixel size make the pixels look too tiny IMO):
![logo3](https://user-images.githubusercontent.com/226123/64444782-3c500780-d0d5-11e9-80bf-11c7ac7f7486.png)
Scaled 3x with pixel grid and light gray background around pixels (looks better on the actual gray background than on white):
![logo4](https://user-images.githubusercontent.com/226123/64444783-3c500780-d0d5-11e9-88a1-0563c05b9e33.png)


[...]

PR #210, opened 2019-11-29
Add new logo

rfuest:
This PR add the new logo discussed in #134. The SVG file should be directly usable as the `html_logo_url`.


rfuest:
> The only issue I had was to add `?sanitize=true` to the raw.githubusercontent URL like this:

Thanks for that info. I've now set the `html_logo_url` to that url.

> Side note: I assume the SVG is the source of truth, so no other proprietary files like the GIMP XCF or whatever Adobe uses are required?

Yes. The SVG is created in Inkscape and contains some additional information to make it editable (like grid settings). But because the file is still tiny and this doesn't conflict with rendering I don't think there is a point in adding two versions for this at the moment. And "&sanitize=true" also seems to remove the additional tags anyway.

PR #215, opened 2019-12-09
Add PrimitiveStyleBuilder

### LINKED ISSUE #212: Builder for styles
Instances of styles are currently often created like this:
```rust
let decoration_style = PrimitiveStyle {
    fill_color: Some(BinaryColor::Off),
    stroke_color: Some(BinaryColor::On),
    stroke_width: 1,
};
```

One problem with this approach is that it isn't possible to add additional fields to the `PrimitityStyle` struct without breaking the user code. Using a builder pattern additional fields could be added without causing a breaking change.

rfuest:

## PR description

This PR adds a builder for `PrimitiveStyle`, which was discussed in #212.

I'm not sure if this requires a changelog entry because `PrimitiveStyle` didn't exist before.

Closes #212


jamwaffles:
> I'm not sure if this requires a changelog entry because PrimitiveStyle didn't exist before.

It should go in **Added** :slightly_smiling_face:

rfuest:
> It should go in **Added** slightly_smiling_face

OK, I noticed that I didn't add an changelog entry when I created the PR and was lazy :wink:. The `PrimitiveStyle` docs should also mention the builder.

> What are your thoughts on adding a `primitive_style!()` macro in this PR to make life a bit easier for end users? It could use `PrimitiveStyleBuilder` internally.

Interesting idea. A macro could also be used to create `const` styles, when it doesn't internally use the builder.
The `primitive_style` macro would duplicate some functionality of the `eg...` macros. Do you think that these macros should also be changed and use the `primitive_style` macro:
```rust
// current code (after changes proposed in #208)
egcircle!(
    radius = CIRCLE_SIZE as u32,
);

// simply use `primtive_style` macro
egcircle!(
    radius = CIRCLE_SIZE as u32
).into_styled(primitive_style!(
);

// or separate the style properties
// and use `primitive_style` internally in the `egcircle` macro
egcircle!(
    radius = CIRCLE_SIZE as u32
    style = {
    }
);

**Comment by jamwaffles** [comment]
If I may throw my hat into the ring:

```rust
egcircle!(
    radius = CIRCLE_SIZE as u32
    style = primitivestyle!(
    )
);
```

It's a nested macro, but I like the syntax. The macro would then just be (untested):

```rust
#[macro_export]
macro_rules! egcircle {
    ($center:expr, $r:expr, $style:expr $(,)?) => {{
        use $crate::style::PrimitiveStyle;

        $crate::primitives::Circle::new($crate::geometry::Point::from($center), $r)
            .into_styled($style)
    }};
}
```

Thoughts?

rfuest:
> Thoughts?

I like it. Should the macros completely replace the builder or would you add both?

The macros currently use the assignment syntax for the parameters, but assignments separated by commas aren't used anywhere else. Maybe it would be better to use colons instead to be similar to struct literals?
```rust
egcircle!(
    center: (CIRCLE_SIZE, CIRCLE_SIZE),
    radius: CIRCLE_SIZE as u32
    style: primitivestyle!(
        stroke_color: Some(BinaryColor::Off),
        fill_color: Some(BinaryColor::Off)
    )
);
```

Should the macros also support building primitives without styling?
```rust
let circle: Styled<Circle, PrimitiveStyle<C>> = egcircle!(
    radius = CIRCLE_SIZE as u32,
    style = primitive_style!(...),
);

let circle: Circle = egcircle!(
    radius = CIRCLE_SIZE as u32,
);
```
(The type annotation in the example are only added to demonstrate the returned type and are not required later.)

jamwaffles:
> Should the macros completely replace the builder or would you add both?

Both, like we have `Circle::new()` and `egcircle!()`, etc. The non-macro pattern also allows building primitives without a style (see my last paragraph in this comment).

> Maybe it would be better to use colons instead to be similar to struct literals? 

:thinking: that's a good point. Much closer to Rust syntax. Other than another major change I can't think of a bad reason to change the syntax.

> Should the macros also support building primitives without styling?

Hm, maybe not? To me, their use case is to add a bit of sugar over the lengthy primitive/style builder patterns so catering to just that niche makes sense. Non-`style` versions can be added later if desired.

rfuest:
Because the macro changes are now implemented in #224 this should be ready to review.

jamwaffles:
Ah, forgot to mention, can you rebase from master? This should fix the build and the conflicts.

rfuest:
> I think we should also have a `TextStyleBuilder` but that can go in either #224 or another PR.

Yes, we should. Do you want to handle this in #224 or should I open an issue (and later a PR) for this?

jamwaffles:
Yeah I’ll handle that. Feel free to open an issue just to track the feature if you like, but it’s gonna happen in #224 either way.

PR #220, opened 2019-12-17
Clean docs dist folder before build

### LINKED ISSUE #219: Clean target directory in CI builds
The CI builds suddenly broke a few days ago without any new commits. This was caused by a wrong link in the docs I had added two weeks ago. The linkchecker wasn't able to catch the error earlier because an old file was still present in the target directory.

Should we add `cargo clean` or `cargo clean --doc` to `build.sh` to make sure this doesn't happen again?

jamwaffles:

- [ ] ~Check that you've added passing tests and documentation~
- [ ] ~Add a simulator example(s) where applicable~
- [ ] ~Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc)~
- [ ] ~Run `rustfmt` on the project~

## PR description

Closes #219

I opted for `cargo clean --doc` as I'd like to keep the build cache
around for faster builds. I don't think there any issues with stale
build cache at the moment (only docs) so I think adding `--doc` is fine.

Not entirely sure how to test this as I've never seen it fail...

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+cargo clean --doc
+
 cargo fmt --all -- --check
 cargo test --release
 cargo test --release --all-features


jamwaffles:
Gave it a test locally and it does indeed pick up missing files now. Thanks for spotting this!

PR #222, opened 2019-12-19
Implement Drawable for Pixel

### LINKED ISSUE #221: Implement Drawable for Pixel
In working with basic pixel examples for teaching I noticed some issues
I started on an embedded device and went with something like this which was agreeable
```
    display.draw((0..100).map(|i| Pixel(Point::new(i, i * 2), RgbColor::RED)));
```

When I wanted to draw pixels in a loop and I had to wrap in an iter, which is fine but I think we could impl for pixel?
```
    use embedded_graphics::pixelcolor::Rgb565;
    use core::iter;
    for x in 0..100 {
        let pixel = Pixel::<Rgb565>(Point::new(x, x * 2), RgbColor::RED);
        display.draw(iter::once(pixel));
    }
```

But then wanted to move to simulator and neither of those work. Simulator has drawtarget for draw_pixel so this does work but device doesnt have it so thats not ideal.
```
    for x in 0..100 {
        display.draw_pixel(Pixel(Point::new(x, x * 2), BinaryColor::On));
    }
```
I think we need to implement into iterator for pixels on simulator maybe? 

Also I wanted to be able to a row column iterator to draw all the pixels in a region, basically like this
```
    display.draw((0..DISP_SIZE_X).rows(0..DISP_SIZE_Y).map(|(x, y)| {
        ///something clever
        Pixel::<Rgb565>(Point::new(xprime as i32, yprime as i32), color)
    }
```
but couldnt find a name for it. I PRed it to itertools to see if some variation can live there, otherwise maybe in here? Any thoughts/comments?
https://github.com/rust-itertools/itertools/pull/390

rfuest:
This PR implements `Drawable` for `Pixel` to allow drawing of single pixels in the same way as styled primitives and texts.

Closes #221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 - #217 tinytga: Added support for TGA files with color map.
 
+- #221 Implemented `Drawable` for `Pixel` to draw single pixels.
+
 ### Fixed
 
 - #143, #209 Circles with no stroke are now drawn correctly
--- a/embedded-graphics/src/drawable.rs
+++ b/embedded-graphics/src/drawable.rs
@@ -3,7 +3,35 @@ use crate::geometry::Point;
 use crate::pixelcolor::PixelColor;
 use crate::DrawTarget;
 
-/// A single pixel
+/// A single pixel.
+///
+/// `Pixel` objects are used to specify the position and color of drawn pixels.
+///
+/// # Examples
+///
+/// The [`Drawable`] trait is implemented for `Pixel` which allows single pixels
+/// to be drawn to a [`DrawTarget`]:
+/// ```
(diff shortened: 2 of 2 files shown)


jacobrosenthal:
Tested in the simulator works great thanks

(though unrelated, I had to peel off a few commits on my local machine to get simulator to work..)
```
$ cargo run --example offscreen
error: failed to resolve patches for `https://github.com/rust-lang/crates.io-index`

Caused by:
  patch for `tinytga` in `https://github.com/rust-lang/crates.io-index` did not resolve to any crates. If this is unexpected, you may wish to consult: https://github.com/rust-lang/cargo/issues/4678

```

rfuest:
> (though unrelated, I had to peel off a few commits on my local machine to get simulator to work..)

Can you to delete `Cargo.lock` and see if this fixes the problem?

jacobrosenthal:
yeah that did it, thanks again

Issue #182, opened 2019-09-14
Coordinate system


rfuest:
We started discussing e-g's coordinate system in issue #142, but the one by off error is only one aspect of this, so I decided to open another issue for a more general discussion.

Quick recap: The issue is to determine how e-g should handle coordinates. Being limited to integer values allows two interpretations: Points can either lie in the center of pixels or on the grid line between them.

In the rest of this post I'll try describe a relatively consistent way of handling coordinates using points in the center of pixels. I'm pretty sure that this might be somewhat confusing (at least I was confused while writing this), but I hope the images make it clearer.

Drawing lines with odd widths works without any special considerations. Lines with even width are drawn with one pixel more on the left side (looking from the start to the end point). By definition lines with zero length will still draw a dot.
![lines](https://user-images.githubusercontent.com/226123/64912309-96804680-d72d-11e9-8456-82f2d1519659.png)

Rectangles are defined by their top left corner and size. The actual shape, that is described by this, lies on the edge of the top left pixel (has a half pixel offset).
To replicate the current behavior, non zero sized rectangles can also be constructed by an additional constructor that takes a top left and bottom right corner. Passing the same point as top left and bottom right corner to this constructor would result in 1px * 1px rectangle.
![primitives](https://user-images.githubusercontent.com/226123/64912393-c8de7380-d72e-11e9-8477-d74af12ff604.png)

Strokes on rectangles are, by default, drawn by using the path shown in the following image. This makes them work consistently with lines:
![rectangles](https://user-images.githubusercontent.com/226123/64912419-47d3ac00-d72f-11e9-936f-ea7efa9b5b97.png)

Strokes on circles work the same way. The path to draw the stroke is not shown in this image, but it is defined to be half a pixel inside the outer edge and going clockwise.

![circles](https://user-images.githubusercontent.com/226123/64912436-8ec1a180-d72f-11e9-9a89-a71abdf3f93d.png)

To replicate the way strokes work currently on rectangles and circles a `stroke_alignment` option can be added to the style. The default centered alignment is consistent with the way strokes work on lines. The stroke alignment only affects closed shapes.

![alignment1](https://user-images.githubusercontent.com/226123/64912471-4c4c9480-d730-11e9-9d93-72253a43723c.png)
![alignment2](https://user-images.githubusercontent.com/226123/64912474-4e165800-d730-11e9-972d-ae75a550cf89.png)

Problems:
[...]

Issue #147, opened 2019-07-25
Circle defined by center and diameter


jamwaffles:
`Circle::new()` creates a circle by center and _radius_. Is it useful to be able to create one by giving a diameter instead? The diameter can then be an odd number, however won't be symmetrical around the circle's center point. Does this matter?

rfuest:
I think it would be easier to use the top left point instead of the center as a parameter. This would for example make it easier to draw a circle inside a rectangle.
Or if we decide to add a `Rect` (see #141) type this could also be used instead.

aurelj:
I've just stumbled on this issue.

Current `Circle` defined by center and radius only allows building circles with an odd diameter (2 * radius + 1). It is impossible to build circles with an even diameter. We can only build half of the possible circles. This is an issue when you want to draw a circle centered on a display with an even width and height (pretty much any display in existence ?).

This is related to #182. With a coordinate system based on the center of pixels and a circle defined by its center, there is no way to build a circle with an even diameter. One solution would be to change the coordinate system to lie on the grid between pixels, but then we couldn't build circles with an odd diameter.

I think the best way to solve this is, as suggested by @rfuest, to define a circle by its top left point and its diameter. This allows building both odd and even diameter circles.

I'm willing to implement this, but I'm wondering what's the best way to do it:
- keep the current `Circle` and add a new `CircleDiameter` (or a better name) ? (I don't really like this)
- change the implementation of `Circle` breaking API compatibility (in 0.7.0 I guess) ?
- something else ?

I'm leaning toward the second solution. What do you think ?

rfuest:
I would also prefer the second solution. Having two types to represent circles would be confusing and unnecessary because all circles can be represented using the new type.

Changing the implementation is a breaking change, but the new type can provide additional constructors and getters to make porting old code easier. Some getters like `center` would only be able to output rounded values, but I guess that can be expected when not using floating point coordinates.

One focus of 0.7 will be geometry (like #182 you mentioned), so this will be a good time to implement this change.

jamwaffles:
I'm on the fence about `Ellipse` being a separate shape or not. Here's my thinking on an API that supports ellipses and circles at the same time:

```rust
let center = Point::new(20, 20);

// Circle
let circle = Ellipse::new(center, Size::new(10, 10));

// Ellipse
let ellipse = Ellipse::new(center, Size::new(25, 10));
```

Thoughts?

aurelj:
[...]

PR #224, opened 2019-12-20
Only use named args in macros

### LINKED ISSUE #208: Only use named parameters in macros
As discussed in PR #206 changing the macros to only use the assignment style named parameters would make macro calls more readable.

```rust
// current version
egcircle!(
    (CIRCLE_SIZE, CIRCLE_SIZE),
    CIRCLE_SIZE as u32,
)

// new version
egcircle!(
    radius = CIRCLE_SIZE as u32,
)
```

This would also allow to define alternative ways of describing the dimension of primitives. The `egcircle`could for example support `radius` and `diameter` as parameters (see #147).

### LINKED ISSUE #231: Add TextStyleBuilder
Add a `TextStyleBuilder` and make sure `TextStyle` isn't instantiated externally by adding the `non_exhaustive` attribute to `TextStyle`.

Similar changes were implemented for `PrimitiveStyle` in PR #215.

jamwaffles:

## PR description

TODO

* [x] `primitive_style!()` macro
* [x] Change primitives macros to use `style = primitive_style!()` syntax
* [x] `text_style!()` macro
* [x] Change `text!()` macro syntax to use nested `text_style!()`
* [x] Add changelog entry/entries for all of this stuff
* [ ] ~Change from assignment to struct-like macro syntax~ - not doing this. The formatter doesn't like it.
* [x] Refactor primitive/text macro documentation
* [x] Add `TextStyleBuilder` now that #215 is merged (closes #231)
* [x] Change macros to remove `Some()`/`None` wrapping; a key being present implies a `Some()` for that value.
* [x] Run nightly format to format examples

Changes macros to use named (as opposed to positional) args:

```rust
// Draw a circle centered at (64, 64) with a radius of 64 and a white 1px stroke
egcircle!(
    center = (64, 64),
    radius = 64,
    style = primitive_style!(stroke_color = Rgb565::WHITE)
)

egline!(
    start = (64, 64),
    end = (0, 64),
    style = primitive_style!(stroke_color = Rgb565::WHITE)
)

egline!(
    start = (64, 64),
    end = (80, 80),
    style = primitive_style!(stroke_color = Rgb565::RED)
)

egrectangle!(
    top_left = (64, 64),
    bottom_right = (80, 80),
    style = primitive_style!(fill_color = Rgb565::BLACK)
)

// Print "Hello world!" in a white 6x8 pixel font with the top left corner positioned at (5, 50)
egtext!(
    text = "Hello world!",
    top_left = (5, 50),
    style = text_style!(font = Font6x8, text_color = Rgb565::WHITE)
)
```

Notes:

1. ~I'm not keen on `pN` for `egtriangle!()`. Because the points can be defined in any order, is there a better naming scheme to use?~ - resolved by passing an array of points.
2. ~`egtext!()` uses `top_left` instead of `position`. This is more accurate, but does it add confusion?~ - resolved
3. ~`egrectangle!()` using `top_left` and `bottom_right` is weird. Should we move over to `top_left` and `size`? I think this was discussed in another issue long ago (or was that circles?) If so, let's do it in another PR after the next alpha release.~ - resolved

Closes #208


CHANGELOG.md
```diff
@@ -69,6 +69,35 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 - **(breaking)** Text rendering is now implemented using `Text`, `TextStyle` and `Font` objects instead of using `Font::render_str`.
 
+- **(breaking)** #224 Embedded Graphics macros now use named instead of positional parameters
+
+  ```rust
+  // OLD
+  egcircle!((CIRCLE_SIZE, CIRCLE_SIZE), CIRCLE_SIZE as u32);`
+  // NEW
+  egcircle!((CIRCLE_SIZE, CIRCLE_SIZE), CIRCLE_SIZE as u32);
```
rfuest: The parameter names are missing and the example should also use number literals instead of `CIRCLE_SIZE` IMO.

CHANGELOG.md
```diff
@@ -69,6 +69,35 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
...
+
+  // OLD
+  egrectangle!((0, 0), (64, 64));`
+  // NEW
+  egrectangle!((0, 0), (64, 64));
```
rfuest: The parameter names are missing,

CHANGELOG.md
```diff
@@ -69,6 +69,35 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
...
+  // NEW
+  egtriangle!(p1 = (32, 0), p2 = (0, 64), p3 = (64, 64));
+
+  // OLD
+  egline!((32, 0), (0, 64));`
+  // NEW
+  egline!(start = (32, 0), end = (0, 64));
+
+  // OLD
+  egtext!("456", (10, 10, font = Font6x8);`
```
rfuest:
```suggestion
  egtext!("456", (10, 10), font = Font6x8);
```

CHANGELOG.md
```diff
@@ -69,6 +69,35 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
...
+  // NEW
+  egtext!(text = "456", top_left = (10, 10, font = Font6x8);
```
rfuest:
```suggestion
  egtext!(text = "456", top_left = (10, 10), font = Font6x8);
```

embedded-graphics/src/lib.rs
```diff
@@ -100,15 +101,25 @@
...
+//!     radius = 8,
+//!     fill_color = Some(Rgb565::RED)
+//! );
+//! let t = egtext!(
+//!     text = "Hello Rust!",
+//!     top_left = Point::zero(),
+//!     font = Font6x8,
+//!     text_color = Some(Rgb565::GREEN)
+//! )
+//! .translate(Point::new(20, 16));
```
rfuest: If this isn't supposed to demonstrate `translate` the `top_left` argument shoud be set to `(20, 16)` instead.

embedded-graphics/src/lib.rs
```diff
@@ -119,16 +130,29 @@
...
+//!             fill_color = Some(Rgb565::RED)
+//!         ))
+//!         .chain(
+//!             &egtext!(
+//!                 text = text,
+//!                 top_left = Point::zero(),
+//!                 font = Font6x8,
+//!                 text_color = Some(Rgb565::GREEN)
+//!             )
+//!             .translate(Point::new(20, 16)),
```
rfuest: See comment above

jamwaffles:
> As a user I would expect to be able to set named parameters in any order.

Is that possible? The macro could be defined as `$(, $key:ident = $value:expr )*` but I don't see a clear way to get that into a defined primitive/font. For example, one of the keys in `egcircle!()` needs to be wrapped in `$crate::geometry::Point::from($center)`. The other way is to have multiple definitions that accept all orderings, but that feels messy. Is there some neat macro trick I'm missing?

> Do you want to change the syntax to be struct literal like, as díscused in https://github.com/jamwaffles/embedded-graphics/pull/215#issuecomment-564223882?

Yeah I'll add that in later. Doesn't make sense to release an alpha with this change in it and then immediately break it later.

> An array like syntax would work:

Good idea. Changed!

>  But maybe text positioning in general should be discussed further

Certainly. Just opened #225.

> See issue #182 and the related #147 for circles.

Thanks for digging those out. I'll leave it alone for this PR as I mentioned though.

rfuest:
> Is there some neat macro trick I'm missing?

I don't think there is an easy way to do this but it would be possible by using some advanced macro magic. If you don't want to change this now the docs should mention that the ordering is important.

> > An array like syntax would work:
> 
> Good idea. Changed!

It would be even nicer if `points` could also be assigned to a real array. Adding an additional `Triangle::with_points` constructor which takes a `[Point; 3]` or perhaps `T: Borrow<[Point; 3]>` parameter should do the trick.

Should I finish #215 first so that the `primitivestyle` macro can use the new builder?

I would expect the `primitivestyle` to be named `primitive_style`, because that is the default camel to snake case conversion. Is there any reason for leaving out the underscore?

jamwaffles:
I was wondering if this might happen...

> Change from assignment to struct-like macro syntax

Introducing this change prevents macro usages from being `rustfmt`-ed. Specifically this syntax:

```rust
{ $($style_key:ident: $style_value:expr ),* } => {{
    // ...
}};

// Usage
primitive_style! { stroke_color: Some(BinaryColor::On), stroke_width: 2 };
```

Formatting also has no effect when going back to parens:

```rust
primitive_style!(stroke_color: Some(BinaryColor::On), stroke_width: 2);
```

Adding whitespace and newlines and stuff doesn't behave quite right. This is using rustfmt `1.4.9-stable` on Rust `1.40.0`. I'd very much like whatever macros e-g exports to be rustfmt-able.

During the refactor, the struct-like syntax is pretty weird to type. It feels like a mix between function syntax and struct syntax. At least with `macro!(foo = bar, ...)` there's a definite syntactic difference. I get that it's a departure from standard Rust property syntax but macros have that freedom available to them. Thoughts?

rfuest:
> Thoughts?

Sorry for the late answer, but hat makes sense, we should keep the assignment syntax.

One thing I'd like to address is the how colors are set in the builders vs the macros. When I wrote the `PrimitiveStyleBuilder` I decided that using `Option` for the color setters was unnecessary, which you also did in `TextStyleBuilder`, but the macros require using `Some(...)`. I would prefer to remove the need for `Some` in the macros as well, because they are also builder like, especially with the assignment syntax. But I would like to hear your thoughts on this.

jamwaffles:
> I would prefer to remove the need for Some in the macros as well, because they are also builder like, especially with the assignment syntax.

Yeah, this has popped into my head a few times as well. There are a couple of arguments either way, but I think it's better to elide the `Some()` wrapping. The macros are convenient sugars over the longer standard syntax so removing the `Some()`s makes sense. I'll get that done asap.

jamwaffles:
@rfuest ready for another round. Macro changes made. It's pretty late so might have missed something dumb...

embedded-graphics/src/primitives/mod.rs
```diff
@@ -30,42 +26,59 @@ pub trait Primitive: Dimensions {
...
+///     prelude::*,
+///     primitive_style,
+///     primitives::Circle,
+///     style::{PrimitiveStyle, Styled},
+/// };
+///
+/// let line_circle: Styled<Circle, PrimitiveStyle<Rgb565>> =
+///     egcircle!(center = (10, 20), radius = 30);
+/// let line_circle: Styled<Circle, PrimitiveStyle<Rgb565>> =
+///     egcircle!(center = Point::new(10, 20), radius = 30);
```
rfuest: This should have a comment to explain that different ways of specifying points are possible.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -30,42 +26,59 @@ pub trait Primitive: Dimensions {
...
-///     fill_color = Some(Rgb565::GREEN)
+///     center = (10, 20),
+///     radius = 30,
+///     style = primitive_style!(stroke_color = Rgb565::RED, fill_color = Rgb565::GREEN)
 /// );
 /// ```
 ///
-/// Style properties like `stroke_color` map to the properties in the [`PrimitiveStyle`] struct.
+/// Style properties like `stroke_color` map to methods on the [`PrimitiveStyleBuilder`] struct.
 /// For example, the following code makes two identical circles:
```
rfuest: The example creates 3 circles and not two as described. I would remove `circle_2` from the example to make the point clear.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -95,36 +114,57 @@ macro_rules! egcircle {
...
+/// };
+///
+/// let line_1: Styled<Line, PrimitiveStyle<Rgb565>> = egline!(
+///     start = Point::new(10, 20),
+///     end = Point::new(30, 40),
+///     style = primitive_style!(
+///         stroke_color = Rgb565::BLUE,
+///         fill_color = Rgb565::YELLOW,
+///         stroke_width = 1
+///     )
```
rfuest: This should be removed. IMO the ability to use different point variants should be explained once in a prominent location and not clutter every example that tries to illustrate something else.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -133,64 +173,89 @@ macro_rules! egcircle {
...
-///     fill_color = Some(Rgb565::GREEN)
+///
+/// let rectangle_2: Styled<Rectangle, PrimitiveStyle<Rgb565>> = egrectangle!(
+///     top_left = Point::new(10, 20),
+///     bottom_right = Point::new(30, 40),
+///     style = primitive_style!(
+///         stroke_color = Rgb565::RED,
+///         fill_color = Rgb565::GREEN,
+///         stroke_width = 1
+///     )
```
rfuest: Remove. See above.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -1,12 +1,16 @@
...
-use super::super::transform::Transform;
-use crate::geometry::{Dimensions, Point, Size};
-use crate::pixelcolor::PixelColor;
-use crate::primitives::Primitive;
-use crate::style::{PrimitiveStyle, Styled};
-use crate::DrawTarget;
+use super::super::{
+    drawable::{Drawable, Pixel},
+    transform::Transform,
+};
```
rfuest: Use `crate::` path.

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -1,13 +1,20 @@
...
-use crate::geometry::{Dimensions, Point, Size};
-use crate::pixelcolor::PixelColor;
-use crate::primitives::line::{Line, StyledLineIterator};
-use crate::primitives::Primitive;
-use crate::style::{PrimitiveStyle, Styled};
-use crate::DrawTarget;
+use super::super::{
+    drawable::{Drawable, Pixel},
+    transform::Transform,
+};
```
rfuest: Use `crate::` paths.

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -75,6 +100,26 @@ impl Triangle {
     pub const fn new(p1: Point, p2: Point, p3: Point) -> Self {
         Triangle { p1, p2, p3 }
     }
+
+    /// Create a new triangle from an array of points
```
```suggestion
    /// Creates a new triangle from an array of points.
```

embedded-graphics/src/style/mod.rs
```diff
@@ -6,4 +6,106 @@ mod text_style;
 
 pub use primitive_style::{PrimitiveStyle, PrimitiveStyleBuilder};
 pub use styled::Styled;
-pub use text_style::TextStyle;
+pub use text_style::{TextStyle, TextStyleBuilder};
+
+/// Create a [`PrimitiveStyle`]
+///
+/// Any properties not specified in the macro call will use the values provided by
```
rfuest: This should also mention what properties can be specified like in the `text_style` macro below.

embedded-graphics/src/style/mod.rs
```diff
@@ -6,4 +6,106 @@ mod text_style;
...
+/// Create a [`TextStyle`]
+///
+/// All properties on [`TextStyle`] are supported. At least `font` is required, and must be the
+/// first property passed to the macro.
+///
+/// ## Examples
+///
+/// ### Create a default text style
+///
+/// This example uses [`Font8x16`] and the [`Rgb565`] colour space to create a default text style.
```
rfuest: Sorry for this, but:
```suggestion
/// This example uses [`Font8x16`] and the [`Rgb565`] color space to create a default text style.
```
(This also applies to the British colours below)

embedded-graphics/src/style/text_style.rs
```diff
@@ -1,12 +1,19 @@
-use crate::fonts::Font;
-use crate::pixelcolor::PixelColor;
+use crate::{fonts::Font, pixelcolor::PixelColor};
 
 /// Style properties for texts.
 ///
 /// `TextStyle` can be applied to a [`Text`] object to define how a text is drawn.
 ///
+/// Because `TextStyle` has the [`non_exhaustive` attribute], it cannot be created using a struct
```
rfuest:
```suggestion
/// Because `TextStyle` has the [`non_exhaustive`] attribute, it cannot be created using a struct
```

embedded-graphics/src/style/text_style.rs
```diff
@@ -1,12 +1,19 @@
...
+/// expression. To create a `TextStyle`, the [`text_style!()` macro] or [`TextStyleBuilder` builder]
```
rfuest:
```suggestion
/// literal. To create a `TextStyle`, the [`text_style!`] macro or [`TextStyleBuilder`] builder
```
Or is builder redundant in this case?
```suggestion
/// literal. To create a `TextStyle`, the [`text_style!`] macro or [`TextStyleBuilder`]
```

embedded-graphics/src/style/text_style.rs
```diff
@@ -35,19 +42,100 @@ where
             background_color: None,
         }
     }
+
+    /// Create a text style with only the font defined
```
rfuest: Is this function necessary? It can easily be replaced by `TextStyleBuilder::new(font).build()` and `TextStyle`s without any set colors are rarely useful.

embedded-graphics/src/style/text_style.rs
```diff
@@ -35,19 +42,100 @@ where
...
+    pub fn with_font(font: F) -> Self {
+        Self {
+            font,
+            text_color: None,
+            background_color: None,
+        }
+    }
+}
+
+/// Primitive style builder.
```
rfuest: This isn't the primitive style builder and it could use some additional docs (which the `PrimitiveStyleBuilder` could also use)

embedded-graphics/src/style/text_style.rs
```diff
@@ -35,19 +42,100 @@ where
...
+{
+    style: TextStyle<C, F>,
+}
+
+impl<C, F> TextStyleBuilder<C, F>
+where
+    C: PixelColor,
+    F: Font,
+{
+    /// Creates a new primitive style builder with a given font.
```
rfuest:
```suggestion
    /// Creates a new text style builder with a given font.
```

embedded-graphics/src/style/text_style.rs
```diff
@@ -35,19 +42,100 @@ where
...
+    }
+
+    /// Sets the background color.
+    pub fn background_color(&mut self, background_color: C) -> &mut Self {
+        self.style.background_color = Some(background_color);
+
+        self
+    }
+
+    /// Builds the primitive style.
```
rfuest:
```suggestion
    /// Builds the text style.
```

embedded-graphics/src/style/text_style.rs
```diff
@@ -35,19 +42,100 @@ where
...
+        assert_eq!(
+            TextStyleBuilder::new(Font12x16)
+                .text_color(BinaryColor::On)
+                .build(),
+            TextStyle::new(Font12x16, BinaryColor::On)
+        );
+    }
+
+    #[test]
+    fn builder_fill() {
```
rfuest:
```suggestion
    fn builder_background_color() {
```

CHANGELOG.md
```diff
@@ -75,6 +71,50 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
...
+    fill = Rgb565::GREEN
+  ));
+
+  // OLD
+  egrectangle!((0, 0), (64, 64), stroke = Some(Rgb565::RED), fill = Some(Rgb565::GREEN));
+  // NEW
+  egrectangle!(top_left = (0, 0), bottom_right = (64, 64), style = primitive_style!(
+    stroke = Rgb565::RED,
+    fill = Rgb565::GREEN
+  ));
```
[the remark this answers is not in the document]
jamwaffles: This was me trying to make the examples a bit shorter. I've replaced the entire block with a rustfmt-ed version. Longer, but a lot easier to read.

embedded-graphics/README.md
```diff
@@ -82,32 +92,36 @@ fn main() {
...
+    .draw(&mut display);
     egrectangle!(
-        (64, 64),
-        (80, 80),
-        stroke_color = None,
-        fill_color = Some(BinaryColor::Off)
-    ).draw(&mut display);
+        top_left = (64, 64),
+        bottom_right = (80, 80),
+        // No stroke, black fill
```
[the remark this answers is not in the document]
jamwaffles: Better comments added, albeit without colours. Not sure how to succinctly explain that `BinaryColor` gets turned into white in most cases. Maybe switch the example to use `Rgb565::WHITE`?

embedded-graphics/src/style/text_style.rs
jamwaffles: Good point. Removed.

embedded-graphics/src/primitives/mod.rs
jamwaffles: I could add an explanation to the top of each macro's doc comment? A bit repetitive but I can't think of anywhere better.

embedded-graphics/src/fonts/mod.rs
```diff
@@ -158,7 +188,7 @@ pub use font8x16::Font8x16;
 use crate::geometry::Size;
 
 /// Monospaced bitmap font.
-pub trait Font {
+pub trait Font: Copy {
```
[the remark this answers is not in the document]
jamwaffles: It's required in the builder `.build()` methods - they return a `TextStyle` which requires that it be `Copy`. That cascades into the `Font` passed into `TextStyle`.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -47,7 +52,7 @@ use crate::DrawTarget;
 ///     .into_styled(PrimitiveStyle::with_fill(Rgb565::BLUE))
 ///     .draw(&mut display);
 /// ```
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
```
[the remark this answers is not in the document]
jamwaffles: I've opened #238 to address adding more derives. This PR is already huge, and the image crates need a few adding too so I'll do them at a later date.

> Should we also consider serde support? 

Definitely behind a feature, if at all. Not really sure how useful serde support would be, but if someone wants it in the future we can add it in.

embedded-graphics/src/style/mod.rs
jamwaffles: :(((((

I did a find and replace through the repo and changed every occurrence I could find, including some test image file names.

embedded-graphics/src/fonts/mod.rs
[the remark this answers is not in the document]
rfuest: If this is only necessary in the builder the builder should get a `Font + Clone` type bound instead. I don't see a need for font type that can't be copied, but IMO its better to keep it generic.

embedded-graphics/src/primitives/mod.rs
rfuest: Adding some module level docs to the `primitives` module could provide an overview of how primitives can be created and drawn. And this could also mention the macros in general and how points can be defined when calling them.

embedded-graphics/README.md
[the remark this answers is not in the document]
rfuest: Yes, changing the example to use colors makes it easier to write shorter comments and explaining `BinaryColor` isn't the point for this example.

embedded-graphics/src/fonts/mod.rs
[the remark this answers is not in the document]
jamwaffles: Agreed, the bound on `Font` is unnecessary. I've added `F: Font + Copy` to `TextStyleBuilder` (as opposed to `Clone`). Is there any reason to require `Clone` over `Copy`? `Copy` is a slightly looser bound so there's a bit more freedom from the `Font` impl side.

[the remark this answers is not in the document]
rfuest: `Clone` is the looser one:
```rust
pub trait Copy: Clone { }
```

embedded-graphics/src/style/text_style.rs
jamwaffles: Added some docs for both. Let me know what you think. Committed whilst tired af so might have missed something important...

embedded-graphics/src/fonts/mod.rs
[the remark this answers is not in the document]
jamwaffles: Oh, huh, TIL. I've replaced the bound with `Clone` which requires calling `.clone()` explicitly in `.build()`.

embedded-graphics/src/style/primitive_style.rs
```diff
@@ -79,6 +79,63 @@ where
 }
 
 /// Primitive style builder.
+///
+/// Use this builder to create styles for [primitives]. For convenience, the [`primitive_style!`]
```
rfuest: > Use this builder to create styles for [primitives].

IMO, this should link to `PrimitiveStyle` instead, which should describe how styled primitives can be used.

> For convenience, the [`primitive_style!`] macro is also provided.

It's not obvious that the function of the `primitive_style` macro is building styles. And I would move this to the end of the paragraph, because the main function of this page is describing `PrimitiveStyleBuilder`.

embedded-graphics/src/style/primitive_style.rs
```diff
@@ -79,6 +79,63 @@ where
...
+/// macro is also provided.
+///
+/// If any properties on the builder are ommitted, the value will default to `None` (transparent).
```
rfuest:
```suggestion
/// If any properties on the builder are omitted, the value will default to `None` (transparent).
```

Not all values default to `None`, like `stroke_width`.

embedded-graphics/src/style/primitive_style.rs
```diff
@@ -79,6 +79,63 @@ where
...
+///
+/// # Examples
+///
+/// ## Build a style with customised stroke and fill
```
rfuest:
```suggestion
/// ## Build a style with customized stroke and fill
```

embedded-graphics/src/style/primitive_style.rs
```diff
@@ -79,6 +79,63 @@ where
...
+///     .stroke_color(Rgb565::RED)
+///     .stroke_width(3)
+///     .fill_color(Rgb565::GREEN)
+///     .build();
+///
+/// let circle = Circle::new(Point::new(20, 20), 10)
+///     .into_styled(style);
+/// ```
+///
+/// ## Build a style with ommitted (transparent) values
```
rfuest: IMO the headline should contain what the example tries to achieve and not how this is done. Something like:
`Build a style with stroke and without fill`.
The description below should then provide how this is achieved.

embedded-graphics/src/style/primitive_style.rs
```diff
@@ -79,6 +79,63 @@ where
...
+///
+/// This example builds a style for a rectangle with a 1px red stroke. Because `.fill_color()` is
+/// not called, the fill is defaulted to `None` (i.e. transparent).
```
rfuest:
```suggestion
/// not called, the fill color remains  the default value of `None` (i.e. transparent).
```

embedded-graphics/src/style/text_style.rs
```diff
@@ -37,17 +44,172 @@ where
     }
 }
 
+/// Text style builder.
+///
+/// Use this builder to create styles for [`Font`]s. For convenience, the [`text_style!`] macro is
```
rfuest: `TextStyle`s are applied to `Text` and not `Font`s.
Like in the `PrimitiveStyleBuilder` docs this should also describe the purpose of the `text_style` macro. And a link to `TextStyle` is missing.

embedded-graphics/src/style/text_style.rs
```diff
@@ -37,17 +44,172 @@ where
...
+/// also provided.
+///
+/// # Examples
+///
+/// ## Render some text with a blue background and yellow text color
```
rfuest:
```suggestion
/// ## Render yellow text on a blue background
```

embedded-graphics/src/style/text_style.rs
```diff
@@ -37,17 +44,172 @@ where
...
+///     text_color = Rgb565::WHITE,
+///     background_color = Rgb565::BLACK
+/// );
+///
+/// let text = Text::new("Hello Rust!", Point::new(0, 0)).into_styled(style);
+/// ```
+///
+/// ## Transparent background
+///
+/// If a property is ommitted, it will default to `None` in the resulting `TextStyle` returned by
```
rfuest: Not all values might default to `None` in the future. I would just mention that they stay at the default value.

embedded-graphics/src/style/text_style.rs
```diff
@@ -37,17 +44,172 @@ where
...
+/// `.build()`. This example draws white text with no background at all.
```
rfuest: The example draws yellow text.

jamwaffles:
Hokay, doc review done. Are there any functional issues remaining with this PR? I'd really like to get this over the line. Can always do a doc update in a future PR.

jamwaffles:
Mad! Thanks!

PR #226, opened 2019-12-25
Trial ~Discord~ Matrix

jamwaffles:
This adds a badge and links to [this Matrix channel](https://matrix.to/#/#rust-embedded-graphics:matrix.org) I don't know if this is the best solution, but let's start somewhere and see if Matrix is right for this project. If not, we can always switch.

This used to be Discord, but switched to Matrix. Thanks to @jacobrosenthal for pointing out that Rust is moving to Matrix.

Cc https://github.com/jamwaffles/embedded-graphics/issues/156#issuecomment-568593897


jamwaffles:
Ignoring the build failures for this PR. Matrix is alright tbh, although haven't used it much.

PR #227, opened 2019-12-26
Fix build from 1.40.0 beakage

jamwaffles:
I removed the offending line in `build_target.sh`. It was somewhat superfluous anyway, but the `--features` arg was doing something weird with dependencies. For some reason, Cargo decided it should build `criterion` which doesn't work for no-std targets. Not sure why this happened, but removing a build cmd with `--features` in it fixes the problem.


jamwaffles:
Looks like this is a symptom of https://github.com/rust-lang/cargo/issues/5730. Didn't know that issue existed.

rfuest:
`criterion` caused some build problems before, but they could be fixed by adding `default-features = false` back then.

We could move the benches (and the `criterion` dependency) into the `simulator` crate, which depends on `std` anyway. Better benchmarking, also for the embedded targets, could then be implemented later.

jamwaffles:
> We could move the benches (and the criterion dependency) into the simulator crate

Thanks for the suggestion. Moved! Appears to work locally but let's see what the build does.

PR #237, opened 2020-01-10
simulator: add game of life adapted from https://github.com/rustwasm/wasm_game_of_life

jacobrosenthal:

## PR description

Not sure where to showcase. I still havent quite figured if maybe it can fit in its original repo without complicating the original example and show how you can have both wasm and no_std embedded code work.. But for now here it is.

The big changes from the upstream codebase were to 
* use two heapless vec for game state, and then `core::mem::swap` between them, instead of using Vec and `clone`. 
* impl `iter` so I didnt have to leak so much internal state, and also because then embedded graphics could consume game state easily.

### DIFF
--- a/simulator/Cargo.toml
+++ b/simulator/Cargo.toml
@@ -33,6 +33,7 @@ required-features = ["bmp"]
 
 [dependencies]
 sdl2 = "0.32.2"
+heapless = "0.5.1"
 
 [dependencies.embedded-graphics]
 version = "0.6.0-alpha.2"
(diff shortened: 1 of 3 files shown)


jamwaffles:
A simulator example makes sense to me if we want to bring the example in-project. The other method would be a hosted WASM demo in a separate repo with a link in the e-g README. Open to other suggestions though.

jamwaffles:
Can you rebase from master to see if that fixes the build?

jamwaffles:
It would be cool to have a GoL example as one of the embedded-graphics demos, but for now I'm closing this PR as stale. Please rebase the source branch and reopen if you'd like to get this merged!

PR #240, opened 2020-01-11
Updated projects list

jamwaffles:

## PR description

* Adds [hub75 driver](https://crates.io/crates/hub75) which has e-g support
* ~Adds version badges next to crates:~
  
  image

  ~After adding them I'm not so sure it's a great idea... thoughts?~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Added
 
-- None
+- Added [hub75](https://crates.io/crates/hub75) driver to list of known supported drivers.
 
 ### Changed
 
--- a/embedded-graphics/README.md
+++ b/embedded-graphics/README.md
@@ -148,6 +148,7 @@ fn main() {
 
 ## Display drivers with embedded-graphics support
 
+- [hub75](https://crates.io/crates/hub75): A rust driver for hub75 rgb matrix displays
 - [ili9341](https://crates.io/crates/ili9341): A platform agnostic driver to interface with the ILI9341 (and ILI9340C) TFT LCD display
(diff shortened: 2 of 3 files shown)


rfuest:
>     * After adding them I'm not so sure it's a great idea... thoughts?

A table could make the list look nicer, but I'm not sure the version information in necessary at that point. But we should consider splitting the list by drivers that support the current e-g version and ones that are still depending on old versions like 0.4 or even older.

jamwaffles:
I did try a table but Github forces headers which looks weird.

> But we should consider splitting the list by drivers that support the current e-g version and ones that are still depending on old versions like 0.4 or even older.

That's quite a bit of maintenance burden on us to keep it fresh (unless it's automated, but then there are two problems) so I don't think that's sensible.

rfuest:
> I did try a table but Github forces headers which looks weird.

Did you only try with markdown or also with HTML?

> That's quite a bit of maintenance burden on us to keep it fresh (unless it's automated, but then there are two problems) so I don't think that's sensible.

I agree, but from a user perspective it's really annoying if you think there is support for your display and it turns out to be only available for an old unmaintained version of e-g.

jamwaffles:
> Did you only try with markdown or also with HTML?

Just markdown. Tbh I don't think it adds much value showing other crate versions in the readme so I'm going to remove it and make this PR just add the hub75 driver.

> from a user perspective it's really annoying if you think there is support for your display and it turns out to be only available for an old unmaintained version of e-g.

Absolutely, but that's not something I'm willing to maintain, at least at the moment.

rfuest:
> Absolutely, but that's not something I'm willing to maintain, at least at the moment.

Fair enough. How about adding a note that some drivers might not work with the latest e-g version and that this can be checked on crates.io?

jamwaffles:
Good idea. Done!

PR #243, opened 2020-01-13
Draw target error type

### LINKED ISSUE #241: Return type for DrawTarget methods
While I was looking at the latest changes in the `ssd1331` crate I noticed that it has a `Ssd1331::flush` method but doesn't implement `DrawTarget::flush`. `DrawTarget::flush`, like all operations in `DrawTarget` doesn't have a return type, so calling `Ssd1331::flush` would only be possible with panicing when an error occurs.

Should we add an associated type to the `DrawTarget` crate, that let's the driver implementations choose an appropriate return type for possibly failing draw operations?

jamwaffles:

## PR description

Gets `Drawable` and `DrawTarget` returning `Result`s allowing them to fail gracefully instead of panicking being the only option.

Was supposed to be a draft PR for now... oops

TODO

* [x] Figure out `Result` and types and stuff
* [x] Squelch all the result warnings
* [x] Should the default `Error` type just be `()` or something provided by e-g?

Relating to https://github.com/jamwaffles/embedded-graphics/issues/241#issuecomment-573866860, I can't see a way to have the default trait impl use a `Result` but allow other return types. There might be some magic I'm missing but it doesn't look possible.

Closes #241


rfuest:
When the never type gets stabilized, it can be used as a `Error` type. Until then the next best think is `core::convert::Infallible`, but this doesn't get rid of the warnings. Should we just accept this inconvenience for the user until `!` gets stabilized? Like it's done in the `digital::v2` traits in `embedded_hal`.

jamwaffles:
Good shout on `core::convert::Infallible`.

> Should we just accept this inconvenience for the user until ! gets stabilized?

I think so yeah. I'll add `.unwrap()` to examples, etc to keep the warning noise down for now. That's how a lot of the embedded-hal examples are so I'll just do the same.

rfuest:
> I think so yeah. I'll add `.unwrap()` to examples, etc to keep the warning noise down for now. That's how a lot of the embedded-hal examples are so I'll just do the same.

I've also seen `.ok()` used instead of `unwrap` which converts `Result` to `Option` and can never panic.

jamwaffles:
I've gone with `.unwrap()` as that's clearer that the method can panic (IMO anyway). It's also what all the embedded examples seem to do.

Could use some guidance on docs/examples. I added some crude words to `type Error` but I think it could use more. What would you suggest?

rfuest:
I didn't have time to review the code yet, but the more I think about this the more it seems to be a bad idea to have the same error type for the `draw...` and `flush` methods. This wouldn't make a difference now, because `unwrap` is necessary everywhere, but if the never type gets stabilized `draw...` could be infallible for drivers that uses a framebuffer and only possible errors caused by flushing would need to be handled.

Or is this a sign that `flush` doesn't belong in the `DrawTarget` trait at all, and should be left to the driver?

jamwaffles:
Hmm interesting. In porting the SSD1331 driver, I was going to attempt to start using the hardware acceleration abilities of the display. However, the hardware accelerated rectangle (for example) has no concept of stroke width, so hardware acceleration could only be used _sometimes_, when the stroke width is 1px. Other times, the framebuffer would be written to, or the iterator would be streamed to the display if the driver supports that. To me, this supports the notion of having a single error type for the entire trait.

> Or is this a sign that flush doesn't belong in the DrawTarget trait at all, and should be left to the driver?

Sounds like we're both not sure about this. Should it be removed for now? It's easier to add things back than take them away...

rfuest:
> To me, this supports the notion of having a single error type for the entire trait.

You could still use the same error type for this special use case, even if both types could be independently defined. IMO pure framebuffer or draw everything directly implementations are more common.

> Sounds like we're both not sure about this. Should it be removed for now? It's easier to add things back than take them away...

I'm leaning towards removing it, but I haven't done much work on the driver side, so you can probably judge this better. Maybe @ReeceStevens could also comment on this, because he added the `flush` method and might have a need for it or a special use case in mind.

jamwaffles:
Ok, let's go with two different return types. They can just be defined as the same type in driver code if that makes sense for the driver to do so.

> I'm leaning towards removing it

I am too.

If there's no other feedback, I'll add the second associated type and remove `flush()`. Any suggestions for naming these two associated types?

rfuest:
I thought that the second associated type would be for `flush` and if we remove `flush` we don't need that any more. What would you use the second type for?

jamwaffles:
`.clear()`, no? Like `.flush()` it can also talk to the hardware.

rfuest:
Hmm, interesting. When would `clear` behave different than the other drawing operations? For direct drawing drivers it would clear the framebuffer inside the display and for drivers with an framebuffer in RAM that would be cleared instead.

ReeceStevens:
Hey @jamwaffles @rfuest -- hope y'all are doing well. Regarding `flush()`, I believe I added that at the behest of @jamwaffles in my previous PR. The hardware I've been working with does not require an explicit flush. For what that's worth!

ReeceStevens:
I believe this was the discussion on the topic: https://github.com/jamwaffles/embedded-graphics/pull/187/files/f9b25deae7e82854affad67635116d5536d64f3d#diff-0afe50bcffb7fa162272291bdadcc14b

jamwaffles:
Ah, thanks for digging that up again. 

I just pushed a commit that removes it from `DrawTarget`. More upgrade pain but it seems to be a point of contention. As I said above, easier to add things to a new release than remove something for a breaking change.

@rfuest No more `flush()` and a single error type now. Should be good for a review when you have some time.

rfuest:
@ReeceStevens Thanks for the quick response and I hope you didn't mind that I did bug you with this.

> I just pushed a commit that removes it from `DrawTarget`. More upgrade pain but it seems to be a point of contention. As I said above, easier to add things to a new release than remove something for a breaking change.

OK, great. The `DrawTarget` trait was added after the latest stable release, so this shouldn't cause any more upgrade pain to users that don't use the bleeding edge alpha versions.

> @rfuest No more `flush()` and a single error type now. Should be good for a review when you have some time.

I didn't have time for a review yet, but I did take a first look while sitting on the train this afternoon. One thing I noticed are all the additional unwraps in the examples. IMO it would be more readable to return `Result` from `main` and use the `?` operator instead. This is also mentioned in the 2018 edition guilde: https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/question-mark-in-main-and-tests.html (this doesn't explicitly mention examples, but I think it applies here)

ReeceStevens:
@rfuest no worries at all, I don’t mind the occasional ping :)

jamwaffles:
>  One thing I noticed are all the additional unwraps in the examples. IMO it would be more readable to return Result from main and use the ? operator instead.

Very good shout. I left a couple of `unwrap()`s around when things like images are used in the examples - otherwise the `.draw()` calls would have to turn into `.draw().map_err(|_| ())?;` which is confusing IMO. I also moved some of the tests over to returning `Result`s.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+///     pub fn flush(&self) -> Result<(), ()> {
+///         self.iface.send_bytes(&self.framebuffer)
+///     }
+/// }
+///
+/// impl DrawTarget<Gray8> for ExampleDisplay {
+///     type Error = core::convert::Infallible;
+///
+///     /// Draw a `Pixel` that has a color defined as `Gray8`.
+///     fn draw_pixel(&mut self, pixel: Pixel<Gray8>) -> Result<(), Self::Error> {
```
rfuest: Doesn't handle negative coordinates correctly. But fixing this could wait until `TryFrom<Point> for (u32, u32)` is merged.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+/// let mut display = ExampleDisplay {
+///     framebuffer: [0; 4096],
+///     iface: SPI1,
+/// };
+///
+/// // Draw a circle centered around `(32, 32)` with a radius of `10` and a white stroke
+/// let circle = egcircle!(
+///     center = (32, 32),
+///     radius = 10,
+///     style = primitive_style!(stroke_color = Gray8::WHITE)
```
rfuest: `stroke_width` isn't set, so no stroke would be drawn.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+///
+/// // Update the display
+/// display.flush().expect("Failed to send data to display");
+/// # Ok::<(), core::convert::Infallible>(())
+/// ```
+///
+/// ## Hardware Acceleration
+///
+/// In addition to defining [`draw_pixel`], an implementation of [`DrawTarget`] can also provide
+/// alternative implementations of drawing methods for graphical primitives. Here is an example of
```
rfuest: Should we use "styled primitives" or "styled graphical primitives" here, because primitives aren't drawable on their own?

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+/// how a display with accelerated methods can implement [`DrawTarget`]:
```
rfuest: > ...accelerated methods...

IMO the display doesn't have methods. Maybe "support for hardware acceleration" or "hardware accelerated drawing operations" would work.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+///     pub fn fast_rectangle(&self, rect: &Styled<Rectangle, PrimitiveStyle<Gray8>>) {
+///         // Does some speedy drawing
+///     }
+/// }
+///
+/// impl DrawTarget<Gray8> for FastExampleDisplay {
+///     type Error = core::convert::Infallible;
+///
+///     /// Draw a `pixel` that has a color defined as `Gray8`
+///     fn draw_pixel(&mut self, pixel: Pixel<Gray8>) -> Result<(), Self::Error> {
```
rfuest: See comment for `draw_pixel` in previous example.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+///         self.framebuffer[index as usize] = color.luma();
+///
+///         Ok(())
+///     }
+///
+///     fn size(&self) -> Size {
+///         Size::new(64, 64)
+///     }
+///
+///     /// Use the accelerated method when drawing rectangles
```
rfuest: This needs to mention that this overrides the default implementation. IMO this should also be mentioned in the into to this example.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+/// let mut display = FastExampleDisplay {
+///     framebuffer: [0; 4096],
+///     iface: SPI1,
+/// };
+///
+/// // Draw a rectangle from (10, 20) to (30, 40) with a white stroke
+/// let rect = egrectangle!(
+///     top_left = (10, 20),
+///     bottom_right = (30, 40),
+///     style = primitive_style!(stroke_color = Gray8::WHITE)
```
rfuest: `stroke_width` is not set -> no stroke drawn

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+///
+/// [`Drawable`]: ../drawable/trait.Drawable.html
+/// [`Pixel`]: ../drawable/struct.Pixel.html
+/// [`draw_pixel`]: ./trait.DrawTarget.html#method.draw_pixel
+/// [`DrawTarget`]: ./trait.DrawTarget.html
+pub trait DrawTarget<C>
+where
+    C: PixelColor,
+{
+    /// Error type to return when a draw or display change operation fails.
```
rfuest: "display change" is very generic and I don't immediately know what this means. IMO all operations in this trait are drawing operations, so I would just remove that part.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+    ///
+    /// This error is returned when a drawing operation fails, or when an error updating/clearing
```
rfuest: Same as above. Now that `flush` is removed all operations manipulate a framebuffer and can be considered drawing operations, even `clear`.

> This error is returned when a drawing operation fails

The drawing operations can't fail. What I mean by that is that it's should not be possible to provide a set of invalid style parameters or coordinates that cause an error. But an error can occur while a draw operation updates the display.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// the display occurs. The examples in this crate use `core::convert::Infallible`, however a
+    /// more descriptive error type should be used with real hardware.
```
rfuest: This should explain why `core::convert::Infallible` is used and that drivers which use an internal framebuffer should probably do the same.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+    type Error;
+
+    /// Draws a pixel on the display.
+    ///
+    /// Note that some displays require a "flush" operation to actually write changes to the
+    /// framebuffer.
```
rfuest: This should be moved to trait-level docs, because it applies to all `draw_...` methods. And it should also mention that e-g don't provide such a method and that the driver docs probably provide more information.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+        Ok(())
+    }
+
+    /// Returns the dimensions of the `DrawTarget` in pixels.
+    fn size(&self) -> Size;
+
+    /// Clears the display with the supplied color.
+    ///
+    /// This default implementation should be replaced if the implementing driver provides an
+    /// accelerated clearing method.
```
rfuest: IMO replacing "should" with "can" sounds better. Something like:
```
/// This default implementation can be replaced by a driver specific implementation to provide
/// support for hardware accelerated clearing of the framebuffer.
```
This also applies to the `draw_...` methods.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+    fn clear(&mut self, color: C) -> Result<(), Self::Error>
+    where
+        Self: Sized,
+    {
+        primitives::Rectangle::new(Point::zero(), Point::zero() + self.size())
+            .into_styled(PrimitiveStyle::with_fill(color))
+            .draw(self)
+    }
+
+    /// Draws a line primitive.
```
rfuest: As mentioned above, in the "Hardware Acceleration" example, we should have a consistent style of referring to styled vs. raw primitives.
This also applies to the other `draw_...` methods.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,304 @@
...
+    ///
+    /// This default trait method should be overridden if a display provides hardware-accelerated
+    /// methods for drawing lines.
+    ///
+    /// # Caution
+    ///
+    /// This method should not be called directly from application code. It is used to define the
```
rfuest: This needs to mention `Styled`. And this also applied to the other `draw_...` methods.

embedded-graphics/src/drawable.rs
```diff
@@ -107,16 +66,60 @@ where
...
+/// [`Drawable`]: trait.Drawable.html
+/// [`DrawTarget`]: ../trait.DrawTarget.html
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct Pixel<C: PixelColor>(pub Point, pub C);
+
+impl<C> Drawable<C> for Pixel<C>
+where
+    C: PixelColor,
+{
+    fn draw<T: DrawTarget<C>>(self, display: &mut T) -> Result<(), T::Error> {
```
rfuest: The type parameter should always be called the same:
```suggestion
    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
```

embedded-graphics/src/fonts/text.rs
```diff
@@ -68,8 +68,8 @@ where
     C: PixelColor,
     F: Font + Copy,
 {
-    fn draw<T: DrawTarget<C>>(self, display: &mut T) {
-        display.draw_iter(self.into_iter());
+    fn draw<T: DrawTarget<C>>(self, display: &mut T) -> Result<(), T::Error> {
```
rfuest:
```suggestion
    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
```

embedded-graphics/src/image/image_bmp.rs
```diff
@@ -246,8 +250,8 @@ mod tests {
     }
 
     macro_rules! test_pattern {
-        ($color_type:ident, $image_data:expr) => {
-            let image: ImageBmp<$color_type> = ImageBmp::new($image_data).unwrap();
+        ($color_type:ident, $image_data:expr) => {{
```
rfuest: I would just keep using `unwrap` here. This way the tests below don't need to deal with `Result`.

embedded-graphics/src/image/image_bmp.rs
```diff
@@ -257,72 +261,82 @@ mod tests {
...
-            ImageBmp::new(include_bytes!("../../tests/colors_grey8.bmp")).unwrap();
+    fn colors_grey8() -> Result<(), ()> {
+        let image: ImageBmp<Gray8> = ImageBmp::new(include_bytes!("../../tests/colors_grey8.bmp"))?;
 
         assert_eq!(image.size(), Size::new(3, 1));
 
         let mut iter = image.into_iter();
 
-        let p = iter.next().unwrap();
+        let p = iter.next().ok_or(())?;
```
rfuest: Mapping all errors to single error type doesn't make sense in a test IMO, just use `unwrap`.

embedded-graphics/src/image/image_tga.rs
```diff
@@ -252,13 +253,15 @@ mod tests {
                 "WKWCMYBGR",
             ])
         );
+
+        Ok(())
     }
 
-    fn test_gray_tga(data: &[u8]) {
-        let image: ImageTga<Gray8> = ImageTga::new(data).unwrap();
+    fn test_gray_tga(data: &[u8]) -> Result<(), ()> {
```
rfuest: Don't return `Result` and `unwrap` instead.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -181,8 +182,8 @@ impl<'a, C: 'a> Drawable<C> for &Styled<Circle, PrimitiveStyle<C>>
 where
     C: PixelColor,
 {
-    fn draw<T: DrawTarget<C>>(self, display: &mut T) {
-        display.draw_circle(self);
+    fn draw<T: DrawTarget<C>>(self, display: &mut T) -> Result<(), T::Error> {
```
```suggestion
    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
```

embedded-graphics/src/primitives/line.rs
```diff
@@ -199,8 +200,8 @@ impl<'a, C: 'a> Drawable<C> for &Styled<Line, PrimitiveStyle<C>>
 where
     C: PixelColor,
 {
-    fn draw<T: DrawTarget<C>>(self, display: &mut T) {
-        display.draw_line(self);
+    fn draw<T: DrawTarget<C>>(self, display: &mut T) -> Result<(), T::Error> {
```
rfuest:
```suggestion
    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
```

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -211,8 +212,8 @@ impl<C> Drawable<C> for &Styled<Rectangle, PrimitiveStyle<C>>
 where
     C: PixelColor,
 {
-    fn draw<T: DrawTarget<C>>(self, display: &mut T) {
-        display.draw_rectangle(self);
+    fn draw<T: DrawTarget<C>>(self, display: &mut T) -> Result<(), T::Error> {
```
rfuest:
```suggestion
    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
```

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -349,8 +350,8 @@ impl<'a, C: 'a> Drawable<C> for &Styled<Triangle, PrimitiveStyle<C>>
 where
     C: PixelColor,
 {
-    fn draw<T: DrawTarget<C>>(self, display: &mut T) {
-        display.draw_triangle(self);
+    fn draw<T: DrawTarget<C>>(self, display: &mut T) -> Result<(), T::Error> {
```
```suggestion
    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
```

simulator/README.md
```diff
@@ -29,18 +29,23 @@ fn main() {
 
     Circle::new(Point::new(64, 64), 64)
         .into_styled(line_style)
-        .draw(&mut display);
+        .draw(&mut display)
+        .unwrap();
```
rfuest: These drawing operations should use the `?` operator.

rfuest:
> Very good shout. I left a couple of `unwrap()`s around when things like images are used in the examples - otherwise the `.draw()` calls would have to turn into `.draw().map_err(|_| ())?;` which is confusing IMO. I also moved some of the tests over to returning `Result`s.

I would probably left in even more `unwrap`s, at least in the tests. Changing one `unwrap` to having to define a return type, using the `?` operator and needing to add a return value to the function doesn't seem to be any easier or shorter IMO. But for the examples and test with multiple drawing ops this definitively make the code nicer.

jamwaffles:
Thanks for the review. I'll address the individual changes when I've got a mo.

> I would probably left in even more unwraps, at least in the tests.

Should I change the tests back? Seeing as the work's already done they're pretty alright. If they become a burden with returning a `Result`, that can be changed at some other point.

rfuest:
> Should I change the tests back?

Sometimes panicing directly can be beneficial in tests, because the backtrace directly points to the location where the error occurred, without having to use a debugger. But just leave them for now, it won't be worth the extra effort.

embedded-graphics/src/draw_target.rs
jamwaffles: Yeah let's wait for that to get merged. Much more elegant solution. Should probably get this PR rebased on a merged #245 before merging into master?

rfuest: Not sure when I have time to review #245. So if you want to merge this PR first just make these changes in #245.

embedded-graphics/src/draw_target.rs
jamwaffles: > drivers which use an internal framebuffer should probably do the same.

Doesn't that go against the idea of using a `Point` and converting that into positive-only coordinates? For example,  in the recent ssd1331 changes that add `.try_into().unwrap()`, that can be converted to return an error on failure as of this PR.

embedded-graphics/src/draw_target.rs
jamwaffles: Sounds good to me :+1:

embedded-graphics/src/draw_target.rs
rfuest: Drawing outside the display area needs to be a noop and not cause an error IMO. How would you handle drawing partially offscreen object otherwise?

jamwaffles: Ah true, I was thinking about converting `Point` to `Size`, but as long as the `Point` is filtered to be positive, the conversion from `i32` to `u32` can't fail.

jamwaffles: Hmm. Change pushed but I'm not that keen on it. Open to suggestions.

rfuest: Rough suggestion:

This error is returned if an error occurred during a drawing operation. This mainly applies
to drivers that need to communicate with the display for each drawing operation, where an
communication error can occur. For drivers that use an internal framebuffer drawing operations
can never fail and `core::covert::Infallible` can instead be used as `Error`.

jamwaffles: That's loads better than my attempt, thank you. I've included it with some very minor tweaks so should be good now.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 ### Changed
 
 - Expose `mock_display::MockDisplay` for use in tests, examples, etc in crates that pull embedded-graphics in as a dependency.
+- **(breaking)** #243 All methods on `DrawTarget` that handle drawing to/updating the display are now fallible and return `Result<(), Self::Error>` to allow for error handling in user code. This also affects `Drawable::draw()`, which now also returns a `Result<(), Self::Error>`.
```
rfuest: The description of the `Drawable::draw` return type is a bit misleading, because `Self::Error` doesn't refer to the `Drawable` but the `DrawTarget` in this case. This could just say that errors `Drawable::draw` propagates errors that occurred while drawing without explicitly mentioning the type.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,322 @@
...
+};
+
+/// Defines a display that can be used to render [`Drawable`] objects.
+///
+/// To to add embedded-graphics support to a display driver, `DrawTarget` must be implemented. Once
+/// a `DrawTarget` is defined, it can be used to render [`Drawable`]s. Note that any iterator over
+/// [`Pixel`]s can be drawn as [`Drawable`] is implemented for `Iterator<Item = Pixel<C:
+/// PixelColor>>`. See the [`Drawable`] trait documentation for more details.
+///
+/// `DrawTarget` also provides a way to define accelerated drawing methods for graphical primitives
```
rfuest: `provides a way` isn't very informative. I would mention that this can be done by overriding the default implementations with hardware accelerated ones.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,322 @@
...
+/// # Ok::<(), core::convert::Infallible>(())
+/// ```
+///
+/// ## Hardware Acceleration
+///
+/// In addition to defining [`draw_pixel`], an implementation of [`DrawTarget`] can also provide
+/// alternative implementations for hardware accelerated drawing operations. This example implements
+/// `DrawTarget` for a display that supports drawing hardware accelerated styled [`Rectangle`]s.
+///
+/// The `draw_rectangle()` method overrides the default implementation.
```
rfuest: IMO this sentence belongs together with the previous one in a single paragraph. The first one describes what is done and the second one how.

embedded-graphics/src/draw_target.rs
rfuest: You missed adding the `stroke_with` in this example, like in the one below.

embedded-graphics/src/draw_target.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// Draws a styled line primitive.
+    ///
+    /// This default trait method can be overridden if a display provides hardware-accelerated
+    /// methods for drawing lines.
+    ///
+    /// # Caution
+    ///
+    /// This method should not be called directly from application code. It is used to define the
+    /// internals of the [`draw`] method used for the [`Styled`] [`Line`] primitive. To draw a line,
+    /// call [`draw`] on a [`Line`] primitive object.
```
rfuest: Better, but not quite there yet. `draw` cannot be called on `Line`, because only `Styled<Line>` has a draw method.

jamwaffles:
Thanks for picking those up. Ready for another check!

jamwaffles:
Whew! Made it :sweat_smile:

PR #245, opened 2020-01-14
All the derives, add fallible point conversions

### LINKED ISSUE #238: Add as many derives as possible
Add as many derives to `embedded-graphics`, `tinytga` and `tinybmp` from [here](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits) as possible.

### LINKED ISSUE #242: Add Point to and from unsigned tuple conversions
Creating points form unsigned coordinates and converting points back into unsigned coordinates is a pretty common operation. Adding `TryFrom` implementations would make this easier:
```rust
impl TryFrom<Point> for (u32, u32) { ... }
impl TryFrom<(u32, u32)> for Point { ... }
```

jamwaffles:

## PR description

TODO

- [x] Add a shit load of derives. This was a find and replace with some manual fixes where some derives couldn't be used, so some might not make sense. Closes #238.
- [x] Add fallible conversions for `(u32, u32)` and `Point`. Closes #242.
- [x] Revert dependency fudging (due to alpha version numbers in `Cargo.toml`s I think)
- [x] ~Decide whether to add the following~ - not doing it. It can be added later if it's a common usage pattern.

  ```rust
  impl TryFrom<Size> for (i32, i32) { ... }
  impl TryFrom<(i32, i32)> for Size { ... }
  ```
- [x] https://github.com/jamwaffles/embedded-graphics/pull/243#discussion_r367593226

### DIFF
--- a/embedded-graphics/Cargo.toml
+++ b/embedded-graphics/Cargo.toml
@@ -26,8 +26,8 @@ circle-ci = { repository = "jamwaffles/embedded-graphics", branch = "master" }
 [dependencies]
 byteorder = { version = "1.3.2", default-features = false }
 nalgebra = { version = "0.19.0", optional = true, default-features = false }
-tinybmp = { version = "0.1.1", optional = true }
-tinytga = { version = "0.2.0", optional = true }
+tinybmp = { version = "0.1.2-alpha.0", optional = true }
+tinytga = { version = "0.2.1-alpha.0", optional = true }
 
 [features]
 default = []
--- a/embedded-graphics/Cargo.toml
+++ b/embedded-graphics/Cargo.toml
@@ -26,8 +26,8 @@ circle-ci = { repository = "jamwaffles/embedded-graphics", branch = "master" }
 [dependencies]
 byteorder = { version = "1.3.2", default-features = false }
 nalgebra = { version = "0.19.0", optional = true, default-features = false }
-tinybmp = { version = "0.1.2-alpha.0", optional = true }
-tinytga = { version = "0.2.1-alpha.0", optional = true }
+tinybmp = { version = "0.1.1", optional = true }
+tinytga = { version = "0.2.0", optional = true }
 
 [features]
 default = []
--- a/embedded-graphics/src/draw_target.rs
+++ b/embedded-graphics/src/draw_target.rs
@@ -70,7 +70,7 @@ use crate::{
 ///         let Pixel(coord, color) = pixel;
 ///
 ///         // Place an (x, y) pixel at the right index in the framebuffer. If the pixel coordinates
-///         // are out of bounds (negative or greater than `(63, 63)`), this operation will be a
+///         // are out of bounds (negative or greater than (63, 63)), this operation will be a
 ///         // noop.
 ///         if let Ok((x @ 0..=63, y @ 0..=63)) = coord.try_into() {
 ///             let index: u32 = x + y * 64;
@@ -107,12 +107,18 @@ use crate::{
 ///
 /// In addition to defining [`draw_pixel`], an implementation of [`DrawTarget`] can also provide
 /// alternative implementations for hardware accelerated drawing operations. This example implements
-/// `DrawTarget` for a display without a framebuffer that supports drawing hardware accelerated
-/// styled [`Rectangle`]s. The `draw_rectangle()` method overrides the default implementation.
+/// `DrawTarget` for a display without a framebuffer that supports hardware accelerated drawing of
+/// styled [`Rectangle`]s.
+///
+/// The default implementation of `draw_rectangle()` (as well as `draw_line()`, `draw_circle()` and
+/// `draw_triangle()`) defer to `draw_iter()` internally. In this example, the default
+/// implementation of `draw_rectangle()` is overridden to allow usage of accelerated draw commands
+/// specific to the targeted hardware.
 ///
 /// As this example doesn't use a framebuffer, a `flush()` method is not required. All draw
-/// operations are performed in "immediate mode" directly on the display. As drawing operations are
-/// now fallible due to hardware issues, a custom error type `CommError` is introduced.
+/// operations are performed in "immediate mode" directly on the display. As each drawing operation
+/// requires communication with the display that may fail, a custom error type `CommError` is
+/// introduced.
 ///
 /// ```rust
 /// # use embedded_graphics::prelude::*;
@@ -188,7 +194,7 @@ use crate::{
 ///     top_left = (10, 20),
 ///     bottom_right = (30, 40),
 ///     style = primitive_style!(stroke_color = Gray8::WHITE, stroke_width = 1)
-/// );
+/// ).draw(&mut display)?;
 ///
 /// // Draw a rectangle on the display using accelerated `draw_rectangle()` function
 /// # Ok::<(), CommError>(())
--- a/embedded-graphics/src/draw_target.rs
+++ b/embedded-graphics/src/draw_target.rs
@@ -30,6 +30,7 @@ use crate::{
 /// is sent to the display over a (simplified) SPI interface.
 ///
 /// ```rust
+/// use core::convert::TryInto;
 /// use embedded_graphics::{
 ///     drawable::Pixel,
 ///     egcircle,
@@ -38,7 +39,7 @@ use crate::{
 ///     prelude::*,
 ///     primitive_style, DrawTarget,
 /// };
-///
+/// #
 /// # struct SPI1;
 /// #
 /// # impl SPI1 {
@@ -47,7 +48,8 @@ use crate::{
 /// #     }
 /// # }
 /// #
-/// /// A fake display 64px x 64px where each pixel is stored as a single `u8`
+///
+/// /// A fake 64px x 64px display where each pixel is stored as a single `u8`
 /// struct ExampleDisplay {
 ///     framebuffer: [u8; 64 * 64],
 ///     iface: SPI1,
@@ -66,9 +68,14 @@ use crate::{
 ///     /// Draw a `Pixel` that has a color defined as `Gray8`.
 ///     fn draw_pixel(&mut self, pixel: Pixel<Gray8>) -> Result<(), Self::Error> {
 ///         let Pixel(coord, color) = pixel;
-///         // Place an (x, y) pixel at the right index in the framebuffer
-///         let index = coord.x + coord.y * 64;
-///         self.framebuffer[index as usize] = color.luma();
+///
+///         // Place an (x, y) pixel at the right index in the framebuffer. If the pixel coordinates
+///         // are out of bounds (negative or greater than `(63, 63)`), this operation will be a
+///         // noop.
+///         if let Ok((x @ 0..=63, y @ 0..=63)) = coord.try_into() {
+///             let index: u32 = x + y * 64;
+///             self.framebuffer[index as usize] = color.luma();
+///         }
 ///
 ///         Ok(())
 ///     }
@@ -100,8 +107,12 @@ use crate::{
 ///
 /// In addition to defining [`draw_pixel`], an implementation of [`DrawTarget`] can also provide
 /// alternative implementations for hardware accelerated drawing operations. This example implements
-/// `DrawTarget` for a display that supports drawing hardware accelerated styled [`Rectangle`]s. The
-/// `draw_rectangle()` method overrides the default implementation.
+/// `DrawTarget` for a display without a framebuffer that supports drawing hardware accelerated
+/// styled [`Rectangle`]s. The `draw_rectangle()` method overrides the default implementation.
+///
+/// As this example doesn't use a framebuffer, a `flush()` method is not required. All draw
+/// operations are performed in "immediate mode" directly on the display. As drawing operations are
+/// now fallible due to hardware issues, a custom error type `CommError` is introduced.
 ///
 /// ```rust
 /// # use embedded_graphics::prelude::*;
@@ -111,6 +122,7 @@ use crate::{
 /// # use embedded_graphics::pixelcolor::{Gray8, GrayColor};
 /// # use embedded_graphics::drawable::Pixel;
 /// # use embedded_graphics::style::{PrimitiveStyle, Styled};
+/// # use core::convert::TryFrom;
 /// #
 /// # struct SPI1;
 /// #
@@ -120,33 +132,35 @@ use crate::{
 /// #     }
 /// # }
 /// #
-/// /// A fake display 64px x 64px where each pixel is stored as a single `u8`
+/// /// SPI communication error
+/// #[derive(Debug)]
+/// struct CommError;
+///
+/// /// A fake display which uses hardware drawing commands instead of a framebuffer
 /// struct FastExampleDisplay {
-///     framebuffer: [u8; 64 * 64],
 ///     iface: SPI1,
 /// }
 ///
 /// impl FastExampleDisplay {
-///     /// Send buffer to the display
-///     pub fn flush(&self) -> Result<(), ()> {
-///         self.iface.send_bytes(&self.framebuffer)
-///     }
+///     /// Draw a rectangle using hardware accelerated commands
+///     pub fn fast_rectangle(
+///         &self,
+///         rect: &Styled<Rectangle, PrimitiveStyle<Gray8>>,
+///     ) -> Result<(), CommError> {
+///         // Send rectangle drawing commands to the display
 ///
-///     /// A HW-accelerated method for drawing rectangles
-///     pub fn fast_rectangle(&self, rect: &Styled<Rectangle, PrimitiveStyle<Gray8>>) {
-///         // Does some speedy drawing
+///         Ok(())
 ///     }
 /// }
 ///
 /// impl DrawTarget<Gray8> for FastExampleDisplay {
-///     type Error = core::convert::Infallible;
+///     type Error = CommError;
 ///
 ///     /// Draw a `pixel` that has a color defined as `Gray8`
 ///     fn draw_pixel(&mut self, pixel: Pixel<Gray8>) -> Result<(), Self::Error> {
 ///         let Pixel(coord, color) = pixel;
-///         // Place an (x, y) pixel at the right index in the framebuffer
-///         let index = coord.x + coord.y * 64;
-///         self.framebuffer[index as usize] = color.luma();
+///
+///         // Send commands directly to the display to set an individual pixel to the given color
 ///
 ///         Ok(())
 ///     }
@@ -157,21 +171,17 @@ use crate::{
 ///
 ///     /// Use the accelerated method when drawing rectangles
 ///     ///
-///     /// This method overrides the default implementation
+///     /// This method overrides the default implementation. If `fast_rectangle()` fails, the error
+///     /// will be propagated through this method.
 ///     fn draw_rectangle(
 ///         &mut self,
 ///         item: &Styled<Rectangle, PrimitiveStyle<Gray8>>,
 ///     ) -> Result<(), Self::Error> {
-///         self.fast_rectangle(item);
-///
-///         Ok(())
+///         self.fast_rectangle(item)
 ///     }
 /// }
 ///
-/// let mut display = FastExampleDisplay {
-///     framebuffer: [0; 4096],
-///     iface: SPI1,
-/// };
+/// let mut display = FastExampleDisplay { iface: SPI1 };
 ///
 /// // Draw a rectangle from (10, 20) to (30, 40) with a white stroke
 /// let rect = egrectangle!(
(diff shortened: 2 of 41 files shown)


embedded-graphics/Cargo.toml
```diff
@@ -26,8 +26,8 @@ circle-ci = { repository = "jamwaffles/embedded-graphics", branch = "master" }
 [dependencies]
 byteorder = { version = "1.3.2", default-features = false }
 nalgebra = { version = "0.19.0", optional = true, default-features = false }
-tinybmp = { version = "0.1.0", optional = true }
-tinytga = { version = "0.2.0", optional = true }
+tinybmp = { version = "0.1.1-alpha.0", optional = true }
+tinytga = { version = "0.2.1-alpha.0", optional = true }
```
jamwaffles: These need to be reverted before merge.

jamwaffles:
I took a stab at https://github.com/jamwaffles/embedded-graphics/pull/243#discussion_r367593226 but I'm not too keen on the examples. They do show what's going on, but they feel quite noisy with the added `DisplayError` enum. Any thoughts on a better approach (if necessary)?

rfuest:
I thought we agreed that drawing out of bounds shouldn't be an error, but a noop instead. Otherwise code like this:
```rust
fn draw_something(display: &mut Display) -> Result<(), Err> {
   SomeDrawable::new(...).draw(display)?;
   OtherDrawable::new(...).draw(display)
}
```
wouldn't even try to draw the second drawable if `SomeDrawable` overlaps the display area.

jamwaffles:
Right, yeah we did. I misunderstood what you meant by https://github.com/jamwaffles/embedded-graphics/pull/243#discussion_r367538339 then. In that case,

1. What should the error type of the examples be? `Infallible`? Keep the custom enum?
2. `draw_pixel()` could be changed to use `unwrap()`:

    ```rust
    let Pixel(coord, color) = pixel;
    
    if coord.x >= 0 && coord.x < 64 && coord.y >= 0 && coord.y < 64 {
        let x = usize::try_from(coord.x).unwrap();
        let y = usize::try_from(coord.y).unwrap();
    
        let index: usize = x + y * 64;
    
        self.framebuffer[index] = color.luma();
    }
  
    Ok(())
    ```
  
    or just use `as` because it'll never fail:
  
    ```rust
    let Pixel(coord, color) = pixel;
    
    if coord.x >= 0 && coord.x < 64 && coord.y >= 0 && coord.y < 64 {
    	  // Coercion from `i32` to `usize` here cannot fail due to bounds check above
        let index: usize = x as usize + y as usize * 64;
    
        self.framebuffer[index] = color.luma();
    }
  
    Ok(())
    ```
  
    I could also just revert the examples back to what's in `master`.

Thoughts?

rfuest:
>     1. What should the error type of the examples be? `Infallible`? Keep the custom enum?

Having at least one example that shows how and when to use an error type other that `Infallible` would be useful. My suggestion would be to change the first example back to using `Infallible` and remove the framebuffer from the hardware accelerated example. This way every draw operation in the second example would need to communicate with the display and could fail. The `Error` type could be something like `SpiError` in this case.

> I misunderstood what you meant by [#243 (comment)](https://github.com/jamwaffles/embedded-graphics/pull/243#discussion_r367538339) then.

What I meant is that if #242 is implemented `draw_pixel` could be implemented as:
```rust
let Pixel(coord, color) = pixel;

if let Some((x, y)) = coord.try_into() {
    if x < 64 && y < 64 {
        let index: u32 = x + y * 64;
        self.framebuffer[index as usize] = color.luma();
    }
}

Ok(())
```
or:
```rust
let Pixel(coord, color) = pixel;

if let Some((x @ 0..=63, y @ 0..=63)) = coord.try_into() {
    let index: u32 = x + y * 64;
    self.framebuffer[index as usize] = color.luma();
}

Ok(())
```

jamwaffles:
Ahh I see now. TIL the `x @ 0..=63` syntax exists - that's awesome! I've updated the `DrawTarget` examples. Might need some doc tweaks as usual 🙂

rfuest:
Is this ready for review? I'm asking because the review request is still open from before the last changes.

jamwaffles:
Yep! Sorry if that wasn’t clear in the comment.

embedded-graphics/src/geometry/point.rs
```diff
@@ -279,6 +299,41 @@ where
...
+        assert_eq!(tuple, (10u32, 20u32));
+    }
+
+    #[test]
+    fn convert_i32_max_to_u32_tuple() {
+        let p = Point::new(i32::max_value(), i32::max_value());
+
+        let tuple: (u32, u32) = p.try_into().unwrap();
+
+        // Literal value taken from [here](https://doc.rust-lang.org/std/primitive.i32.html#method.max_value)
```
rfuest: Because this isn't a doc comment there is no benefit of using the markdown syntax here IMO.

embedded-graphics/src/geometry/point.rs
```diff
@@ -279,6 +299,41 @@ where
...
+        assert_eq!(tuple, (2147483647, 2147483647));
+    }
+
+    #[test]
+    #[should_panic]
```
rfuest: This test and the next one shouldn't use `unwrap` together with the `should_panic` attribute, because this way the tests would also pass if, for example, `Point::new` panics. Explicitly testing the return value of `try_into` would remove this possibility for false positives.

embedded-graphics/src/geometry/point.rs
```diff
@@ -279,6 +299,41 @@ where
 mod tests {
     use super::*;
 
+    #[test]
+    fn convert_positive_to_u32_tuple() {
+        let p = Point::new(10, 20);
+
+        let tuple: (u32, u32) = p.try_into().unwrap();
+
+        assert_eq!(tuple, (10u32, 20u32));
```
rfuest: Nit: The type annotation isn't necessary.

embedded-graphics/src/geometry/point.rs
jamwaffles: I first tried an `assert_eq!(tuple, Err(core::num::TryFromIntError))` but `TryFromIntError` can't be constructed. I went with `assert!(tuple.is_err())` instead. Is there a better check?

jamwaffles:
Comments addressed 👍

embedded-graphics/src/geometry/point.rs
rfuest: > Is there a better check?

IMO `assert!(tuple.is_err())` is a good check in this case. There is only one possible error type which doesn't have any parameters, so that both asserts in your comment are equivalent.

rfuest:
> Comments addressed

These changes look good. Did you also notice the comment about `TryFrom` for arrays and slices?

embedded-graphics/src/geometry/point.rs
jamwaffles: Sweet 👌

jamwaffles:
> Did you also notice the comment about TryFrom for arrays and slices?

Oop, missed that. Do you mean `impl TryFrom<[u32; 2]> for Point` and `impl TryFrom<&[u32; 2]> for Point`?

rfuest:
> Do you mean `impl TryFrom<[u32; 2]> for Point` and `impl TryFrom<&[u32; 2]> for Point`?

Yes, that's what I meant. We did decide to keep the `From` array conversions while renaming `Coord` to `Point`. Adding the same kind of conversions for `u32` arrays would improve consistency.

I also just noticed that the opposite conversion `From<Point> for [i32; 2]` doesn't exist. Should we add these too (also for `TryFrom`)?

jamwaffles:
> Yes, that's what I meant.

Cool, added. I doubled up the tests at the bottom of `point.rs` instead of adding new array-specific ones as the file is already hella long.

> I also just noticed that the opposite conversion From<Point> for [i32; 2] doesn't exist. Should we add these too (also for TryFrom)?

Makes sense to me, so I've added it.

I gotta be honest all these conversion impls are frying my brain, so apologies in advance for any stupid mistakes/omissions. I'm starting to think about impl macros to make life easier, but not in this PR :)

rfuest:
Great, I'll review the changes together with the derives, but this will probably have to wait until next week.

> I gotta be honest all these conversion impls are frying my brain, so apologies in advance for any stupid mistakes/omissions. I'm starting to think about impl macros to make life easier, but not in this PR :)

No problem, I can understand that and I'm not looking forward to reviewing the code :wink:. Should you have some motivation left you could also add `impl From<Size> for [u32; 2]`.

jamwaffles:
> this will probably have to wait until next week.

No problem, thanks for the time spent on all these reviews. I should note that the derive additions were very much a copypasta all the derives everywhere, then remove anything that errored in the compiler. There might be some derives lying around that don't make sense.

> Should you have some motivation left you could also add impl From<Size> for [u32; 2].

Good idea, added.

jamwaffles:
I realise there's a lot of code to review @rfuest but if you have time could you look through this too?

rfuest:
> I realise there's a lot of code to review @rfuest but if you have time could you look through this too?

Yes, I hope that I'll find some spare time to do the code reviews on the weekend.

embedded-graphics/src/style/primitive_style.rs
```diff
@@ -14,7 +14,7 @@ use core::convert::TryFrom;
 /// [`PrimitiveStyleBuilder`]: ../style/struct.PrimitiveStyleBuilder.html
 /// [`non_exhaustive`]: https://blog.rust-lang.org/2019/12/19/Rust-1.40.0.html#[non_exhaustive]-structs,-enums,-and-variants
 /// [`primitive_style!`]: ../macro.primitive_style.html
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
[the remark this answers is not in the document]
jamwaffles: Replacing the manual impl with a derive leads to some odd results. I had to pepper some extra `+ Default` trait bounds on `C` and it got messy fast so I gave up. I've swapped the `new()` and `default()` method bodies as I think that makes a bit more sense - `PrimitiveStyle::new()` now calls `PrimitiveStyle::default()` instead of the other way round. Not a huge change either way tbh.

embedded-graphics/src/draw_target.rs
```diff
@@ -100,8 +107,12 @@ use crate::{
 ///
 /// In addition to defining [`draw_pixel`], an implementation of [`DrawTarget`] can also provide
 /// alternative implementations for hardware accelerated drawing operations. This example implements
-/// `DrawTarget` for a display that supports drawing hardware accelerated styled [`Rectangle`]s. The
-/// `draw_rectangle()` method overrides the default implementation.
+/// `DrawTarget` for a display without a framebuffer that supports drawing hardware accelerated
```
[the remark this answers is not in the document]
jamwaffles: Take a look at this whole example intro section. I've split it into 3 paragraphs which are hopefully a bit clearer.

embedded-graphics/src/draw_target.rs
```diff
@@ -100,8 +107,12 @@ use crate::{
...
+/// styled [`Rectangle`]s. The `draw_rectangle()` method overrides the default implementation.
```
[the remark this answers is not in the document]
jamwaffles: See above comment

embedded-graphics/src/draw_target.rs
```diff
@@ -100,8 +107,12 @@ use crate::{
...
+///
+/// As this example doesn't use a framebuffer, a `flush()` method is not required. All draw
+/// operations are performed in "immediate mode" directly on the display. As drawing operations are
+/// now fallible due to hardware issues, a custom error type `CommError` is introduced.
```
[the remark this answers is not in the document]
jamwaffles: See above comment

embedded-graphics/src/image/image_raw.rs
```diff
@@ -93,7 +93,7 @@ pub type ImageBE<'a, C> = Image<'a, C, BigEndian>;
 /// [`ImageLE`]: type.ImageLE.html
 /// [`PixelColor`]: ../pixelcolor/trait.PixelColor.html
 /// [`ByteOrder`]: ../pixelcolor/raw/trait.ByteOrder.html
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```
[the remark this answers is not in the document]
jamwaffles: Nah it doesn't, I've removed it.

embedded-graphics/src/pixelcolor/binary_color.rs
```diff
@@ -39,7 +39,7 @@ use crate::pixelcolor::{
 /// };
 /// assert_eq!(color, BinaryColor::On);
 /// ```
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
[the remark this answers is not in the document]
jamwaffles: Makes sense, added with a test.

jamwaffles:
Cheers for the review. I've removed nearly all the `Default` derives you picked out - the blanket copypasta was a bit overreaching and most of them certainly don't make sense. Ready for another pass when you're free.

embedded-graphics/src/style/primitive_style.rs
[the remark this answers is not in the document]
rfuest: Interesting, I thought deriving would work because `Default` is implemented for every `Option<T>` and not only `Option<T>`s where `T: Default`.
I only suggested this as a minor cleanup and I'm OK with leaving the code the way it is now.

embedded-graphics/src/draw_target.rs
```diff
@@ -100,8 +107,18 @@ use crate::{
 ///
 /// In addition to defining [`draw_pixel`], an implementation of [`DrawTarget`] can also provide
 /// alternative implementations for hardware accelerated drawing operations. This example implements
```
rfuest: This might be misleading, because "alternative implementations" could be interpreted as alternative implementations of `draw_pixel`.

embedded-graphics/src/draw_target.rs
```diff
@@ -100,8 +107,18 @@ use crate::{
...
-/// `DrawTarget` for a display that supports drawing hardware accelerated styled [`Rectangle`]s. The
-/// `draw_rectangle()` method overrides the default implementation.
+/// `DrawTarget` for a display without a framebuffer that supports hardware accelerated drawing of
+/// styled [`Rectangle`]s.
+///
+/// The default implementation of `draw_rectangle()` (as well as `draw_line()`, `draw_circle()` and
+/// `draw_triangle()`) defer to `draw_iter()` internally. In this example, the default
+/// implementation of `draw_rectangle()` is overridden to allow usage of accelerated draw commands
+/// specific to the targeted hardware.
```
rfuest: IMO the function references should be links and the `()` removed.
Explicitly listing all `draw_...` functions increases the risk of this comment to become out of sync with the code, when additional methods are added.

embedded-graphics/src/draw_target.rs
```diff
@@ -100,8 +107,18 @@ use crate::{
...
+///
+/// As this example doesn't use a framebuffer, a `flush()` method is not required. All draw
```
rfuest: The way the framebuffer is flushed and the name of the method isn't relevant here. The previous example uses the more generic `"flush" operation` which could also be used here.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -51,7 +51,7 @@ use crate::{
 ///     .draw(&mut display)?;
 /// # Ok::<(), core::convert::Infallible>(())
 /// ```
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```

embedded-graphics/src/primitives/circle.rs
```diff
@@ -122,7 +122,7 @@ impl Transform for Circle {
 }
 
 /// Pixel iterator for each pixel in the circle border
-#[derive(Debug, Copy, Clone)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```

embedded-graphics/src/primitives/line.rs
```diff
@@ -37,7 +37,7 @@ use crate::{
 ///     .draw(&mut display)?;
 /// # Ok::<(), core::convert::Infallible>(())
 /// ```
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
rfuest:
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```

embedded-graphics/src/primitives/line.rs
```diff
@@ -149,7 +149,7 @@ where
 }
 
 /// Pixel iterator for each pixel in the line
-#[derive(Debug, Clone, Copy)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```
rfuest:
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -43,7 +43,7 @@ use crate::{
 ///     .draw(&mut display)?;
 /// # Ok::<(), core::convert::Infallible>(())
 /// ```
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
rfuest:
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -136,7 +136,7 @@ where
 }
 
 /// Pixel iterator for each pixel in the rect border
-#[derive(Debug, Clone, Copy)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```
rfuest:
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -60,7 +60,7 @@ use core::borrow::Borrow;
 /// # assert_eq!(tri, Triangle::new(p1, p2, p3));
 /// # assert_eq!(tri_ref, Triangle::new(p1, p2, p3));
 /// ```
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -221,7 +221,7 @@ enum IterState {
 }
 
 /// Pixel iterator for each pixel in the triangle border
-#[derive(Debug, Clone, Copy)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
```
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```

jamwaffles:
Oh! From the _iterator_, not the primitive itself... you can blame Github's lack of context in review comments on the main thread for that lol. Fixed along with other suggestions.

embedded-graphics/src/draw_target.rs
jamwaffles: Yeah maybe it's a bit vague, but IMO reading on to the next paragraphs and looking at the code example clarifies things. Can we leave this as is for now? Docs are a simple addition to future releases. Non breaking changes from e-g? Who'd've thought???!!1/!?!?1/1/1 :grin:

embedded-graphics/src/draw_target.rs
jamwaffles: Good shout. I've tweaked it a bit. I think the intent is pretty clear now so let's leave this doc section as is.

embedded-graphics/src/draw_target.rs
jamwaffles: Tweaked as well, thanks for the suggestion

embedded-graphics/src/draw_target.rs
rfuest: > Can we leave this as is for now?

OK.

> Non breaking changes from e-g? Who'd've thought???!!1/!?!?1/1/1 grin

That's a novel idea, but I promise I'll find more ways to break the code in future releases :wink:

jamwaffles: > That's a novel idea

I'm gonna patent it

> I promise I'll find more ways to break the code in future releases 

Wahey but that's what semver pre-1.0.0 is for :+1:

rfuest:
LGTM :+1:

jamwaffles:
Beaut! Thank you!

PR #246, opened 2020-01-14
Public mock display

### LINKED ISSUE #239: Make MockDisplay public
`MockDisplay` is currently hidden in the docs, but used in the README. To make it easier to follow the examples `MockDisplay` should be public.

Making `MockDisplay` publicly available also allows to use it to write tests in other crates.

(originally posted in: https://github.com/jamwaffles/embedded-graphics/issues/236#issuecomment-573056612)

jamwaffles:

## PR description

Un-hides `mock_display` allowing it to be used in other crates for testing, examples, etc. I also moved the module docs to the `MockDisplay` trait and added a couple of examples.

Closes #239

### DIFF
--- a/embedded-graphics/src/mock_display.rs
+++ b/embedded-graphics/src/mock_display.rs
@@ -1,4 +1,110 @@
-//! Mock display for use in tests
+//! Mock display for use in tests.
+//!
+//! `MockDisplay` can be used to replace a real display in tests. The internal
+//! framebuffer wraps the color values in `Option` to be able to test which
+//! pixels were modified by drawing operations.
+//!
+//! The [`from_pattern`] method provides a convenient way of creating expected
+//! test results. The same patterns are used by the implementation of `Debug`
+//! and will be shown in failing tests.
+//!
+//! The display is internally capped at 64x64px.
+//!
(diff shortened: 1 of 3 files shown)


jamwaffles:
Hmm, I thought it would be more discoverable when directly linking to `MockDisplay`, but the docs homepage links only to the `mockdisplay` module so I'll move them back. I'll also add a `See module docs...` line to `MockDisplay` itself.

embedded-graphics/src/mock_display.rs
```diff
@@ -1,15 +1,14 @@
 //! Mock display for use in tests.
 //!
-//! [`MockDisplay`] can be used to replace a real display in tests. The internal
+//! `MockDisplay` can be used to replace a real display in tests. The internal
```
rfuest: This now needs to be a link again.

jamwaffles: Good catch. Should've checked the diff...

rfuest:
Or use `&#8942;`

jamwaffles:
I've left it as is. I think it's handy having the literal char in the doc comment as opposed to an escape code. The rendered docs show up fine as well so it's good enough :)

Issue #211, opened 2019-11-30
Remove non mutable reference Transform::translate method


rfuest:
This was discussed in #162 and #198 before.

If I remember correctly we decided to only implement `fn translate(&mut self, offset: Point)`, because the old behavior can still be accessed using `object.clone().translate(offset)`.

jamwaffles:
Hit a bit of a snag trying to get a PR going for this issue. The trait should look like this, right?

```rust
trait Transform {
    fn translate(&mut self, by: Point) -> &mut Self;
}
```

This breaks lines like this:

```rust
let line = Line::new(start, end).translate(Point::new(10, 10));

let line2 = Line::new(start, end)
    .translate(Point::new(10, 10))
    .into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 1));
```

If I change the trait to `fn translate(self, by: Point) -> Self` then the above _do_ work, but the following breaks (I think because this version copies `self` instead of moving it?):

```rust
let mut line = Line::new(start, end);

line.translate(Point::new(10, 10));
```

Any ideas? I really want to support both patterns. I can put what I have in a PR for a review if you like.

rfuest:
This problem was the reason I decided not to to try to implement this in PR #198 and I think there won't be a good solution to support both patterns with one method.

A `LineBuilder` would make the API cleaner, by separating owned vs. borrowed, but also more verbose:
```rust
let line =
    LineBuilder::new(start, end)
    .translate(Point::new(10, 10))
    .build();

 // equivalent to

let mut line = Line::new(start, end);
line.translate(Point::new(10, 10))
```
```rust
let line2 =
    LineBuilder::new(start, end)
    .translate(Point::new(10, 10))
    .build_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 1));

 // equivalent to

let mut line2 = Line::new(start, end);
line2.translate(Point::new(10, 10))
let styled_line2 = line2.into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 1));
```
```rust
trait Transform {
    fn translate(self, by: Point) -> Self;
}

// translate owns self for LineBuilder
impl Transform for LineBuilder {
    ...
}

// translate borrows self for Line
impl Transform for &mut Line {
    ...
}
 ```

rfuest:
After thinking about this some more, adding a builder which only exists to make primitives translatable seems weird.

[...]

PR #247, opened 2020-01-24
"Reverse" integration of tinytga and tinybmp crates

### LINKED ISSUE #229: tinytga and tinybmp dependencies
TGA and BMP support in e-g consists of code in the `tinytga` and `tinybmp` crates and additional code inside e-g itself. This way e-g needs to depend on every supported image format which isn't very scalable if other formats get added in the future.

I would suggest to move all format  depended code out of e-g into the `tinytga` and `tinybmp` crates and add an `embedded-graphics` feature to them. E-g would need to provide some interface that these external crates can implement. This would also serve as an template for other image format crates.

### LINKED ISSUE #230: Split changelog
Somewhat related to #229, `CHANGELOG.md` should probably be split up now. My current idea:

* Refactor `/CHANGELOG.md` to provide a list of relative links to each crate's `CHANGELOG.md`
* Add a `CHANGELOG.md` to the root of every crate which is a reduction of the current changelog in the project root

I initially considered having a single changelog and adding crate/version subheadings, however this gets messy pretty quickly I think when there are multiple crate releases in a different cadence. The benefit of having a single changelog is the discoverability of all changes for all crates. I don't think that's particularly useful IRL though.

jamwaffles:

## PR description

Closes #229 

TODO

* [x] Split changelogs - closes #230 
* [x] Tinybmp
* [x] Tinytga
* [x] Delete e-g builtins for TGA and BMP
* [x] Add documentation to e-g `lib.rs` with a list of external crates with support. e-g-sim and [embedded-picofont](https://crates.io/crates/embedded-picofont) should probably be added to this list too.
* [x] Move image tests in e-g to respective tiny* crates
* [x] Revert/fix Cargo.toml dependency imports that use `path`s
* [x] Add documentation for removed/new items


jamwaffles:
Very much a draft at the moment. Let's get tinybmp solid then I'll do the same to tinytga. Some points:

1. I've gone with a simple `((u32, u32), u32)` `Item` type for the BMP iterator. This is pretty confusing as to which `u32` represents what. Should this be a struct instead? Something that's basically e-g's `Pixel` like `BmpPixel(pub (u32, u32), u32)`? Something with named fields?
2. I've added an associated error type to `ImageFile`. Seems like a good idea to me to allow customisation of load errors for various crates. Thoughts?
3. `Transform` is implemented for `Bmp` using a kludgy feature flag to add an `offset` field to `Bmp` which I don't like. Is there a better way to do this?
4. Please ignore the failing builds for now lol

rfuest:
>  1. I've gone with a simple `((u32, u32), u32)` `Item` type for the BMP iterator. This is pretty confusing as to which `u32` represents what. Should this be a struct instead? Something that's basically e-g's `Pixel` like `BmpPixel(pub (u32, u32), u32)`? Something with named fields?

Not sure about this one, but named x, y and color fields could be the least confusing option.

Related to this, I thought about moving the `Point` and `Size` types into a separate small crate. This way crates that have optional e-g support could still use the same geometry types internally without a huge impact on compile times. This is outside the scope of this PR, but is this something you think could be useful?

> 2\. I've added an associated error type to `ImageFile`. Seems like a good idea to me to allow customisation of load errors for various crates. Thoughts?

That is a good idea!

> 3\. `Transform` is implemented for `Bmp` using a kludgy feature flag to add an `offset` field to `Bmp` which I don't like. Is there a better way to do this?

That's one of the reasons I wanted to split `Image` and `ImageData`. `Image` would be contained in e-g and handle the offset and implement `Drawable`, `Transform` and `Dimensions`. The external crates would only need to implement `ImageData` which provides a pixel iterator (starting at `(0, 0)`) and the size of the image.

jamwaffles:
> Not sure about this one, but named x, y and color fields could be the least confusing option.

Sounds good to me

> but is this something you think could be useful?

Hmm, yeah I think it could be but I don't see a pressing need for this right now. It can be added to e-g as a non breaking change I think, so let's tackle it when the need arises.

> That's one of the reasons I wanted to split Image and ImageData.

I like that. I'll prototype what that would look like in this PR and we can go from there?

jamwaffles:
Hey @rfuest, lots of rough edges still on this PR, but if you could take a look at the new [`ImageData`](https://github.com/jamwaffles/embedded-graphics/pull/247/files#diff-a7f43429d0e9bf7e40f73103411e6c2d) trait (and its [impl in tinybmp](https://github.com/jamwaffles/embedded-graphics/pull/247/files#diff-8b9653f0e084dd574a457db2e797935a) and [tinytga](https://github.com/jamwaffles/embedded-graphics/pull/247/files#diff-d58353461ba8491a12ed09f5f5e14ccb)) and tell me if I'm going in the right direction that would be great.

embedded-graphics/src/image/mod.rs
```diff
@@ -3,31 +3,151 @@
...
     /// Get the height in pixels of an image
     fn height(&self) -> u32;
+
+    /// Get an iterator over the pixels of the image
+    fn pixel_iter(&self) -> Self::PixelIterator;
+}
+
+/// TODO: Docs
+#[derive(Debug, Clone, Copy)]
+pub struct Image<I: Clone, C> {
```
jamwaffles: The `Clone` requirement is required for `Transform::translate_mut`. This can be fixed and removed in #211, but let's do it after this PR is merged.

jamwaffles: Also, I renamed `Image` in `image_raw.rs` to `ImageRaw` and reused the `Image` name here for the wrapping type. Now that I write this, this will make migration a pain so do we leave it, or put it in the changelog and just stomach it? I'm erring on the latter seeing as this release is full of so many breaking changes already.

tinybmp/src/lib.rs
```diff
@@ -184,9 +230,75 @@ impl<'a> Iterator for BmpIterator<'a> {
...
+
+    /// TODO: Docs
+    impl<'a, C> ImageData<C> for &'a Bmp<'a>
+    where
+        C: PixelColor + From<<C as PixelColor>::Raw>,
+    {
+        type PixelIterator = EgPixelIterator<'a, C>;
+
+        fn width(&self) -> u32 {
+            self.header.image_width
```
jamwaffles: If I call `self.width()` here instead I get this weird error, likely getting confused between `ImageData::width` and `Bmp::width`:

```
error[E0283]: type annotations needed: cannot resolve `_: embedded_graphics::pixelcolor::PixelColor`
```

I thought this might happen with the trait methods having the same name as methods on the image. I'm pretty ok just using `self.header.image_width` here, but it feels like a bit of a hack and will confuse anybody adding other image libraries in the future for sure. Any ideas?

jamwaffles: Ah never mind, figured it out!

```rust
fn width(&self) -> u32 {
    Bmp::width(&self)
}
```

I thought I tried this before but oh well, tried it again and it works now :ok_hand:

rfuest:
> Hey @rfuest, lots of rough edges still on this PR, but if you could take a look at the new [`ImageData`](https://github.com/jamwaffles/embedded-graphics/pull/247/files#diff-a7f43429d0e9bf7e40f73103411e6c2d) trait (and its [impl in tinybmp](https://github.com/jamwaffles/embedded-graphics/pull/247/files#diff-8b9653f0e084dd574a457db2e797935a) and [tinytga](https://github.com/jamwaffles/embedded-graphics/pull/247/files#diff-d58353461ba8491a12ed09f5f5e14ccb)) and tell me if I'm going in the right direction that would be great.

Yes, that looks like it's going in the right direction. I assume that `image_bmp.rs` and `image_tga.rs` are just leftovers and not used anymore?

> Also, I renamed `Image` in `image_raw.rs` to `ImageRaw` and reused the `Image` name here for the wrapping type. Now that I write this, this will make migration a pain so do we leave it, or put it in the changelog and just stomach it? I'm erring on the latter seeing as this release is full of so many breaking changes already.

IMO the raw image support should be changed to implement `ImageData` instead. This way all image types can be used in the same way by using the new `Image` struct. This would make it easier to use any image in a higher level crate like a UI libaray.

rfuest:
Sorry, I just noticed that `ImageRaw` implements `ImageData`. I was confused by the `offset` field in `ImageRaw`.

jamwaffles:
> Yes, that looks like it's going in the right direction. 

Wahey, nice

> I assume that image_bmp.rs and image_tga.rs are just leftovers and not used anymore?

Yeah I just pushed a cleanup commit :)

I'll go through the rest of the checkboxes in the top comment and spam you for review when I'm done :ok_hand:

rfuest:
> I'll go through the rest of the checkboxes in the top comment and spam you for review when I'm done ok_hand

Great! I'm looking forward to seeing this merged.

jamwaffles:
Ready for a first round of review. Sorry it's another long one! I need to get out of this habit...

I'm pretty happy with the design of it all, but the docs could use some work for sure.

embedded-graphics/src/image/image_raw.rs
```diff
@@ -233,17 +192,6 @@ where
...
-    BO: ByteOrder,
-    RawDataIter<'a, C::Raw, BO>: Iterator<Item = C::Raw>,
-{
-    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
-        display.draw_iter(self.into_iter())
-    }
-}
-
 #[derive(Debug)]
 pub struct ImageIterator<'a, C, BO>
```
rfuest:
```suggestion
 pub struct ImageRawIterator<'a, C, BO>
```

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,209 @@
...
+where
+    I: ImageData<C> + Clone,
+    C: PixelColor + From<<C as PixelColor>::Raw>,
+{
+    /// Create a new `Image` with a given [`ImageData`]
+    ///
+    /// The passed [`ImageData`] provides a source of pixel data from the original image.
+    ///
+    /// [`ImageData`]: ./trait.ImageData.html
+    pub fn new(image: I) -> Self {
```
rfuest: The `Text::new` function has an additional `position` parameter. IMO it would be a good idea to also add this parameter here.

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,209 @@
...
+        ImageIterator {
+            it: self.image.pixel_iter(),
+            image: self,
+        }
+    }
+}
+
+/// Pixel iterator over `Image` objects
+#[derive(Debug)]
+pub struct ImageIterator<'a, I: ImageData<C> + Clone, C: PixelColor + From<<C as PixelColor>::Raw>>
```
rfuest: The `where` syntax is used everywhere else, so this should do the same.

jamwaffles:
> Do you think that's a good idea?

I don't think it's a bad idea :grin: but could we add it later? Easier to add than remove :slightly_smiling_face:.

Review changes addressed, ready for another pass!

rfuest:
Currently `Image` owns the `ImageData`. Wouldn't it be better if `Image` would use a reference to the image data instead? This way the struct that implements `ImageData` could be stored in a constant.

jamwaffles:
> Wouldn't it be better if Image would use a reference to the image data instead?

Certainly, but I'm struggling with some lifetime stuff. I uploaded a breaking commit - could you take a look at it and see where my naive approach is going wrong?

I'm getting errors like

```
error[E0277]: the trait bound `tinybmp::Bmp<'_>: embedded_graphics::image::ImageData<_>` is not satisfied
  --> simulator/examples/bmp.rs:15:17
   |
15 |     let image = Image::new(&bmp, Point::zero());
   |                 ^^^^^^^^^^ the trait `embedded_graphics::image::ImageData<_>` is not implemented for `tinybmp::Bmp<'_>`
   |
   = help: the following implementations were found:
             <&'a tinybmp::Bmp<'a> as embedded_graphics::image::ImageData<C>>
   = note: required by `embedded_graphics::image::Image::<'a, I, C>::new`
```

And 

```
error[E0277]: the trait bound `image::image_raw::ImageRaw<'_, pixelcolor::binary_color::BinaryColor>: image::ImageData<_>` is not satisfied
   --> embedded-graphics/src/image/image_raw.rs:265:21
    |
265 |         let image = Image::new(&image, Point::zero()).translate(Point::new(-1, -1));
    |                     ^^^^^^^^^^ the trait `image::ImageData<_>` is not implemented for `image::image_raw::ImageRaw<'_, pixelcolor::binary_color::BinaryColor>`
    |
   ::: embedded-graphics/src/image/mod.rs:66:5
    |
66  |     pub fn new(image: &'a I, position: Point) -> Self {
    |     ------------------------------------------------- required by `image::Image::<'a, I, C>::new`
    |
    = help: the following implementations were found:
              <&'a image::image_raw::ImageRaw<'a, C, BO> as image::ImageData<C>>
```

jamwaffles:
On top of all the weird errors, how could we work around cases like this (simplified) example?

```rust
struct Image<'a> {
    buf: &'a [u8]
}

struct WrappedImage<'a> {
    image: &'a Image<'a>
}

fn load_image<'a>(buf: &'a [u8]) -> WrappedImage<'a> {
    let image = Image { buf };
    
    WrappedImage { image: &image }
}
```

This throws this familiar error:

```
error[E0515]: cannot return value referencing local variable `image`
  --> src/lib.rs:14:5
   |
14 |     WrappedImage { image: &image }
   |     ^^^^^^^^^^^^^^^^^^^^^^------^^
   |     |                     |
   |     |                     `image` is borrowed here
   |     returns a value referencing data owned by the current function
```

jamwaffles:
I'm starting to have second thoughts about keeping images as a ref inside `Image`

1. The current approach is to use `include_bytes!()` to store the image _data_ as a constant, and load it in at runtime. This incurs some parsing overhead, but parsing headers and stuff in an image is not a huge amount of work, at least for tinytga and tinybmp. Most processing is done in the iterator, so storing the image as a `const` wouldn't do much to help this I think.
2. My take on `Image` is that the `Bmp`, `Tga`, etc passed to it "becomes" an `Image` which is pretty ok IMO. Are there use cases where the programmer would want to reuse the original image later? We could add an `.into_inner()` or `.free()` or whatever method to facilitate this if required?
3. It breaks the pattern described in the above comment (https://github.com/jamwaffles/embedded-graphics/pull/247#issuecomment-582414744).

Are there concrete use cases you can think of where keeping the image as a ref is better than owning it?

rfuest:
Look at the changes in this commit for a possible solution: https://github.com/rfuest/embedded-graphics/commit/0dac951eb2859fdaf1c89a66329052cdbb69d243
It splits the `ImageData` trait into two traits to make the lifetimes easier to manage. `ImageData` is very similar to `IntoIter` and `ImageDimensions` is a basic trait without the need to consider lifetimes.

rfuest:
> This incurs some parsing overhead, but parsing headers and stuff in an image is not a huge amount of work, at least for tinytga and tinybmp.

But for other formats, especially compressed ones, this could be expensive in both memory and runtime.

> 2. My take on `Image` is that the `Bmp`, `Tga`, etc passed to it "becomes" an `Image` which is pretty ok IMO.

I think of `Image` like a small one time use object that is build to specify where a image is drawn.

> Are there use cases where the programmer would want to reuse the original image later?

If it is expensive to create the `ImageData` it would be beneficial to be able to reuse it when the same image is displayed multiple times, like an icon in a GUI.

jamwaffles:
> Look at the changes in this commit for a possible solution:

Thanks for figuring that out. Pretty gnarly! I've renamed the traits to `ImageDataIter` and `ImageData` which I think makes a bit more sense, especially when:

> ... `ImageData` is very similar to `IntoIter`

Both traits could use a doc review. Slept like crap last night so the words aren't coming to me. 

> I think of `Image` like a small one time use object that is build to specify where a image is drawn.

I see, that makes sense :)

jamwaffles:
Right, build is passing. Ready for another pass when you have a mo :)

rfuest:
The names for the traits were temporary, but I'm not sure I like the new names. If `ImageDataIter` is similar to `IntoIter` the name should reflect this IMO. Maybe `IntoPixelIter`?

The `ImageData` trait doesn't provide any functions that access the image data. I did choose `ImageDimenstions` because the trait only contains functions to access the dimensions at the moment. Additional functions, like a `get_pixel` function, could later be added as a separate traits. This way the external crates can implement different subsets of functions.

jamwaffles:
> Maybe `IntoPixelIter`?

Ah yeah, much better name.

> Additional functions, like a get_pixel function, could later be added as a separate traits.

Ok, I originally thought they could go in `ImageData` but I think we discussed this a while back to allow for varying levels of integration for different image formats.

jamwaffles:
Just reread my other comment which made no mention of this PR being ready for a re-review. When you have time :)

rfuest:
> Just reread my other comment which made no mention of this PR being ready for a re-review. When you have time :)

Just to let you know, I haven't forgotten this review yet but won't have time to review it until probably next week.

jamwaffles:
Cheers for the update. Let me know if there's anything I can do to accelerate the process. I might end up self-reviewing and self-merging this if I get bored so we might have to do some smaller (whoa, yay!) PRs to clean it up :)

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,262 @@
...
+    C: PixelColor + From<<C as PixelColor>::Raw>,
+{
+    /// Iterator over pixels in the image
+    type PixelIterator: Iterator<Item = Pixel<C>>;
+
+    /// Get an iterator over the pixels of the image
+    fn pixel_iter(self) -> Self::PixelIterator;
+}
 
+/// Image data trait
```
rfuest: There doesn't seem to be one default way to write these summaries for trait docs (At least the `std` docs uses different ways). I would like them to be somewhat consistent across e-g. This would by one variant:
```suggestion
/// A trait to get the dimensions of an image.
```
What is your preferred way of writing these?

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,262 @@
...
+///
+/// This trait provides an interface to common image metadata. It should be implemented along with
```
rfuest: IMO this trait will only ever be used for image dimensions and not any other metadata.

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,262 @@
...
     /// Get the width in pixels of an image
     fn width(&self) -> u32;
 
     /// Get the height in pixels of an image
     fn height(&self) -> u32;
 }
+
+/// Image support for embedded-graphics
+///
+/// The `Image` struct serves as a wrapper around other image types that provide pixel data decoded
```
rfuest: This should also mention why such an wrapper is needed.

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,262 @@
...
+    c: PhantomData<C>,
+}
+
+impl<'a, I, C> Image<'a, I, C>
+where
+    &'a I: IntoPixelIter<C>,
+    I: ImageDimensions,
+    C: PixelColor + From<<C as PixelColor>::Raw>,
+{
+    /// Create a new `Image` with a given [`IntoPixelIter`]
```
rfuest: The parameter doesn't only need to implement `IntoPixelIter` but also the `ImageDimensions` trait. So maybe the docs could use a more generic way of referring to the parameter by using `... the given image data` instead. An additional description of the required trait impls can then be added in the next paragraph.

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,262 @@
...
+
+impl<'a, 'b, I, C> fmt::Debug for ImageIterator<'a, 'b, I, C>
+where
+    &'b I: IntoPixelIter<C>,
+    C: PixelColor + From<<C as PixelColor>::Raw>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // The Debug derive didn't work anymore without additional trait bounds
+        // TODO: add fields
+        f.debug_struct("ImageIterator").finish()
```
rfuest: I did add this comment to my lifetime fix. But I think this should be fixed and the comment should be removed before this is merged.

embedded-graphics/src/image/mod.rs
```diff
@@ -1,33 +1,262 @@
...
+    }
+}
+
+impl<'a, 'b, I, C> Drawable<C> for &'a Image<'b, I, C>
+where
+    &'b I: IntoPixelIter<C>,
+    C: PixelColor + From<<C as PixelColor>::Raw>,
+{
+    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
+        display.draw_iter(
```
rfuest: I can't remember if I tried this when fixing the lifetimes, but can this use the `ImageIterator` defined below to reduce code duplication?

embedded-graphics/src/lib.rs
```diff
@@ -1,27 +1,35 @@
...
+//! You can also add your own objects by implementing [`Drawable`] on them. Additionally, all
+//! iterators over pixels (`Iterator<Item = Pixel<C>>`) have a default [`Drawable`] implementation
+//! already created.
 //!
-//! A core goal is to do the above without using any buffers; the crate should work without a
-//! dynamic memory allocator and without pre-allocating large chunks of memory. To achieve this, it
-//! takes an `Iterator` based approach, where pixel values and positions are calculated on the fly,
-//! with the minimum of saved state. This allows the consuming application to use far less RAM at
-//! little to no performance penalty.
+//! A core goal of embedded-graphics is to draw graphics without using any buffers; the crate should
```
rfuest: See notes above for README.md.

embedded-graphics/src/pixelcolor/raw/mod.rs
```diff
@@ -65,7 +65,7 @@
 //!
 //! fn main() {
 //!     // Create new image with RGBI colors.
-//!     let image: Image<RGBI> = Image::new(IMAGE_DATA, 2, 2);
+//!     let image: ImageRawLE<RGBI> = ImageRawLE::new(IMAGE_DATA, 2, 2);
```
rfuest: Why did you change this to `ImageRawLE` instead of `ImageRaw`? A `RawU4` image shouldn't be depended on the byte order.

simulator/examples/bmp.rs
```diff
@@ -2,22 +2,18 @@
...
 use embedded_graphics_simulator::{SimulatorDisplay, WindowBuilder};
+use tinybmp::Bmp;
 
 fn main() -> Result<(), core::convert::Infallible> {
-    let mut display = SimulatorDisplay::new(Size::new(128, 128));
+    let mut display: SimulatorDisplay<Rgb565> = SimulatorDisplay::new(Size::new(128, 128));
+
+    let bmp = Bmp::from_slice(include_bytes!("./rust-pride.bmp")).unwrap();
+
+    let image = Image::new(&bmp, Point::zero());
```
rfuest: IMO the examples should always annotate `Image` with a color type, because this defines the color format that's stored inside the image. The type inferred by the compiler equals the display color type, which might not be correct in all cases and can cause portability issues.

simulator/examples/tga.rs
```diff
@@ -2,22 +2,18 @@
...
 use embedded_graphics_simulator::{SimulatorDisplay, WindowBuilder};
+use tinytga::Tga;
 
 fn main() -> Result<(), core::convert::Infallible> {
-    let mut display = SimulatorDisplay::new(Size::new(128, 128));
+    let mut display: SimulatorDisplay<Rgb888> = SimulatorDisplay::new(Size::new(128, 128));
+
+    let tga = Tga::from_slice(include_bytes!("./rust-pride.tga")).unwrap();
+
+    let image = Image::new(&tga, Point::zero());
```
rfuest: See note above for the BMP example.

tinybmp/Cargo.toml
```diff
@@ -18,6 +18,20 @@ exclude = [
 [badges]
 circle-ci = { repository = "jamwaffles/embedded-graphics", branch = "master" }
 
+[[test]]
+name = "embedded_graphics"
+required-features = ["postgres", "tools"]
```
rfuest: I didn't know that e-g has database support :wink: Did you maybe copy this from somewhere else?

tinybmp/tests/coordinates.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[test]
+fn coordinates() {
+    let bmp = Bmp::from_slice(include_bytes!("./chessboard-8px-color-16bit.bmp"))
+        .expect("Failed to parse");
+
+    let pixels: Vec<_> = bmp.into_iter().map(|Pixel { x, y, .. }| (x, y)).collect();
+
+    // 8px x 8px image. Check that iterator returns all pixels in it
+    assert_eq!(pixels.len(), 8 * 8);
```
rfuest: This assertion is redundant.

tinytga/Cargo.toml
```diff
@@ -18,6 +18,20 @@ exclude = [
 [badges]
 circle-ci = { repository = "jamwaffles/embedded-graphics", branch = "master" }
 
+[[test]]
+name = "embedded_graphics"
+required-features = ["postgres", "tools"]
```
rfuest: See comment for `tinybmp/Cargo.toml`

embedded-graphics/CHANGELOG.md
```diff
@@ -0,0 +1,406 @@
+# Changelog
+
+Embedded Graphics is a `no_std` library for adding graphics features to display drivers. It aims to use the minimum amount of memory for builtin graphics objects by leveraging Rust's iterators to avoid large allocations. It targets embedded environments, but can run anywhere like a Raspberry Pi up to full desktop machines.
```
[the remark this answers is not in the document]
jamwaffles: It can't hurt? It's handy for people directly visiting the changelog to know what the project is about, similar to Rust announcement blog posts where there's a "Rust is a ..." preamble.

embedded-graphics/CHANGELOG.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Unreleased
+
+### Added
+
+- Added [hub75](https://crates.io/crates/hub75) driver to list of known supported drivers.
```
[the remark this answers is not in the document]
jamwaffles: Hm good point, I'll remove it. It's noise from the perspective of "what changed in e-g that broke my damn code".

embedded-graphics/src/pixelcolor/raw/mod.rs
jamwaffles: Good point. I think this was from a blanket find and replace.

simulator/examples/bmp.rs
jamwaffles: Fair, which one is better do you think?

```rust
let image: Image<_, Rgb565> = Image::new(&bmp, Point::zero());
```

or 

```rust
let image: Image<Bmp, Rgb565> = Image::new(&bmp, Point::zero());
```

I'll go with the latter for now but let me know if that's a stupid idea.

tinybmp/Cargo.toml
jamwaffles: Oh my dog what was that about. Turns out this was also hiding the broken tiny* tests, so I've fixed those up now as well.

embedded-graphics/src/image/mod.rs
jamwaffles: Yes it can, good catch :)

jamwaffles:
Thanks for the review and catching the dumb mistakes. Ready for another pass. I've changed the docs around a bit. They might not be perfect, but I'd prefer to get this PR over the line now so unless there's anything egregiously misleading in the docs or the functionality is incorrect can we fix-forward in new PRs? Good practice for smaller PRs as well lol.

embedded-graphics/CHANGELOG.md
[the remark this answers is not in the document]
rfuest: No, it can't really hurt, but it is one more description that needs to be maintained. And I think it's uncommon to include this in a changelog (but I might be wrong).

IMO an announcement blog post is quite different from a changelog. Do you think many people will read the changelog before knowing anything about the project?

embedded-graphics/CHANGELOG.md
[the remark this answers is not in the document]
rfuest: > "what changed in e-g that broke my damn code".

Answer for 0.6: A lot :stuck_out_tongue_winking_eye:

simulator/examples/bmp.rs
rfuest: Both variants are OK and I don't have a preference.

embedded-graphics/src/image/image_raw.rs
```diff
@@ -62,7 +62,7 @@ pub type ImageRawBE<'a, C> = ImageRaw<'a, C, BigEndian>;
 /// // raw data gets converted into.
 /// let raw_image: ImageRaw<BinaryColor> = ImageRaw::new(DATA, 12, 5);
 ///
-/// let image = Image::new(&raw_image, Point::zero());
+/// let image: Image<_, BinaryColor> = Image::new(&raw_image, Point::zero());
```
rfuest: I'm not sure about this one. There are 3 possible ways of handling type annotations for images with `ImageRaw` data:
1. Annotate both `Image` and `ImageRaw`
   This is a bit redundant.
2. Only annotate `ImageRaw`
   Not consistent with the BMP and TGA examples, but having the annotation at this point _feels_ right IMO, because it's the place where the format of the stored data is defined.
3. Only annotate `Image`
   Consistent with the BMP and TGA examples.

Any opinions on this or should we just leave it as it is now?

jamwaffles:
Sweeeeeeet!

Also...

image

wowser

PR #251, opened 2020-02-08
Liberate triangle primitive from unwrap()

therealprof:
Signed-off-by: Daniel Egger <daniel@eggers-club.de>

Hi! Thank you for helping out with Embedded Graphics development! Please:

- [ ] Check that you've added passing tests and documentation
- [ ] Add a simulator example(s) where applicable
- [x] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc)
- [x] Run `rustfmt` on the project
- [x] Run `./build.sh` (Linux/macOS only) and make sure it passes. If you use Windows, check that CI passes once you've opened the PR.

## PR description

Building up an array, iterator, iterating over it with the min/max function and calling unwrap() on the result just to find the minimum or maxmim of three values seemed like overkill so I replace them with two comparisons each instead.

PR #253, opened 2020-02-08
Change PR template to mention public API

jamwaffles:
From https://github.com/jamwaffles/embedded-graphics/pull/251#pullrequestreview-355543845, the changelog from now on should probably just have any publicly changing items added to it.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@ Hi! Thank you for helping out with Embedded Graphics development! Please:
 
 - [ ] Check that you've added passing tests and documentation
 - [ ] Add a simulator example(s) where applicable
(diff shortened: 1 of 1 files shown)


rfuest:
"fixing the public API" might not be clear enough. Bug fixes that don't change the API at all, but fix bugs that can be triggered using the public API should also be mentioned in the changelog IMO.

jamwaffles:
How about 

> [ ] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) if you're adding/fixing/changing something that may affect the **public** API

?

rfuest:
That's pretty long. How about
> [ ] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) if your changes affect the **public API**

?

jamwaffles:
Much better, let's see how that goes for future contributions :)

PR #255, opened 2020-02-10
0.5 -> 0.6 migration guide

### LINKED ISSUE #137: Add a migration guide from 0.5.x to 0.6.0
There are pretty fundamental changes to how colours work in 0.6.0, so there should be an easy to follow guide to help people upgrade.

Probably want to look through the changelog instead of the list here.

Some more notes to go on top of the changelog:

* `PrimitiveStyle::with_fill()` and `PrimitiveStyle::with_stroke()` only create a fill/stroke respectively. To create a style with both, use the struct construction method instead (`PrimitiveStyle { /* ... */ }`) - #198.

jamwaffles:

- [x] ~Check that you've added passing tests and documentation~
- [x] ~Add a simulator example(s) where applicable~
- [x] ~Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc)~
- [x] ~Run `rustfmt` on the project~
- [x] ~Run `./build.sh` (Linux/macOS only) and make sure it passes. If you use Windows, check that CI passes once you've opened the PR.~

## PR description

TODO

* [x] Add section/link to README(s)
* [x] Finish the migration guide

Closes #137

### DIFF
--- a/embedded-graphics/MIGRATING-0.5-0.6.md
+++ b/embedded-graphics/MIGRATING-0.5-0.6.md
@@ -1,0 +1,170 @@
+# Migrating from 0.5.x to 0.6.0
+
+## Pixel colours
+
+TODO for both driver authors and library consumers
+
+## For driver authors
+
+Adding embedded-graphics support to a display driver has changed somewhat. The `Drawing` trait has been replaced by the `DrawTarget` trait which provides a slightly different interface.
+
+An associated error type is also added to `DrawTarget` to allow for better error handling than a `panic!()`.
+
+```diff
+// TODO: Mad example
+```
+
+## For library consumers
+
+### General drawing operations
+
+Drawing operations are "reversed" in 0.6.0. Instead of calling `display.draw(thing)`, call `thing.draw(&mut display)`:
+
+```diff
+- let rect = Rectangle::new(Coord::new(50, 20), Coord::new(60, 35))
+-     .stroke(Some(5u8))
+-     .stroke_width(3)
+-     .fill(Some(10u8));
+-
+- display.draw(rect);
+
++ let style = PrimitiveStyleBuilder::new()
++     .stroke_color(Rgb565::RED)
++     .stroke_width(3)
++     .fill_color(Rgb565::GREEN)
++     .build();
++
++ Rectangle::new(Point::new(50, 20), Point::new(60, 35))
++     .into_styled(style)
++     .draw(&mut display)?;
+```
+
+Drawing operations are now fallible, with `.draw()` calls returning a `Result`. This allows for error handling if an error occurs during a drawing operation.
+
+### Coordinates and positioning
+
+The `Coord` and `UnsignedCoord` have been renamed to `Point` and `Size` respectively. Both items are no longer tuple structs, but structs with the named fields `x` and `y`. `Point`s can store negative coordinates, whereas `Size`s must be positive up to `u32::MAX_VALUE` as with `Coord` and `UnsignedCoord` before.
+
+The `icoord` and `ucoord` macros have been removed. Instead, use `Point::new(x, y)` and `Size::new(x, y)` respectively.
+
+`Point::zero()` and `Size::zero()` were also added to get zero-filled coordinates.
+
+```diff
+- let c = Coord::new(-10, 20);
++ let c = Point::new(-10, 20);
+
+- println!("X: {}, Y: {}", c.0, c.1);
++ println!("X: {}, Y: {}", c.x, c.y);
+
+- let p = UnsignedCoord::new(5, 15);
++ let p = Size::new(5, 15);
+
+- println!("X: {}, Y: {}", p.0, p.1);
++ println!("X: {}, Y: {}", p.x, p.y);
+```
+
+### Text
+
+Text is now drawn with the `Text` struct. A `TextStyle` must be provided which selects which font to draw the text with.
+
+```diff
+- use embedded_graphics::{
+-     prelude::*,
+-     fonts::Font6x8,
+- };
+-
+- display.draw(
+-     Font6x8::render_str("Hello Rust!")
+-         .translate(Coord::new(20, 30))
+-         .fill(Some(1u8))
+-         .stroke(Some(0u8)),
+- );
+
++ use embedded_graphics::{
++     fonts::{Font6x8, Text},
++     pixelcolor::BinaryColor,
++     prelude::*,
++     style::TextStyleBuilder,
++ };
++
++ // Create a new text style
++ let style = TextStyleBuilder::new(Font6x8)
++     .text_color(BinaryColor::On)
++     .background_color(BinaryColor::Off)
++     .build();
++
++ // Create a text at position (20, 30) and draw it using style defined above
++ Text::new("Hello Rust!", Point::new(20, 30))
++     .into_styled(style)
++     .draw(&mut display)?;
+```
+
+Macro usage has also changed:
+
+```diff
+- use embedded_graphics::{prelude::*, text_6x8};
+-
+- display.draw(
+-     text_6x8!(
+-         "Hello Rust!",
+-         stroke = Some(1u8.into()),
+-         fill = Some(0u8.into())
+-     )
+-     .translate(Coord::new(20, 30))
+- );
+
++ use embedded_graphics::{egtext, text_style, fonts::Font6x8, pixelcolor::BinaryColor, prelude::*};
++
++ egtext!(
++     text = "Hello Rust!",
++     top_left = (20, 30),
++     style = text_style!(
++         font = Font6x8,
++         text_color = BinaryColor::On,
++         background_color = BinaryColor::Off,
++     )
++ )
++ .draw(&mut display)?;
+```
+
+### Primitives
+
+Primitives changes:
+
+- Note: The `Triangle` primitive was added. If you were using 3 separate lines, this primitive is suggested instead.
+
+- Primitive styles now default to a stroke width of `0` instead of `1`. If nothing is drawn to the display, check that you set the `stroke_width` in `PrimitiveStyleBuilder`.
+
+### Images
+
+The `ImageBMP` and `ImageTGA` built-in image types are removed in favour of using the new `graphics` features of the [tinybmp](https://crates.io/crates/tinybmp) and [tinytga](https://crates.io/crates/tinytga) crates. These must be wrapped in the `Image` struct for use with embedded-graphics.
+
+First, change the imports in `Cargo.toml`:
+
+```diff
+[dependencies]
+- embedded-graphics = { version = "0.5.1", features = [ "tga", "bmp" ] }
+
++ embedded-graphics = "0.6.0"
++ tinytga = { version = "0.3.0", features = [ "graphics" ] }
++ tinybmp = { version = "0.2.0", features = [ "graphics" ] }
+```
+
+Then any image code:
+
+```diff
+- use embedded_graphics::{prelude::*, image::ImageBmp};
+-
+- let image = ImageBmp::new(include_bytes!("../../../assets/patch_16bpp.bmp")).unwrap();
+-
+- display.draw(&image);
+
++ use embedded_graphics::{prelude::*, image::Image, geometry::Point}
++ use tinybmp::Bmp;
++
++ let bmp = Bmp::from_slice(include_bytes!("../../../assets/patch_16bpp.bmp")).unwrap();
++
++ let image = Image::new(&bmp, Point::zero());
++
++ image.draw(&mut display)?;
+```
--- a/embedded-graphics/MIGRATING-0.5-0.6.md
+++ b/embedded-graphics/MIGRATING-0.5-0.6.md
@@ -377,23 +377,24 @@ If the display's pixel order is reversed (BGR instead of RGB) and cannot be chan
 
 ## For font authors
 
-- The `embedded_graphics::fonts::font_builder` module along with its exported `FontBuilderConf` and `FontBuilder` is removed. Now only `Font` needs to be implemented.
-- The `Font` trait (used to be `FontBuilderConf`) implementation has changed slightly, replacing `CHART_WIDTH` and `CHAR_HEIGHT` with a single `CHARACTER_SIZE` constant, using the `Size` struct:
+The `FontBuilderConf` trait is renamed to `Font`. This is the only trait impl that is required for embedded-graphics integration. `FontBuilder` is now removed, along with the `embedded_graphics::fonts::font_builder` module.
 
-  ```diff
-  - use embedded_graphics::fonts::font_builder::{FontBuilder, FontBuilderConf};
-  + use embedded_graphics::fonts::Font;
+The `Font` trait implementation has changed slightly, replacing `CHART_WIDTH` and `CHAR_HEIGHT` with a single `CHARACTER_SIZE` constant, using the `Size` struct:
 
-  + use embedded_graphics::geometry::Size;
+```diff
+- use embedded_graphics::fonts::font_builder::{FontBuilder, FontBuilderConf};
++ use embedded_graphics::fonts::Font;
 
-    #[derive(Debug, Copy, Clone)]
-    pub enum MyFont {}
++ use embedded_graphics::geometry::Size;
 
-    impl Font for MyFont {
-        const FONT_IMAGE: &'static [u8] = include_bytes!("../data/my_font.raw");
-  -     const CHAR_WIDTH: u32 = 12;
-  -     const CHAR_HEIGHT: u32 = 22;
-  +     const CHARACTER_SIZE: Size = Size::new(12, 22);
-        const FONT_IMAGE_WIDTH: u32 = 480;
-    }
-  ```
+  #[derive(Debug, Copy, Clone)]
+  pub enum MyFont {}
+
+  impl Font for MyFont {
+      const FONT_IMAGE: &'static [u8] = include_bytes!("../data/my_font.raw");
+-     const CHAR_WIDTH: u32 = 12;
+-     const CHAR_HEIGHT: u32 = 22;
++     const CHARACTER_SIZE: Size = Size::new(12, 22);
+      const FONT_IMAGE_WIDTH: u32 = 480;
+  }
+```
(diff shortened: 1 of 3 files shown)


embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,248 @@
+# Migrating from embedded-graphics 0.5.x to 0.6.0
+
+## Pixel colours
+
+TODO for both driver authors and library consumers
```
jamwaffles: @rfuest As you did most of the colour changes, do you have any suggestions for what to add here? A bulleted list is fine - I'd just like a starting point to work from.

jamwaffles:
I've marked this as ready for review but it's incomplete. It would be great to get some feedback on anything that's unclear or missing.

embedded-graphics/MIGRATING-0.5-0.6.md
rfuest: Some initial suggestions:

* List of color types.
* Primitive types (like `u8` or `u16`) can't directly be used as colors anymore.
  The primitive types didn't allow to specify the exact color format on a type level, which could lead to errors when colors weren't used in different places without being converted explicitly (e.g. when displaying an image).
* Added distinction between the pixel color type and the underlying storage.
  Multiple color types can now share the same storage size, but still have distinct types.
* Added some color conversions:
  * Any RgbColor <-> Any other RgbColor
  * Any GrayColor -> Any RgbColor
  * BinaryColor -> Any RgbColor or GrayColor

For display driver authors:
* Use `BinaryColor` when implementing a driver for a display with 2 states.
* When implementing a color display driver `DrawTarget` can be implemented to automatically convert colors:
  ```rust
  impl<C> DrawTarget<C> for SomeRgb565Display
  where
    C: PixelColor + Into<Rgb565>,
  {
  
      fn draw_pixel(&mut self, pixel: Pixel<C>) -> Result<(), Self::Error> {
          let Pixel(point, color) = pixel;
  
          self.set_pixel(point.x, point.y, color.into());
  
          Ok(())
      }
  
      fn size(&self) -> Size {
          self.size
      }
  }
  ```
  This way, for example, a `Rgb565` or `Rgb888` image can be drawn the same way, without any explicit color conversion.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,248 @@
...
++ Rectangle::new(Point::new(50, 20), Point::new(60, 35))
++     .into_styled(style)
++     .draw(&mut display)?;
+```
+
+Drawing operations are now fallible, with `.draw()` calls returning a `Result`. This allows for error handling if an error occurs during a drawing operation. The error type of this `Result` is dependent on the associated `Error` type as defined by the display driver.
+
+### Coordinates and positioning
+
+The `Coord` and `UnsignedCoord` have been renamed to `Point` and `Size` respectively. Both items are no longer tuple structs, but structs with the named fields `x` and `y` for `Point` and `width` and `height` for `Size`. `Point`s can store negative coordinates, whereas `Size`s must be positive up to `u32::MAX_VALUE`.
```
rfuest: IMO they weren't just renamed but replaced by the new types and gained some semantic meaning.

An important aspect that is missing here is the change to the `Pixel` type. It used `UnsignedCoord` in 0.5 and now uses the signed `Point` instead.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,248 @@
...
+- let p = UnsignedCoord::new(5, 15);
++ let p = Size::new(5, 15);
+
+- println!("X: {}, Y: {}", p.0, p.1);
++ println!("X: {}, Y: {}", p.width, p.height);
+```
+
+### Text
+
+Text is now drawn with the `Text` struct. A `TextStyle` must be provided which selects which font to draw the text with.
```
rfuest: `TextStyle` also selects the colors. We should also mention that `TextStyle` no longer uses stoke and fill colors, but text and background.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,248 @@
...
++     .build();
++
++ // Create a text at position (20, 30) and draw it using style defined above
++ Text::new("Hello Rust!", Point::new(20, 30))
++     .into_styled(style)
++     .draw(&mut display)?;
+```
+
+- A new `Text` struct is introduced. Instead of `YourFont::render_str("text")`, use `Text::new("text")`.
+- Text must be given a `TextStyle` for it to be drawn on a display. Create a style with `TextStyleBuilder` and add it with `.into_styled(style)`.
```
rfuest: Simple text styles can also be created using `TextStyle::new(font, text_color)`

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+- All built in `text_*!()` macros are removed and replaced with the `egtext!()` macro.
+- `egtext!()` should be coupled with the `text_style!()` macro to create styled texts with a chosen font.
+
+### Primitives
+
+Primitives changes:
+
+- Note: The `Triangle` primitive was added. If you were using 3 separate lines, this primitive is suggested instead.
```
rfuest: The `Triangle` primitive was already added in `0.4.8`.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,248 @@
...
+
+- Primitive styles now default to a stroke width of `0` instead of `1`. If nothing is drawn to the display, check that you set the `stroke_width` in `PrimitiveStyleBuilder`.
```
rfuest: We need to mention that `PrimitiveStyle` is a replacement for the old `Style` and that drawing primitives now requires creating a `Styled`.

jamwaffles:
@rfuest This is ready for another go round. Soz for the delay in picking up your comments.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,422 @@
...
+  ```rust
+  impl DrawTarget<Rgb565> for DisplayDriver {
+      // ...
+  }
+  ```
+
+- The `draw()` method is replaced by `draw_pixel()`. This method should handle setting of individual pixels on the display. How it does this is at the descretion of the display driver (pixel buffer, immediate mode, etc).
+- A new `size()` method is now a required item. This should return the width and height of the display as a `Size`.
+- The implementation of `DrawTarget` must now provide an associated `Error` type. In the above example `core::convert::Infallible` is used, however a better error type should be used to communicate hardware failures, etc to the user.
+- Any pixels that are offscreen (negative coordinates or greater than display dimensions) should result in a noop. In the example above, `self.set_pixel()` (defined by the SSD1306 driver) will not attempt to set any pixels beyond the positive screen limits.
```
rfuest:
```suggestion
- Any pixels drawn to offscreen coordinates need to be ignored by `DrawTarget` implementations and not result in panics or any modification of the display content. In the example above, `self.set_pixel()` (defined by the SSD1306 driver) will not attempt to set any pixels beyond the positive screen limits.
```

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,422 @@
...
+- `draw_pixel()` now returns `Result<(), Self::Error>` to account for driver error handling.
```
rfuest: The other `draw_...` methods and the reason for their existence  should also be mentioned.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,422 @@
...
+
+Below are some common use cases to help choose the right pixel color for a given display module. The full list of pixel colors is described in [the table above](#pixel-colors).
+
+#### `BinaryColor`
+
+If the display only supports two states, use `BinaryColor`. This is applicable to monochrome OLED displays like the SSD1306 or SH1106, character/bitmap LCDs and even LED matrices.
+
+#### `Gray2`
+
+Use for displays that can represent 2 or 3 color states. For example, tricolor epaper displays that can show white, black or red like the SSD1675 should use this type.
```
rfuest: I don't think this is correct. `Gray2` is a 2 bpp color, so there are 4 color states and this shouldn't be used for 2 color states. Tricolor epaper display could also use an eum, as shown in the example: https://github.com/jamwaffles/embedded-graphics/blob/272346369e312068cc504c079f0b4ce7ffa32a8a/embedded-graphics/src/pixelcolor/mod.rs#L7-L75

Maybe `Gray4` is a better example, because it could be used for `SSD132x` displays.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,422 @@
...
+
+#### `Rgb888`
+
+24 bit color, most useful in the embedded graphics simulator, or for high color display modules.
+
+## For font authors
+
+The `FontBuilderConf` trait is renamed to `Font`. This is the only trait impl that is required for embedded-graphics integration. `FontBuilder` is now removed, along with the `embedded_graphics::fonts::font_builder` module.
+
+The `Font` trait implementation has changed slightly, replacing `CHART_WIDTH` and `CHAR_HEIGHT` with a single `CHARACTER_SIZE` constant, using the `Size` struct:
```
rfuest:
```suggestion
The `Font` trait implementation has changed slightly, replacing `CHAR_WIDTH` and `CHAR_HEIGHT` with a single `CHARACTER_SIZE` constant, using the `Size` struct:
```
`CHARACTER_SPACING` and `VARIABLE_WIDTH` were also added.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,422 @@
...
+use embedded_graphics::{
+    pixelcolor::{BinaryColor, Rgb565, Rgb888},
+    prelude::*,
+};
+
+// Binary color (off/on)
+let on = BinaryColor::On;
+
+// Red with a small amount of green creates a deep orange colour
+let rust = Rgb565::new(0xff, 0x07, 0x00);
```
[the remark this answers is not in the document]
jamwaffles: The original colour looks good on a physical OLED so I'll leave it as is.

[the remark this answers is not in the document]
rfuest: It should be the same color, but the original value relies on the masking that is applied in the `new` method.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,422 @@
...
+
+        Size::new(w as u32, h as u32)
+    }
+}
+```
+
+Some notes on the above:
+
+- The `Drawing` trait is renamed to `DrawTarget`.
+- This implementation takes a `C: PixelColor + Into<BinaryColor>` allowing items using other color types to be used on the monochrome display. If this is not possible or undesired, use the appropriate pixel type directly. For example:
```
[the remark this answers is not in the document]
jamwaffles: Ok, fair. I've reverted back to explicitly declaring `DrawTarget<PixelColor>`. This guide should serve as a description of patterns anyway, so that's good enough for now.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,422 @@
...
++ )
++ .draw(&mut display)?;
+```
+
+- All built in `text_*!()` macros are removed and replaced with the `egtext!()` macro.
+- `egtext!()` should be coupled with the `text_style!()` macro to create styled texts with a chosen font.
+
+## Primitives and their styling
+
+Note: The `Triangle` primitive was added in 0.4.8. If you were using 3 separate lines, this primitive is suggested instead.
```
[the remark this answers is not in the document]
jamwaffles: Can't hurt

embedded-graphics/MIGRATING-0.5-0.6.md
jamwaffles: I ended up removing this example. I couldn't think of a better one to replace it with.

jamwaffles:
Comments addressed! Would like to get this merged as there's very little else left to do for a 0.6.0 release I think?

embedded-graphics/MIGRATING-0.5-0.6.md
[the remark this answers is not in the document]
rfuest: IMO if you want to keep the value it should be mentioned that the red component will get masked.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,444 @@
...
+    .chain(
+        &Text::new("Hello World!", Point::new(5, 50))
+            .into_styled(TextStyle::new(Font6x8, BinaryColor::On)),
+    )
+    .draw(&mut display);
+```
+
+Drawing operations are now fallible, with `.draw()` calls returning a `Result`. This allows for error handling if an error occurs during a drawing operation. The error type of this `Result` is dependent on the associated `Error` type as defined by the display driver's `DrawTarget` impl.
+
+## Coordinates and positioning
```
rfuest: Maybe
```suggestion
## Coordinates and sizing
```
?

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,444 @@
...
+
+        Size::new(w as u32, h as u32)
+    }
+}
+```
+
+Some notes on the above:
+
+- The `Drawing` trait is renamed to `DrawTarget`.
+- This implementation takes a `C: PixelColor + Into<BinaryColor>` allowing items using other color types to be used on the monochrome display. If this is not possible or undesired, use the appropriate pixel type directly. For example:
```
rfuest: It doesn't take `C: PixelColor + Into<BinaryColor>`  anymore.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,444 @@
...
+- A new `DrawTarget::size` method is now a required, which should return the width and height of the display in pixels as a `Size`.
+- The implementation of `DrawTarget` must now provide an associated `Error` type. If drawing operations can't fail, for example if the driver uses a framebuffer in RAM, `core::convert::Infallible` can used. If drawing operations can fail a better error type should be used to communicate hardware failures, etc to the user.
+- Any pixels drawn to offscreen coordinates need to be ignored by `DrawTarget` implementations and not result in panics or any modification of the display content. In the example above, `self.set_pixel()` (defined by the SSD1306 driver) will not attempt to set any pixels beyond the positive screen limits.
+- `draw_pixel()` now returns `Result<(), Self::Error>` to account for driver error handling.
+
+New methods have been added to `DrawTarget` to enable accelerated drawing operations if the hardware supports it:
```
rfuest: This is in the driver authors section, but IMO this could still be misinterpreted that these methods should be called on a `DrawTarget` impl.
```suggestion
New methods have been added to `DrawTarget` that can be overridden by hardware accelerated implementations:
```

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,444 @@
...
+
+- `DrawTarget::draw_circle` - draw a circle using hardware accelerated operations.
+- `DrawTarget::draw_rectangle` - draw a rectangle using hardware accelerated operations.
+- `DrawTarget::draw_triangle` - draw a triangle using hardware accelerated operations.
+- `DrawTarget::draw_line` - draw a line using hardware accelerated operations.
+- `DrawTarget::draw_image` - draw an image using hardware accelerated operations.
```
rfuest: The description before the list should be enough and the method names are pretty self explanatory. And these methods aren't always hardware accelerated, so the description isn't totally accurate.

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,444 @@
...
+
+By default these delegate to `DrawTarget::draw_iter` but can be overridden with custom implementations for improved draw performance.
```
rfuest: If the description of the list above is change, this could be changed to:
```suggestion
Methods that aren't overridden by the `DrawTarget` impl delegate to `DrawTarget::draw_iter` by default.
```

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,444 @@
...
+  impl Font for MyFont {
+      const FONT_IMAGE: &'static [u8] = include_bytes!("../data/my_font.raw");
+-     const CHAR_WIDTH: u32 = 12;
+-     const CHAR_HEIGHT: u32 = 22;
++     const CHARACTER_SIZE: Size = Size::new(12, 22);
+      const FONT_IMAGE_WIDTH: u32 = 480;
+  }
+```
+
+The `CHARACTER_SPACING` and `VARIABLE_WIDTH` constants were also added, creating support for basic variable width fonts.
```
rfuest: `CHARACTER_SPACING` can also be used for fixed fonts, like in the `custom-font' example.

embedded-graphics/MIGRATING-0.5-0.6.md
jamwaffles: I went with `Coordinates, sizing and positioning` because why not cover all the bases ;)

embedded-graphics/MIGRATING-0.5-0.6.md
```diff
@@ -0,0 +1,444 @@
...
+- [General drawing operations](#general-drawing-operations)
+- [Coordinates and positioning](#coordinates-and-positioning)
+- [Text](#text)
+- [Primitives and their styling](#primitives-and-their-styling)
+- [Images](#images)
+- [For driver authors](#for-driver-authors)
+  - [Choosing the right pixel color type](#choosing-the-right-pixel-color-type)
+- [For font authors](#for-font-authors)
+
+## Pixel colors
```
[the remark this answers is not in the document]
jamwaffles: Ah yeah it did. I added a small section. Not sure if it needs much else.

jamwaffles:
Final (maybe? :D) comments addressed. Thanks for the laser sharp review! This is looking like a pretty solid guide now.

jamwaffles:
Haha the nags are useful, I always make some stupid mistakes. Thanks for spending the time on this as usual!

PR #256, opened 2020-02-17
Make builders consuming

### LINKED ISSUE #248: Consuming builders for styles
`PrimitiveStyleBuilder` and `TextStyleBuilder` are currently implemented as "Non-consuming builders" (https://doc.rust-lang.org/1.0.0/style/ownership/builders.html). But it might be beneficial to change them to consuming builders.

Consuming builders can be used to generate `const`s, which would be pretty useful IMO. This won't be a advantage immediately, because trait bounds aren't yet supported for `const fn`s, but this way this won't be a breaking change later.

It could also generate more efficient code, because `build` doesn't need to `clone` the style anymore. (this is just an assumption and not tested)

Are there any advantages of leaving the builders as non-consuming variants?

jamwaffles:

## PR description

Closes #248

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -20,6 +20,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Changed
 
+- **(breaking)** [#256](https://github.com/jamwaffles/embedded-graphics/pull/256) Make `PrimitiveStyleBuilder` and `TextStyleBuilder` consuming when `.build()` is called.
 - **(breaking)** #247 The raw data `Image` struct is renamed to `ImageRaw`. The `Image` name is reused as a wrapper for other image types that implement the `ImageData` trait.
 - **(breaking)** #247 `ImageLE` and `ImageBE` are renamed to `ImageRawLE` and `ImageRawBE` respectively.
 - **(breaking)** #247 `ImageTga` and `ImageTga` are removed, along with their respective enabling features `bmp` and `tga`. To use BMP or TGA images, add `tinybmp` or `tinytga` to your dependencies with the `graphics` feature enabled. This allows usage of `Bmp` and `Tga` in embedded-graphics contexts. Tinybmp or tinytga usage now looks like this:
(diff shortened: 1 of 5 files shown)


rfuest:
All builder methods need to be changed from `&mut self` to `self` for this to work as expected. This currently only works because the builder and style structs are both implementing `Copy`, but probably won't work for creating constants in the future.

embedded-graphics/src/style/mod.rs
```diff
@@ -28,9 +28,9 @@ pub use text_style::{TextStyle, TextStyleBuilder};
 macro_rules! primitive_style {
     ($($style_key:ident = $style_value:expr ),* $(,)?) => {{
         #[allow(unused_mut)]
-        let mut builder = $crate::style::PrimitiveStyleBuilder::new();
+        let  builder = $crate::style::PrimitiveStyleBuilder::new();
```
rfuest:
```suggestion
        let builder = $crate::style::PrimitiveStyleBuilder::new();
```

jamwaffles:
Oh, right I see what you mean now. I've pushed some changes that closely follows the [consuming builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html#consuming-builders:).

embedded-graphics/src/style/mod.rs
rfuest: Maybe we should try to run `rustfmt` nightly with the appropriate options to format the macros.

jamwaffles: I turned on `format_macro_matchers` and `format_macro_bodies` in the nightly config but it didn't catch this issue specifically. I've left them in because they'll be handy later anyway. I fixed the double space manually.

embedded-graphics/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ### Changed
 
+- **(breaking)** [#256](https://github.com/jamwaffles/embedded-graphics/pull/256) Make `PrimitiveStyleBuilder` and `TextStyleBuilder` consuming when `.build()` is called.
```
rfuest: This isn't accurate anymore, because all methods now take ownership of the builder.
Will you also add links for the other issue numbers in the changelog?

jamwaffles: > This isn't accurate anymore, because all methods now take ownership of the builder.

True, fixed.

> Will you also add links for the other issue numbers in the changelog?

Yeah, before release probably. Should be a simple find/replace.

PR #257, opened 2020-02-19
Color conversion ergo

jamwaffles:

## PR description

I haven't thought about this too much, so if there's a reason not to add it let me know. At any rate, @almindor on Matrix was asking about making colour conversions easier.

Adds `.into_raw()` for `PixelColor`. Allows easier conversion from a colour type to its raw value. This is mostly sugar for the `RawU1::from(color).into_inner()` dance which is a bit of a pain. It can now be replaced by `color.into_raw()`.

I'm not keen on the name `into_raw()`. Better ideas welcome! `into_storage()`? `into_inner()` like the `Raw` colours?

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,9 +10,13 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Added
 
-- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added `.into_raw()` method on `PixelColor`s to allow for easier conversion from a colour to its raw storage type.
+- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added the `.into_raw()` method on `PixelColor`s via the `PixelColorExt` trait to allow for easier conversion from a colour to its raw storage type.
 
   ```diff
+  // PixelColorExt is included in the prelude.
+  // You can also import it with use embedded_graphics::pixelcolor::PixelColorExt.
+  + use embedded_graphics::prelude::*;
+
   - RawU1::from(color).into_inner()
   + color.into_raw()
   ```
@@ -417,6 +421,6 @@ A big release, focussed on ergonomics. There are new macros to make drawing and 
   ```
 
 <!-- next-url -->
-[unreleased]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.0-beta.1...HEAD
 
+[unreleased]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.0-beta.1...HEAD
 [0.6.0-beta.1]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.0-alpha.3...embedded-graphics-v0.6.0-beta.1
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,7 +10,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Added
 
-- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added the `.into_storage()` method on `PixelColor`s via the `IntoStorage` trait to allow for more ergonomic conversion from a colour to its raw storage type.
+- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added the `.into_storage()` method on `PixelColor`s via the `IntoStorage` trait to allow for more ergonomic conversion from a color to its raw storage type.
 
   ```diff
   // IntoStorage is included in the prelude.
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,15 +10,15 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Added
 
-- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added the `.into_raw()` method on `PixelColor`s via the `PixelColorExt` trait to allow for easier conversion from a colour to its raw storage type.
+- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added the `.into_storage()` method on `PixelColor`s via the `IntoStorage` trait to allow for more ergonomic conversion from a colour to its raw storage type.
 
   ```diff
-  // PixelColorExt is included in the prelude.
-  // You can also import it with use embedded_graphics::pixelcolor::PixelColorExt.
+  // IntoStorage is included in the prelude.
+  // You can also import it with use embedded_graphics::pixelcolor::IntoStorage.
   + use embedded_graphics::prelude::*;
 
   - RawU1::from(color).into_inner()
-  + color.into_raw()
+  + color.into_storage()
   ```
 
 - Added as many `#[derive()]`s as possible to all embedded-graphics, tinybmp and tinytga types.
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,6 +10,13 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Added
 
+- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added `.into_raw()` method on `PixelColor`s to allow for easier conversion from a colour to its raw storage type.
+
+  ```diff
+  - RawU1::from(color).into_inner()
+  + color.into_raw()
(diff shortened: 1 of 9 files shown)


embedded-graphics/src/pixelcolor/rgb_color.rs
```diff
@@ -295,4 +299,11 @@ mod tests {
         let color: Bgr555 = RawU16::from(0xFFFF).into();
         assert_eq!(RawU16::from(color).into_inner(), 0x7FFF);
     }
+
+    #[test]
+    fn convert_to_raw() {
+        let color = Rgb888::new(0xaa, 0xbb, 0xcc);
```
jamwaffles: I only added `Rgb888` here as I'm banking on the macro impls all doing the same/right thing internally.

embedded-graphics/src/pixelcolor/gray_color.rs
```diff
@@ -101,4 +105,11 @@ mod tests {
         assert_eq!(RawU4::from(Gray4::new(0x6)), RawU4::new(0x6));
         assert_eq!(RawU8::from(Gray8::new(0x7E)), RawU8::new(0x7E));
     }
+
+    #[test]
+    fn convert_to_raw() {
+        let color = Gray8::new(0xaa);
```
jamwaffles: See other macro banking comment.

almindor:
Looks good, `into_raw` is a bit funny name but I can't come up with anything that sounds any better. `into_native` comes to mind but that has other conotations, same for `into_primitive`...

rfuest:
> I haven't thought about this too much, so if there's a reason not to add it let me know.

I'm not 100% convinced that this is necessary. Sure the long form isn't very user friendly, but conversion between colors and raw data isn't a very common operation either, at least in my experience. But adding this shouldn't harm anything and I'm totally OK with adding it, if you and @almindor think it's useful.

But I would like to change the way this is implemented: By adding an additional method to the `PixelColor` trait each implementation has to deal with this separately, even if the color can't be converted to `RawData` at all. By using an extension trait this function could be implemented for all `PixelColor`s that can be converted to `RawUx`, with the additional benefit of not breaking the API:
```rust
pub trait PixelColorExt {
    type Storage;
    fn into_raw(self) -> Self::Storage;
}

impl<C> PixelColorExt for C
where
    C: PixelColor,
    C::Raw: From<C>,
{
    type Storage = <<C as PixelColor>::Raw as RawData>::Storage;

    fn into_raw(self) -> Self::Storage {
        C::Raw::from(self).into_inner()
    }
}
```
And `into_raw` should really get another name, because it doesn't convert into `PixelColor::Raw`.

almindor:
Shouldn't it be possible to implement standard `From` for `PixelColor<Each>` in the macro? So you'd just do `u16::from(color)` in the end.

As for use cases this is pretty much required in drivers. It's a minor issue, just ergonomics really.

jamwaffles:
> By using an extension trait this function could be implemented for all PixelColors that can be converted to RawUx, with the additional benefit of not breaking the API:

Ah yes that's what I was looking for, thanks. I've updated the PR with this approach.

> Shouldn't it be possible to implement standard From for PixelColor<Each> in the macro?

Yes, but I think I'd prefer an explicit `into_raw()` (or whatever we call it) method to make it clear that the conversion is between the `PixelColor` and its underlying storage.

> And into_raw should really get another name, because it doesn't convert into PixelColor::Raw.

Yeah agreed, was hoping we could bikeshed some alternatives in this thread. `into_storage` might work? Although I'm not sure how much e-g internals one needs to know about to understand that `storage` means a `u16` or whatever. The doc example on the trait helps though tbf.

rfuest:
> Yeah agreed, was hoping we could bikeshed some alternatives in this thread. `into_storage` might work? Although I'm not sure how much e-g internals one needs to know about to understand that `storage` means a `u16` or whatever. The doc example on the trait helps though tbf.

`into_storage` isn't ideal, but I can't come up with something better. Should we also rename `PixelColorExt` to into `IntoStorage` then, because it only contains methods and types related to storage?
```rust
pub trait IntoStorage {
    type Storage;
    fn into_storage(self) -> Self::Storage;
}
```
This would leave the `PixelColorExt` name free for further extensions that don't require the `C::Raw: From<C>` trait bound.

rfuest:
Do you think we should also simplify the opposite conversion from storage to `PixelColor`?

jamwaffles:
>  Should we also rename PixelColorExt to into IntoStorage then

Yeah I like that. I've pushed a new commit with the renamed trait/method.

> Do you think we should also simplify the opposite conversion from storage to PixelColor?

Maybe at some point, but the point of this PR was to address the currently-used code in various driver crates that converts `PixelColor`s to their raw counterparts. Let's leave it for a future PR if we need it.

embedded-graphics/src/pixelcolor/mod.rs
```diff
@@ -108,5 +109,49 @@ pub trait PixelColor: Copy + PartialEq {
...
+///
+/// The [`Rgb565`] color type is stored in a `u16` internally. This example converts an [`Rgb565`]
+/// color into its underlying `u16` represenation.
+///
+/// ```rust
+/// use embedded_graphics::{prelude::*, pixelcolor::Rgb565};
+///
+/// let color = Rgb565::new(0xff, 0x00, 0xaa);
+///
+/// let raw: u16 = color.into_storage();
```
rfuest: Can this be misinterpreted that the type annotation is necessary? Perhaps adding an explicit type to the `0b..` constant below would be clearer.
```suggestion
/// let storage = color.into_storage();
```

embedded-graphics/src/pixelcolor/rgb_color.rs
```diff
@@ -295,4 +296,11 @@ mod tests {
...
```
rfuest: For consistency with the above tests.
```suggestion
        let color = Rgb888::new(0xAA, 0xBB, 0xCC);
```

embedded-graphics/src/pixelcolor/rgb_color.rs
```diff
@@ -295,4 +296,11 @@ mod tests {
...
+
+        assert_eq!(color.into_storage(), 0x00aabbcc);
```
rfuest:
```suggestion
        assert_eq!(color.into_storage(), 0x00AABBCC);
```

embedded-graphics/src/pixelcolor/gray_color.rs
```diff
@@ -101,4 +102,11 @@ mod tests {
...
```
rfuest:
```suggestion
        let color = Gray8::new(0xAA);
```

embedded-graphics/src/pixelcolor/mod.rs
```diff
@@ -108,5 +109,49 @@ pub trait PixelColor: Copy + PartialEq {
...
+/// Convert a [`PixelColor`] into its underlying storage type
+///
+/// This trait provides the `into_storage()` method for implementors of [`PixelColor`]. This method
+/// exposes the underlying storage value of a pixel color type.
+///
+/// # Examples
+///
+/// ## Get the `u16` representing an `Rgb565` color
+///
+/// The [`Rgb565`] color type is stored in a `u16` internally. This example converts an [`Rgb565`]
```
rfuest: Isn't the first part of this paragraph redundant? The second sentence already states that the color is represented by a `u16`.

embedded-graphics/CHANGELOG.md
```diff
@@ -10,6 +10,17 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ### Added
 
+- [#257](https://github.com/jamwaffles/embedded-graphics/pull/257) added the `.into_storage()` method on `PixelColor`s via the `IntoStorage` trait to allow for more ergonomic conversion from a colour to its raw storage type.
```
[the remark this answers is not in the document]
jamwaffles: We must regain our sovereignty! #brexit

(joke)

embedded-graphics/src/pixelcolor/mod.rs
jamwaffles: Ah yep, just some copypasta noise. Removed :)

rfuest:
> Maybe at some point, but the point of this PR was to address the currently-used code in various driver crates that converts `PixelColor`s to their raw counterparts. Let's leave it for a future PR if we need it.

If we _needed_ to implement this, IMO the opposite should also be implemented to provide some symmetry for the API. The opposite conversion is used in image crates like the `tiny...` crates (in a slightly different form because of the more generic `u32` conversion). But I agree that this doesn't belong in this PR.

jamwaffles:
Thanks for the review!

PR #258, opened 2020-02-20
add draw_image to allow optimizations

almindor:

## PR description

Individual primitives have `draw_<primitive>` calls in the `DrawTarget` trait to allow optimizing where applicable. Images however always just call `draw_pixel` which results in extremely sub-par performance with most drivers.

This adds `draw_image` to the `DrawTarget` trait to enable drivers to optimize rednering, especially given that images are rectangular.

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- Added `draw_image` to `DrawTarget` trait with default implementation.
+
 ## [0.6.0-beta.1] - 2020-02-17
 
 ### Added
--- a/embedded-graphics/src/draw_target.rs
+++ b/embedded-graphics/src/draw_target.rs
@@ -1,6 +1,7 @@
 use crate::{
     drawable::{self, Drawable},
-    geometry::{Point, Size},
+    geometry::{Dimensions, Point, Size},
+    image::{Image, ImageDimensions, IntoPixelIter},
     pixelcolor::PixelColor,
     primitives::{self, Primitive},
     style::{PrimitiveStyle, Styled},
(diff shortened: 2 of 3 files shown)


almindor:
Here are the results for an ST7789 on PineTime:

* [Ferris image optimized rendering](http://objdump.katona.me/ferris_fast.mp4)
* [Ferris image slow rendering](http://objdump.katona.me/ferris_slow.mp4)

jamwaffles:
Is there a PR in a driver crate you can link to that demonstrates this change being faster? It also raises the question of whether to add a generally usable `SizedDrawTarget` (or something) as opposed to a method that only deals with images.

almindor:
I've already merged to master in the [ST7789 display driver crate](https://github.com/almindor/st7789/commit/b7992b387f1f3b6f4727c13f248eb39b2a2b1b83) but you can check the [main commit here](https://github.com/almindor/st7789/commit/b7992b387f1f3b6f4727c13f248eb39b2a2b1b83)

The reason why I think `draw_image` is appropriate at this time is that this way it's a non-breaking change, purely additive and optional.

If we wanted to do something like `draw_sized(generic_item: I) where I: IntoIterator<C> + Dimensions` I feel it'd currently conflict with all the existing `draw_<primitive>`.

There is another consideration to take into account here which is "the bounds frame is exactly same as the pixel data" information which is cruicial here.

When I implement `draw_rectangle` and `draw_image` I safely know that the bounds of the drawn item is 100% filled with pixels I will get. This makes working with window frames in the driver trivial.

If I only knew the bounds but not what the shape/contents were I couldn't use this optimization at all especially without a known pixel ordering.

I think for `0.6.0` `draw_image` is a good enough intermediate "fix" to get 100% optimal paths for rectangular renders but `0.7.0` should probably redesign `DrawTarget` a bit. (not to say that `DrawTarget` is bad, it's definitely a step in the right direction compared to `draw_sized` being exposed).

jamwaffles:
Apologies for the delay in getting to this. I think this is a fine addition, but yes definitely a stopgap until a better solution for drawing contiguous data is found.

@rfuest What do you think? There's quite a bit of context behind this from some Matrix chats which I can attempt to summarise here if you need it.

rfuest:
Adding special support for 100% filled rectangular areas was discussed before and is definitely a good idea. And now that all images are handled equally adding support for images as a first step to better performance seems like a good short term solution.

> When I implement `draw_rectangle` and `draw_image` I safely know that the bounds of the drawn item is 100% filled with pixels I will get.

Not really relevant to this PR, but this isn't true for all `draw_rectangle` calls. A rectangle could only be styled using a stroke, which wouldn't fill all pixels.

PR #259, opened 2020-02-20
Update README.md

rfuest:
I tried to restructure the README a bit. It's far from perfect but and would like some feedback on the changes and if this is going in the right direction.

Summary of the changes:
* The readme now starts with an overview of e-g
   The screenshots are visually nice but don't provide that much actual information, so I decided to move them down a bit.
* I've added the logo to the readme
  Having a centerd logo seems to be very common on GitHub and looks good IMO. I think crates.io strips the alignment tag, but having the tags and logo left aligned there should still look OK.
* The external crates and display drivers are now listed below each other.
  IMO this gives a better overview of all additional features.
* The simulator wasn't mentioned before
   The readme didn't really refer to the simulator before. I've mentioned it in the overview and also added it to the display drivers list.
* The features section was outdated

TODO:
* [ ] Add screenshots to example
   Seeing the output of the example code would make the example more useful.
* [ ] Decide if we need two versions of the example
   I'm not sure if the readme is the right place to provide to equivalent examples which are implemented slightly differently. I would prefer to use both drawing variants in one example (e.g. drawing the first line without and the second with macro) and link to the docs for further information. The freed up space could better be used to add another example IMO.
* [ ] Add more examples
  If we decide to remove one of the example implementation the additional space could be used to add another example. Maybe one that use `ImageRaw`?


embedded-graphics/README.md
```diff
@@ -1,11 +1,46 @@
...
+</p>
+
 # Embedded graphics
 
-[![Build Status](https://circleci.com/gh/jamwaffles/embedded-graphics/tree/master.svg?style=shield)](https://circleci.com/gh/jamwaffles/embedded-graphics/tree/master)
-[![Crates.io](https://img.shields.io/crates/v/embedded-graphics.svg)](https://crates.io/crates/embedded-graphics)
-[![Docs.rs](https://docs.rs/embedded-graphics/badge.svg)](https://docs.rs/embedded-graphics)
-[![embedded-graphics on Matrix](https://img.shields.io/matrix/rust-embedded-graphics:matrix.org)](https://matrix.to/#/#rust-embedded-graphics:matrix.org)
+Embedded-graphics is a 2D graphics library that is focused on memory constrained embedded devices.
+It contains build in items that make it easy to draw 2D graphics primitives:
```
jamwaffles:
```suggestion
It contains built in items that make it easy to draw 2D graphics primitives:
```

embedded-graphics/README.md
```diff
@@ -15,41 +50,41 @@ If you need more deeper/more personalised help, please check out the [embedded-g
...
-- Raw data images
-- Primitives
-  - Lines
-  - Rectangles (and squares)
-  - Circles
-  - Triangles
-- Text with multiple fonts
-
-Additional functionality provided by external crates:
+## Additional functions provided by external crates
```
jamwaffles: Can we move this section closer/underneath the built in features list? I think it's useful to make it easier to discover extensions to e-g.

embedded-graphics/README.md
```diff
@@ -154,26 +189,6 @@ fn main() {
 
 - `nalgebra_support` - use the [Nalgebra](https://crates.io/crates/nalgebra) crate with `no_std`
   support to enable conversions from `nalgebra::Vector2` to `Coord` and `UnsignedCoord`.
-- `bmp` - use the [TinyBMP](https://crates.io/crates/tinybmp) crate for BMP image support.
-- `tga` - use the [TinyTGA](https://crates.io/crates/tinytga) crate for TGA image support.
```
jamwaffles: Ahhh! Can't believe I forgot to remove these lol, thank you

embedded-graphics/README.md
rfuest: I've now tried to incorporate the additional features provided by the external creates in the main feature overview. Do you think that would work?

If we move the entire section the fancy screenshot image would move further down and maybe not as many people would notice it. One solution might be to use multiple smaller screenshots throughout the readme instead of one huge screenshot collage, that fills the entire screen.

jamwaffles: How about moving the two photos into their own wide and short image that can go under the `Embedded Graphics` heading? Or even centered underneath the badges at the top of the page.

The simulator screenshots could be moved/replaced with individual outputs on each example near the bottom of the page.

jamwaffles: > Do you think that would work?

Hm, not really. I was thinking of two separate bullet lists, one after the other. It should take up nearly the same amount of vertical space as your change just now and we can go back to having a nice list of links for external crates.

rfuest:
I've added a new hello world example, that is based on the `graphics_i2c.rs` example in the `ssd1306` crate, but with some additional features of e-g used. If we keep both examples in sync we could also add a picture of the example running on real hardware and a link to the `ssd1306` example.

> 2\. I'd like to show off the macro usage, but maybe there could be two smaller examples that do different things, one using macros and one not?

Sure macros should be demonstrated in the readme. Do you have an idea for a simple macro example?

rfuest:
The README now contains the entire example, that can be run using the simulator. This obviously breaks CI on `no_std` platforms. Do you think we can simply disable the readme check for `no_std` tagets?

embedded-graphics/README.md
rfuest: > How about moving the two photos into their own wide and short image that can go under the `Embedded Graphics` heading? Or even centered underneath the badges at the top of the page.

That sounds like a good idea. Can you prepare that image, because I don't have access to the full size versions of the images?

rfuest: > Hm, not really. I was thinking of two separate bullet lists, one after the other. It should take up nearly the same amount of vertical space as your change just now...

I get your point, but having a link list inside the overview of the crate isn't ideal either. The 4 links aren't to disturbing, but if the lists gets longer this isn't the ideal place for it IMO.

> ... and we can go back to having a nice list of links for external crates.

We still have that, only a bit further down the page. But if we remove the large screenshot we could move the "Additional functions provided by external crates" sections directly after the introduction.

jamwaffles: > Can you prepare that image

I've pushed a banner here: https://github.com/jamwaffles/embedded-graphics/blob/readme-images/assets/banner-photo.png it's a 2x scaling of the 100% width of a Github readme so should hopefully fit nicely. I added drop shadows because I'm not sure what else to do to make it look good tbh. Looks alright though.

If you pull it in, please take the GIMP XCF too: https://github.com/jamwaffles/embedded-graphics/blob/readme-images/assets/banner-photo.xcf it's massive, but that gives us scope to reuse the images later.

jamwaffles: As for the other layout stuff, I'll leave it to your better judgement and see what it looks like with the images split/moved around a bit.

embedded-graphics/Cargo.toml
```diff
@@ -34,3 +34,4 @@ nalgebra_support = [ "nalgebra" ]
 [dev-dependencies]
 arrayvec = "0.5.1"
 tinytga = { version = "0.3.1", features = [ "graphics" ] }
+embedded-graphics-simulator = "0.2.0-beta.2"
```
jamwaffles: I got this to "work" with:

```toml
[target.'cfg(target_arch = "x86_64")'.dependencies]
embedded-graphics-simulator = "0.2.0-beta.2"
```

However now Rust complains about a cyclic dependency :(

rfuest:
```
[target.'cfg(target_arch = "x86_64")'.dev-dependencies]
embedded-graphics-simulator = "0.2.0-beta.2"
```
gets rid of the cyclic dependency, but somehow tries to build some crates with `std` support, even if the arch doesn't match.

Any ideas how to solve this? Checking the e-g README while building the simulator could work, but is a hack.

jamwaffles: Seems to be the `--all-features` flag in `build_target.sh`. Try changing `set -e` to set -ex` at the top and it'll show the bailing command. I can repro if I run this:

```bash
cargo build -p embedded-graphics --target thumbv7m-none-eabi --features nalgebra_support
```

This builds fine:

```bash
cargo build -p embedded-graphics --target thumbv7m-none-eabi
```

Very weird because this works in master 🤔 

> Any ideas how to solve this? Checking the e-g README while building the simulator could work, but is a hack.

We could add another, unpublished crate to the workspace that pulls in all the readme(s) as examples. Would give a bit more control I think, but yeah still feels quite hacky.

rfuest: Enabling the new feature resolver (https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#features) seems to fix this problem. But until this gets stabilized we need another solution, so I decided to move all README checks into the simulator crate for now. Are you OK with leaving it like this?

embedded-graphics/README.md
rfuest: > If you pull it in, please take the GIMP XCF too: https://github.com/jamwaffles/embedded-graphics/blob/readme-images/assets/banner-photo.xcf it's massive, but that gives us scope to reuse the images later.

I don't like the idea of doubling the size of the repo just for one image, but it would be useful to have these available. Can we add a separate repo just for assets?

embedded-graphics/Cargo.toml
jamwaffles: Yeah that's fine, as long as they get tested in CI at some point that's cool. I don't think a `cargo publish` runs tests so we're not missing any safety there. Is it worth opening a tracking issue for reverting the README move when the new resolver stabilised? If you think so, please open one, otherwise it works as is so let's leave it.

embedded-graphics/README.md
jamwaffles: A separate repo seems a bit overkill. I could just keep it in the branch I linked to, or delete it? It's not difficult to recreate that banner tbh. How about I export a high-ish resolution PNG and replace the XCF with it? That gives a decent tradeoff between size and future usability, no?

rfuest: I did tweak your image a bit, because I didn't like the look of images with different heights in one row. And I also removed the shadow to gain some additional with and have the left and right edges of the image be aligned with the rest of the content. It might look a little more plain now, but I think it's OK. Or do you prefer to change it back?

> A separate repo seems a bit overkill.

Sure, having a repo just for this one image is overkill. My idea was to have a place to store nice e-g related images that can be reused for different purposes like docs or if someone wants to write a blog post about e-g. But that's probably also overkill if we only have a few images.

jamwaffles: > but I think it's OK

Me too haha, thanks for fixing my botch job.

rfuest:
I've reverted the changes made in PR #261 to make the hello-world example work again. This change doesn't really belong in this PR, but was required for this. If you prefer I could also create another PR for this change.

This PR was originally only intended to add the logo and grew in scope after that. Because of that I would like to merge this now and we can improve the README later in other PRs. The images are still linked to my fork of the repo to make the preview work and I will fix the links before merging.

PR #261, opened 2020-02-26
Add automatic color conversion for SimulatorDisplay

rfuest:

## PR description

The simulator used to implement `DrawTarget` for all `C`s that can be converted `Into` the color type used to build the `SimulatorDisplay`. This feature was somehow lost during the refactoring of the simulator some time ago. This PR reimplements that feature.

After this changes an explicit type annotation on the `SimulatorDisplay` variables is required. But IMO that's beneficial because this way the type of the framebuffer data is always unambiguously defined.

### DIFF
--- a/simulator/src/display.rs
+++ b/simulator/src/display.rs
@@ -62,14 +62,14 @@ where
     }
 }
 
-impl<CP, CD> DrawTarget<CP> for SimulatorDisplay<CD>
+impl<PixelC, DisplayC> DrawTarget<PixelC> for SimulatorDisplay<DisplayC>
 where
-    CP: PixelColor + Into<CD>,
-    CD: PixelColor,
+    PixelC: PixelColor + Into<DisplayC>,
+    DisplayC: PixelColor,
 {
     type Error = core::convert::Infallible;
 
-    fn draw_pixel(&mut self, pixel: Pixel<CP>) -> Result<(), Self::Error> {
+    fn draw_pixel(&mut self, pixel: Pixel<PixelC>) -> Result<(), Self::Error> {
         let Pixel(point, color) = pixel;
 
         if let Some(index) = self.point_to_index(point) {
--- a/simulator/src/display.rs
+++ b/simulator/src/display.rs
@@ -62,17 +62,18 @@ where
     }
 }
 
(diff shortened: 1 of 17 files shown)


simulator/src/display.rs
```diff
@@ -62,17 +62,18 @@ where
     }
 }
 
-impl<C> DrawTarget<C> for SimulatorDisplay<C>
+impl<CP, CD> DrawTarget<CP> for SimulatorDisplay<CD>
```
jamwaffles: Minor nit, but can we change these to `PC` for Pixel Color` and `DC` for Display Color? If you'd prefer to keep them as is, please add a (non-doc) comment above with a quick explanation - the names and usages are a bit confusing even to me.

jamwaffles:
Ah, some missed doc examples broke the build - `lib.rs` and `check_readme.rs`

simulator/src/display.rs
rfuest: I know they aren't great, but I'm not sure that reversing the letters will make them better. They both are colors so I decided to use a common `C` prefix for both because all other color generics use `C`.

 Should we just use `PixelColor` and `DisplayColor` instead? Longer names for generics aren't very common but I think they would make this code more readable.

jamwaffles: > Should we just use PixelColor and DisplayColor instead? Longer names for generics aren't very common but I think they would make this code more readable.

That crossed my mind too. In this case yeah I think that's the best option so let's do that.

rfuest: Hmm, I just noticed that `PixelColor` is obviously not a good name. Maybe `DrawableColor`?

rfuest: Or `PixelC` and `DisplayC`?

jamwaffles: > Hmm, I just noticed that `PixelColor` is obviously not a good name.

Oh... yeah oops

> Or `PixelC` and `DisplayC`?

Yeah that works, it fixes the ambiguities imo

rfuest:
> Ah, some missed doc examples broke the build - `lib.rs` and `check_readme.rs`

Oops, I forgot to run the tests again, I should really start reading the checklist before adding the checkmarks :wink:. But the tests and other nits should now be fixed.

PR #262, opened 2020-02-26
Unstyled iterator for lines

### LINKED ISSUE #260: Filled triangles with no stroke fail to draw
- Version of embedded-graphics in use (if applicable): 0.6.0-beta.1
- Target: All

`Triangle`s with no stroke colour but a set fill don't draw anything due to the `Line` iterator bailing if `stroke_color == None`. 

Stroke _width_ can remain unset, only stroke _colour_ is required to make the triangle show up again.

## Test case (if applicable)

Open `simulator/examples/fill.rs` and add the following before the `let mut window ...` line near the bottom:

```rust
 let fill_only = PrimitiveStyleBuilder::new()
    .fill_color(BinaryColor::On)
    .build();

Triangle::new(Point::new(32, 0), Point::new(0, 64), Point::new(64, 64))
    .translate(Point::new(96 * 2 + 48, 48))
    .into_styled(fill_only)
    .draw(&mut display)?;
```

A fourth, filled triangle should be drawn but nothing appears. Adding `.stroke_color(BinaryColor::Off)` to the `fill_only` style fixes this.

This may be relevant to #156 in some way, or we fix lines to iterate over transparent pixels anyway.

Cc @almindor

jamwaffles:

## PR description

* Adds `IntoIterator` impl for `&Line`, alongside the impl for `Styled<Line>`
* Uses unstyled line iterator in `Triangle` so triangles without stroke can be rendered - closes #260 

Questions:

1. [ ] Is this the right direction, or should we add a `points()` or `iter()` method to the `Primitive` trait? 
1. [ ] If we continue down this path of adding `IntoIterator` impls, should we add a bound to `Primitive` so it becomes `Primitive: IntoIter`? I can't remember if this is possible with all the associated type stuff
1. [ ] Should we do all the other primitives, ~and if so should we do it in this PR?~ - if yes, let's do it in another PR so this one can focus on closing #260

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) `Triangle`s with a fill color but no stroke color are now rendered instead of not rendered at all.
+
 ### Added
 
 - Added `draw_image` to `DrawTarget` trait with default implementation.
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -9,6 +9,8 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 ### Fixed
 
 - [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) `Triangle`s with a fill color but no stroke color are now rendered instead of not rendered at all.
+- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) A triangle with a different stroke and fill color now renders its left-most border with the stroke color, instead of fill color
+- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) Triangles with a stroke width of 0 no longer draw the triangle outline and draw nothing as intended.
 
 ### Added
 
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,7 +10,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 - [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) `Triangle`s with a fill color but no stroke color are now rendered instead of not rendered at all.
 - [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) A triangle with a different stroke and fill color now renders its left-most border with the stroke color, instead of fill color
-- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) Triangles with a stroke width of 0 no longer draw the triangle outline and draw nothing as intended.
+- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) Triangles and lines with a stroke width of 0 (and no fill for triangles) are no longer drawn.
 
 ### Added
 
--- a/embedded-graphics/src/primitives/line.rs
+++ b/embedded-graphics/src/primitives/line.rs
@@ -37,21 +37,6 @@ use crate::{
 ///     .draw(&mut display)?;
 /// # Ok::<(), core::convert::Infallible>(())
 /// ```
-///
-/// ## Get an iterator over all points on the line
-///
-/// Call `.into_iter()` on a `Line` to get an iterator over all points along it.
-///
-/// ```rust
-/// use embedded_graphics::{prelude::*, primitives::Line};
-///
-/// let line = Line::new(Point::new(10, 10), Point::new(20, 20));
-///
-/// let mut line_it = line.into_iter();
-///
-/// // Diagonal line length is 14.141... according to Pythagoras' theorem
-/// assert_eq!(line_it.next(), Some(Point::new(10, 10)));
-/// ```
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
 pub struct Line {
     /// Start point
@@ -84,12 +69,22 @@ impl Line {
     }
 }
 
-impl<'a> IntoIterator for &'a Line {
-    type Item = Point;
-    type IntoIter = LineIterator;
+/// Pixel iterator for each pixel in the line
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub(crate) struct LineIterator {
+    start: Point,
+    end: Point,
+    delta: Point,
+    /// in which quadrant is the line drawn (upper-left=(-1, -1), lower-right=(1, 1), ...)
+    direction: Point,
+    err: i32,
+    stop: bool,
+}
 
-    fn into_iter(self) -> Self::IntoIter {
-        let mut delta = self.end - self.start;
+impl LineIterator {
+    /// Create a new line iterator from a `Line`
+    pub(crate) fn new(line: &Line) -> Self {
+        let mut delta = line.end - line.start;
 
         if delta.x < 0 {
             delta = Point::new(-delta.x, delta.y);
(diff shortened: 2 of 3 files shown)


embedded-graphics/src/primitives/line.rs
```diff
@@ -37,6 +37,23 @@ use crate::{
...
+///     primitives::Line, prelude::*
+/// };
+///
+/// let line = Line::new(Point::new(10, 10), Point::new(20, 20));
+///
+/// let mut line_it = line.into_iter();
+///
+/// // Diagonal line length is 14.141... according to Pythagoras' theorem
+/// assert_eq!(line_it.next(), Some(Point::new(10, 10)));
+/// ```
```
jamwaffles: Any ideas for a better example? This one is pretty lame.

rfuest:
Are you sure this should be included in 0.6? I don't like the idea of adding this as a last minute addition and would prefer to fix #260 some other way for the 0.6 release.

jamwaffles:
Hm, perhaps it's best to keep it out. 

That said, I think the fix here is pretty elegant (and we're going to do it anyway). Hiding it from the user and/or disallowing usage of `Line.into_iter` for now would be enough I think? Adding `#[doc(hidden)]` to `impl<'a> IntoIterator for &'a Line` would help. Could we use a sealed trait trick maybe?

rfuest:
If you want to keep `LineIterator` I would suggest to add a crate public `LineIterator::new` function and move the code there for now. Otherwise some kind of dummy style could be used to fix #260.

jamwaffles:
I've gone with the `LineIterator::new` suggestion - thanks for that :)

embedded-graphics/src/primitives/line.rs
```diff
@@ -156,43 +203,20 @@ where
...
 }
 
 // [Bresenham's line algorithm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm)
 impl<C: PixelColor> Iterator for StyledLineIterator<C> {
     type Item = Pixel<C>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        // return none if stroke color is none
+        // Return none if stroke color is none
         let stroke_color = self.style.stroke_color?;
```
rfuest: This problem probably existed before this refactoring, but to be consistent with rectangles and circles, lines should also not be drawn if `stroke_width` is `0`.

This also applies to triangles. Setting one of the stroke width of one of the styles in the `stroke.rs` example to `0` is a nice demo of this problem.

jamwaffles: Good catch! This also uncovered another problem with triangles and their borders which we haven't uncovered until now. All the triangle demos/tests use `BinaryColor` which masks this overdraw bug:

image

I've pushed some half-done debugging to this branch which I'll finish off later.

jamwaffles: And another example: the top-left line is overlapped by the border.

Screenshot from 2020-02-27 14-36-57

rfuest: Also a good catch!

almindor:
Alternative thin fix-workaround: https://github.com/jamwaffles/embedded-graphics/issues/260

NOTE: I only checked triangles with `PrimitiveStyle::with_fill` so far on that PR.

jamwaffles:
Ok, looks like I fixed the triangle overdraw issue:

image

image

There's now a test in `triangle.rs` that does a pixel check on a drawn triangle in `Rgb888` using `MockDisplay` so hopefully we'll catch stuff like this in the future.

@rfuest can you throw another review my way when you've got time?

embedded-graphics/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) `Triangle`s with a fill color but no stroke color are now rendered instead of not rendered at all.
+- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) A triangle with a different stroke and fill color now renders its left-most border with the stroke color, instead of fill color
+- [#262](https://github.com/jamwaffles/embedded-graphics/pull/262) Triangles with a stroke width of 0 no longer draw the triangle outline and draw nothing as intended.
```
rfuest: This also applies to lines.

Should a change like this be considered breaking? It doesn't break the API itself, but the changed drawing behavior can cause problems. If someone used a style with only the `stroke_color` set to draw lines they would disappear after applying this PR.

embedded-graphics/src/primitives/line.rs
```diff
@@ -250,6 +279,17 @@ mod tests {
...
 
+    #[test]
+    fn no_stroke_width_no_line() {
+        let start = Point::new(2, 3);
+        let end = Point::new(3, 2);
+
+        let line =
+            Line::new(start, end).into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 0));
+
+        assert_eq!(line.into_iter().next(), None);
```
rfuest: Not really necessary to change this, but this made me think about what the nicest looking and most semantic way to test if an iterator is empty is:
```suggestion
        assert!(line.into_iter().eq(core::iter::empty()));
```
```suggestion
        assert_eq!(line.into_iter().count(), 0);
```

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -319,8 +315,14 @@ where
             match self.points() {
                 IterState::Border(point) => {
                     // Draw edges of the triangle
-                    if let Some(color) = self.style.stroke_color.or_else(|| self.style.fill_color) {
-                        if point.x >= 0 && point.y >= 0 {
+                    if point.x >= 0 && point.y >= 0 {
+                        if self.style.stroke_width > 0 {
+                            if let Some(color) = self.style.stroke_color {
```
rfuest: At first glance I thought both `return` statements would use the same `color`. By changing
```suggestion
                            if let Some(stroke_color) = self.style.stroke_color {
```
it would be immediately obvious that the colors are different.

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -319,8 +315,14 @@ where
...
+                                self.x += 1;
+                                return Some(Pixel(point, color));
+                            }
+                        } else if let Some(color) = self.style.fill_color {
```
```suggestion
                        } else if let Some(fill_color) = self.style.fill_color {
```

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -406,6 +452,27 @@ mod tests {
         assert_eq!(tri.next(), None);
     }
 
+    #[test]
+    fn it_draws_filled_strokeless_tri() {
+        let mut tri = Triangle::new(Point::new(2, 2), Point::new(2, 4), Point::new(2, 4))
+            .into_styled(
+                PrimitiveStyleBuilder::new()
+                    .fill_color(BinaryColor::On)
+                    .build(),
```
```suggestion
                PrimitiveStyle::with_fill(BinaryColor::On)
```

embedded-graphics/src/primitives/triangle.rs
```diff
@@ -406,6 +452,27 @@ mod tests {
...
+
+        // Nodes are returned twice. first line a and b yield the same point.
+        // After that line a ends where line c starts.
+        assert_eq!(tri.next(), Some(Pixel(Point::new(2, 2), BinaryColor::On)));
+        assert_eq!(tri.next(), Some(Pixel(Point::new(2, 2), BinaryColor::On)));
+        assert_eq!(tri.next(), Some(Pixel(Point::new(2, 3), BinaryColor::On)));
+        assert_eq!(tri.next(), Some(Pixel(Point::new(2, 3), BinaryColor::On)));
+        assert_eq!(tri.next(), Some(Pixel(Point::new(2, 4), BinaryColor::On)));
+        assert_eq!(tri.next(), Some(Pixel(Point::new(2, 4), BinaryColor::On)));
+        assert_eq!(tri.next(), None);
```
rfuest: Does this test (and the comment) make sense for the fill? Maybe a `MockDisplay` test would be better suited.

embedded-graphics/CHANGELOG.md
jamwaffles: I added a mention to lines in here too. We could bikeshed whether it's breaking or not, but IMO this is fixing a bug so let's leave it non-breaking. Relying on buggy behaviour isn't the smartest thing to do anyhow ;)

embedded-graphics/src/primitives/triangle.rs
jamwaffles: Changed to a MockDisplay test. The other tests like this in the file can be migrated some other time :)

jamwaffles:
Nits addressed :)

jamwaffles:
Swag!

Issue #223, opened 2019-12-20
Add table of drawing operations to docs


rfuest:
From https://github.com/jamwaffles/embedded-graphics/pull/222#pullrequestreview-335126128:
>Forgot we had DrawTarget::draw_pixel! Looks good to me, but please add an entry to the changelog under Added - it's a public API change that someone out there might want to know about.

The ability to be able to draw single pixels by using `Pixel(...).draw(&mut display)` isn't very discoverable IMO. I've added some examples to the docs, but they are hard to find inside the `drawable` module.

I suggest to add a table to the front page of the docs, which provides a visual reference of the supported drawing operations:

| Image | Example |
|--------|-----------|
| (Image of a single pixel) | <p>Link to Pixel</p> **Example:** <br> `Pixel(...).draw(&mut display)` |
| (Image of a rectangle) | <p>Link to Rectangle</p> **Example:** <br> `Rectangle(...).draw(&mut display)` |
| ... | ... |

jamwaffles:
A table is a really good idea, particularly with screenshots. Visual docs for a visual library 👍.

If they get added to the main page docs, I think we need to do something like the hack in `check_readme.md` and include that huge doc comment from an external markdown file. `lib.rs` is getting pretty long already, and editing lots of text in `///` comments gets painful fast.

jamwaffles:
To add: the screenshots should be of the simulator with window title. In the example above, `(Image of a single pixel)` is going to look very strange without any context.

jamwaffles:
This is the last meaty issue before we can put 0.6.0 out I think. Do we still want to get it done before release? If so I'm happy to take a stab at it. Do you have any suggestions on how to automatically generate the screenshots?

rfuest:
I wouldn't mind if this was postponed. And this might also be a good candidate for the appendix of an eventual future e-g book.

But if you want to take a stab at it go ahead. I would create a program that generates the screenshots using multiple `display.to_image_buffer(&output_settings).save('some.png')` calls (and not the environment variable). This program could be called manually, because the output doesn't change very often and automating this further won't be worth the effort IMO.

jamwaffles:
I made a start in #279 but it's becoming a bit of a can of worms to do it in a maintainable way I think. I'm going to postpone this as well as #279 for after 0.6.0 release, which I'll do later today (:tada:)

PR #266, opened 2020-03-03
Add PNG export to simulator

rfuest:

## PR description

This PR adds PNG export to the simulator (or any other format supported by image.rs).

Two different methods of using the PNG export are supported by specifying one of two flags:
- The `image-rs` flag adds a `SimulatorDisplay::to_image_buffer` method to convert the display content to a image-rs `ImageBuffer` that can, for example, be saved to any supported format.
  This mode of operation is useful for applications that only generate PNG images. One use case I had in mind for this is automatically generating the images for #223.
- The `dump-png` flag changes the way the simulator works. Instead of showing a window the first drawn frame is dumped to a `dump.png` file and then the process is exited. This is useful for creating screenshots of the examples, to include in the docs, without cluttering the code of each example with conditional compilation. The `generate_example_screenshots.sh` script is an example of using this feature to generate screenshots of all examples.


simulator/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- **(breaking)** #266 Added image-rs support and PNG export. See the `README.md` for information about how to enable this features. The API for creating windows was changed to make the output settings independent of the `Window` type. The pixel scaling and theme settings were moved to a new `OutputSettings` struct, that can be built using the `OutputSettingsBuilder`. `WindowBuilder` was removed and replaced by a `Window::new(title, &output_settings)` function.
```
jamwaffles:
```suggestion
- **(breaking)** #266 Added [image](https://crates.io/crates/image) support and PNG export. See the `README.md` for information about how to enable these features. The API for creating windows was changed to make the output settings independent of the `Window` type. The pixel scaling and theme settings were moved to a new `OutputSettings` struct that can be built using the `OutputSettingsBuilder`. `WindowBuilder` was removed and replaced by a `Window::new(title, &output_settings)` function.
```

simulator/examples/chaining.rs
```diff
@@ -29,12 +31,11 @@ fn main() -> Result<(), core::convert::Infallible> {
...
     objects.draw(&mut display)?;
 
-    let mut window = WindowBuilder::new(&display)
-        .title("Chained drawing")
+    let output_settings = OutputSettingsBuilder::new()
         .theme(BinaryColorTheme::OledBlue)
         .build();
 
-    window.show_static(&display);
+    Window::new("Chained drawing", &output_settings).show_static(&display);
```
jamwaffles:
```suggestion
    Window::new("Chained drawing", &OutputSettings::default()).show_static(&display);
```

simulator/examples/chaining.rs
```diff
@@ -29,12 +31,11 @@ fn main() -> Result<(), core::convert::Infallible> {
 
     objects.draw(&mut display)?;
 
-    let mut window = WindowBuilder::new(&display)
-        .title("Chained drawing")
+    let output_settings = OutputSettingsBuilder::new()
```
jamwaffles:
```suggestion
```

simulator/examples/hello-world.rs
```diff
@@ -51,11 +53,10 @@ fn main() -> Result<(), std::convert::Infallible> {
         .into_styled(text_style)
         .draw(&mut display)?;
 
-    let mut window = WindowBuilder::new(&display)
-        .title("Hello World")
+    let output_settings = OutputSettingsBuilder::new()
```
jamwaffles:
```suggestion
```

simulator/examples/hello-world.rs
```diff
@@ -51,11 +53,10 @@ fn main() -> Result<(), std::convert::Infallible> {
...
         .theme(BinaryColorTheme::OledBlue)
         .build();
-    window.show_static(&display);
+    Window::new("Hello World", &output_settings).show_static(&display);
```
jamwaffles:
```suggestion
    Window::new("Hello World", &OutputSettings::default()).show_static(&display);
```

simulator/examples/text-custom-font.rs
```diff
@@ -36,11 +38,10 @@ fn main() -> Result<(), core::convert::Infallible> {
     )
     .draw(&mut display)?;
 
-    let mut window = WindowBuilder::new(&display)
-        .title("Custom font")
+    let output_settings = OutputSettingsBuilder::new()
         .theme(BinaryColorTheme::OledBlue)
         .build();
-    window.show_static(&display);
+    Window::new("Custom font", &output_settings).show_static(&display);
```
jamwaffles:
```suggestion
    Window::new("Custom font", &OutputSettings::default()).show_static(&display);
```

simulator/examples/text-custom-font.rs
```diff
@@ -36,11 +38,10 @@ fn main() -> Result<(), core::convert::Infallible> {
     )
     .draw(&mut display)?;
 
-    let mut window = WindowBuilder::new(&display)
-        .title("Custom font")
+    let output_settings = OutputSettingsBuilder::new()
```
jamwaffles:
```suggestion
```

simulator/generate_example_screenshots.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+OUTPUT_DIR=screenshots
```
jamwaffles: This should probably go in `../target/screenshots` (relative to the script dir), or at least be added to `.gitignore`.

simulator/src/display.rs
```diff
@@ -32,21 +33,19 @@ where
...
-        let Point { x, y } = point;
-
-        if x >= 0 && y >= 0 && (x as u32) < self.size.width && (y as u32) < self.size.height {
-            Some(x as usize + y as usize * self.size.width as usize)
-        } else {
-            None
+        if let Ok((x, y)) = <(u32, u32)>::try_from(point) {
+            if x < self.size.width && y < self.size.height {
+                return Some((x + y * self.size.width) as usize);
+            }
```
jamwaffles: Nice!

simulator/src/framebuffer.rs
```diff
@@ -0,0 +1,143 @@
...
+    style::{PrimitiveStyle, Styled},
+    DrawTarget,
+};
+use std::convert::TryFrom;
+
+/// Rgb888 framebuffer
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Framebuffer {
+    size: Size,
+    pub(crate) data: Box<[u8]>,
```
jamwaffles: I assume this is here for good reason, but why can't it be a `&'a [u8]`?

simulator/src/output_settings.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    /// Sets the pixel scale.
+    ///
+    /// A scale of `2` or higher is useful for viewing the simulator on high DPI displays.
+    pub fn scale(mut self, scale: u32) -> Self {
+        if scale == 0 {
+            panic!("scale must be >= 0");
+        }
```
jamwaffles: [`NonZeroU32`](https://doc.rust-lang.org/beta/std/num/struct.NonZeroU32.html)?

simulator/src/output_settings.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Sets the binary color theme for the display to use.
+    pub fn theme(mut self, theme: BinaryColorTheme) -> Self {
+        self.theme = theme;
+
+        // Most binary color displays are small and individual pixels are hard to recognize
+        // on higher resolution screens. So apply some default scaling when no scaling was explicitly
+        // set.
+        self.scale.get_or_insert(3);
+        self.pixel_spacing.get_or_insert(1);
```
jamwaffles: Nice, I like it.

For a future PR, SDL seems to support getting pixel density with [`display_dpi`](https://rust-sdl2.github.io/rust-sdl2/sdl2/struct.VideoSubsystem.html#method.display_dpi) which could help here.

embedded-graphics/README.md
```diff
@@ -145,11 +147,10 @@ fn main() -> Result<(), std::convert::Infallible> {
         .into_styled(text_style)
         .draw(&mut display)?;
 
-    let mut window = WindowBuilder::new(&display)
-        .title("Hello World")
+    let output_settings = OutputSettingsBuilder::new()
```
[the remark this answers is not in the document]
rfuest: Applying a color theme breaks the output of `SimulatorDisplay<RgbXXX>`, so I don't think thats a good default.

simulator/src/framebuffer.rs
rfuest: `Framebuffer` needs to own data. If a slice would be used the `new` method wouldn't be able to allocate `data`, because it would only be valid inside the `new` method.

simulator/src/output_settings.rs
rfuest: I did think about this, but having to use `.scale(NonZeroU32::new(2).unwrap())` to call the method seems a bit cumbersome IMO.

simulator/src/output_settings.rs
rfuest: This was already done before, but in a less documented way.

This PR changes the way the images are scaled and uses the e-g API for everything. I thought it was a good idea to do some dogfooding and have some place in e-g that uses the acceleration API. And this way the scaling is independent of SDL and the same code can be used for SDL and image output.

simulator/README.md
```diff
@@ -8,6 +8,11 @@
 
 The simulator can be used to test and debug [embedded-graphics](https://crates.io/crates/embedded-graphics) code, or produce snazzy examples for people to try drivers out without needing physical hardware to run on.
 
+# Cargo Features
+
+* `image-rs` - Enables support for converting a `SimulatorDisplay` to an `image` crate `ImageBuffer`.
+* `dump-png` - Changes the behavior of the `Window` struct to export the first rendered frame to a `dump.png` file instead of displaying a window.
```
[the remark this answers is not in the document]
jamwaffles: Yeah, an environment variable would be a better option I think. Better than abusing feature flags, and it allows setting the output file path.

embedded-graphics/README.md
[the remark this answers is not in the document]
jamwaffles: Noted. Should I resolve this and the related suggestions in other files too then?

simulator/src/output_settings.rs
jamwaffles: Yeah that API change is a bit ugly. You could convert it inside the method though. Not bothered either way though. Feel free to resolve this thread if you cba to change it :)

simulator/src/output_settings.rs
jamwaffles: Sounds good to me :)

simulator/src/framebuffer.rs
jamwaffles: Ah yeah sure. This can be a `Vec<u8>` if it's owned then, no?

embedded-graphics/README.md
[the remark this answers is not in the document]
rfuest: I'll resolve the others and leave this one open to remind me to add some docs that applying a color theme is a bad idea when using a color display.

simulator/src/framebuffer.rs
rfuest: Would that have any advantage? I did choose a boxed slice because `data` is a fixed size array and shouldn't change in size during the lifetime of a `Framebuffer`.

jamwaffles: Nah I guess not, the `Box<>` just looks weird :)

simulator/README.md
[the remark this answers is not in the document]
rfuest: I've now changed the code to use a `EG_SIMULATOR_DUMP` variable and I think that's much better that using a feature.

simulator/generate_example_screenshots.sh
rfuest: Writing files to a directory that isn't a subdirectory of the `simulator` directory, but of the parent directory, didn't "feel" right. But now that creating the screenshots doesn't require enabling a cargo feature anymore I was able to move the script into the root directory of the repository and use `target/screenshots` as an output directory.

simulator/src/output_settings.rs
rfuest: After trying to use `NonZeroU32` I decided to not use it at the moment and instead just document that the function panics if the scale is set to `0`.

Issue #225, opened 2019-12-22
Text positioning improvements


jamwaffles:
From https://github.com/jamwaffles/embedded-graphics/pull/224#pullrequestreview-335608934

> IMO top_left is OK for the current way e-g handles text positioning. But maybe text positioning in general should be discussed further, because most drawing libraries use the baseline to specify text positions to make it easier to mix different font sizes on a single line.

I'd like to be able to position text relative to a given point, not just it's `top_left` coordinate. For example, I might want to position text relative to the bottom-center point of the bounding box. Nine positions (denoted by blue dots) would be supported:

![FontPositioning](https://user-images.githubusercontent.com/800791/71327984-c9459200-2507-11ea-9f68-8b3cfe872501.png)

This could be specified with a new `origin` property added to `TextStyle` defined by the following enum:

```rust
enum TextOrigin {
    TopLeft,
    TopCenter,
    TopRight,
    CenterLeft,
    Center,
    CenterRight,
    BottomLeft,
    BottomCenter,
    BottomRight,
}
```

The `top_left` property should probably be renamed to something like `position` as it will now define where the text origin is put, not the top left corner.

To retain current behaviour:

```rust
egtext!(
    text = "Hello!", 
    position = Point::zero(10, 10), 
    font = Font6x8, 
    origin = TextOrigin::TopLeft
);
```

I'm open to bikeshedding the `origin` and `position` names - I'm sure there's a better pair of words out there.

rfuest:
Could we postpone these changes to `0.7`?

I would like to take a closer look at font rendering in general and add support for external font rendering libraries. But this will take a while and shouldn't block the `0.6` release IMO.

jamwaffles:
Sounds good to me. I've switched the project to 0.7.0.

bugadani:
For some reason I've missed this issue. Horizontal alignment is already implemented by embedded-text, with vertical alignment also planned for a future release.

jamwaffles:
I think it might be best to leave the behaviour originally described to an external crate, particularly now that the ball is rolling on that. I'll close this issue.

PR #270, opened 2020-03-11
Add support for horizontal and vertical text alignment

aurelj:

- [X] Add a simulator example(s) where applicable
- [X] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) and appropriate crate (`embedded-graphics`, `simulator`, `tinytga`, `tinybmp`) if your changes affect the **public API**
- [X] Run `./build.sh` (Linux/macOS only) and make sure it passes. If you use Windows, check that CI passes once you've opened the PR.

## PR description

Add support for horizontal and vertical text alignment

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -8,6 +8,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Added
 
+- [#270](https://github.com/jamwaffles/embedded-graphics/pull/270) Support horizontal and vertical text alignment.
 - [#269](https://github.com/jamwaffles/embedded-graphics/pull/269) Support for multiline text.
 - [#267](https://github.com/jamwaffles/embedded-graphics/pull/267) Added a `6x6` hand-drawn variable width font.
 - [#267](https://github.com/jamwaffles/embedded-graphics/pull/267) Support for variable width fonts.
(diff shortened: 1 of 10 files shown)


jamwaffles:
Thanks for the PR and I appreciate the time you've taken to write this, but this behaviour is outside the scope of fonts/text in e-g. I'd suggest creating your own wrapper struct for layout in your app code. The functionality proposed here would be added by something like a UI/layout library at a higher level, using e-g components as building blocks.

Both I and @rfuest want to take a deeper look at how fonts and text works in 0.7.0 (including whatever comes out of #225 which might help your use case), but I don't think padding like this PR adds will be included in fonts directly.

rfuest:
IMO we should consider adding this in 0.7. Drawing backgrounds this way works much better that our current implementation and this would be useful for simple applications that don't need any higher level UI library.

But it should be implemented as a separate drawable, like `FramedText`, and not mixed in with the current `Text`.

aurelj:
It seemed quite the obvious way to do it, but with your feedback I now totally see that it shouldn't be part of `Text`. I still think it is worth having in embedded-graphics somewhere. `FramedText` sounds like a good idea. Maybe even something more versatile like a `Layout` module that would allow positioning and aligning any kind of entity ?

Oh, and I also perfectly understand that you don't want to include such a change in 0.6, it seems sensible to wait for 0.7.

Anyway, I will put this idea at rest for a bit and think about it again in a few days with a fresh mind.

bugadani:
I've published a preview version of [embedded-text](https://crates.io/crate/embedded-text) that you may find useful @aurelj 
No vertical alignment support yet, though

jamwaffles:
This PR is pretty stale now, and I think `embedded-text` or `embedded-layout` could be good options for better text layout than a simpler solution included in e-g so I'm going to close this PR in lieu of those crates.

PR #271, opened 2020-03-15
Thick lines another go

### LINKED ISSUE #46: Line thickness support
#45 adds stroke width support to all primitives except lines. This is unexpected for the user and bad from a feature perspective so should be fixed.

### LINKED ISSUE #301: Line always has stroke width 1
No matter which value I pass to `.stroke_width()`, it always gets rendered as if the stroke width is 1. Any idea why?
```rust
Line::new(Point::new(x, 0), Point::new(x, h))
	.into_styled(
		PrimitiveStyleBuilder::new().stroke_color(Rgb888::new(255, 165, 0))
			.stroke_width(10).build()
	)
	.draw(&mut display)?;
```
image

jamwaffles:

## PR description

Finally, thick lines! This only handles single line segments, not joined lines like triangles or (later) polylines - those need to have end/join caps figured out.

This can/should probably wait till after 0.6.0 is released.

Closes #46 
Closes #301 

TODO

* [x] Clean up `line-thickness` example so it can be used for future debugging
* [x] Add some `MockDisplay` tests to `Line` to check thick drawing
* [x] Some cleanup in the algo itself, although I'm starting to miss the forest for the trees at this point.

Questions:

* [x] The `LineIterator` (unstyled) iterates over every pixel in a thick line. Should this be the case, or should it revert to only iterating over the 1px wide "ideal" line between `start` and `end`?


jamwaffles:
@rfuest This is a draft still, but could you take a look at [`line.rs`](https://github.com/jamwaffles/embedded-graphics/pull/271/files#diff-3cabc14b766e6af41ee4a0b04201a4d2) and point out anything you see that could make the code cleaner? As long as it's not fundamentally broken, I'd like to keep the shitty algorithm as is for this PR even if it's not perfect. Thick lines has kicked my ass enough recently and it would be great to see it released at some point :grin:

rfuest:
Have a look at https://github.com/rfuest/embedded-graphics/tree/thick-lines-another-go

I didn't really know what I was doing, without looking at the algorithm in more detail. But I think I've reduced some code duplication and at least the tests still work.

jamwaffles:
Ah that looks a bit better, thanks! Ready for review now I think :)

jamwaffles:
@rfuest Just a bump in case this slipped off the table. There's one unresolved question but otherwise I think this is good to go:

> The `LineIterator` (unstyled) iterates over every pixel in a thick line. Should this be the case, or should it revert to only iterating over the 1px wide "ideal" line between `start` and `end`?

Ignore the nightly failure, I guess the latest release build failed.

rfuest:
`LineIterator` couldn't be constructed for an unstyled line, because it doesn't have a stroke width. I would like to have a `LineIterator` that uses the ideal line behavior you described, but for this PR I would suggest to simply make `LineIterator` private.

jamwaffles:
> for this PR I would suggest to simply make `LineIterator` private. 

I believe it was private before (couldn't find it in the doc search), but I've changed it now to `pub(crate)` just to make sure.

> The "ideal" `LineIterator` could be added later with similar iterators for other primitives.

Sounds like a good idea to do all the primitives at the same time, yeah.

jamwaffles:
Woohoo, passing tests. Can you give this another look over?

rfuest:
What I don't like about this is how thick 45° lines are drawn, which might be a common operation. One side is always one pixel longer than the other.

Should we merge this regardless and fix this issue later? If you want to merge this now, please create an issue for this problem.

Do you want to remove the `line-thickness` example before merging? It is described as a debugging tool and might not be useful for the user of e-g, but is fun to play with.

jamwaffles:
> What I don't like about this is how thick 45° lines are drawn, which might be a common operation. One side is always one pixel longer than the other.

Hm, that shouldn't be the case. I'll take a look when I'm on my Linux machine again in a bit.

> Do you want to remove the line-thickness example before merging?

It's not an example per se, but it's been incredibly useful for debugging thick lines, so much so that I think I'll need it again in the future.

> but is fun to play with.

So let's leave it in ;)

jamwaffles:
@rfuest Fixed the 45 degree issue and added a test. Thanks for spotting that. This whole thing has been really finicky. Anything else left to fix?

rfuest:
I would really like to say there aren't any more things to fix, but I think I've found yet another problem.

If you set the stroke width to an even number of pixels and, for example, move the line between `DY = 0` and `DY = -1` the start of the line jumps. This seems to be caused by a problem with the sides, which aren't always on the left or right but somehow get mirrored in some octants.

jamwaffles:
Haha no problem, I'd rather spend the time fixing bugs :). Lemme take a look...

jamwaffles:
Alright, I think I fixed the last issue you found. I'm not too happy with the code but the tests pass and the line in the debugger demo looks good. Can you give it a look over for any more bugs or obvious cleanups/optimisations please?

jamwaffles:
Just gave this a rebase. Is this good to merge?

rfuest:
Should this be merged into `master` or the `0.7` branch? It doesn't break the API, but could be considered a breaking change because it changes the way lines are drawn.

jamwaffles:
Master IMO as it's _sort of_ a bugfix. It adds "new" (expected) behaviour. I can see why someone would treat it as a breaking change though as any > 1px thick lines will now draw correctly.

rfuest:
This could cause problems if someone used the same style with stroke width > 1px for drawing lines and other shapes and relied on the current behavior that lines are always 1 px wide.

But because that's very unlikely I'm OK with merging this into master. But maybe the changelog entry should make this clearer. Suggestion:
```
-  [#271](https://github.com/jamwaffles/embedded-graphics/pull/271)
   Styled `Line`s are now drawn using the `stroke_with` specified in the
   `PrimitiveStyle` and no longer default to 1 px width.
```

 I'll take another look at the code and will let you know if it's ready to be merged.

jamwaffles:
> Suggestion:

Thanks, I've modified the changelog entry with that text.

embedded-graphics/src/primitives/line.rs
```diff
@@ -65,76 +60,318 @@ impl Dimensions for Line {
...
+        Self { start, end }
+    }
+}
+
+/// Which side of the center line to draw on
+///
+/// Imagine standing on `start`, looking ahead to where `end` is. `Left` is to your left, `Right` to
+/// your right.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum Side {
```
rfuest: Does this need to be `pub`?

embedded-graphics/src/primitives/line.rs
```diff
@@ -65,76 +60,318 @@ impl Dimensions for Line {
...
+    /// DOCS LOL
```
rfuest: This might not be the best doc comment :wink:. If `Side` isn't `pub` anymore these comments could be removed.

embedded-graphics/src/primitives/line.rs
```diff
@@ -65,76 +60,318 @@ impl Dimensions for Line {
...
+        }
+    }
+}
+
+/// Current state of each parallel line drawn
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+struct ParallelLineState {
+    start: Point,
+    error: i32,
+    dx_accum: u32,
```
rfuest: These should also have doc comments, like `SideState`.

embedded-graphics/src/primitives/line.rs
```diff
@@ -65,76 +60,318 @@ impl Dimensions for Line {
...
+            Self::Left => *self = Self::Right,
+            Self::Right => *self = Self::Left,
+        }
+    }
+}
+
+/// Current state of each parallel line drawn
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+struct ParallelLineState {
+    start: Point,
```
rfuest: `start` is a misleading name for this field, because it's the current point and not the start of the line.

embedded-graphics/src/primitives/line.rs
```diff
@@ -65,76 +60,318 @@ impl Dimensions for Line {
...
+            dx_accum,
+        }
+    }
+}
+
+/// Current side state
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+struct SideState {
+    /// Parallel line start point
+    start: Point,
```
rfuest: See comment above.

jamwaffles: Changed to `parallel_start`, because it's used as the starting point of a `ParallelLineState`. Open to better suggestions :)

embedded-graphics/src/primitives/line.rs
jamwaffles: Renamed to `current_point`.

rfuest:
I've tried to refactor the code a bit, let me know if you think that this is any better: https://github.com/rfuest/embedded-graphics/tree/thick-lines-another-go

jamwaffles:
Oh that's loads better, thanks! I've rebased those changes in and force pushed my branch. I like the split between all the thick line stuff and the styled line. I didn't look too deeply at the code but it feels quite a bit more understandable.

jamwaffles:
Woohoo!

> it's finished enough

High praise from mr fuest ;)

rfuest:
> High praise from mr fuest ;)

You need to remember that I'm german, so this was really euphoric ;)

PR #279, opened 2020-03-20
Add example generation harness

### LINKED ISSUE #223: Add table of drawing operations to docs

jamwaffles:

## PR description

* [x] Adds example screenshot generator
* [x] Adds table of drawing operations to `lib.rs`

Closes #223

### DIFF
--- a/embedded-graphics/src/lib.rs
+++ b/embedded-graphics/src/lib.rs
@@ -191,7 +191,10 @@
 //! <div style="flex-grow: 1;">
 //!
 //! ```rust
-//! Pixel(Point::new(32, 32), Rgb888::GREEN).draw(&mut display)
+//! # let mut display = embedded_graphics::mock_display::MockDisplay::default();
+//! use embedded_graphics::{pixelcolor::Rgb888, prelude::*};
+//! Pixel(Point::new(32, 32), Rgb888::GREEN).draw(&mut display)?;
+//! # Ok::<(), core::convert::Infallible>(())
 //! ```
 //!
 //! </div>
@@ -206,7 +209,10 @@
 //! <div style="flex-grow: 1;">
 //!
 //! ```rust
-//! use embedded_graphics::{primitives::Line, style::PrimitiveStyleBuilder};
+//! # let mut display = embedded_graphics::mock_display::MockDisplay::default();
+//! use embedded_graphics::{
+//!     pixelcolor::Rgb888, prelude::*, primitives::Line, style::PrimitiveStyleBuilder,
+//! };
 //! Line::new(Point::new(16, 24), Point::new(48, 40))
 //!     .into_styled(
 //!         PrimitiveStyleBuilder::new()
@@ -214,7 +220,8 @@
 //!             .stroke_color(Rgb888::RED)
 //!             .build(),
 //!     )
-//!     .draw(&mut display)
+//!     .draw(&mut display)?;
+//! # Ok::<(), core::convert::Infallible>(())
 //! ```
 //!
 //! </div>
@@ -229,7 +236,10 @@
 //! <div style="flex-grow: 1;">
 //!
 //! ```rust
-//! use embedded_graphics::{primitives::Rectangle, style::PrimitiveStyleBuilder};
+//! # let mut display = embedded_graphics::mock_display::MockDisplay::default();
+//! use embedded_graphics::{
+//!     pixelcolor::Rgb888, prelude::*, primitives::Rectangle, style::PrimitiveStyleBuilder,
+//! };
 //! Rectangle::new(Point::new(16, 24), Point::new(48, 40))
 //!     .into_styled(
 //!         PrimitiveStyleBuilder::new()
@@ -238,7 +248,8 @@
 //!             .fill_color(Rgb888::CYAN)
 //!             .build(),
 //!     )
-//!     .draw(&mut display)
+//!     .draw(&mut display)?;
+//! # Ok::<(), core::convert::Infallible>(())
 //! ```
 //!
 //! </div>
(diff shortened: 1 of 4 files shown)


rfuest:
Why do you add the examples to a `HashMap` before running them? Did you plan to reuse the hash map for anything but running the examples?

Here is a test I did using some macro magic to generate images and prints rust docs to stdout: https://gist.github.com/rfuest/d6c7a1f365ae5e90a77d4e6c27c9df6d
Note the changed run command, because stringifying code blocks gives ugly results and needs rustfmt to make the output nicer.

jamwaffles:
> Why do you add the examples to a HashMap before running them? Did you plan to reuse the hash map for anything but running the examples?

No other use, it was just easier to loop through them than copy and paste the boilerplate.

> Here is a test I did

I didn't know about `stringify!()`. That's awesome. I'll fold in the macro magic and see what else needs doing.

jamwaffles:
Here's the output so far:

image

The homepage is gonna get very long very quickly if there are a lot of examples to be added... thoughts? Submodule?

Also, do you have a list of examples you'd like to see included?

rfuest:
> The homepage is gonna get very long very quickly if there are a lot of examples to be added... thoughts? Submodule?

I don't really have a good idea for this and this was the reason I mentioned an e-g book in #223. But until someone wants to start writing one we need another solution. Maybe a markdown file in the repo and a link to that in the docs?

> Edit: I could create a literal table of all the examples with images next to each. That would shorten the content significantly

That was my initial idea. IMO we should try if that works.

> Also, do you have a list of examples you'd like to see included?

Let's start with one example per drawable (primitives, text and image). That should give a good first impression and additional examples can easily be added later.

jamwaffles:
image

Getting there!

rfuest:
That looks much better!

jamwaffles:
Hokay, this has turned out to be a gigantic pain in the arse.

I went with flexbox instead of tables because the latter are irritating af to tweak well. Still looks good IMO, and it doesn't increase the `lib.rs` doc length by too much.

The last issue is the image example. I went with tinytga which is fine, but including the file with a relative path only works in either one of the generator _or_ `lib.rs`, because they're in different depths relative to the image. Any suggestions?

Screen Shot 2020-03-25 at 17 16 15-fullpage

rfuest:
> The last issue is the image example. I went with tinytga which is fine, but including the file with a relative path only works in either one of the generator _or_ `lib.rs`, because they're in different depths relative to the image. Any suggestions?

Could you use a path relative to `CARGO_MANIFEST_DIR`? 
https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

jamwaffles:
Buh I new there was an env var somewhere. Thanks for finding that.

I fixed it in latest commit as far as I can see. You'll notice I have to go up one dir then back down into `simulator` to normalise the path - `CARGO_MANIFEST_DIR` includes the crate path instead of just pointing at the workspace.

simulator/src/bin/generate-example-screenshots.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        // Note: empty lines must remain between HTML elements and inner Markdown for the Markdown
+        // to be parsed correctly.
+        println!(
+            r#"//! ## {}
+            //!
+            //! {}
+            //!
+            //! <div style="display: flex">
+            //! <img style="width: 128px; height: 128px; display: inline-block; margin-right: 8px;" alt="{} example screenshot" src="data:image/png;base64,{}" />
```
rfuest: Why do you use `inline-block` for this? I thought this wouldn't make sense when used with flexbox.

simulator/src/bin/generate-example-screenshots.rs
```diff
@@ -0,0 +1,216 @@
...
+fn main() {
+    let output_base = "./target/drawing-ops";
+    std::fs::create_dir_all(output_base).expect("Output directory could not be created");
+
+    let mut display: SimulatorDisplay<pixelcolor::Rgb888> =
+        SimulatorDisplay::new(Size::new(64, 64));
+
+    op!(
+        display,
+        "Draw a single pixel",
```
rfuest: Should we add a note here that this isn't a preferred way of drawing multiple pixels, but users should instead implement their own iters?

If a display driver implements some form of caching for `draw_iter` and doesn't use an internal framebuffer, drawing pixels this way might be much slower that using an iter. But for prototypes or when performance isn't too important drawing individual pixels this way is fine.

simulator/src/bin/generate-example-screenshots.rs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            use embedded_graphics::{
+                pixelcolor::Rgb888, prelude::*, primitives::Circle, style::PrimitiveStyleBuilder,
+            };
+
+            Circle::new(Point::new(22, 22), 20)
+                .into_styled(
+                    PrimitiveStyleBuilder::new()
+                        .fill_color(Rgb888::BLUE)
+                        .build(),
```
rfuest: Should we use `PrimitiveStyle::with_fill(Rgb888::BLUE)` instead? IMO this would be recommended for simple styles.

simulator/src/bin/generate-example-screenshots.rs
```diff
@@ -0,0 +1,216 @@
...
+            use embedded_graphics::{
+                pixelcolor::Rgb888, prelude::*, primitives::Triangle, style::PrimitiveStyleBuilder,
+            };
+
+            Triangle::new(Point::new(32, 16), Point::new(16, 48), Point::new(48, 48))
+                .into_styled(
+                    PrimitiveStyleBuilder::new()
+                        .stroke_width(1)
+                        .stroke_color(Rgb888::MAGENTA)
+                        .build(),
```
rfuest:
```suggestion
                    PrimitiveStyle::with_stroke(Rgb888::MAGENTA, 1),
```

simulator/src/bin/generate-example-screenshots.rs
```diff
@@ -0,0 +1,216 @@
...
+                fonts::{Font6x8, Text},
+                pixelcolor::Rgb888,
+                prelude::*,
+                style::TextStyleBuilder,
+            };
+
+            // Create a new text style
+            let style = TextStyleBuilder::new(Font6x8)
+                .text_color(Rgb888::GREEN)
+                .build();
```
rfuest: This could also be replaced by `TextStyle::new`.

simulator/src/bin/generate-example-screenshots.rs
jamwaffles: Yeah that might be a good idea. Can you come up with a sentence or two and I can put that in?

jamwaffles: I've added a small note to this example so let me know what you think. I'm wary of adding too much text here as I don't want to make the page too long.

rfuest: Sorry I didn't reply earlier. Maybe something like this could work?

```
For cases where many pixels are drawn it is preferable to implement
a custom iterator instead of calling `Pixel::draw` for each pixel, because
some display drivers implement accelerated drawing of iterators.
```

jamwaffles: SGTM :) I've added that in so this is good for another look over :ok_hand:

rfuest:
Can you also update `lib.rs` with the changes made in the last commit?

jamwaffles:
Argh missed it again. Done!

rfuest:
I didn't like that the `stringify` macro removes all newlines from the examples. I tried to fix this by adding a hack that replaced `{}` with a newline when generating the output: https://github.com/rfuest/embedded-graphics/tree/examples-table
The repo also contains a new `rustfmt.examples.toml` which sets `imports_layout = "Vertical"` to reduce the width of the examples and prevents overflow in the `Rectangle` example.

jamwaffles:
I noticed that too >:C

I've rebased your changes onto this branch like the thick lines PR. The formatting changes are a neat trick. Nice!

embedded-graphics/src/lib.rs
```diff
@@ -171,16 +177,214 @@
...
+//!     },
+//!     pixelcolor::Rgb888,
+//!     prelude::*,
+//!     style::TextStyle,
+//! };
+//!
+//! let style = TextStyle::new(Font6x8, Rgb888::GREEN);
+//!
+//! Text::new("Hello,\nRust!", Point::new(2, 28))
+//!     .into_styled(style)
```
rfuest, line 349: To be consistent with the examples that use `PrimitiveStyle` this shouldn't use a `style` variable, but directly call `TextStyle::new` inside the `into_styled` call.

jamwaffles: Well caught. I've made this change and rebased from master.

rfuest:
You've accidentally added `xclip` to the repo.

jamwaffles:
No I didn't! History is written by ~the winners~ git rebase :sunglasses:

simulator/src/bin/generate-example-screenshots.rs
```diff
@@ -0,0 +1,211 @@
...
+                pixelcolor::Rgb888, prelude::*, primitives::Line, style::PrimitiveStyleBuilder,
+            };
+            {}
+            Line::new(Point::new(16, 24), Point::new(48, 40))
+                .into_styled(
+                    PrimitiveStyleBuilder::new()
+                        .stroke_width(1)
+                        .stroke_color(Rgb888::RED)
+                        .build(),
+                )
```
rfuest, line 111: I missed this one.
```suggestion
                .into_styled(PrimitiveStyle::with_stroke(Rgb888::RED, 1))
```

Also, should we change this to something other than 1px width, now that the fabulous PR #271 is merged?

jamwaffles: Good idea. I've updated this example to use an 8px thick line. I moved the end point slightly to give the line a better look for the example's sake.

PR #285, opened 2020-03-24
Add scalar div and mul for Point and Size

rfuest:

## PR description

This PR adds multiplication and division by a scalar for `Point` and `Size`.

@jamwaffles  I think we've discussed adding this before, but I'm not sure if you thought that this is a good idea. But in my recent experience with trying to implementing code using the geometry types I've noticed that this could often be useful.

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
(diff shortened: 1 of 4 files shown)


rfuest:
> For the sake of completeness, should we add `Mul<u32> for Point` and `Div<u32> for Point`? If you don't think that's a good idea, feel free to merge without.

This might be useful sometimes, but I wouldn't add it at the moment, because rust also doesn't implement `i32 * u32`. But it's independent of the changes in this PR and can be added later if we realize that it's commonly used.

PR #286, opened 2020-03-24
Change Rectangle to use top_left and size

rfuest:

## PR description

This PR changes `Rectangle` to use `top_left` and `size` instead of `top_left` and `bottom_right`. A similar change for `Circles` was recently implemented in #147.

As a part of the changes for #182 some of the code in this PR isn't final. Some problems like the bottom right corner problem mentioned in #182 will need to be addressed, but this will require changes in the `Dimensions` trait and is outside the scope of this PR IMO.

Questions:
- [ ] To ease porting to the new API, I've added `Rectangle::with_corners`. I've used `corner_1` and `corner_2` instead of `top_left` and `bottom_right` as parameter names, because any two corners can be used to construct a rectangle. To reflect this change the `egrectangle` macro now uses a `corner` parameter with a points array, like the `egtriangle` macro. Is this a good idea?

### DIFF
--- a/embedded-graphics/src/primitives/line.rs
+++ b/embedded-graphics/src/primitives/line.rs
@@ -54,7 +54,7 @@ impl Dimensions for Line {
     }
 
     fn bottom_right(&self) -> Point {
-        self.top_left() + self.size()
+        self.top_left() + (self.size() - Size::new(1, 1))
     }
 
     fn size(&self) -> Size {
@@ -257,18 +257,18 @@ mod tests {
     #[test]
     fn bounding_box() {
         let start = Point::new(10, 10);
-        let end = Point::new(20, 20);
+        let end = Point::new(19, 29);
 
         let line: Line = Line::new(start, end);
         let backwards_line: Line = Line::new(end, start);
 
         assert_eq!(line.top_left(), start);
         assert_eq!(line.bottom_right(), end);
-        assert_eq!(line.size(), Size::new(10, 10));
+        assert_eq!(line.size(), Size::new(10, 20));
 
         assert_eq!(backwards_line.top_left(), start);
         assert_eq!(backwards_line.bottom_right(), end);
-        assert_eq!(backwards_line.size(), Size::new(10, 10));
+        assert_eq!(backwards_line.size(), Size::new(10, 20));
     }
 
     #[test]
--- a/embedded-graphics/src/primitives/mod.rs
+++ b/embedded-graphics/src/primitives/mod.rs
@@ -202,11 +202,11 @@ macro_rules! egline {
 /// };
 ///
 /// let empty_rect: Styled<Rectangle, PrimitiveStyle<Rgb565>> =
-///     egrectangle!(top_left = (10, 20), bottom_right = (30, 40));
+///     egrectangle!(top_left = (10, 20), size = (20, 20));
 ///
 /// let filled_rect: Styled<Rectangle, PrimitiveStyle<Rgb565>> = egrectangle!(
 ///     top_left = (10, 20),
-///     bottom_right = (30, 40),
+///     size = (20, 20),
 ///     style = primitive_style!(stroke_color = Rgb565::RED, fill_color = Rgb565::GREEN)
 /// );
 /// ```
@@ -228,7 +228,7 @@ macro_rules! egline {
 ///
 /// let rectangle_1: Styled<Rectangle, PrimitiveStyle<Rgb565>> = egrectangle!(
 ///     top_left = (10, 20),
-///     bottom_right = (30, 40),
+///     size = (20, 20),
 ///     style = primitive_style!(
 ///         stroke_color = Rgb565::RED,
 ///         fill_color = Rgb565::GREEN,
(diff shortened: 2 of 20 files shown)


jamwaffles:
> To reflect this change the egrectangle macro now uses a corner parameter with a points array, like the egtriangle macro. Is this a good idea?

I think that's ok. What reservations do you have? If there's a method that can create a rectangle from two points still, then the macro should probably support the same.

This PR needs a rebase from master. I'll give it a review now.

embedded-graphics/src/primitives/line.rs
```diff
@@ -54,7 +54,7 @@ impl Dimensions for Line {
     }
 
     fn bottom_right(&self) -> Point {
-        self.top_left() + self.size()
+        self.top_left() + (self.size() - Size::new(1, 1))
```
jamwaffles: Will this off-by-one be resolved with the other work around #182?

embedded-graphics/src/primitives/mod.rs
```diff
@@ -243,23 +243,36 @@ macro_rules! egline {
...
+            style = $crate::style::PrimitiveStyle::default()
+        )
+    }};
+    (corners = [$corner_1:expr, $corner_2:expr], style = $style:expr $(,)?) => {{
+        $crate::primitives::Rectangle::with_corners(
+            $crate::geometry::Point::from($corner_1),
+            $crate::geometry::Point::from($corner_2),
+        )
+        .into_styled($style)
+    }};
```
jamwaffles: Would `box` or `bounding_box` be a better name perhaps? The `fn` version is called `from_bounding_box` so some consistency with that might be good. Naming this param is pretty hard.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -60,20 +61,35 @@ impl Dimensions for Rectangle {
     }
 
     fn bottom_right(&self) -> Point {
-        self.bottom_right
+        if self.size == Size::zero() {
+            // FIXME: The bottom right corner isn't valid if the size of the rectangle is zero,
+            //        because creating a rectangle from two corner points creates a rectangle that
+            //        is at least 1x1 pixels large.
```
jamwaffles: Fixed by #182 as well?

embedded-graphics/src/primitives/line.rs
rfuest: Subtracting `(1, 1)` should make this consistent with #182.

embedded-graphics/src/primitives/mod.rs
rfuest: I want to reserve the `bounding_box` name for something else, so I went with corners for this.

embedded-graphics/src/primitives/rectangle.rs
rfuest: This is one of the problems mentioned in #182 that needs to be fixed later.

rfuest:
> I think that's ok. What reservations do you have?

Just that it's a breaking change. The old version used `top_left` and `bottom_right` instead, but also supported using any two corners, which wasn't clear by looking at the names.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -60,20 +61,35 @@ impl Dimensions for Rectangle {
...
 impl Rectangle {
-    /// Create a new rectangle from the top left point to the bottom right point with a given style
-    pub const fn new(top_left: Point, bottom_right: Point) -> Self {
+    /// Creates a new rectangle from the top left point and the size.
+    pub const fn new(top_left: Point, size: Size) -> Self {
+        Rectangle { top_left, size }
+    }
+
+    /// Creates a new rectangle from two corners.
+    pub fn with_corners(corner_1: Point, corner_2: Point) -> Self {
```
jamwaffles:
```suggestion
    pub fn from_corners(corner_1: Point, corner_2: Point) -> Self {
```

Fits Rust API guidelines better, and we already have stuff like `Triangle::from_points`.

rfuest: Does it? I guess this is a case that's not 100% defined. This were my thoughts when choosing this name:
`from_...` is a conversion constructor and IMO two points can't be converted into a rectangle, but  a rectangle can be constructed by specifying two of it's corners. And as a general constructor with additional details it's supposed to use a `with_` prefix.

jamwaffles: Could be argued either way I spose. Let's leave it as `with_corners` then :). Once the changelog item is added I'll approve the PR. Needs (another) rebase too 😰

jamwaffles:
I've changed the merge branch to `v0.7.0`. I was also going to rebase it from `master`, but that feels dangerous if you're doing things with it as well. Thoughts on how to proceed?

embedded-graphics/src/primitives/rectangle.rs
rfuest: I did just another look at this and maybe we should find a more consistent naming scheme.

`Triangle::from_points` is different than this, because it uses the same points to define a triangle as `Triangle::new` but as an array instead of three arguments. So maybe `Triangle::from_points` shouldn't have been named this way.

Suggestion:
* Use `vertex` instead of `corner`
* Change `Triangle` to also use `vertex` instead of the generic `points`.
  This would also make it possible to use `points()` for the iterator suggested in #156, with a lower risk of confusion between vertices and points.

jamwaffles: If I understand correctly:

* Change `Triangle::from_points` to `Triangle::from_vertices`. This continues to take a list of 3 points. This might also wanna be `Triangle::with_vertices` if we're going for consistency?
* Change `Rectangle::with_corners` to `Rectangle::with_vertices`. Does this remain as two arguments or should we change it to take a list to be consistent with `Triangle`?

?

rfuest: > If I understand correctly:

I'm not quite sure myself and open to better suggestions.

> * Change `Triangle::from_points` to `Triangle::from_vertices`. This continues to take a list of 3 points. This might also wanna be `Triangle::with_vertices` if we're going for consistency?

Maybe we should make `Triangle` behave consistently with `Polygon`. This would probably mean that `Triangle::new` and `Poygon::new` would have a slice as an argument. `Triangle::with_vertices` could then be used for specifying three individual points.

> Change `Rectangle::with_corners` to `Rectangle::with_vertices`. Does this remain as two arguments or should we change it to take a list to be consistent with `Triangle`?

If we change `Triangle` like in the above comment `Rectangle::with_vertices` could still use two arguments. Creating a rectangle from a slice of two points isn't a common operation IMO.

jamwaffles:
Argh! I can't change the branch now that `v0.7.0` is deleted! Feel free to create a new PR if you can't fix it. Oops.

rfuest:
No, I don't think I can fix it and I'll create a new one.

jamwaffles:
Thanks, apologies for the extra work

Issue #288, opened 2020-03-26
0.6.x releases in tandem with 0.7.x work


jamwaffles:
Before we get too far ahead, there will be fixes and improvements that can be released as part of the 0.6.x series whilst 0.7.x breaking changes happen in parallel. To this end, should we create a 0.7 feature branch to merge breaking PRs into? There are already breaking changes in master, so I'd have to branch off an earlier commit and force push master. Is there a better/preferred way?

rfuest:
Force pushing master seems like a bad idea IMO. For now I would suggest to keep doing 0.7 development in master and create an 0.6 branch for 0.6 patch releases. We can reconsider changing the way we use branches after 0.7 is released.

jamwaffles:
Sure, that makes sense. Cherry picking stuff from master into 0.6 should work ok.

jamwaffles:
On second thoughts, I would prefer to keep releasing from master, and work in a 0.7 feature branch. I think we'll run into problems with git tags and `cargo-release` and probably other stuff later if we don't.

Splitting off 0.7.0 into a feature branch would require one force push to move the `Circle` changes in master, which I think is acceptable. Force pushing isn't great, but the benefits outweigh the drawbacks this time I think?

rfuest:
I haven't used `cargo-release` before, so you can better judge if this will cause any problems.

We could use `git revert` to add a new commit that removes the circle changes instead of force pushing. IMO that would be a cleaner solution.

jamwaffles:
Ah, yeah a revert would work. Let me rustle a PR up.

rfuest:
Can you also create the development branch and push the circle changes to that branch? This would allow me to finish the rectangle PR.

jamwaffles:
Yep, I just pushed the [`v0.7.0`](https://github.com/jamwaffles/embedded-graphics/tree/v0.7.0) branch. I reverted the revert for the circle PR on it which is messy but should get us back to the correct state and (hopefully) make it easy to rebase master onto v0.7.0 to keep everything in sync.

jamwaffles:
Do we want to create a long lived PR for that branch or just leave it in git for now? Might useful for us and other watchers to see what's going to be in 0.7.0 if we do create it.

rfuest:
[...]

PR #274, opened 2020-03-18
circle: define by top-left + diameter instead of center + radius

aurelj:
Hi! Thank you for helping out with Embedded Graphics development! Please:

- [X] Check that you've added passing tests and documentation
- [X] Add a simulator example(s) where applicable
- [x] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) and appropriate crate (`embedded-graphics`, `simulator`, `tinytga`, `tinybmp`) if your changes affect the **public API**
- [X] Run `rustfmt` on the project
- [X] Run `./build.sh` (Linux/macOS only) and make sure it passes. If you use Windows, check that CI passes once you've opened the PR.

## PR description

This changes the API of `Circle` so that it is defined by `top-left` and `diameter` rather than by `center` and `radius`.

This is a breaking change. It should be 0.7.0 material.

This fixes #147

PR #290, opened 2020-03-27
Revert "Define circles by top-left + diameter instead of center + rad…

jamwaffles:
This reverts the merge of #274 into master so that it may be cherry picked onto a 0.7.0 feature branch. 

Cc #288

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,10 +10,6 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 - [#285](https://github.com/jamwaffles/embedded-graphics/pull/285) Add multiplication and division by a scalar for `Point` and `Size`.
 
-### Changed
-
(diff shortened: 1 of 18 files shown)


jamwaffles:
@rfuest We should probably merge this before anything else that's currently open.

rfuest:
> @rfuest We should probably merge this before anything else that's currently open.

Yes and I guess we also need to create new PRs, because I don't think there is a way to change the target branch.

jamwaffles:
> I don't think there is a way to change the target branch.

The Github UI allows the base branch of the PR to be changed if that's what you mean.

rfuest:
I just wanted to compare the `revert-breaking-master-change` branch with the `0.6.0` tag, but there is no `0.6.0` tag. The latest tagged version is `beta.2`

jamwaffles:
Huh oh yeah. Let me create one and push it

jamwaffles:
Pushed!

PR #292, opened 2020-03-27
Add Polyline primitive

### LINKED ISSUE #77: Unfilled polygon support
It should be possible to create an open and closed line loop. Filled polygons can be added in a separate issue.

jamwaffles:

## PR description

This PR adds the `Polyline` primitive, allowing non-filled shapes to be drawn.

Polylines won't yet support stroke thickness even after #271 is merged as we need a good solution for line caps/joiners which I'll tackle later on.

Closes #77

### DIFF
--- a/embedded-graphics/src/primitives/polyline.rs
+++ b/embedded-graphics/src/primitives/polyline.rs
@@ -257,7 +257,11 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{geometry::Size, pixelcolor::Rgb565, prelude::*};
+    use crate::{
+        mock_display::MockDisplay,
+        pixelcolor::{BinaryColor, Rgb565},
+        prelude::*,
+    };
 
     // A "heartbeat" shaped polyline
     const HEARTBEAT: [Point; 10] = [
@@ -271,6 +275,14 @@ mod tests {
         Point::new(110, 84),
         Point::new(120, 64),
         Point::new(300, 64),
+    ];
+
+    // Smaller test pattern for mock display
+    const SMALL: [Point; 4] = [
+        Point::new(2, 5),
+        Point::new(5, 2),
+        Point::new(10, 5),
+        Point::new(15, 2),
     ];
 
     #[test]
@@ -330,5 +342,50 @@ mod tests {
         let thin = polyline.into_styled(PrimitiveStyle::with_stroke(Rgb565::RED, 1));
 
         assert!(thick.into_iter().eq(thin.into_iter()));
+    }
+
+    #[test]
+    fn mock_display() {
+        let mut display: MockDisplay<BinaryColor> = MockDisplay::new();
+
+        Polyline::new(&SMALL)
+            .into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 1))
+            .draw(&mut display)
+            .unwrap();
+
+        assert_eq!(
+            display,
+            MockDisplay::from_pattern(&[
+                "                ",
+                "                ",
+                "     #         #",
+                "    # ##     ## ",
+                "   #    ## ##   ",
+                "  #       #     ",
+            ])
+        );
+    }
+
+    // Ensure that consecutive points are always different
+    #[test]
+    fn no_duplicate_points() {
+        let expected: [Point; 14] = [
+            Point::new(2, 5),
+            Point::new(3, 4),
+            Point::new(4, 3),
+            Point::new(5, 2),
+            Point::new(6, 3),
+            Point::new(7, 3),
+            Point::new(8, 4),
+            Point::new(9, 4),
+            Point::new(10, 5),
+            Point::new(11, 4),
+            Point::new(12, 4),
+            Point::new(13, 3),
+            Point::new(14, 3),
+            Point::new(15, 2),
+        ];
+
+        assert!(Polyline::new(&SMALL).points().eq(expected.iter().copied()))
     }
 }
--- a/embedded-graphics/src/primitives/polyline.rs
+++ b/embedded-graphics/src/primitives/polyline.rs
@@ -5,7 +5,10 @@ use crate::{
     drawable::{Drawable, Pixel},
     geometry::{Dimensions, Point},
     pixelcolor::PixelColor,
-    primitives::{line::Line, thick_line_iterator::ThickLineIterator, Primitive, Rectangle},
+    primitives::{
+        line::{self, Line},
+        Primitive, Rectangle,
+    },
     style::{PrimitiveStyle, Styled},
     transform::Transform,
 };
@@ -156,7 +159,7 @@ impl<'a> Transform for Polyline<'a> {
 pub struct Points<'a> {
     vertices: &'a [Point],
     translate: Point,
-    segment_iter: ThickLineIterator,
+    segment_iter: line::Points,
 }
 
 impl<'a> Points<'a> {
@@ -167,7 +170,7 @@ impl<'a> Points<'a> {
         Points {
             vertices: polyline.vertices,
             translate: polyline.translate,
-            segment_iter: ThickLineIterator::new(&Line::new(Point::zero(), Point::zero()), 1),
+            segment_iter: Line::new(Point::zero(), Point::zero()).points(),
         }
     }
 }
@@ -188,10 +191,7 @@ impl<'a> Iterator for Points<'a> {
 
             self.vertices = rest;
 
-            self.segment_iter = ThickLineIterator::new(
-                &Line::new(*start + self.translate, *end + self.translate),
-                1,
-            );
+            self.segment_iter = Line::new(*start + self.translate, *end + self.translate).points();
 
             // Skip first point of next line, otherwise we overlap with the previous line
             self.nth(1)
--- a/embedded-graphics/src/primitives/polyline.rs
+++ b/embedded-graphics/src/primitives/polyline.rs
@@ -1,0 +1,235 @@
+//! The line primitive
+
+// TODO: Group imports
+use crate::draw_target::DrawTarget;
+use crate::drawable::Drawable;
+use crate::drawable::Pixel;
+use crate::geometry::Dimensions;
+use crate::geometry::Size;
+use crate::pixelcolor::PixelColor;
+use crate::primitives::Primitive;
+use crate::style::PrimitiveStyle;
+use crate::style::Styled;
+use crate::{
+    geometry::Point,
+    primitives::{line::Line, thick_line_iterator::ThickLineIterator},
+};
+
+/// Polyline primitive
+///
+/// Creates an unfilled chained line shape
+///
+/// # Examples
+///
+/// ## Create some lines with different styles
+///
+/// ```rust
+/// use embedded_graphics::{
+///     pixelcolor::Rgb565, prelude::*, primitives::Line, style::PrimitiveStyle,
+/// };
+/// # use embedded_graphics::mock_display::MockDisplay;
+/// # let mut display = MockDisplay::default();
+///
+/// // TODO: Example
+/// # Ok::<(), core::convert::Infallible>(())
+/// ```
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
+pub struct Polyline<'a> {
+    /// All vertices in the line
+    pub vertices: &'a [Point],
+}
+
+impl<'a> Polyline<'a> {
+    /// Create a new polyline from a list of vertices
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the number of vertices is less than 2
+    pub fn new(vertices: &'a [Point]) -> Self {
+        if vertices.len() < 2 {
+            panic!("Polyline must contain at least two vertices");
+        }
+
+        Self { vertices }
+    }
+}
+
+impl<'a> Primitive for Polyline<'a> {}
+
+impl<'a> Dimensions for Polyline<'a> {
+    fn top_left(&self) -> Point {
+        self.vertices
+            .iter()
+            .fold(Point::new(core::i32::MAX, core::i32::MAX), |accum, v| {
+                Point::new(accum.x.min(v.x), accum.y.min(v.y))
+            })
+    }
+
+    fn bottom_right(&self) -> Point {
+        self.vertices
+            .iter()
+            .fold(Point::new(core::i32::MIN, core::i32::MIN), |accum, v| {
+                Point::new(accum.x.max(v.x), accum.y.max(v.y))
+            })
+    }
+
+    fn size(&self) -> Size {
+        Size::from_bounding_box(self.top_left(), self.bottom_right())
+    }
+}
+
+/// TODO: Docs
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct PolylineIterator<'a> {
+    stop: bool,
+    vertices: &'a [Point],
+    segment_iter: ThickLineIterator,
+}
+
+impl<'a> Iterator for PolylineIterator<'a> {
+    type Item = Point;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.stop {
+            return None;
+        }
+
+        if let Some(p) = self.segment_iter.next() {
+            Some(p)
+        } else {
+            let (start, rest) = self.vertices.split_first()?;
+            let end = rest.get(0)?;
+
+            self.vertices = rest;
+
+            self.segment_iter = ThickLineIterator::new(&Line::new(*start, *end), 1);
+
+            Self::next(self)
+        }
+    }
+}
+
+impl<'a> IntoIterator for Polyline<'a> {
+    type Item = Point;
+    type IntoIter = PolylineIterator<'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.vertices
+            .split_first()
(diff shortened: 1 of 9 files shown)


embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+impl<'a> Polyline<'a> {
+    /// Create a new polyline from a list of points or an iterator
+    pub fn new(points: &'a [Point]) -> Self {
+        Self { points }
+    }
+
+    /// Get the total number of points in this polyline
+    pub fn len(&self) -> usize {
```
rfuest, line 47: `len` could be misinterpreted as the length of the polyline in pixels. If `points` is a public field this function might not be necessary at all.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,189 @@
...
+    line: Polyline<'a>,
+    segment_iter: LineIterator,
+    start_idx: usize,
+    end_idx: usize,
+}
+
+impl<'a> Iterator for PolylineIterator<'a> {
+    type Item = Point;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
rfuest, line 81: It might be possible to simplify this by using https://doc.rust-lang.org/std/primitive.slice.html#method.windows

jamwaffles: I tried this originally, but the code ends up something like this:

```rust
self.points
    .windows(2)
    .map(|[start, end, ..]| ThickLineIterator::new(&Line::new(*start, *end), 1))
    .flatten()
```

Storing that type in the iterator struct is _gnarly_, so I went with the hand made method instead. If you can figure a way to use `.windows()` instead that would be great.

rfuest: I think that stroing a closure inside a struct doesn't work at all. You would need to implement the mapping as a type:
```rust
struct LinesFromSlices<I> {
    iter: I,
}

impl<'a, I> Iterator for LinesFromSlices<I>
where
    I: Iterator<Item = &'a [Point]>,
{
    type Item = ThickLineIterator;

        self.iter
            .next()
            .map(|[start, end, ..]| ThickLineIterator::new(&Line::new(*start, *end), 1))
    }
}
```
The type stored in `PolylineIterator` would be: `Flatten<LinesFromSlices<Windows<'a, Point>>>`

jamwaffles: That kind of works, however I have to do this:

```rust
#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
struct LinesFromSlices<I>(I);

impl<'a, I> Iterator for LinesFromSlices<I>
where
    I: Iterator<Item = &'a [Point]>,
{
    type Item = ThickLineIterator;

        self.0.next().map(|parts| match parts {
            [start, end] => ThickLineIterator::new(&Line::new(*start, *end), 1),
            _ => unreachable!(),
        })
    }
}
```

I don't like having to add `unreachable!()` in there. Somehow `Windows` fixes that but I haven't looked into how/why.

We also lose nearly every derive except `Clone` and `Debug` when storing `Flatten<LinesFromSlices<Windows<'a, Point>>>` in a struct. Doesn't matter a great deal, but this is the only primitive with an iter that has this few derives IIRC.

IMO this change doesn't give much benefit over the "custom windows" implementation so I'd rather leave it as is. Is there a reason the existing code is dangerous or unclean that you'd like to see fixed?

rfuest: > IMO this change doesn't give much benefit over the "custom windows" implementation so I'd rather leave it as is. Is there a reason the existing code is dangerous or unclean that you'd like to see fixed?

No, I just thought it might result in easier to understand code.

Another way of simplifying would be by using a points slice in the iterator instead of the `Polyline`. The next line segment could always be constructed from the first two elements in the slice and then the first point could be removed from the slice. IMO that would be easier to understand than using the two index variables.

jamwaffles:
Ready for a first review @rfuest :pray:

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,358 @@
...
+
+    #[test]
+    fn positive_dimensions() {
+        let polyline = Polyline::new(&HEARTBEAT);
+
+        let bb = polyline.bounding_box();
+
+        assert_eq!(bb.top_left, Point::new(10, 10));
+        assert_eq!(bb.top_left + bb.size, Point::new(301, 85));
+        assert_eq!(bb.size, Size::new(291, 75));
```
jamwaffles, line 310: Is this correct? I'm not sure if it's an off-by-one error, or something related with #182.

rfuest: Yes, the size is correct. But I would remove the `bb.top_left + bb.size` test, because that's not the correct way to calculate the bottom right corner of the bounding box.

jamwaffles: Cool, thanks for confirming. I see `bottom_right` referenced in a few places, but it's not a method or property on `Rectangle`. What's the correct way?

rfuest: The correct way is to use the soon to be added `bottom_right` getter.

But you could also use `assert_eq(bb, Rectangle::with_corners(Point::new(10, 10), Point::new(300, 84))`.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+/// An iterator over all pixel positions on the polyline
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct PolylineIterator<'a> {
+    stop: bool,
+    vertices: &'a [Point],
+    translate: Point,
+    segment_iter: ThickLineIterator,
```
rfuest, line 166: This should use `line::Points` instead of `ThickLineIterator`. They are equivalent at the moment, but `Points` will be optimized at some point (#314).

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+}
+
+impl<'a> IntoIterator for Polyline<'a> {
+    type Item = Point;
+    type IntoIter = PolylineIterator<'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.vertices
```
rfuest, line 203: Could you always initialize `segment_iter` with an empty iterator and move all conditional code into `next`? IMO that might result in cleaner code.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+/// Pixel iterator for each pixel in the line
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct StyledPolylineIterator<'a, C>
+where
+    C: PixelColor,
+{
+    style: PrimitiveStyle<C>,
```
rfuest, line 251: Using `stroke_color` and `stroke_width` instead of the whole `style` saves some memory.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,358 @@
...
+    #[test]
+    fn one_px_wide_only() {
+        let polyline = Polyline::new(&HEARTBEAT);
+
+        let thick = polyline.into_styled(PrimitiveStyle::with_stroke(Rgb565::RED, 10));
+        let thin = polyline.into_styled(PrimitiveStyle::with_stroke(Rgb565::RED, 1));
+
+        assert!(thick.into_iter().eq(thin.into_iter()));
+    }
+}
```
rfuest, line 358: IMO some tests are missing:
* Drawing a polyline to `MockDisplay`
* Test that the `Points` iter doesn't return the points at the vertices twice

embedded-graphics/src/primitives/polyline.rs
jamwaffles: Great idea. I thought I was being all clever with those `.map()`s...

jamwaffles: I changed it, but changed it back (although with the `ThickLineIterator` -> `line::Points` change) as the first segment need to render its first point, but subsequent segments skip their first point in the `next()` impl. I couldn't think of a clean way to check for segment index in `next()` so left this approach here.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,358 @@
...
+/// let line_style = PrimitiveStyle::with_stroke(Rgb565::GREEN, 1);
+///
+/// Polyline::new(&points)
+///     .into_styled(line_style)
+///     .draw(&mut display)?;
+/// # Ok::<(), core::convert::Infallible>(())
+/// ```
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
+pub struct Polyline<'a> {
+    translate: Point,
```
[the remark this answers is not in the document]
jamwaffles: Hm, I'm on the fence with this but I've made it `pub` because it can't hurt :)

jamwaffles:
Thanks for all the good suggestions. Comments applied/addressed :)

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,402 @@
...
+impl<'a> Primitive for Polyline<'a> {
+    type PointsIter = Points<'a>;
+
+    fn points(&self) -> Self::PointsIter {
+        Points::new(self)
+    }
+}
+
+impl<'a> Dimensions for Polyline<'a> {
+    fn bounding_box(&self) -> Rectangle {
```
rfuest, line 85: I think this doesn't handle `vertices.len() < 2` correctly.
For `len = 0` the bounding box doesn't have a defined `top_left` corner, but should be `Size::zero()`.
For `len = 1` the correct behavior depends on how we decide to handle single vertex polylines (see comment above). Depending on that the size of the bounding box would be `Size::zero()` or `Size::new(1, 1)`.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,402 @@
...
+                rest.get(0).map(|end| Points {
+                    vertices: rest,
+                    translate: polyline.translate,
+                    segment_iter: Line::new(*start + polyline.translate, *end + polyline.translate)
+                        .points(),
+                })
+            })
+            .unwrap_or_else(||
+                // Polyline is less than 2 vertices long. Return a dummy iterator that will short
+                // circuit due to `stop: true`
```
rfuest, line 185: `stop` doesn't exist anymore

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+impl<'a, C: PixelColor> Iterator for StyledPolylineIterator<'a, C> {
+    type Item = Pixel<C>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Break if stroke width is zero
+        if self.stroke_width == 0 {
+            return None;
+        }
```
rfuest, line 248: Can this be handled outside the `next` function, by using a dummy iter for this unlikely case?

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        assert_eq!(
+            bb,
+            Rectangle::with_corners(Point::new(-90, -90), Point::new(200, -16))
+        );
+    }
+
+    #[test]
+    fn one_point() {
+        let _polyline = Polyline::new(&[Point::zero()]);
```
rfuest, line 343: You could also add a test if the points iterator works correctly for this case.

embedded-graphics/src/primitives/polyline.rs
jamwaffles: I've moved it out to `into_iter` by kludging the stroke colour based on width. Can you think of a better way?

embedded-graphics/src/primitives/polyline.rs
jamwaffles: Added!

embedded-graphics/src/primitives/polyline.rs
jamwaffles: > For len = 0 the bounding box doesn't have a defined top_left corner

Hm. Should we just set it to `Point::zero()` in this case? It could also use the value of the `translate` field on `Polyline`.

rfuest: Yes, setting it to `Point::zero()` seems to be the best option IMO.

jamwaffles:
> A single vertex polyline could be handled the same way as a line with two copies of the same vertex

IMO it would be better in this case to not produce any output. I think it's a pain for implementation - it's not possible to return a slice created in `Polyline::new()`, so I think there would have to be special cases dotted around `polyline.rs` to deal with 1-length polylines.

I've changed the `Dimensions` impl with behaviour specific to empty and 1-length lines. Let me know what you think.

embedded-graphics/src/primitives/polyline.rs
rfuest: I think that's OK, because we need to check for `stroke_color != None` in the `next` function anyway. It would be nice if we could also get rid of the `let stroke_color = stroke_color?`, but I don't think that's possible.

jamwaffles: Certainly possible but would probably lead to messier code. I'll leave it as is for now. Perf-wise, I imagine `self.line_iter().next()` has a lot more overhead than the stroke width check.

rfuest: Sure it's fine, the comment was meant for iterators in general. But creating an empty dummy iterator wouldn't be too messy.
Changing `pixel_color: Option<C>` to `pixel_color: C`  isn't possible, because you can't create a dummy `stroke_color` for any `C: PixelColor` when creating a dummy iter.

rfuest:
> IMO it would be better in this case to not produce any output. I think it's a pain for implementation - it's not possible to return a slice created in `Polyline::new()`, so I think there would have to be special cases dotted around `polyline.rs` to deal with 1-length polylines.

OK. I don't think that it  would add much complexity at the moment, but it could cause problems later on when line joins are added.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+impl<'a, C> IntoIterator for &'a Styled<Polyline<'a>, PrimitiveStyle<C>>
+where
+    C: PixelColor,
+{
+    type Item = Pixel<C>;
+    type IntoIter = StyledPolylineIterator<'a, C>;
+
+    fn into_iter(self) -> Self::IntoIter {
```
rfuest, line 225: I started to move the code from the `into_iter` methods to `StyledPolylineIterator::new`. This helps if we decide to split the primitives files if they get too long by keeping related code together in one place.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,426 @@
...
+    }
+}
+
+/// Pixel iterator for each pixel in the line
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct StyledPolylineIterator<'a, C>
+where
+    C: PixelColor,
+{
+    stroke_width: u32,
```
rfuest, line 248: `stroke_width` isn't used anymore.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,426 @@
...
+    type IntoIter = StyledPolylineIterator<'a, C>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        // If the stroke width is zero, short circuit the iterator by setting `stroke_color` to
+        // none.
+        let stroke_color = if self.style.stroke_width == 0 {
+            None
+        } else {
+            self.style.stroke_color
+        };
```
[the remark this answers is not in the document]
jamwaffles: Ah nice use of `.filter()`. Just when I thought I had all these methods under my belt...

I've added a method to `PrimitiveStyle` and changed the styled line iterator to use it as well.

embedded-graphics/src/primitives/polyline.rs
```diff
@@ -0,0 +1,426 @@
...
+                        .points(),
+                })
+            })
+            .unwrap_or_else(||
+                // Polyline is less than 2 vertices long. Return a dummy iterator that will short
+                // circuit 
+                Points {
+                    vertices: &[],
+                    translate: Point::zero(),
+                    segment_iter: Line::new(Point::zero(), Point::zero()).points()
```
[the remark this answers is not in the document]
jamwaffles: I've added a test and an extra item to the end of #314 :+1:

embedded-graphics/src/primitives/polyline.rs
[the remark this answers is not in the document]
rfuest: > Ah nice use of `.filter()`. Just when I thought I had all these methods under my belt...

I've read the `Iterator`, `Result` and `Option` docs more times than I care to admit, but still seem to find new interesting methods each time.

jamwaffles:
Ready for final (hopefully) review!

jamwaffles:
Vielen Dank!

PR #298, opened 2020-03-28
Make SDL2 an optional simulator dependency

### LINKED ISSUE #294: Please make sdl dependency optional for the simulator
Thanks for this great crate. I'm using `embedded-graphics-simulator` to render data into ImageBuffers and then turn them into glium textures for use with imgui-rs.
But currently I get linking errors because I don't have SDL (and I don't want to have to link against SDL, to keep dependencies minimal).
Please make sdl an optional dependency for `embedded-graphics-simulator` :)

jamwaffles:

## PR description

This is a non-breaking change, so can be merged into `master` for a future 0.6 release.

Makes `sdl2` an optional dependency for `simulator`. It is enabled by default, but can be disabled by adding `default-features = false` to the dependency import.

Closes #294

### DIFF
--- a/simulator/README.md
+++ b/simulator/README.md
@@ -73,16 +73,17 @@ any format supported by `image`.
 
 # Usage without SDL2
 
-The `with-sdl` feature can be disabled by adding `default-features = false` to the dependency. This
-will remove the requirement on SDL2 build files being present and removes the `OutputSettings::output_to_display` and `OutputSettings::pixel_pitch` methods. For example:
+When the simulator is used in headless/CI environments that don't require showing a window, SDL2
+support can be disabled. This removes the requirement of SDL2 being installed on the target machine,
+but still allows the simulator to be used to generate images.
+
+The `with-sdl` feature is enabled by default and can be disabled by adding `default-features = false` to the dependency:
 
 ```toml
-[dependencies]
-embedded-graphics-simulator = "0.2.0"
+[dependencies.embedded-graphics-simulator]
+version = "0.2.0"
 default-features = false
 ```
-
-This may be useful for screenshot generation in headless/CI environments where extra dependencies are undesirable.
 
 See the [Choosing
 Features](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choosing-features)
--- a/simulator/README.md
+++ b/simulator/README.md
@@ -74,13 +74,15 @@ any format supported by `image`.
 # Usage without SDL2
 
 The `with-sdl` feature can be disabled by adding `default-features = false` to the dependency. This
-will remove the requirement on SDL2 build files being present. For example:
+will remove the requirement on SDL2 build files being present and removes the `OutputSettings::output_to_display` and `OutputSettings::pixel_pitch` methods. For example:
 
 ```toml
 [dependencies]
 embedded-graphics-simulator = "0.2.0"
 default-features = false
 ```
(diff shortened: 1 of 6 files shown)


simulator/src/output_settings.rs
```diff
@@ -31,11 +31,13 @@ impl OutputSettings {
     }
 
     /// Translates a output coordinate to the corresponding display coordinate.
+    #[allow(unused)]
```
rfuest: Would it be cleaner to move these two functions into a second `impl OutputSettings` block and use the `#[cfg(feature = "with-sdl")]` on that block?

simulator/README.md
```diff
@@ -70,3 +70,18 @@ If a program doesn't require to display a window and only needs to export one or
 `SimulatorDisplay` can also be converted to an `image` crate `ImageBuffer` by using the
 `to_image_buffer` method. The resulting buffer can then be used to save the display content to
 any format supported by `image`.
+
+# Usage without SDL2
+
+The `with-sdl` feature can be disabled by adding `default-features = false` to the dependency. This
```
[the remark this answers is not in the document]
jamwaffles: > IMO this should also mention what features are missing if you disable SDL and how the simulator could still be useful.

Added some more text.

> Regarding the SDL2 build files being present: Should we support this https://github.com/Rust-SDL2/rust-sdl2#bundled-feature?

Maybe, but that can happen in a separate PR if the need arises.

simulator/src/output_settings.rs
jamwaffles: Ah yes, I forgot you can have multiple impl blocks, thanks!

simulator/README.md
```diff
@@ -70,3 +70,20 @@ If a program doesn't require to display a window and only needs to export one or
...
+will remove the requirement on SDL2 build files being present and removes the `OutputSettings::output_to_display` and `OutputSettings::pixel_pitch` methods. For example:
```
rfuest: `OutputSettings::output_to_display` and `OutputSettings::pixel_pitch` are internal functions and don't need to be mentioned here IMO.
```suggestion
When the simulator is used in headless/CI environments, that don't require showing a window, the SDL2 support can be disabled. This removes the requirement of SDL2 being installed on the target machine, but still allows the simulator to be used to generate images.

The `with-sdl` feature is enabled by default and can be disabled by adding `default-features = false` to the dependency:
```

simulator/README.md
```diff
@@ -70,3 +70,20 @@ If a program doesn't require to display a window and only needs to export one or
...
+
+```toml
+[dependencies]
+embedded-graphics-simulator = "0.2.0"
+default-features = false
```
rfuest: Shouldn't this be like this?
```suggestion
[dependencies.embedded-graphics-simulator]
version = "0.2.0"
default-features = false
```

simulator/README.md
```diff
@@ -70,3 +70,20 @@ If a program doesn't require to display a window and only needs to export one or
...
+```
+
+This may be useful for screenshot generation in headless/CI environments where extra dependencies are undesirable.
```
rfuest: This isn't necessary anymore, if you use the text suggested above.

jamwaffles:
Thanks for the wording fix. Sounds much better. I copied your text nearly verbatim. Let me know if there's anything else that needs doing :)

PR #306, opened 2020-04-04
Change Rectangle to use top_left and size

rfuest:

## PR description

Continuation of PR #286.

The original comment for the old PR:

This PR changes Rectangle to use top_left and size instead of top_left and bottom_right. A similar change for Circles was recently implemented in #147.

As a part of the changes for #182 some of the code in this PR isn't final. Some problems like the bottom right corner problem mentioned in #182 will need to be addressed, but this will require changes in the Dimensions trait and is outside the scope of this PR IMO.

### DIFF
--- a/embedded-graphics/src/lib.rs
+++ b/embedded-graphics/src/lib.rs
@@ -163,7 +163,7 @@
 //! let mut display: MockDisplay<Rgb565> = MockDisplay::default();
 //!
 //! fn build_thing(text: &'static str) -> impl Iterator<Item = Pixel<Rgb565>> {
-//!     egrectangle!(top_left = (0, 0), bottom_right = (40, 40))
+//!     egrectangle!(top_left = (0, 0), size = (40, 40))
 //!         .into_iter()
 //!         .chain(&egcircle!(
 //!             top_left = (12, 12),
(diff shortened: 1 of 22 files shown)


rfuest:
There were some unresolved naming issues in the old PR:

@jamwaffles :
> If I understand correctly:
> 
> * Change `Triangle::from_points` to `Triangle::from_vertices`. This continues to take a list of 3 points. This might also wanna be `Triangle::with_vertices` if we're going for consistency?
> 
> * Change `Rectangle::with_corners` to `Rectangle::with_vertices`. Does this remain as two arguments or should we change it to take a list to be consistent with `Triangle`?
> 
> 
> ?

@rfuest :
> > If I understand correctly:
> 
> I'm not quite sure myself and open to better suggestions.
> 
> > * Change `Triangle::from_points` to `Triangle::from_vertices`. This continues to take a list of 3 points. This might also wanna be `Triangle::with_vertices` if we're going for consistency?
> 
> Maybe we should make `Triangle` behave consistently with `Polygon`. This would probably mean that `Triangle::new` and `Poygon::new` would have a slice as an argument. `Triangle::with_vertices` could then be used for specifying three individual points.
> 
> > Change `Rectangle::with_corners` to `Rectangle::with_vertices`. Does this remain as two arguments or should we change it to take a list to be consistent with `Triangle`?
> 
> If we change `Triangle` like in the above comment `Rectangle::with_vertices` could still use two arguments. Creating a rectangle from a slice of two points isn't a common operation IMO.

jamwaffles:
I'll try and distill the above with a few of my thoughts/opinions added:

* Rename the current `Triangle::from_points` to `Triangle::from_slice` which takes a 3-element slice. I think this name is more idiomatic Rust, and frees up the `points` identifier for other uses.
* `Triangle::new()` stays as is, because the number of vertices is fixed. This is in contrast to a polygon/polyline where there is naturally a variable number of vertices, so I don't think we need consistency here.
* Discuss `Polyline` (and polygon) naming separately as it's muddying the waters here.
* `Rectangle::new()` **remains** as `Rectangle::new(top_left: Point, size: Size) -> Self` as authored in this PR.
* Also keep `Rectangle::with_corners()` as is in this PR. Or call it `with_vertices()`, I'm on the fence.

Interested to hear your thoughts.

Also the build is failing due to some unrebased code :)

rfuest:
SGTM. But IMO `Triangle::from_points` can be renamed in another PR, because it's not directly related to this change.

> Also the build is failing due to some unrebased code :)

That should already have been fixed when you wrote your comment. And I think this PR is ready for a final review.

embedded-graphics/src/lib.rs
```diff
@@ -283,7 +283,7 @@
 //!     style::PrimitiveStyle,
 //! };
 //!
-//! Circle::new(Point::new(32, 32), 20)
+//! Circle::new(Point::new(16, 16), 40)
```
jamwaffles, line 286: Not sure why this changed, but the circle is no longer centered in the screenshot.

simulator/src/framebuffer.rs
```diff
@@ -116,7 +113,7 @@ impl DrawTarget<Rgb888> for Framebuffer {
             let color = &[fill_color.r(), fill_color.g(), fill_color.b()];
 
             let tl = item.primitive.top_left;
-            let br = item.primitive.bottom_right;
+            let br = item.primitive.bottom_right();
```
jamwaffles, line 116: Shouldn't these be either both method calls or both property accesses?

embedded-graphics/src/lib.rs
rfuest: This problem was caused because the examples weren't updated after the recent circle changes. I just quickly adjusted the values to be closer to the original, but if you want I can also center it perfectly. But a centered circle should probably be constructed with the `with_center` constructor instead.

simulator/src/framebuffer.rs
rfuest: They probably should at some point, but these methods are still work in progress, as mentioned in the comment in `Rectangle::bottom_right`.

rfuest:
Changelog entry is added.

PR #307, opened 2020-04-05
Add points iterators for primitives

rfuest:

## PR description

This PR adds `Primitive::points` which returns an iterator over all points inside a closed shape or on a line.

To keep the changes in this PR small the new iterators are mostly copies of the existing ones without the styling features. Future PRs can reduce the code duplication, but some of these changes depend on other features that aren't implemented yet.

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- [#307](https://github.com/jamwaffles/embedded-graphics/pull/307) Added `Primitive::points` to get an iterator over all points inside a primitive.
+
 ### Changed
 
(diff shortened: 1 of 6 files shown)


jamwaffles:
This PR looks good to me, but can you talk about about how it will support https://github.com/jamwaffles/embedded-graphics/issues/182#issuecomment-602697037? I'd just like to get a picture of where this is headed.

rfuest:
The points iterators aren't directly influenced by a `PrimitiveStyle` because they work on "raw" `Primitive`s and not styled primitives. The same behavior will also replicated by the hit testing function, which will essentially return `true` if a point is contained in `primitive::points()`.

To make it easier to use the same functions for styled primitives additional functions can be added. For example, `Rectangle::stroke_area(&self) -> Rectangle` and `Rectangle::fill_area(&self) -> Rectangle`, would return offset versions of the original rectangle that define the outer edge of the styled primitives fill and stroke. Having this functions would make it easy to implement custom drawing routines:
```rust
let c = Circle::new(...).into_styled(...);
let stroke_area = c.stroke_area();
let fill_area = c.fill_area();
stroke_area.points()
   .map(|p| if fill_area.contains(p) { FILL_COLOR }  else { STROKE_COLOR })
   .draw(&mut display);
```

jamwaffles:
> For example, Rectangle::stroke_area(&self) -> Rectangle and Rectangle::fill_area(&self) -> Rectangle, would return offset versions of the original rectangle that define the outer edge of the styled primitives fill and stroke.

That's really elegant, I like it :)

PR #309, opened 2020-04-05
Infinite triangle

### LINKED ISSUE #308: Some filled triangles cause endless loop
While preparing #307 I noticed a problem with some filled triangles:
```rust
Triangle::new(Point::new(10, 10), Point::new(20, 30), Point::new(30, -10))
    .into_styled(PrimitiveStyle::with_fill(BinaryColor::On))
    .draw(&mut display)?
```
Trying to execute this will cause an endless loop. The problem only occurs if a fill color is set.

jamwaffles:

## PR description

Triangles were doing the off-screen check back from 0.4/0.5. This PR removes that check, aligning `Triangle` behaviour with other primitives.

Closes #308 

As an aside, this is a candidate for a 0.6 bugfix release at some point. Do you have any suggestions on keeping track of this stuff? I could:

1. Create a tracking issue with each PR to cherry pick tagged in it
2. Add a `v0.6` label and tag PRs with that
3. Both
4. Something else that's smarter.

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -13,8 +13,11 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 ### Changed
 
 - **(breaking)** [#274](https://github.com/jamwaffles/embedded-graphics/pull/274) The `Circle` is now defined by its bounding box top-left corner and its diameter instead of its center and its radius. To convert your code, you can replace `Circle::new(point, radius)` by `Circle::with_center(point, 2 * radius + 1)`.
-
 - **(breaking)** [#306](https://github.com/jamwaffles/embedded-graphics/pull/306) The `Rectangle` is now defined by its top-left corner and its size instead of the top-left and bottom-right corner. To convert your code, you can replace `Rectangle::new` by `Rectangle::with_corners`.
+
+### Fixed
+
+- [#309](https://github.com/jamwaffles/embedded-graphics/pull/309) Prevent triangles with off-screen vertices from infinitely looping.
 
 ## [0.6.1] - 2020-04-01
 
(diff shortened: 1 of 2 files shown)


rfuest:
Can you resolve the conflict and also apply the fix to the new `Points` iterator?

jamwaffles:
All done :+1:

rfuest:
> As an aside, this is a candidate for a 0.6 bugfix release at some point. Do you have any suggestions on keeping track of this stuff? I could:
> 
> 1. Create a tracking issue with each PR to cherry pick tagged in it
> 2. Add a `v0.6` label and tag PRs with that
> 3. Both
> 4. Something else that's smarter.

I don't have a better idea at the moment. As long as you handle the previous releases you should use whatever method works for you. But a tracking issue sounds like a reasonable idea IMO.

But this PR also demonstrates another problem with this approach. If you squash merge this PR it will be harder to cherry pick the appropriate changes for `0.6`, because the `Points` iterator doesn't exist in the previous version. Should we handle PRs differently in the future, by not squashing them and rebasing them before merging to generate a cleaner history? This PR, for example, could be interactively rebased into two commits, one with the changes for the styled iter and one for the points iter.

jamwaffles:
> Should we handle PRs differently in the future, by not squashing them and rebasing them before merging to generate a cleaner history?

I would prefer to squash by default, but yes if it makes cherry picking to 0.6 easier, a PR should be rebased with sensibly split commits. Shall we handle this on a case-by-case basis?

> This PR, for example, could be interactively rebased into two commits, one with the changes for the styled iter and one for the points iter.

That's a good idea which I've just done - I'll rebase this PR once the build passes.

rfuest:
> I would prefer to squash by default, but yes if it makes cherry picking to 0.6 easier, a PR should be rebased with sensibly split commits. Shall we handle this on a case-by-case basis?

OK.

Issue #284, opened 2020-03-22
Add filled/unfilled arc primitive


jamwaffles:
Add an `Arc` primitive. An arc is defined by a start position, end position and center point. 

It may also be useful to add the ability to specify an arc by its center point, start/end number of degrees from 0 (which points to the top screen edge) and radius.

```rust
// Create the top-left quadrant of an ellipse
// (center, start, end)
let arc = Arc::new(Point::new(20, 20), Point::new(5, 20), Point::new(20, 10));

// Create a circular segment in the top-right quadrant
// (center, start angle, end angle, radius)
let segment = Arc::from_degrees(Point::new(20, 20), 0, 90, 10)
```

rfuest:
Should filled and unfilled arcs use the same primitive? In #77 we decided to use different types for filled and unfilled polygons. So we might want to do the same for arcs.

There are two common variants of how a filled arcs are handles. One results in a pie shape and the other just connects the two arc ends with a straight line:
![rect843](https://user-images.githubusercontent.com/226123/77255535-97f47380-6c68-11ea-99f4-b71eea79bacf.png)
Should we support both?

jamwaffles:
> Should filled and unfilled arcs use the same primitive?

In #77, I think we're going to disallow fills on `Polyline`s which is why they're separate from `Polygon`s, whereas here an `Arc` has an optional fill, so I think filled and unfilled arcs should probably be the same type. That said, if we use the same type for `Arc`s, maybe the `Polyline`/`Polygon` distinction should be reduced into a single `Polygon` type too?

> Should we support both?

Hmm. Yes, ideally, although if I had to choose one it'd be the "pie" still fill. If we can do both, how would you suggest choosing the fill style?

rfuest:
> maybe the `Polyline`/`Polygon` distinction should be reduced into a single `Polygon` type too?

I would like to keep them distinct, because a polygon encloses an area an can, for example, be used for hit testing, which doesn't make sense for an open path like a polyline

> Hmm. Yes, ideally, although if I had to choose one it'd be the "pie" still fill. If we can do both, how would you suggest choosing the fill style?

I don't know yet. Setting the fill style in `PrimitiveStyle` would work for drawing, but unstyled `Arc`s couldn't be used for hit testing anymore. One solution would be to use distinct `Arc`, `Segment` and `Sector` primitives.

IMO it should also be possible to set if the stroke is also applied to the linear segments.

jamwaffles:
[...]

PR #310, opened 2020-04-06
Add support for Arc and Sector primitive

aurelj:

## PR description

This is a first draft implementation of the Arc primitive.
For now it supports filled Arcs, but I plan to implement Sector and remove fill support from Arc, as described in #284. I think Arc and Sector will use a common `StyledIterator` but I'm not sure yet how to do this cleanly.
There is probably no need for a full detailed review right now, but I would like to know if this is overall going into the right direction.

Here are a few technical details of the current implementation:
- Angles are specified as integer degrees (this is plenty enough for me right now, and it shouldn't be very hard to add support for radiants)
- Center + Angles are converted to linear equation using `tan()`
- `tan()` is implemented with a LUT of `FixedPoint` values
- I introduced a basic implementation of `FixedPoint` numbers

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -27,6 +27,8 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 - [#366](https://github.com/jamwaffles/embedded-graphics/pull/366) Allow usage of all grayscale color types in `MockDisplay` patterns.
 - [#342](https://github.com/jamwaffles/embedded-graphics/pull/342) Added `Rectangle::intersection` method.
 - [#363](https://github.com/jamwaffles/embedded-graphics/pull/363) Export `primitives::ContainsPoint` trait in prelude.
+- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Arc` primitive.
+- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Sector` primitive.
 
 ### Changed
 
--- a/embedded-graphics/README.md
+++ b/embedded-graphics/README.md
@@ -19,6 +19,7 @@ It contains built in items that make it easy to draw 2D graphics primitives:
   - Circles
   - Ellipses
   - Arcs
+  - Sectors
   - Triangles
   - Polylines
   - Rounded rectangles
--- a/embedded-graphics/README.md
+++ b/embedded-graphics/README.md
@@ -18,6 +18,7 @@ It contains built in items that make it easy to draw 2D graphics primitives:
   - Rectangles
   - Circles
   - Ellipses
+  - Arcs
   - Triangles
   - Polylines
   - Rounded rectangles
--- a/embedded-graphics/README.md
+++ b/embedded-graphics/README.md
@@ -176,6 +176,9 @@ Additional examples can be found in the [simulator](./simulator/examples) crate.
 - `nalgebra_support` - use the [Nalgebra](https://crates.io/crates/nalgebra) crate with `no_std`
   support to enable conversions from `nalgebra::Vector2` to `Coord` and `UnsignedCoord`.
 
+- `fixed_point` - use fixed point arithmetic instead of floating point for all trigonometric
+  calculation.
+
 ## Migrating from 0.5 to 0.6
 
 Please read [the migration guide](embedded-graphics/MIGRATING-0.5-0.6.md).
--- a/embedded-graphics/src/geometry/angle.rs
+++ b/embedded-graphics/src/geometry/angle.rs
@@ -1,0 +1,486 @@
+use super::real;
+use crate::geometry::Real;
+use core::f32::consts::PI;
+use core::ops::{Add, AddAssign, Neg, Sub, SubAssign};
+use float_cmp::{ApproxEq, F32Margin};
+#[cfg(not(feature = "fixed_point"))]
+#[allow(unused_imports)]
+use micromath::F32Ext;
+
+pub(crate) mod angle_consts {
+    use super::{real, Angle};
+
+    pub(crate) const ANGLE_90DEG: Angle = Angle(real::FRAC_PI_2);
+    pub(crate) const ANGLE_180DEG: Angle = Angle(real::PI);
+    pub(crate) const ANGLE_360DEG: Angle = Angle(real::TAU);
+}
+
+/// Angle.
+///
+/// `Angle` is used to define the value of an angle.
+///
+/// # Examples
+///
+/// ## Create an `Angle` from a value
+///
+/// ```rust
+/// use embedded_graphics::geometry::{Angle, AngleUnit};
+/// use core::f32::consts::PI;
+///
+/// // Create an angle using the `from_degrees` constructor method
+/// let angle_a = Angle::from_degrees(10.0);
+/// let angle_b = Angle::from_radians(PI);
+///
+/// // Angles can also be created using the [AngleUnit](./trait.AngleUnit.html) trait
+/// let angle_c = 30.0.deg();
+/// let angle_d = PI.rad();
+/// ```
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+pub struct Angle(Real);
+
+impl Angle {
+    /// Creates an angle defined in degrees.
+    pub fn from_degrees(angle: f32) -> Self {
+        Angle((angle * PI / 180.0).into())
+    }
+
+    /// Creates an angle defined in radians.
+    pub fn from_radians(angle: f32) -> Self {
+        Angle(angle.into())
+    }
+
+    /// Creates a zero degree angle.
+    pub fn zero() -> Self {
+        Angle(0.into())
+    }
+
+    /// Compute the absolute value of the angle.
+    pub fn abs(self) -> Self {
+        Angle(self.0.abs())
+    }
+
+    /// Normalize the angle to less than one full rotation (ie. in the range 0..360).
+    pub fn normalize(self) -> Self {
+        Angle(self.0.rem_euclid((2.0 * PI).into()))
+    }
+
+    /// Normalize the angle to less than one full rotation, starting from angle.
+    /// (ie. in the range angle..(angle+360)).
+    pub(crate) fn normalize_from(self, angle: Self) -> Self {
+        Angle((self.0 - angle.0).rem_euclid((2.0 * PI).into()) + angle.0)
+    }
+
+    /// Return numerical value of the angle in degree
+    pub fn to_degrees(self) -> f32 {
+        let angle: f32 = self.0.into();
+        180.0 * angle / PI
+    }
+
+    /// Return numerical value of the angle in radian
+    pub fn to_radians(self) -> f32 {
+        self.0.into()
+    }
+}
+
+/// AngleUnit trait.
+///
+/// `AngleUnit` is a trait to convert numbers into angle by appending .deg() or .rad()
+/// to the number as if it was a unit.
+///
+/// # Examples
+///
+/// ## Create an `Angle` from a value using `AngleUnit`
+///
+/// ```rust
+/// use embedded_graphics::geometry::AngleUnit;
+/// use core::f32::consts::PI;
+///
+/// // Create an angle using the `AngleUnit` methods
+/// let angle_a = 30.0.deg();
+/// let angle_b = PI.rad();
+/// ```
+pub trait AngleUnit {
+    /// Convert a number (interpreted as degrees) to an `Angle`.
+    fn deg(self) -> Angle;
+
+    /// Convert a number (interpreted as radians) to an `Angle`.
+    fn rad(self) -> Angle;
+}
+
+impl AngleUnit for f32 {
+    fn deg(self) -> Angle {
+        Angle::from_degrees(self)
+    }
+
+    fn rad(self) -> Angle {
+        Angle::from_radians(self)
+    }
+}
+
+pub(crate) trait Trigonometry {
+    /// Get the sine of the angle.
+    fn sin(self) -> Real;
+
+    /// Get the cosine of the angle.
+    fn cos(self) -> Real;
+
+    /// Get the tangent of the angle.
+    fn tan(self) -> Option<Real>;
+}
+
+#[cfg(not(feature = "fixed_point"))]
+impl Trigonometry for Angle {
+    fn sin(self) -> Real {
+        let angle: f32 = self.0.into();
+        angle.sin().into()
+    }
+
+    fn cos(self) -> Real {
+        let angle: f32 = self.0.into();
+        angle.cos().into()
+    }
+
+    fn tan(self) -> Option<Real> {
+        let angle: f32 = self.0.into();
+        let tan = angle.tan();
+        // FRAC_PI_2.tan() has no value, but the approximate method used by micromath actually return a huge
+        // value which is > 20000000.0, so we check for this to decide that the angle was approximately
+        // FRAC_PI_2 and that tan() has actually no value.
+        if tan.is_nan() || tan.abs() > 20000000.0 {
+            None
+        } else {
+            Some(tan.into())
+        }
+    }
+}
+
+#[cfg(feature = "fixed_point")]
+impl Trigonometry for Angle {
+    fn sin(self) -> Real {
+        use fixed::types::I16F16;
+        const SIN: [I16F16; 91] = [
+            // Ideally we could make the compiler generate those values, but for now sin() is not a const fn,
+            // so it can't be used here. Here is how it would look like:
+            //   I16F16::from_bits((0f64.sin() * (1 << 16) as f64).round()))),
+            I16F16::from_bits(0),
+            I16F16::from_bits(1144),
+            I16F16::from_bits(2287),
+            I16F16::from_bits(3430),
+            I16F16::from_bits(4572),
+            I16F16::from_bits(5712),
+            I16F16::from_bits(6850),
+            I16F16::from_bits(7987),
+            I16F16::from_bits(9121),
+            I16F16::from_bits(10252),
+            I16F16::from_bits(11380),
+            I16F16::from_bits(12505),
+            I16F16::from_bits(13626),
+            I16F16::from_bits(14742),
+            I16F16::from_bits(15855),
+            I16F16::from_bits(16962),
+            I16F16::from_bits(18064),
+            I16F16::from_bits(19161),
+            I16F16::from_bits(20252),
+            I16F16::from_bits(21336),
+            I16F16::from_bits(22415),
+            I16F16::from_bits(23486),
+            I16F16::from_bits(24550),
+            I16F16::from_bits(25607),
+            I16F16::from_bits(26656),
+            I16F16::from_bits(27697),
+            I16F16::from_bits(28729),
+            I16F16::from_bits(29753),
+            I16F16::from_bits(30767),
+            I16F16::from_bits(31772),
+            I16F16::from_bits(32768),
+            I16F16::from_bits(33754),
+            I16F16::from_bits(34729),
+            I16F16::from_bits(35693),
+            I16F16::from_bits(36647),
+            I16F16::from_bits(37590),
+            I16F16::from_bits(38521),
+            I16F16::from_bits(39441),
+            I16F16::from_bits(40348),
+            I16F16::from_bits(41243),
+            I16F16::from_bits(42126),
+            I16F16::from_bits(42995),
+            I16F16::from_bits(43852),
+            I16F16::from_bits(44695),
+            I16F16::from_bits(45525),
+            I16F16::from_bits(46341),
+            I16F16::from_bits(47143),
+            I16F16::from_bits(47930),
+            I16F16::from_bits(48703),
+            I16F16::from_bits(49461),
+            I16F16::from_bits(50203),
+            I16F16::from_bits(50931),
+            I16F16::from_bits(51643),
+            I16F16::from_bits(52339),
+            I16F16::from_bits(53020),
+            I16F16::from_bits(53684),
+            I16F16::from_bits(54332),
+            I16F16::from_bits(54963),
+            I16F16::from_bits(55578),
+            I16F16::from_bits(56175),
+            I16F16::from_bits(56756),
+            I16F16::from_bits(57319),
+            I16F16::from_bits(57865),
+            I16F16::from_bits(58393),
+            I16F16::from_bits(58903),
+            I16F16::from_bits(59396),
+            I16F16::from_bits(59870),
+            I16F16::from_bits(60326),
+            I16F16::from_bits(60764),
+            I16F16::from_bits(61183),
+            I16F16::from_bits(61584),
+            I16F16::from_bits(61966),
+            I16F16::from_bits(62328),
+            I16F16::from_bits(62672),
+            I16F16::from_bits(62997),
+            I16F16::from_bits(63303),
+            I16F16::from_bits(63589),
+            I16F16::from_bits(63856),
+            I16F16::from_bits(64104),
+            I16F16::from_bits(64332),
+            I16F16::from_bits(64540),
+            I16F16::from_bits(64729),
+            I16F16::from_bits(64898),
+            I16F16::from_bits(65048),
+            I16F16::from_bits(65177),
+            I16F16::from_bits(65287),
+            I16F16::from_bits(65376),
+            I16F16::from_bits(65446),
+            I16F16::from_bits(65496),
+            I16F16::from_bits(65526),
+            I16F16::from_bits(65536),
+        ];
+        let degree: i32 = (Real::from(180) * self.0 / real::PI).round().into();
+        let degree = degree.rem_euclid(360) as usize;
+        let sin = if degree <= 90 {
+            SIN[degree]
+        } else if degree <= 180 {
+            SIN[180 - degree]
+        } else if degree <= 270 {
+            -SIN[degree - 180]
+        } else {
+            -SIN[360 - degree]
+        };
+        sin.into()
+    }
+
+    fn cos(self) -> Real {
+        (self + angle_consts::ANGLE_90DEG).sin()
+    }
+
+    fn tan(self) -> Option<Real> {
+        let cos = self.cos();
+        if cos != Real::zero() {
+            Some(self.sin() / cos)
+        } else {
+            None
+        }
+    }
+}
+
+impl Add for Angle {
+    type Output = Angle;
+
+    fn add(self, other: Angle) -> Angle {
+        Angle(self.0 + other.0)
+    }
+}
+
+impl AddAssign for Angle {
+    fn add_assign(&mut self, other: Angle) {
+        self.0 += other.0;
+    }
+}
+
+impl Sub for Angle {
+    type Output = Angle;
+
+    fn sub(self, other: Angle) -> Angle {
+        Angle(self.0 - other.0)
+    }
+}
+
+impl SubAssign for Angle {
+    fn sub_assign(&mut self, other: Angle) {
+        self.0 -= other.0;
+    }
+}
+
+impl Neg for Angle {
+    type Output = Angle;
+
+    fn neg(self) -> Angle {
+        Angle(-self.0)
+    }
+}
+
+impl ApproxEq for Angle {
+    type Margin = F32Margin;
+
+    fn approx_eq<M: Into<Self::Margin>>(self, other: Self, margin: M) -> bool {
+        self.0.approx_eq(other.0, margin.into())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use float_cmp::approx_eq;
+
+    #[test]
+    fn angles_can_be_added() {
+        let left = Angle::from_degrees(10.0);
+        let right = Angle::from_degrees(30.0);
+
+        assert!(approx_eq!(
+            Angle,
+            left + right,
+            Angle::from_degrees(40.0),
+            epsilon = 0.0001
+        ));
+    }
+
+    #[test]
+    fn angles_can_be_subtracted() {
+        let left = Angle::from_degrees(30.0);
+        let right = Angle::from_degrees(10.0);
+
+        assert!(approx_eq!(
+            Angle,
+            left - right,
+            Angle::from_degrees(20.0),
+            epsilon = 0.0001
+        ));
+    }
+
+    #[test]
+    fn angles_can_be_absoluted() {
+        let angle = Angle::from_degrees(30.0).abs();
+        assert_eq!(angle, Angle::from_degrees(30.0));
+
+        let angle = Angle::from_degrees(-30.0).abs();
+        assert_eq!(angle, Angle::from_degrees(30.0));
+    }
+
+    #[test]
+    fn angle_unit() {
+        assert_eq!(180.0.deg(), Angle::from_degrees(180.0));
+        assert_eq!(PI.rad(), Angle::from_radians(PI));
+    }
+
+    #[test]
+    fn from_radians() {
+        assert_eq!(Angle(PI.into()), Angle::from_radians(PI));
+    }
+
+    #[test]
+    fn to_radians() {
+        let angle = Angle(PI.into()).to_radians();
+        assert!(approx_eq!(f32, angle, PI, epsilon = 0.0001));
+    }
+
+    #[test]
+    fn from_degrees() {
+        let angle = Angle::from_degrees(180.0);
+        assert!(approx_eq!(f32, angle.0.into(), PI, epsilon = 0.0001));
+    }
+
+    #[test]
+    fn to_degrees() {
+        let angle = Angle(PI.into()).to_degrees();
+        assert!(approx_eq!(f32, angle, 180.0, epsilon = 0.001));
+    }
+
+    #[test]
+    fn sin_correct() {
+        let degree_sin_pairs = [
+            (-90.0, -1.0),
+            (-60.0, -0.86602540),
+            (-45.0, -0.70710678),
+            (-30.0, -0.5),
+            (0.0, 0.0),
+            (30.0, 0.5),
+            (45.0, 0.70710678),
+            (60.0, 0.86602540),
+            (90.0, 1.0),
+            (120.0, 0.86602540),
+            (135.0, 0.70710678),
+            (150.0, 0.5),
+            (180.0, 0.0),
+            (210.0, -0.5),
+            (225.0, -0.70710678),
+            (240.0, -0.86602540),
+            (270.0, -1.0),
+        ];
+
+        for (angle, sin) in &degree_sin_pairs {
+            assert!(approx_eq!(
+                Real,
+                angle.deg().sin(),
+                (*sin).into(),
+                epsilon = 0.0001
+            ));
+        }
+    }
+
+    #[test]
+    fn cos_correct() {
+        let degree_cos_pairs = [
+            (-90.0, 0.0),
+            (-60.0, 0.5),
+            (-45.0, 0.70710678),
+            (-30.0, 0.86602540),
+            (0.0, 1.0),
+            (30.0, 0.86602540),
+            (45.0, 0.70710678),
+            (60.0, 0.5),
+            (90.0, 0.0),
+            (120.0, -0.5),
+            (135.0, -0.70710678),
+            (150.0, -0.86602540),
+            (180.0, -1.0),
+            (210.0, -0.86602540),
+            (225.0, -0.70710678),
+            (240.0, -0.5),
+            (270.0, -0.0),
+        ];
+
+        for (angle, cos) in &degree_cos_pairs {
+            assert!(approx_eq!(
+                Real,
+                angle.deg().cos(),
+                (*cos).into(),
+                epsilon = 0.0001
+            ));
+        }
+    }
+
+    #[test]
+    fn tan_correct() {
+        let degree_tan_pairs = [
+            (-60.0, -1.73205080),
+            (-45.0, -1.0),
+            (-30.0, -0.57735026),
+            (0.0, 0.0),
+            (30.0, 0.57735026),
+            (45.0, 1.0),
+            (60.0, 1.73205080),
+        ];
+
+        for (angle, tan) in &degree_tan_pairs {
+            assert!(approx_eq!(
+                Real,
+                angle.deg().tan().unwrap(),
+                (*tan).into(),
+                epsilon = 0.0001
+            ));
+        }
+
+        assert_eq!((-90.0.deg()).tan(), None);
+        assert_eq!(90.0.deg().tan(), None);
+    }
+}
--- a/embedded-graphics/src/geometry/real.rs
+++ b/embedded-graphics/src/geometry/real.rs
@@ -1,0 +1,214 @@
+use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+use float_cmp::{ApproxEq, F32Margin};
+#[cfg(not(feature = "fixed_point"))]
+#[allow(unused_imports)]
+use micromath::F32Ext;
+
+/// Real.
+///
+/// `Real` is used to store real numbers as either floating point or 32 bit fixed point if the
+/// `fixed_point` cargo feature is enabled.
+/// ```
+pub(crate) use real_impl::{Real, FRAC_PI_2, PI, TAU};
+
+#[cfg(not(feature = "fixed_point"))]
+mod real_impl {
+    use core::f32;
+
+    pub(crate) const FRAC_PI_2: Real = Real(f32::consts::FRAC_PI_2);
+    pub(crate) const PI: Real = Real(f32::consts::PI);
+    pub(crate) const TAU: Real = Real(2.0 * f32::consts::PI);
+
+    #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+    pub(crate) struct Real(pub(super) f32);
+
+    impl From<f32> for Real {
+        fn from(src: f32) -> Self {
+            Self(src)
+        }
+    }
+
+    impl From<i32> for Real {
+        fn from(src: i32) -> Self {
+            Self(src as f32)
+        }
+    }
+
+    impl From<u32> for Real {
+        fn from(src: u32) -> Self {
+            Self(src as f32)
+        }
+    }
+
+    impl From<Real> for f32 {
+        fn from(src: Real) -> Self {
+            src.0
+        }
+    }
+
+    impl From<Real> for i32 {
+        fn from(src: Real) -> Self {
+            src.0 as i32
+        }
+    }
+
+    impl From<Real> for u32 {
+        fn from(src: Real) -> Self {
+            src.0 as u32
+        }
+    }
+}
+
+#[cfg(feature = "fixed_point")]
+mod real_impl {
+    use fixed::types::I16F16;
+
+    pub(crate) const FRAC_PI_2: Real = Real(I16F16::from_bits(102944));
+    pub(crate) const PI: Real = Real(I16F16::from_bits(205887));
+    pub(crate) const TAU: Real = Real(I16F16::from_bits(411775));
+
+    #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+    pub(crate) struct Real(pub(super) I16F16);
+
+    impl Real {
+        pub(crate) const fn zero() -> Self {
+            Self(I16F16::from_bits(0))
+        }
+    }
+
+    impl From<I16F16> for Real {
+        fn from(src: I16F16) -> Self {
+            Self(src)
+        }
+    }
+
+    impl From<f32> for Real {
+        fn from(src: f32) -> Self {
+            Self(I16F16::from_num(src))
+        }
+    }
+
+    impl From<i32> for Real {
+        fn from(src: i32) -> Self {
+            Self(I16F16::from_num(src))
+        }
+    }
+
+    impl From<u32> for Real {
+        fn from(src: u32) -> Self {
+            Self(I16F16::from_num(src))
+        }
+    }
+
+    impl From<Real> for f32 {
+        fn from(src: Real) -> Self {
+            src.0.to_num::<f32>()
+        }
+    }
+
+    impl From<Real> for i32 {
+        fn from(src: Real) -> Self {
+            src.0.round_to_zero().to_num::<i32>()
+        }
+    }
+
+    impl From<Real> for u32 {
+        fn from(src: Real) -> Self {
+            src.0.to_num::<u32>()
+        }
+    }
+}
+
+impl Add for Real {
+    type Output = Real;
+
+    fn add(self, other: Real) -> Real {
+        Self(self.0 + other.0)
+    }
+}
+
+impl AddAssign for Real {
+    fn add_assign(&mut self, other: Real) {
+        self.0 += other.0;
+    }
+}
+
+impl Sub for Real {
+    type Output = Real;
+
+    fn sub(self, other: Real) -> Real {
+        Self(self.0 - other.0)
+    }
+}
+
+impl SubAssign for Real {
+    fn sub_assign(&mut self, other: Real) {
+        self.0 -= other.0;
+    }
+}
+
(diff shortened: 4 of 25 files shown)


jamwaffles:
I had a (very) quick scroll and this stood out:

1. I would prefer to support the `(center: Point, start: Point, end: Point)` definition first for my immediate use case (joiners and end caps for polylines) over the current implementation using angles. This seems at odds with your requirements, so it would be great if this PR supported both.
2. That's a lot of lookup table, and a lot of code for `FixedPoint`. If we do have to use that approach, is there an existing crate we can use? I'm moderately strongly against adding this functionality into e-g.
3. I haven't thought about this too deeply, but is it possible to use Bresenham-like integer maths to replace the `LinearEquation` structs? I haven't taken the time to understand the algo so this could be a misguided comment.
4. Are all the tiny `combine_*` functions required?

aurelj:
> I had a (very) quick scroll and this stood out:
> 
> 1. I would prefer to support the `(center: Point, start: Point, end: Point)` definition first for my immediate use case (joiners and end caps for polylines) over the current implementation using angles. This seems at odds with your requirements, so it would be great if this PR supported both.

Indeed, we don't have the same need at all. I understand your use case and I agree that it would be great to support both. I will see what I can do.
But I have an issue with your `(center: Point, start: Point, end: Point)` definition. It looks over-constrained to me. What if the distance between `center` and `start` is not the same as the distance between `center` and `stop` ? We couldn't draw a Circle or Arc that passes thru both `start` and `stop`.
Maybe the definition should be `(start: Point, end: Point, radius: u32)` with a way to specify on which side to position the center (it could be for example at the right of the `start` -> `end` vector).
I also guess that for this use case you will always want the minor Arc and never the major one.

> 2. That's a lot of lookup table, and a lot of code for `FixedPoint`. If we do have to use that approach, is there an existing crate we can use? I'm moderately strongly against adding this functionality into e-g.

Maybe we could use the `fixed` crate to replace my `FixedPoint` implementation. I will have a look at it.
But we will still need our `tan()` implementation (ie. LUT).

> 3. I haven't thought about this too deeply, but is it possible to use Bresenham-like integer maths to replace the `LinearEquation` structs? I haven't taken the time to understand the algo so this could be a misguided comment.

I don't think Bresenham has anything to do with what I need here.
This part of the algorithm is basically:
- get the equations of both lines delimiting the Arc/Sector
- for each point of the Circle, check if it is on the required side of both lines to decide if it is part of the Arc or not.

> 4. Are all the tiny `combine_*` functions required?

They are actually not strictly required. I did it this way to avoid having lots of comparisons / logic in the hot path (ie. pixel iterator). So basically I decide the logic to apply while constructing the iterator by just choosing which of the `combine_*` function to use, and then I can simply call the appropriate `combine_*` logic in the iterator itself.
But honestly, that might be premature optimization. I could try to move all the logic into the iterator.

jamwaffles:
> But I have an issue with your (center: Point, start: Point, end: Point) definition. It looks over-constrained to me. What if the distance between center and start is not the same as the distance between center and stop ? We couldn't draw a Circle or Arc that passes thru both start and stop.

That's a great point. I've added a comment [here](https://github.com/jamwaffles/embedded-graphics/issues/284#issuecomment-609991263) to discuss it further. I think this PR may be a bit premature as there are some bits of #284 that haven't been finalised yet.

aurelj:
Here is a much improved version of this patch set.

It now has two separate Primitives for `Arc` and `Sector` (`Arc` having no fill). There is some code duplication between the two, but we can get rid of this later on.

I replaced my `FixedPoint` implementation using the `fixed` crate.

I also got rid of the `combine_*` function and split the algorithm into two separate Iterator, one that return the points of the bounding box that are in the desired sector of the plane, and one that return the distance from each of this points to the center (in the same spirit as the new `DistanceIterator` in `Circle`).

Now I have a question regarding `Arc`. As it is not a closed shape and it has no surface (except for the stroke width), I guess it should not implement `ContainsPoint`, but I wonder if its `Points` iterator should return the points covered by the stroke width or nothing at all ? The documentation for `PointsIter` state "all points inside the primitive", but since `Arc` has no inside (ie. it is not a closed shape), I choose to not return any point. Let me know if I should change this.

Overall, I think this patch set is in a pretty good shape now.

rfuest:
I haven't had time to look at the code, but I can at least answer your question.

> Now I have a question regarding `Arc`. As it is not a closed shape and it has no surface (except for the stroke width), I guess it should not implement `ContainsPoint`, but I wonder if its `Points` iterator should return the points covered by the stroke width or nothing at all ? The documentation for `PointsIter` state "all points inside the primitive", but since `Arc` has no inside (ie. it is not a closed shape), I choose to not return any point. Let me know if I should change this.

You are correct that `CointainsPoint' should not be implemented for `Arc`.

The `PointsIter` docs need an update to mention non closed shapes. For non closed shapes `PointsIter` behaves differently by iterating over all points along the shape. This means that it should return the same points as the corresponding styled iter when drawing a 1px wide stroke.

aurelj:
> The `PointsIter` docs need an update to mention non closed shapes. For non closed shapes `PointsIter` behaves differently by iterating over all points along the shape. This means that it should return the same points as the corresponding styled iter when drawing a 1px wide stroke.

Oh, I see. I've now implemented `PointsIter` for `Arc` to match this definition.

I think this PR is now ready for review.

rfuest:
Before reviewing this we should decide how e-g should handle angles in general. Fixed point arithmetic could be beneficial on some architectures, but because microcontrollers with floating point units are getting more common I would also like to support floating point angles. We will need to decide if we want to support both or if one is enough.

@aurelj: Do you plan to use e-g on a target platform that would greatly benefit from using the fixed point implementation? Or could you also use something like the `micromath` crate with `f32` angles without a huge impact on performance or code size?

aurelj:
First note that in my implementation, `Angle` does not expose any floating or fixed point in its public API. It is only used internally, so it is easy to change at any time.
I guess it shouldn't be too hard to have 2 different internal implementations that can be chosen at build time with a feature flag. That would help comparing both implementations on different platforms. I will give it a try.

I'm working on various STM32 platforms. For now I'm doing some e-g testing on a STM32F4 which does support hardware floating point, but many of the popular STM32 don't have a FPU (STM32F0, STM32F1 and STM32F2). This includes the Blue Pill (STM32F103) which is one of the most common beginner platform for embedded-rust.

I don't want to be "forced" to choose a high end MCU when using e-g just because of the choice of using floating point for some internal computation. So I think if we choose to only have the floating point implementation, it first has to be seriously evaluated on MCU without FPU first.

rfuest:
> First note that in my implementation, `Angle` does not expose any floating or fixed point in its public API. It is only used internally, so it is easy to change at any time.

But it also only allows integer degrees, which should be enough resolution for smaller displays but larger displays (>= 320x240 px) could benefit from < 1 degree steps.

Can we always use `f32` for the user facing API, even if we will use fixed point internally? I would prefer not to have different APIs when changing a feature.

> I don't want to be "forced" to choose a high end MCU when using e-g just because of the choice of using floating point for some internal computation. So I think if we choose to only have the floating point implementation, it first has to be seriously evaluated on MCU without FPU first.

Sure this needs to be considered and lower end MCUs need be supported well. But not using advanced features, if present, isn't great either.

@jamwaffles: What are your thoughts on this subject?

jamwaffles:
> Can we always use f32 for the user facing API, even if we will use fixed point internally?

This would be ideal yeah, even if the first thing that happens is to convert the `f32` to fixed point. It gives us the freedom to redefine the internal implementation.

Unfortunately `micromath` doesn't have any benchmarks but it might be a good compromise between angle accuracy and performance so I'd be leaning towards that, at least for an initial implementation.

Using `f32` and `micromath` I think would allow the removal of `Angle` and the associated boilerplate as well?

rfuest:
> Using `f32` and `micromath` I think would allow the removal of `Angle` and the associated boilerplate as well?

Keeping `Angle` provides some type safety for mixing radians and degrees.
An extension trait for `f32` it could make `Angle` nicer to use: `90.0.deg()`, `3.141.rad()`

jamwaffles:
> Keeping Angle provides some type safety for mixing radians and degrees.

True, that sounds like a useful feature.

Shall we settle on `micromath` and a trait on `f32` that adds `.angle()` for now? I'm not sure if we need to explore anything else at the moment. Radian support and faster/better internal implementations can always be added later on.

rfuest:
> Shall we settle on `micromath` [...] for now?

If @aurelj agrees I think that's a good solution.

>  [...] and a trait on `f32` that adds `.angle()` for now?

IMO `.angle()` doesn't make much sense, because it doesn't define the unit of the value. I would suggest to use `fn rad(self) -> Angle` and `fn deg(self) -> Angle`, where `rad` is a basic constructor and `deg` converts the given value into radians (the internal unit).

jamwaffles:
> IMO .angle() doesn't make much sense

Argh, yes you're right! I was flying through issues too fast between things at work. I meant to say `.deg()` and `.rad()` as you originally suggested :facepalm:

aurelj:
OK, here is a new revision with, I think, all of your concerns taken into account.

The `Angle` facing API is now using `f32` to support higer precision. It also support specifying angles in degree or radian. They are stored internally as radian.
I also added an `f32` extension trait to be able to use 90.0.deg() or PI.rad().

Internally, all the values are stored in a `Real` wrapper type (better name ?). `Real` is implemented as either `f32` or `I16F16` fixed point from the `fixed` crate. To use the fixed point version, you need to use `cargo --features fixed_point`. By default the `f32` implementation is used.

The trigonometric functions are implemented using micromath for the `f32` version and using a LUT for the `I16F16` version.

Let me know if this matches a bit better what you had in mind.

aurelj:
I've just fixed the `float-cmp` crate so that it builds with no_std.
Until the next `float-cmp` release, I've pointed the dependency to upstream git.

aurelj:
And... `float-cmp` has a new release, so no more dependency on the git master.

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,634 @@
...
+    line_b: LinearEquation,
+    draw_above_a: bool,
+    draw_above_b: bool,
+    sweep: Angle,
+}
+
+impl PlaneSector {
+    fn new(arc: &Arc) -> Self {
+        let angle_start = arc.angle_start.normalize_from(-90.0.deg());
+        let angle_end = arc.angle_end().normalize_from(-90.0.deg());
```
rfuest, line 314: Why does this subtract 90 degree from the angle?

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,634 @@
...
+        let negative_sweep = arc.angle_sweep < Angle::zero();
+
+        Self {
+            line_a: LinearEquation::from_point_angle(arc.center(), arc.angle_start),
+            line_b: LinearEquation::from_point_angle(arc.center(), arc.angle_end()),
+            draw_above_a: (angle_start < 90.0.deg()) ^ negative_sweep,
+            draw_above_b: (angle_end >= 90.0.deg()) ^ negative_sweep,
```
rfuest, line 321: Adding constants for `90.0.deg()` (and maybe other important angles) could reduce the number of floating point operations.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -41,6 +44,111 @@ pub trait ContainsPoint {
...
+///     .stroke_width(1)
+///     .build();
+///
+/// let arc_2: Styled<Arc, PrimitiveStyle<Rgb565>> =
+///     Arc::new(Point::new(10, 20), 30, Angle::from_degree(90.0), Angle::from_degree(60.0)).into_styled(style);
+///
+/// assert_eq!(arc_1, arc_2);
+/// ```
+#[macro_export]
+macro_rules! egarc {
```
rfuest, line 113: IMO the macro shouldn't convert from degree to `Angle` internally, but take `Angle` as an argument instead. This way the user can use `180.0.deg()`, `PI.rad()` and even `some_angle_variable + 123.0.deg()` as angles.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -387,6 +495,111 @@ macro_rules! egrectangle {
...
+///     .stroke_width(1)
+///     .build();
+///
+/// let sector_2: Styled<Sector, PrimitiveStyle<Rgb565>> =
+///     Sector::new(Point::new(10, 20), 30, Angle::from_degree(90.0), Angle::from_degree(60.0)).into_styled(style);
+///
+/// assert_eq!(sector_1, sector_2);
+/// ```
+#[macro_export]
+macro_rules! egsector {
```
rfuest, line 564: The note about angles for `egarc` also applies to this macro.

simulator/examples/pacman.rs
```diff
@@ -0,0 +1,68 @@
...
+
+// the number of steps of the animation
+const STEPS: i32 = 10;
+
+fn main() -> Result<(), std::convert::Infallible> {
+    // Create a new simulator display with 65x65 pixels.
+    let mut display: SimulatorDisplay<Rgb565> = SimulatorDisplay::new(Size::new(65, 65));
+
+    // Create styles used by the drawing operations.
+    let sector_stroke = PrimitiveStyleBuilder::new()
```
rfuest, line 22: These styles apply both stroke and fill:
```suggestion
    let sector_style = PrimitiveStyleBuilder::new()
```

simulator/examples/pacman.rs
```diff
@@ -0,0 +1,68 @@
...
+        .stroke_color(Rgb565::BLACK)
+        .stroke_width(2)
+        .fill_color(Rgb565::YELLOW)
+        .build();
+    let eye_stroke = PrimitiveStyleBuilder::new()
```
rfuest, line 27:
```suggestion
    let eye_style = PrimitiveStyleBuilder::new()
```

tinytga/src/lib.rs
```diff
@@ -289,7 +289,7 @@ impl<'a> Iterator for TgaIterator<'a> {
 
         self.x += 1;
 
-        if self.x >= self.tga.width().into() {
+        if self.x >= self.tga.width() as u32 {
```
rfuest, line 292: Please remove the changes to `tinytga` because they aren't related to this PR.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+    }
+
+    /// Normalize the angle to less than one full rotation (ie. in the range 0..360).
+    pub fn normalize(self) -> Self {
+        Angle(self.0.rem_euclid((2.0 * PI).into()))
+    }
+
+    /// Normalize the angle to less than one full rotation, starting from angle.
+    /// (ie. in the range angle..(angle+360)).
+    pub fn normalize_from(self, angle: Self) -> Self {
```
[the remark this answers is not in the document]
aurelj: Indeed, I agree. Done.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+            I16F16::from_bits(65048),
+            I16F16::from_bits(65177),
+            I16F16::from_bits(65287),
+            I16F16::from_bits(65376),
+            I16F16::from_bits(65446),
+            I16F16::from_bits(65496),
+            I16F16::from_bits(65526),
+            I16F16::from_bits(65536),
+        ];
+        let degree: i32 = (Real::from(180.0) * self.0 / Real::from(PI)).round().into();
```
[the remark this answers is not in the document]
aurelj: Should be better now with added constant for PI and converting from int for 180.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+        } else if degree <= 270 {
+            -SIN[degree - 180]
+        } else {
+            -SIN[360 - degree]
+        };
+        sin.into()
+    }
+
+    fn cos(self) -> Real {
+        (self + 90.0.deg()).sin()
```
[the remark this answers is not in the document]
aurelj: Using a constant now, so since this is fixed point code only, this is a simple integer addition.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+    }
+
+    fn tan(self) -> Option<Real> {
+        let cos = self.cos();
+        if cos != 0.0.into() {
```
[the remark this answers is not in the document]
aurelj: Right. Done.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+    fn from(other: f32) -> Self {
+        Angle(other.into())
+    }
+}
+
+impl From<Angle> for f32 {
+    fn from(other: Angle) -> f32 {
+        other.0.into()
+    }
+}
```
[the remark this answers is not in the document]
aurelj: Indeed. That's some remaining code from initial version which supported only integer degree that I converted to f32 without thinking about it. Removed now.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+    }
+
+    #[test]
+    fn to_f32() {
+        let angle: f32 = Angle::from_radian(PI).into();
+        assert!(approx_eq!(f32, angle, PI, epsilon = 0.0001));
+    }
+
+    #[test]
+    fn sin_correct() {
```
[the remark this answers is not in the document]
aurelj: Yes, good idea. That looks much better.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+        assert!(approx_eq!(
+            Real,
+            270.0.deg().sin(),
+            (-1.0).into(),
+            epsilon = 0.0001
+        ));
+    }
+
+    #[test]
+    fn cos_correct() {
```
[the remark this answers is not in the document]
aurelj: Done.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+        assert!(approx_eq!(
+            Real,
+            270.0.deg().cos(),
+            0.0.into(),
+            epsilon = 0.0001
+        ));
+    }
+
+    #[test]
+    fn tan_correct() {
```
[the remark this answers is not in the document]
aurelj: Done.

embedded-graphics/src/geometry/real.rs
```diff
@@ -0,0 +1,206 @@
...
+#[allow(unused_imports)]
+use micromath::F32Ext;
+
+/// Real.
+///
+/// `Real` is used to store real numbers as either floating point or fixed
+/// point depending on feature flag.
+/// ```
+
+#[cfg(not(feature = "fixed_point"))]
```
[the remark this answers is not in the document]
aurelj: Good idea as well. This looks cleaner now with those private mod.

embedded-graphics/src/geometry/real.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+impl DivAssign for Real {
+    fn div_assign(&mut self, other: Real) {
+        self.0 /= other.0
+    }
+}
+
+use float_cmp::{ApproxEq, F32Margin};
```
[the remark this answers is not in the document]
aurelj: Oh, my mistake. Put it there at first for a quick test of float_cmp and forgot to move it up when I finally seriously adopted float_cmp.

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,634 @@
...
+    type Item = Point;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        for (point, distance) in &mut self.iter {
+            if distance < self.outer_threshold && distance >= self.inner_threshold {
+                return Some(point);
+            }
+        }
+
+        None
```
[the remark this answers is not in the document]
aurelj: I like it better this way indeed.

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,634 @@
...
+pub struct StyledArcIterator<C>
+where
+    C: PixelColor,
+{
+    iter: DistanceIterator,
+
+    outer_threshold: u32,
+    inner_threshold: u32,
+
+    color: Option<C>,
```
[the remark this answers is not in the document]
aurelj: Ok, fine with me.

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,634 @@
...
+        for (point, distance) in &mut self.iter {
+            if distance < self.outer_threshold && distance >= self.inner_threshold {
+                if let Some(color) = self.color {
+                    return Some(Pixel(point, color));
+                }
+            }
+        }
+
+        None
+    }
```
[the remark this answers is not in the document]
aurelj: Right, I like it.

embedded-graphics/src/primitives/arc.rs
aurelj: First note that it doesn't subtract 90 degree from the angle, it selects the normalization starting point to -90 rather than 0.

Basically, to decide whether we should draw on the part of the plane which is below or above line_a (variable draw_above_a), we need to know if the angle is either between -90 and 90, or if it is between 90 and 270. To do this, the easiest way is to normalize the angle between -90 and 270 rather than 0 to 360.

embedded-graphics/src/primitives/arc.rs
aurelj: Yep, I kind of thought about adding some consts at some point without actually doing it...
Done now, for 90deg, 180deg and 360deg.
No more use of float.deg() conversion in the code base (except in tests and examples).

embedded-graphics/src/primitives/mod.rs
aurelj: Yes, you're right. I didn't put much thought about those macros after adding support for floating point angles and radiant. Changed to take `Angle` argument now.

embedded-graphics/src/primitives/mod.rs
aurelj: Done.

simulator/examples/pacman.rs
aurelj: OK.

simulator/examples/pacman.rs
aurelj: OK.

tinytga/src/lib.rs
aurelj: In fact the 3 changes to `tinytga` are related to this PR. When the `fixed` crate is used, some of those types can't be inferred anymore. Here is the error you get with this code with the `fixed` crate in use:

```
error[E0283]: type annotations needed for `u32`
   --> tinytga/src/lib.rs:292:19
    |
250 |         let mut pixel_value = {
    |             --------------- consider giving `pixel_value` a type
...
292 |         if self.x >= self.tga.width().into() {
    |                   ^^ cannot infer type for type `u32`
    |
    = note: cannot satisfy `u32: core::cmp::PartialOrd<_>`

error: aborting due to previous error
```

rfuest: OK, interesting, I didn't expect that. Can you change this to:
```suggestion
        if self.x >= u32::from(self.tga.width()) {
```
which is a little safer than a cast if the return type of `tga.with` ever changes.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,636 @@
...
+
+    fn cos(self) -> Real {
+        let angle: f32 = self.0.into();
+        angle.cos().into()
+    }
+
+    fn tan(self) -> Option<Real> {
+        let angle: f32 = self.0.into();
+        let tan = angle.tan();
+        if tan.is_nan() || tan.abs() > 20000000.0 {
```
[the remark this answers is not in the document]
aurelj: tan(90) has no value (tending toward INFINITY...). So we need a return value for this case, which works with Real whether it is implemented as floating point or fixed point. That's why I chose to return `Option<Real>` so tan(90) can return `None`.

This `None` return value is then used in `LinearEquation` to generate the equation of a vertical line (in the form of `x + c = 0`).

The issue here is the approximation returned by `micromath` for 90.0.tan() (a big value slightly greater than 20000000.0). So to be consistent with the fixed point code, I chose to detect when tan() is huge, and consider that it means that the angle is (+/-)90 degree and thus consider that its tan() has no actual numerical value.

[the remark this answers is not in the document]
rfuest: OK, thanks for the explanation. The main reason for the question was to know how the magic value is derived. I'll take another look at this when I have looked at the drawing algorithm.

Can you add a comment to the code that describes why the value `20000000.0` was chosen?

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,797 @@
...
+        let Styled { primitive, style } = styled;
+
+        let stroke_area = primitive.expand(style.outside_stroke_width());
+        let fill_area = primitive.shrink(style.inside_stroke_width());
+
+        let inner_threshold = circle::diameter_to_threshold(fill_area.diameter);
+        let outer_threshold = circle::diameter_to_threshold(stroke_area.diameter);
+
+        let line_a = stroke_area.line_from_angle(styled.primitive.angle_start);
+        let line_b = stroke_area.line_from_angle(styled.primitive.angle_end());
```
jamwaffles, line 308: Won't this draw pixels near the center of the arc multiple times? I'm ok to leave it as is for this PR, but we should maybe open an issue to optimise arc filling. Also, other primitives AFAIK make an effort to not draw any two pixels twice so it would be good to have `Sector` mirror this behaviour.

jamwaffles:
This is looking really promising. The demo is cool!

But... I found a small bug with `Sector` with a stroke:

image

The bottom green line of the sector is 1px too long. In the code below, if I set `progress` to `35` the line renders correctly.

It also looks like the vertical starting line is off by one also. Is the sector too small, or the line too long maybe?

Code to reproduce:

```rust
let progress = 34;

let sweep = progress as f32 * 360.0 / 100.0;

Sector::new(Point::new(16, 16), 64 - 32, 90.0.deg(), sweep.deg())
    .into_styled(
        PrimitiveStyleBuilder::new()
            .stroke_color(Rgb565::GREEN)
            .fill_color(Rgb565::RED)
            .stroke_width(1)
            .stroke_alignment(StrokeAlignment::Inside)
            .build(),
    )
    .draw(&mut display)?;
```

The effect is a bit more pronounced with a stroke width of `5`:

image

embedded-graphics/src/geometry/angle.rs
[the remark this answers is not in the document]
aurelj: Yes, adding comment now.

tinytga/src/lib.rs
aurelj: OK.

aurelj:
> This is looking really promising. The demo is cool!
> 
> But... I found a small bug with `Sector` with a stroke:
> 
> [...]
> 
> The bottom green line of the sector is 1px too long.

Oh, right. There seems to be some kind of off by one in the length of the 2 stroke lines delimiting the sector. And this is actually also visible in the pacman example. I think it wasn't present in some previous version of this patchset.

I guess this is related to some changes I made to accommodate the new .stroke_alignment() property recently, but off the top of my head I'm not sure what. I will try to have a look at it.

aurelj:
> > This is looking really promising. The demo is cool!
> > But... I found a small bug with `Sector` with a stroke:
> > [...]
> > The bottom green line of the sector is 1px too long.
> 
> Oh, right. There seems to be some kind of off by one in the length of the 2 stroke lines delimiting the sector. And this is actually also visible in the pacman example. I think it wasn't present in some previous version of this patchset.

This was a simple rounding issue when computing the `Point` at the end of the `Line`.
Fixed now.

jamwaffles:
Just gave this another test. The end line is indeed fixed, but the vertical start looks to be 1px off still:

image

The arc has a radius of 16px but the line is 17px long.

aurelj:
> Just gave this another test. The end line is indeed fixed, but the vertical start looks to be 1px off still:

Fixed now.

aurelj:
> 1. `lib.rs` and `embedded-graphics/README.md` both have a list of primitives available. Can you add `Arc` and `Sector` to them please?

OK, done.

> 2. I've started adding benchmarks at the beginning of new primitive development to `simulator/benches/primitives.rs` so we can track performance over time.

Oh, I didn't notice this. Nice !

> Can you add one or two for `Arc` and `Sector` to that suite?

Sure, done.

> It might also be worth adding a separate bench for the `fixed_point` feature toggle with something like this in `Cargo.toml` [...]

Hum... OK, I tried to add this but I'm not sure it is such a great thing. With my last commit, you can now use `cargo bench --features fixed_point` which will indeed run the benches that are in `primitive_fixed_point.rs`, but it will also run all the other benches (including the identical `primitive.rs` benches) compiled with a fixed_point version of e-g. So I'm not sure if it is useful to have a specific `primitive_fixed_point.rs`, and it may even be a good thing to test everything with fixed_point, just in case a regression is introduced later on, in some code that wasn't supposed to make use of fixed_point...

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,632 @@
...
+
+    /// Angle at which the arc starts
+    pub angle_start: Angle,
+
+    /// Angle defining the arc sweep starting at angle_start
+    pub angle_sweep: Angle,
+}
+
+impl Arc {
+    /// Create a new arc delimited with a top-left point with a specific diamter and start and sweep angles
```
rfuest, line 63:
```suggestion
    /// Create a new arc delimited with a top-left point with a specific diameter and start and sweep angles
```

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,632 @@
...
+    }
+
+    /// Create a new arc centered around a given point with a specific diameter and start and sweep angles
+    pub fn with_center(
+        center: Point,
+        diameter: u32,
+        angle_start: Angle,
+        angle_sweep: Angle,
+    ) -> Self {
+        let top_left = center - Size::new(diameter, diameter).center_offset();
```
rfuest, line 85:
```suggestion
        let top_left = center - Size::new_equal(diameter).center_offset();
```

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,632 @@
...
+            } else {
+                true
+            }
+        })
+    }
+}
+
+/// Iterator that returns the squared distance to the center for all points in the bounding box.
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+struct DistanceIterator {
```
rfuest, line 382: We should merge this with `circle::DistanceIter`, by replacing `circle::DistanceIter` with:
(I didn't test the code, but it hope it's working enough to describe what I mean)
```rust
pub(in crate::primitives) struct DistanceIterator<I> {
    center: Point,
    points: I,
}

impl<I> DistanceIterator<I>
where
    I: Iterator<Item = Point>
{
    pub(in crate::primitives) fn new(center: Point, points: I) -> Self {
        Self {
            center,
            points,
        }
    }
}

impl<I> Iterator for DistanceIterator<I>
where
    I: Iterator<Item = Point>
{
    type Item = (Point, u32);

        self.points.next().map(|p| {
            let delta = self.center - p * 2;
            let distance = delta.x.pow(2) as u32 + delta.y.pow(2) as u32;

            (p, distance)
        })
    }
}
```
Which could then be used for `Circle`, `Arc` and `Sector`.

embedded-graphics/src/primitives/arc.rs
```diff
@@ -0,0 +1,632 @@
...
+        })
+    }
+}
+
+/// Define one side of a line
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+enum LineSide {
+    ABOVE,
+    BELLOW,
+}
```
rfuest, line 420: Enum variants need to be camel case and there is a typo:
```suggestion
enum LineSide {
    Above,
    Below,
}
```

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+
+    fn points(&self) -> Self::PointsIter {
+        Points::new(self)
+    }
+}
+
+impl ContainsPoint for Sector {
+    fn contains(&self, point: Point) -> bool {
+        let delta = self.center_2x() - point * 2;
+        let distance = delta.x.pow(2) as u32 + delta.y.pow(2) as u32;
```
rfuest, line 162: After rebasing you can use the new: `delta.length_squared() as u32`.

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+        let negative_sweep = self.angle_sweep < Angle::zero();
+
+        let line_a = LinearEquation::from_point_angle(self.center(), self.angle_start);
+        let line_b = LinearEquation::from_point_angle(self.center(), self.angle_end());
+        let draw_above_a = (angle_start < ANGLE_90DEG) ^ negative_sweep;
+        let draw_above_b = (angle_end >= ANGLE_90DEG) ^ negative_sweep;
+        let sweep = self.angle_sweep.abs();
+
+        let side_a = line_a.side(&point);
+        let side_b = line_b.side(&point);
```
rfuest, line 181: Couldn't you instead use `PlaneSector::new`?

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+            correct_a_side || correct_b_side
+        } else {
+            true
+        }
+    }
+}
+
+impl Dimensions for Sector {
+    fn bounding_box(&self) -> Rectangle {
+        Rectangle::new(self.top_left, Size::new(self.diameter, self.diameter))
```
rfuest, line 198:
```suggestion
        Rectangle::new(self.top_left, Size::new_equal(self.diameter))
```

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+            .map(|(point, _)| point)
+    }
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+enum IterState {
+    ARC,
+    LINES,
+    DONE,
+}
```
rfuest, line 272: Enum variants need to be camel case:
```suggestion
enum IterState {
    Arc,
    Lines,
    Done,
}
```

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+        let correct_b_side = draw_above_b ^ (side_b == LineSide::BELLOW);
+
+        if sweep < ANGLE_180DEG {
+            correct_a_side && correct_b_side
+        } else if sweep < ANGLE_360DEG {
+            correct_a_side || correct_b_side
+        } else {
+            true
+        }
+    }
```
rfuest, line 193: This is the same code used in `PlaneSectorIterator::next`. I would suggest to add this to `PlaneSector`:
```rust
fn contains(&self, point: Point) -> bool {
    let side_a = self.line_a.side(point);
    let side_b = self.line_b.side(point);

    let correct_a_side = self.draw_above_a ^ (side_a == LineSide::BELLOW);
    let correct_b_side = self.draw_above_b ^ (side_b == LineSide::BELLOW);

    if self.sweep < ANGLE_180DEG {
        correct_a_side && correct_b_side
    } else if self.sweep < ANGLE_360DEG {
        correct_a_side || correct_b_side
    } else {
        true
    }
}
```
And use it in `PlaneSectorIterator::next` and `ContainsPoint::contains`.

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+    type Item = Pixel<C>;
+    type IntoIter = StyledSectorIterator<C>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        StyledSectorIterator::new(self)
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+struct PlaneSector {
```
rfuest, line 398: Am I right that this identical to `arc::PlaneSector`? If it is, it shouldn't be duplicated but imported.

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+            draw_above_a: true,
+            draw_above_b: true,
+            sweep: Angle::zero(),
+        }
+    }
+}
+
+/// Iterator that returns only the points which are inside a plane sector defined by two lines.
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+struct PlaneSectorIterator {
```
rfuest, line 434: See comment for `PlaneSector`.

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+            } else {
+                true
+            }
+        })
+    }
+}
+
+/// Iterator that returns the squared distance to the center for all points in the bounding box.
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+struct DistanceIterator {
```
rfuest, line 480: See comment for `arc::DistanceIterator`.

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,798 @@
...
+enum LineSide {
+    ABOVE,
+    BELLOW,
+}
+
+/// Linear equation representation
+///
+/// The equation is stored as the a, b and c coefficients of the ax + by + c = 0 equation
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+struct LinearEquation {
```
rfuest, line 524: This also seems to be identical to `arc::LinearEquation`.

jamwaffles:
> If you like you could also apply the new code layout for primitives to Arc and Sector

I'm happy to tackle this over in #361 so we can land this PR sooner. It's been open quite a while and would be great to get merged in.

embedded-graphics/README.md
```diff
@@ -18,6 +18,8 @@ It contains built in items that make it easy to draw 2D graphics primitives:
   - Rectangles
   - Circles
   - Ellipses
+  - Arc
+  - Sector
```
jamwaffles, line 22: Pedantic nit:

```suggestion
  - Arcs
  - Sectors
```

embedded-graphics/src/lib.rs
```diff
@@ -7,6 +7,8 @@
 //!     * [Rectangles (and squares)](./primitives/rectangle/struct.Rectangle.html)
 //!     * [Circles](./primitives/circle/struct.Circle.html)
 //!     * [Ellipses](./primitives/ellipse/struct.Ellipse.html)
+//!     * [Arc](./primitives/arc/struct.Arc.html)
+//!     * [Sector](./primitives/sector/struct.Sector.html)
```
jamwaffles, line 11:
```suggestion
//!     * [Arcs](./primitives/arc/struct.Arc.html)
//!     * [Sectors](./primitives/sector/struct.Sector.html)
```

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    /// Normalize the angle to less than one full rotation, starting from angle.
+    /// (ie. in the range angle..(angle+360)).
+    pub(crate) fn normalize_from(self, angle: Self) -> Self {
+        Angle((self.0 - angle.0).rem_euclid((2.0 * PI).into()) + angle.0)
+    }
+
+    /// Return numerical value of the angle in degree
+    pub fn to_degree(self) -> f32 {
```
jamwaffles, line 68:
```suggestion
    pub fn to_degrees(self) -> f32 {
```

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+        let angle: f32 = self.0.into();
+        180.0 * angle / PI
+    }
+
+    /// Return numerical value of the angle in radian
+    pub fn to_radian(self) -> f32 {
```
jamwaffles, line 74:
```suggestion
    pub fn to_radians(self) -> f32 {
```

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+///
+/// # Examples
+///
+/// ## Create an `Angle` from a value using `AngleUnit`
+///
+/// ```rust
+/// use embedded_graphics::geometry::AngleUnit;
+///
+/// // Create an angle using the `AngleUnit` method
+/// let angle_b = 30.0.deg();
```
jamwaffles, line 92:
```suggestion
/// let angle_b = 30.0.deg();
/// ```
```

jamwaffles, line 92: For completeness' sake, it would be great if this example showed usage of `.rad()` as well.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+pub trait AngleUnit {
+    /// Convert a number (interpreted as degree) to an `Angle`.
```
jamwaffles, line 94:
```suggestion
    /// Convert a number (interpreted as degrees) to an `Angle`.
```

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+    fn deg(self) -> Angle;
+
+    /// Convert a number (interpreted as radian) to an `Angle`.
```
jamwaffles, line 97:
```suggestion
    /// Convert a number (interpreted as radians) to an `Angle`.
```

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+pub struct Angle(Real);
+
+impl Angle {
+    /// Creates an angle defined in degrees.
+    pub fn from_degree(angle: f32) -> Self {
+        Angle((angle * PI / 180.0).into())
+    }
+
+    /// Creates an angle defined in radian.
+    pub fn from_radian(angle: f32) -> Self {
```
jamwaffles, line 42:
```suggestion
    /// Creates an angle defined in radians.
    pub fn from_radians(angle: f32) -> Self {
```

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+///
+/// // Create an angle using the `from_degree` constructor method
+/// let angle_a = Angle::from_degree(10.0);
+/// ```
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+pub struct Angle(Real);
+
+impl Angle {
+    /// Creates an angle defined in degrees.
+    pub fn from_degree(angle: f32) -> Self {
```
jamwaffles, line 37:
```suggestion
    pub fn from_degrees(angle: f32) -> Self {
```

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+        180.0 * angle / PI
+    }
+
+    /// Return numerical value of the angle in radian
+    pub fn to_radian(self) -> f32 {
+        self.0.into()
+    }
+}
+
+/// AngleUnit trait.
```
jamwaffles, line 79: I think this should be reexported in `prelude.rs` unless @rfuest disagrees.

embedded-graphics/src/geometry/angle.rs
```diff
@@ -0,0 +1,477 @@
...
+///
+/// # Examples
+///
+/// ## Create an `Angle` from a value
+///
+/// ```rust
+/// use embedded_graphics::geometry::Angle;
+///
+/// // Create an angle using the `from_degree` constructor method
+/// let angle_a = Angle::from_degree(10.0);
```
jamwaffles, line 30: Similar to my other comment, it would be useful to show `Angle::from_radian(s)` here as well.

Another example showing usage of the `AngleUnit` (with a doc link to it as well) would be handy too.

embedded-graphics/src/geometry/real.rs
```diff
@@ -0,0 +1,214 @@
+use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+use float_cmp::{ApproxEq, F32Margin};
+#[cfg(not(feature = "fixed_point"))]
+#[allow(unused_imports)]
+use micromath::F32Ext;
+
+/// Real.
+///
+/// `Real` is used to store real numbers as either floating point or fixed
+/// point depending on feature flag.
```
jamwaffles, line 10: This should mention the feature in question:

```suggestion
/// `Real` is used to store real numbers as either floating point or 32 bit fixed point if the
/// `fixed_point` cargo feature is enabled.
```

embedded-graphics/src/geometry/angle.rs
rfuest: No, I don't disagree.

aurelj:
> Sorry for the long delay in reviewing this, I would like to get this merged soon. Can you rebase the code one more time? The macros were recently removed and `DrawTarget` doesn't need a `draw_arc` and `draw_sector` method anymore.

Rebased against latest master, with all the small needed adaptations.

> If you like you could also apply the new code layout for primitives to `Arc` and `Sector`, that @jamwaffles is currently working on for the other primitives: #363. But this can also be done after this PR is merged.

@jamwaffles, if you don't mind, I would happily let you split the `Arc` and `Sector` primitives into multiple files.

aurelj:
> Please also add the `fixed_point` feature with a short description to the features list in both `lib.rs` and `README.md`.

Done as well.

embedded-graphics/src/geometry/angle.rs
jamwaffles: Ah so it is, my bad.

embedded-graphics/CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#366](https://github.com/jamwaffles/embedded-graphics/pull/366) Allow usage of all grayscale color types in `MockDisplay` patterns.
 - [#342](https://github.com/jamwaffles/embedded-graphics/pull/342) Added `Rectangle::intersection` method.
 - [#363](https://github.com/jamwaffles/embedded-graphics/pull/363) Export `primitives::ContainsPoint` trait in prelude.
+- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Arc` primitive.
+- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added `DrawTarget::draw_arc`.
+- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Sector` primitive.
+- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added `DrawTarget::draw_sector`.
```
jamwaffles, line 33: The `DrawTarget` methods are not longer added

```suggestion
- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Arc` primitive.
- [#310](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Sector` primitive.
```

embedded-graphics/src/primitives/sector.rs
```diff
@@ -0,0 +1,646 @@
...
+
+    /// Angle at which the sector starts
+    pub angle_start: Angle,
+
+    /// Angle defining the sector sweep starting at angle_start
+    pub angle_sweep: Angle,
+}
+
+impl Sector {
+    /// Create a new sector delimited with a top-left point with a specific diamter and start and sweep angles
```
jamwaffles, line 72:
```suggestion
    /// Create a new sector delimited with a top-left point with a specific diameter and start and sweep angles
```

embedded-graphics/CHANGELOG.md
aurelj: Oh, of course... Done.

embedded-graphics/src/primitives/sector.rs
aurelj: Fixed.

PR #317, opened 2020-04-07
Fix Dimension impl for Circle

rfuest:

- [ x Run `./build.sh` (Linux/macOS only) and make sure it passes. If you use Windows, check that CI passes once you've opened the PR.

## PR description

This PR fixes the `Dimension` impl for `Circle`s. The size of the bounding box was off by one before and was inconsistent with the impl for `Rectangle`.

In addition the PR also reduces the code duplication in the iterators impls by adding an internal `DistanceIterator`.

### DIFF
--- a/embedded-graphics/src/primitives/circle.rs
+++ b/embedded-graphics/src/primitives/circle.rs
@@ -146,7 +146,7 @@ impl Transform for Circle {
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 pub struct Points {
     iter: DistanceIterator,
-    threshold: i32,
+    threshold: u32,
 }
 
 impl Points {
@@ -179,8 +179,8 @@ where
 {
     iter: DistanceIterator,
     style: PrimitiveStyle<C>,
-    outer_threshold: i32,
-    inner_threshold: i32,
+    outer_threshold: u32,
+    inner_threshold: u32,
 }
 
 impl<C> StyledCircleIterator<C>
@@ -244,11 +244,11 @@ where
     }
 }
 
-fn diameter_to_threshold(diameter: u32) -> i32 {
+fn diameter_to_threshold(diameter: u32) -> u32 {
     if diameter <= 4 {
-        (diameter.pow(2) - diameter / 2) as i32
+        diameter.pow(2) - diameter / 2
     } else {
-        diameter.pow(2) as i32
+        diameter.pow(2)
     }
 }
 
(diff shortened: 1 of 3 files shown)


embedded-graphics/src/primitives/circle.rs
```diff
@@ -274,11 +244,11 @@ where
     }
 }
 
-fn diameter_to_threshold(diameter: i32) -> i32 {
+fn diameter_to_threshold(diameter: u32) -> i32 {
```
jamwaffles, line 247: I think this should return a `u32`, i.e. `fn diameter_to_threshold(diameter: u32) -> u32` along with the other `i32` to `u32` fixes in this file to make it compile. The only change beyond type fixes would be that `DistanceIterator` becomes something like:

```diff
impl Iterator for DistanceIterator {
    type Item = (Point, u32);

    fn next(&mut self) -> Option<Self::Item> {
        self.points.next().map(|p| {
            let delta = self.center - p * 2;
-           let distance = delta.x.pow(2) + delta.y.pow(2);
+           let distance = delta.x.pow(2) as u32 + delta.y.pow(2) as u32;

            (p, distance)
        })
    }
}
```

(or maybe that should be `let distance = (delta.x.pow(2) as u32).saturating_add(delta.y.pow(2) as u32);` idk)

This makes it clearer to the reader that thresholds can never be negative, which percolates through the rest of the file.

rfuest: That's a good point. I've changed all thresholds and distances to `u32`.

rfuest:
I've addressed your comment and also tweaked the `StyledCircleIterator` a bit. You can review it whenever you have some time and energy left, I don't mind waiting a while.

embedded-graphics/src/primitives/circle.rs
```diff
@@ -177,91 +163,84 @@ impl Points {
...
 
-    // https://stackoverflow.com/a/1237519/383609
-    // https://stackoverflow.com/questions/1201200/fast-algorithm-for-drawing-filled-circles#comment80182898_1237519
     fn next(&mut self) -> Option<Self::Item> {
         // If the fill and stroke colors are `None`, treat entire object as transparent and exit
         // early.
-        if self.style.stroke_color.is_none() && self.style.fill_color.is_none() {
+        if self.outer_color.is_none() && self.inner_color.is_none() {
             return None;
         }
```
jamwaffles, line 227: This might be a bit more pleasant:

```suggestion
        self.outer_color.or(self.inner_color)?;
```

This also optimises to like 2 fewer asm instructions so yay I guess :grinning:

rfuest: That's much nicer. But I've decided to remove it completely, which should remove some additional asm instructions, but i haven't counted :wink:

rfuest:
> Interestingly, `cargo bench circle` is roughly 50% faster in this branch compared to `master`. This is on a laptop so results are a bit noisy but it's an interesting positive side effect. YMMV of course.

That's a nice surprise, I had feared that this code might be slower than the old version. But I get the same ~50% improvement on my PC.

jamwaffles:
Good stuff!

PR #318, opened 2020-04-09
Add ContainsPoint trait

rfuest:

## PR description

This PR adds a `contains(point: Point)` methods to all primitives that are closed shapes. The added `Shape` trait can be extended in the future to contain additional functions that don't make sense on line like primitives.

@jamwaffles Any suggestions for a better name, or do you think `Shape` is OK?

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,6 +10,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 - [#307](https://github.com/jamwaffles/embedded-graphics/pull/307) Added `Primitive::points` to get an iterator over all points inside a primitive.
 - [#317](https://github.com/jamwaffles/embedded-graphics/pull/317) Added `Rectangle::center` to get the center point of a rectangle.
+- [#318](https://github.com/jamwaffles/embedded-graphics/pull/317) Added `ContainsPoint` trait to check if a point is inside a closed shape.
 
 ### Changed
 
--- a/embedded-graphics/src/primitives/circle.rs
+++ b/embedded-graphics/src/primitives/circle.rs
@@ -4,7 +4,7 @@ use crate::{
     drawable::{Drawable, Pixel},
     geometry::{Dimensions, Point, Size},
     pixelcolor::PixelColor,
-    primitives::{Primitive, Rectangle, Styled},
+    primitives::{ContainsPoint, Primitive, Rectangle, Styled},
     style::PrimitiveStyle,
     transform::Transform,
     DrawTarget,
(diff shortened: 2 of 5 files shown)


jamwaffles:
Maybe call it `ClosedShape`? It hints that there's something "special" about the trait (in this case requiring that the shape be closed), as opposed to `Shape` which is very general.

If the trait is going to only handle collision detection, it might be better to give it a name suiting that purpose instead.

rfuest:
> If the trait is going to only handle collision detection, it might be better to give it a name suiting that purpose instead.

I'm not sure yet. Maybe it's better to have more traits with fewer methods each to have fine grained control of which primitive supports which methods. But this would also mean that trait bounds might get more complex.

Should I rename the trait to `Contains` or maybe `ContainsPoint` at the moment and rename it later (before the `0.7` release) if additional methods are added?

jamwaffles:
> Should I rename the trait to Contains or maybe ContainsPoint at the moment and rename it later (before the 0.7 release) if additional methods are added?

Yeah, `ContainsPoint` sounds like a good idea. Can you open a tracking issue so we remember to check before release?

rfuest:
> Yeah, `ContainsPoint` sounds like a good idea.

Changed. Should be ready for review now.

> Can you open a tracking issue so we remember to check before release?

I'll add this to my existing tracking issue, because it's directly related IMO.

jamwaffles:
Looks good to me, but... how do you feel about the semantics of lines? Can't a line contain a point, even with a thickness of 1? That might be an edge case, but a thick line can certainly contain a point, so shouldn't it also implement `ContainsPoint`?

rfuest:
`ContainsPoint` is only implemented for unstyled primitives and I would consider an unstyled line to have 0 width.

But testing if a point is on the stroke of a line is a useful feature that could be added later. I'm not sure if it should use `ContainsPoint` or something specific for lines. Lines could also benefit from line specific features, like testing if a point is to the left or right of the line.

jamwaffles:
Makes sense to me. Will you implement the same trait for styled primitives too? I don't think that needs a different trait, right?

rfuest:
> Will you implement the same trait for styled primitives too?

No, I haven't plan to implement `ContainsPoint` for styled primitives. Instead there will be different ways of testing if a point is inside a styled:
- `styled.primitive.contains(point)`
- `styled.fill_area().contains(point)`
- `styled.stroke_area().contains(point)`
- `styled.bounding_box().contains(point)`

One useful addition might be a method to test if a point is on the stroke:
`styled.is_on_stroke(point)` (this needs a better name) which would equal `styled.stroke_area().contains(point) && !styled.fill_area().contains(point)` for closed shapes, but could also be implemented for lines and arcs.

PR #319, opened 2020-04-10
Reduce code duplication in rectangle.rs

rfuest:

- [x] ~~Add a simulator example(s) where applicable~~
- [x] ~~Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) and appropriate crate (`embedded-graphics`, `simulator`, `tinytga`, `tinybmp`) if your changes affect the **public API**~~

## PR description

This PR reduces the code duplication in rectangle.rs by using some of the recently added features.

### DIFF
--- a/embedded-graphics/src/style/primitive_style.rs
+++ b/embedded-graphics/src/style/primitive_style.rs
@@ -70,6 +70,23 @@ where
     pub(crate) fn stroke_width_i32(&self) -> i32 {
         i32::try_from(self.stroke_width).unwrap_or(i32::max_value())
     }
+
+    /// Returns the effective stroke width.
+    ///
+    /// The effective stroke width takes the stroke color into account. If `stroke_color` is `None`
+    /// the effective stroke width is `0` independent of `stroke_width`.
+    pub(crate) fn effective_stroke_width(&self) -> u32 {
+        if self.stroke_color.is_some() {
+            self.stroke_width
+        } else {
+            0
+        }
+    }
+
+    /// Returns if a primitive drawn with this style is completely transparent.
+    pub(crate) fn is_transparent(&self) -> bool {
+        (self.stroke_color.is_none() || self.stroke_width == 0) && self.fill_color.is_none()
+    }
 }
 
 impl<C> Default for PrimitiveStyle<C>
--- a/embedded-graphics/src/geometry/size.rs
+++ b/embedded-graphics/src/geometry/size.rs
@@ -80,6 +80,17 @@ impl Size {
         }
     }
 
+    /// Saturating subtraction.
+    ///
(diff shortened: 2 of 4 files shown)


embedded-graphics/src/style/primitive_style.rs
```diff
@@ -70,6 +70,23 @@ where
...
+    /// Returns the effective stroke width.
+    ///
+    /// The effective stroke width takes the stroke color into account. If `stroke_color` is `None`
+    /// the effective stroke width is `0` independent of `stroke_width`.
+    pub(crate) fn effective_stroke_width(&self) -> u32 {
+        if self.stroke_color.is_some() {
+            self.stroke_width
+        } else {
+            0
+        }
```
jamwaffles, line 83: This could be:

```suggestion
        self.stroke_color.map(|_| self.stroke_width).unwrap_or(0)
```

rfuest:
> Curiously, this PR reduces `cargo bench 'unfilled rectangle'` performance by 50% compared to master. I'm not _too_ bothered because this crate never makes performance guarantees, but it would be good to keep it at the current level. Can you take a quick look and see where it might be lagging? If there are no easy wins we can leave it as is.

I did notice that, but haven't looked into it yet. It might be caused by the `Rectangle::contains` function. I'll add a note to the tracking issue and revisit this in another PR (if possible).

Did the `filled rectangle` performance increase on your PC?

embedded-graphics/src/style/primitive_style.rs
rfuest: TBH I think the original version is more readable.

jamwaffles:
> I'll add a note to the tracking issue and revisit this in another PR (if possible).

SGTM, let's keep the ball rolling.

> Did the filled rectangle performance increase on your PC?

It did indeed, a few tens of percent. Not particularly accurate due to laptop and thermal rubbish, but the number is always green ;)

embedded-graphics/src/style/primitive_style.rs
jamwaffles: Fair! Not worth bikeshedding over :)

rfuest:
> It did indeed, a few tens of percent. Not particularly accurate due to laptop and thermal rubbish, but the number is always green ;)

Interesting, I got a much larger improvement:
`filled rectangle`: -42.302%
`unfilled rectangle`: +28.579%

jamwaffles:
That's about what I got, just the percentages I get are pretty variable from run to run so I kept my numbers vague at "a fews tens of percent".

rfuest:
Ahh, I should have read your comment more carefully, I thought that you meant "a few tenth".

jamwaffles:
Hah sorry, English is a stupid language

PR #320, opened 2020-04-13
Ellipse

### LINKED ISSUE #283: Add ellipse primitive
Similar to `Circle`, defined by top left as per #274, but with a width/height given as a `Size`. E.g.:

```rust
// Create an ellipse 10px wide, 20px tall
let ellipse = Ellipse::new(Point::zero(), Size::new(10, 20));

// Create a circle 20px in diameter
let ellipse = Ellipse::new(Point::zero(), Size::new(20, 20));
```

Questions:

1. [x] ~Should `Circle` become a simple wrapping struct to `Ellipse`, or should it be removed entirely? IMO `Circle` should be kept in.~ `Circle` will become a thin wrapper over `Ellipse`.

jamwaffles:

## PR description

Adds support for the `Ellipse` primitive.

Closes #283

### DIFF
--- a/embedded-graphics/src/primitives/ellipse.rs
+++ b/embedded-graphics/src/primitives/ellipse.rs
@@ -1,0 +1,542 @@
+//! The ellipse primitive
+
+use crate::{
+    drawable::{Drawable, Pixel},
+    geometry::{Dimensions, Point, Size},
+    pixelcolor::PixelColor,
+    primitives::{circle, ContainsPoint, Primitive, Rectangle, Styled},
+    style::PrimitiveStyle,
+    transform::Transform,
+    DrawTarget,
+};
+
+/// Ellipse primitive
+///
+/// # Examples
+///
+/// The [macro examples](../../macro.egellipse.html) make for more concise code.
+///
+/// ## Create some ellipses with different styles
+///
+/// ```rust
+/// use embedded_graphics::{
+///     pixelcolor::Rgb565,
+///     prelude::*,
+///     primitives::Ellipse,
+///     style::{PrimitiveStyle, PrimitiveStyleBuilder},
+/// };
+/// # use embedded_graphics::mock_display::MockDisplay;
+/// # let mut display = MockDisplay::default();
+///
+/// // Ellipse with 1 pixel wide white stroke with top-left point at (10, 20) with a size of (30, 50)
+/// Ellipse::new(Point::new(10, 20), Size::new(30, 50))
+///     .into_styled(PrimitiveStyle::with_stroke(Rgb565::WHITE, 1))
+///     .draw(&mut display)?;
+///
+/// // Ellipse with styled stroke and fill with top-left point at (50, 20) with a size of (40, 30)
+/// let style = PrimitiveStyleBuilder::new()
+///     .stroke_color(Rgb565::RED)
+///     .stroke_width(3)
+///     .fill_color(Rgb565::GREEN)
+///     .build();
+///
+/// Ellipse::new(Point::new(50, 20), Size::new(40, 30))
+///     .into_styled(style)
+///     .draw(&mut display)?;
+///
+/// // Ellipse with blue fill and no stroke with a translation applied
+/// Ellipse::new(Point::new(10, 20), Size::new(20, 40))
+///     .translate(Point::new(65, 35))
+///     .into_styled(PrimitiveStyle::with_fill(Rgb565::BLUE))
+///     .draw(&mut display)?;
+/// # Ok::<(), core::convert::Infallible>(())
+/// ```
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
+pub struct Ellipse {
+    /// Top-left point of ellipse's bounding box
+    pub top_left: Point,
+
+    /// Size of the ellipse
+    pub size: Size,
+}
+
+impl Ellipse {
+    /// Create a new ellipse delimited with a top-left point with a specific size
+    pub const fn new(top_left: Point, size: Size) -> Self {
+        Ellipse { top_left, size }
+    }
+
+    /// Create a new ellipse centered around a given point with a specific size
+    pub fn with_center(center: Point, size: Size) -> Self {
+        let offset = size.saturating_sub(Size::new(1, 1)) / 2;
+        let top_left = center - offset;
+
+        Ellipse { top_left, size }
+    }
+
+    /// Return the center point of the ellipse
+    pub fn center(&self) -> Point {
+        self.bounding_box().center()
+    }
+
+    /// Return the center point of the ellipse scaled by a factor of 2
+    ///
+    /// This method is used to accurately calculate the outside edge of the ellipse.
+    /// The result is not equivalent to `self.center() * 2` because of rounding.
+    fn center_2x(&self) -> Point {
+        let radius = self.size.saturating_sub(Size::new(1, 1));
+
+        self.top_left * 2 + radius
+    }
+}
+
+impl Primitive for Ellipse {
+    type PointsIter = Points;
+
+    fn points(&self) -> Self::PointsIter {
+        Points::new(self)
+    }
+}
+
+impl ContainsPoint for Ellipse {
+    fn contains(&self, point: Point) -> bool {
+        let (size_sq, threshold) = compute_threshold(self.size);
+
+        is_point_inside_ellipse(size_sq, point * 2 - self.center_2x(), threshold)
+    }
+}
+
+impl Dimensions for Ellipse {
+    fn bounding_box(&self) -> Rectangle {
+        Rectangle::new(self.top_left, self.size)
+    }
+}
+
+impl Transform for Ellipse {
+    /// Translate the ellipse from its current position to a new position by (x, y) pixels,
+    /// returning a new `Ellipse`. For a mutating transform, see `translate_mut`.
+    ///
+    /// ```
+    /// # use embedded_graphics::primitives::Ellipse;
+    /// # use embedded_graphics::prelude::*;
+    /// let ellipse = Ellipse::new(Point::new(5, 10), Size::new(10, 15));
+    /// let moved = ellipse.translate(Point::new(10, 10));
+    ///
+    /// assert_eq!(moved.top_left, Point::new(15, 20));
+    /// ```
+    fn translate(&self, by: Point) -> Self {
+        Self {
+            top_left: self.top_left + by,
+            ..*self
+        }
+    }
+
+    /// Translate the ellipse from its current position to a new position by (x, y) pixels.
+    ///
+    /// ```
+    /// # use embedded_graphics::primitives::Ellipse;
+    /// # use embedded_graphics::prelude::*;
+    /// let mut ellipse = Ellipse::new(Point::new(5, 10), Size::new(10, 15));
+    /// ellipse.translate_mut(Point::new(10, 10));
+    ///
+    /// assert_eq!(ellipse.top_left, Point::new(15, 20));
+    /// ```
+    fn translate_mut(&mut self, by: Point) -> &mut Self {
+        self.top_left += by;
+
+        self
+    }
+}
+
+/// Iterator over all points inside the ellipse
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct Points {
+    iter: super::rectangle::Points,
+    center_2x: Point,
+    size_sq: Size,
+    threshold: u32,
+}
+
+impl Points {
+    fn new(ellipse: &Ellipse) -> Self {
+        let (size_sq, threshold) = compute_threshold(ellipse.size);
+
+        Self {
+            iter: ellipse.bounding_box().points(),
+            center_2x: ellipse.center_2x(),
+            size_sq,
+            threshold,
+        }
+    }
+
+    fn empty() -> Self {
+        Self {
+            iter: Rectangle::new(Point::zero(), Size::zero()).points(),
+            center_2x: Point::zero(),
+            size_sq: Size::zero(),
+            threshold: 0,
+        }
+    }
+}
+
+impl Iterator for Points {
+    type Item = Point;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        for point in &mut self.iter {
+            if is_point_inside_ellipse(self.size_sq, point * 2 - self.center_2x, self.threshold) {
+                return Some(point);
+            }
+        }
+
+        None
+    }
+}
+
+/// Pixel iterator for each pixel in the ellipse border
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct StyledEllipseIterator<C>
+where
+    C: PixelColor,
+{
+    iter: Points,
+    outer_color: Option<C>,
+    inner_size_sq: Size,
+    inner_color: Option<C>,
+    center: Point,
+    threshold: u32,
+}
+
+impl<C> StyledEllipseIterator<C>
+where
+    C: PixelColor,
+{
+    fn new(styled: &Styled<Ellipse, PrimitiveStyle<C>>) -> Self {
+        // Always use a stroke width of 0 if no stroke color was set.
+        let stroke_width = styled.style.effective_stroke_width();
+
+        let outer_size = styled.primitive.size;
+        let inner_size = outer_size.saturating_sub(Size::new(2 * stroke_width, 2 * stroke_width));
+        let iter = if !styled.style.is_transparent() {
+            Points::new(&styled.primitive)
+        } else {
+            Points::empty()
+        };
+
+        let (inner_size_sq, threshold) = compute_threshold(inner_size);
+
+        Self {
+            iter,
+            outer_color: styled.style.stroke_color,
+            inner_size_sq,
+            inner_color: styled.style.fill_color,
+            center: styled.primitive.center_2x(),
+            threshold,
+        }
+    }
+}
+
+impl<C> Iterator for StyledEllipseIterator<C>
+where
+    C: PixelColor,
+{
+    type Item = Pixel<C>;
+
+    fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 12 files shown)


jamwaffles:
@rfuest I think this is ready for review now. Some notes:

1. I didn't replace `Circle`s implementation with `Ellipse` as `Ellipse` is about 200% slower. It would be good if you could confirm this by running `cargo bench circle`, then changing the bench to the following code and running it again:

    ```diff
    fn filled_circle(c: &mut Criterion) {
        c.bench_function("filled circle", |b| {
            let style = PrimitiveStyleBuilder::new()
                .fill_color(Gray8::new(1))
                .stroke_color(Gray8::new(10))
                .stroke_width(1)
                .build();

    -       let object = &Circle::new(Point::new(100, 100), 100).into_styled(style);
    +       let object = &Ellipse::new(Point::new(100, 100), Size::new(100, 100)).into_styled(style);

            b.iter(|| object.into_iter().collect::<Vec<Pixel<Gray8>>>())
        });
    }
    ```

    Optimisations accepted ;)

2. I think this is squashable and can be cherry-picked into the `v0.6` branch without splitting anything out, but I'm not sure what's been added recently that's breaking. If something is in this PR that's breaking for 0.6 I'll split it into two or more commits and we can rebase this change instead of squashing.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub const fn new(top_left: Point, size: Size) -> Self {
+        Ellipse { top_left, size }
+    }
+
+    /// Create a new ellipse centered around a given point with a specific size
+    pub fn with_center(center: Point, size: Size) -> Self {
+        let offset = Size::new(
+            size.width.saturating_sub(1) / 2,
+            size.height.saturating_sub(1) / 2,
+        );
```
rfuest, line 74:
```suggestion
        let offset = size.saturating_sub(Size::new(1, 1)) / 2;
```

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub fn center(&self) -> Point {
+        self.bounding_box().center()
+    }
+
+    /// Return the center point of the ellipse scaled by a factor of 2
+    ///
+    /// This method is used to accurately calculate the outside edge of the ellipse.
+    /// The result is not equivalent to `self.center() * 2` because of rounding.
+    fn center_2x(&self) -> Point {
+        let radius = self.size - Size::new(1, 1);
```
rfuest, line 90:
```suggestion
        let radius = self.size.saturating_sub(Size::new(1, 1));
```

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,518 @@
...
+    fn new(styled: &Styled<Ellipse, PrimitiveStyle<C>>) -> Self {
+        let has_fill_color = styled.style.fill_color.is_some();
+        let has_stroke_color = styled.style.stroke_color.is_some();
+
+        // Always use a stroke width of 0 if no stroke color was set.
+        let stroke_width = if has_stroke_color {
+            styled.style.stroke_width
+        } else {
+            0
+        };
```
rfuest, line 219:
```suggestion
        let stroke_width = styled.style.effective_stroke_width();
```

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,518 @@
...
+
+        let outer_size = styled.primitive.size;
+        let inner_size = Size::new(
+            outer_size.width.saturating_sub(2 * stroke_width),
+            outer_size.height.saturating_sub(2 * stroke_width),
+        );
```
rfuest, line 225:
```suggestion
        let inner_size = outer_size.saturating_sub_(Size::new(2 * stroke_width, 2 * stroke_width));
```

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,518 @@
...
+        let iter = if has_stroke_color || has_fill_color {
```
rfuest, line 226:
```suggestion
        let iter = if !style.is_transparent() {
```

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,518 @@
...
+        let threshold = if diameter <= 4 {
+            diameter.pow(2) - diameter
+        } else {
+            diameter.pow(2)
+        } as i32;
+
+        (x + y) - threshold
+    } else {
+        let a = a.pow(2) as i32;
+        let b = b.pow(2) as i32;
```
rfuest, line 308: These don't change and could be stored in the iter instead of `size` to improve performance.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,518 @@
...
+{
+    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
+        display.draw_iter(self)
+    }
+}
+
+/// Uses the ellipse equation b^2 * x^2 + a^2 * y^2 - a^2 * b^2 to return a value signifying whether
+/// a given point lies inside (negative), on (0) or outside (positive) an ellipse centered around
+/// `(0, 0)` with width and height defined by the `size` parameter.
+fn size_to_threshold(size: &Size, point: &Point) -> i32 {
```
rfuest, line 284: Is there a reason to use references here? Most other code in e-g doesn't use references for `Size` and `Point`.

rfuest:
> 1. I didn't replace `Circle`s implementation with `Ellipse` as `Ellipse` is about 200% slower. It would be good if you could confirm this by running `cargo bench circle`, then changing the bench to the following code and running it again:

I get a change of +157% in runtime.

rfuest:
Did you consider this https://github.com/jamwaffles/embedded-graphics/issues/283#issuecomment-610284009?

jamwaffles:
Thanks for the quick review! I'll address those comments shortly.

> Did you consider this https://github.com/jamwaffles/embedded-graphics/issues/283#issuecomment-610284009?

I did, and I'd like to leave that for another day. We can't rotate rectangles either and I'd like to find a general solution.

jamwaffles:
> I get a change of +157% in runtime.

Ouch

rfuest:
> I did, and I'd like to leave that for another day. We can't rotate rectangles either and I'd like to find a general solution.

OK, good point. Can you open an issue for that?

jamwaffles:
Sure, I've opened #321 with some spitball comments

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: This would require storing and passing in the non-squared version of `a` as that's used in the optimised circle path. I tried adding that in but it's not very clean.

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: Thought it might improve performance but benchmarking this compared to passing by copy didn't change much. I've removed the references.

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: Ok, I've pushed a change that adds a `diameter` argument to the function. A slight performance improvement. Let me know what you think.

jamwaffles:
Changes addressed, ready for another look :+1:

rfuest:
Instead of copying the tests from `circle.rs` it would be easier to simply compare the result of drawing a circle and an ellipse: https://gist.github.com/rfuest/2c6918cce6dc9c8a1ac92c5ab28968c0

This makes it easier to test different styles and that revealed that there seems to be a difference for the stroked styles.

Could you also add tests for the `Points` iterator and the `ContainsPoint` impl?

jamwaffles:
> it would be easier to simply compare the result of drawing a circle and an ellipse

Very neat, thanks for the suggestion. I've copy-pasta'd your gist into the PR along with adding the other tests as requested.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,459 @@
...
+        } else {
+            diameter.pow(2)
+        } as i32;
+
+        (x + y) < threshold
+    } else {
+        let a = a as i32;
+        let b = b as i32;
+
+        b * x + a * y - b * a <= 0
```
rfuest, line 299: I think this should also be `< 0`.
If you set a wide border in the demo app and reduce the width or hight to `< 2 * stroke_width` you'll see a glitch appear if this is `<= 0`.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,459 @@
...
+    let x = x.pow(2);
+    let y = y.pow(2);
+
+    // Special case for circles, where width and height are equal
+    if a == b {
+        let threshold = if diameter <= 4 {
+            diameter.pow(2) - diameter / 2
+        } else {
+            diameter.pow(2)
+        } as i32;
```
rfuest, line 292: You can refactor this code by computing this in the constructor and pass `threshold` instead of `diameter`. For the non circle case you can rearrange the formula to ` b * x + a * y <  threshold` and pass `a * b` as `threshold`.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,459 @@
...
+{
+    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
+        display.draw_iter(self)
+    }
+}
+
+/// Uses the ellipse equation b^2 * x^2 + a^2 * y^2 - a^2 * b^2 to return a value signifying whether
+/// a given point lies inside (`true`) or outside (`false`) an ellipse centered around `(0, 0)` with
+/// width and height defined by the `size` parameter.
+fn size_to_threshold(size: Size, point: Point, diameter: u32) -> bool {
```
rfuest, line 275: Now that the function returns a `bool` the function name doesn't seem right anymore.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,459 @@
...
+        })
+    }
+}
+
+impl<'a, C: 'a> Drawable<C> for &Styled<Ellipse, PrimitiveStyle<C>>
+where
+    C: PixelColor,
+{
+    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
+        display.draw_iter(self)
```
rfuest, line 268: To be consistent we should probably add `DrawTarget::draw_ellipse`. Do you want to do this in this PR, later or not at all?

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let Self {
+            inner_size,
+            inner_color,
+            outer_color,
+            center,
+            inner_width,
+            ..
+        } = *self;
```
rfuest, line 247: I think this will copy more values than necessary per `next` call, which could impact performance. Using `for point in self.iter.by_ref() { ... }` isn't as nice as using `find_map` but you can use `self` inside the for loop.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+}
+
+impl Iterator for Points {
+    type Item = Point;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let ellipse = self.ellipse;
+
+        self.iter.find(|point| ellipse.contains(*point))
```
rfuest, line 182: The performance of this could be improved by storing `center_2x` and the squared size in the `Points` struct and using `size_to_threshold`, like in the `StyledEllipseIterator`.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,459 @@
...
+}
+
+/// Pixel iterator for each pixel in the ellipse border
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct StyledEllipseIterator<C>
+where
+    C: PixelColor,
+{
+    iter: Points,
+    outer_size: Size,
```
rfuest, line 193: This field is never used.

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: I've added it but in a separate commit with it's own changelog item. I haven't seen a display that can natively draw an ellipse, but it's here if people want it. I'll rebase this branch into master when it's done to preserve both commits.

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: Reproduced and fixed, thanks for catching that. I've added another test to catch any regressions.

jamwaffles:
Those improvements look good. I've integrated them into the PR.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,512 @@
...
+impl<'a, C: 'a> Drawable<C> for &Styled<Ellipse, PrimitiveStyle<C>>
+where
+    C: PixelColor,
+{
+    fn draw<D: DrawTarget<C>>(self, display: &mut D) -> Result<(), D::Error> {
+        display.draw_ellipse(self)
+    }
+}
+
+fn compute_threshold(size: Size, diameter: u32) -> u32 {
```
rfuest, line 281: The function could be changed to `fn compute_threshold(size: Size) -> (Size, u32)` by doing the squaring of size in this function. This would remove the need to pass a squared and non squared version of diameter and reduce some code duplication.

A slightly fancier version of this would be to use a struct instead of `(Size, u32)` which could also be used to implement the `is_point_inside_ellipse` function.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,512 @@
...
+        height: b,
+    } = size;
+
+    // Special case for circles, where width and height are equal
+    if a == b {
+        if diameter <= 4 {
+            diameter.pow(2) - diameter / 2
+        } else {
+            diameter.pow(2)
+        }
```
rfuest, line 293: We could share this code with `circle.rs` by making `diameter_to_threshold` in `circle.rs` `pub(crate)`. I'm not sure if that's an huge improvement, but it would make sure that they stay equal.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,512 @@
...
+    } = size;
+
+    let Point { x, y } = point;
+
+    let x = x.pow(2) as u32;
+    let y = y.pow(2) as u32;
+
+    // Special case for circles, where width and height are equal
+    if a == b {
+        (x + y) < threshold
```
rfuest, line 315:
```suggestion
        x + y < threshold
```

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,512 @@
...
+        Ellipse::new(Point::new(0, 0), size)
+            .into_styled(style)
+            .draw(&mut display)
+            .unwrap();
+
+        assert_eq!(display, MockDisplay::from_pattern(pattern));
+    }
+
+    fn test_circles(style: PrimitiveStyle<BinaryColor>) {
+        for diameter in 1..50 {
```
rfuest, line 353: Why did you change this to start from `1` instead of `0`? IMO diameter `0` should also match.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,512 @@
...
+    type Item = Pixel<C>;
+    type IntoIter = StyledEllipseIterator<C>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        StyledEllipseIterator::new(self)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
rfuest, line 334: If I haven't missed something, all tests seem to use `top_left = Point::zero()`. IMO at least one should use a different value for `top_left`.
This could maybe be combined with a test for `translate`, which isn't tested yet.

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: Good idea. I went with `pub(in crate::primitives)` in `circle.rs` to keep it more tightly scoped.

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: Good suggestion, changed. Adding a struct sure is fancy but this is a pretty small internal method so not worth the effort IMO.

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: Oops, that was a debugging artifact left over. It should indeed by `0..50`. Changed!

embedded-graphics/src/primitives/ellipse.rs
jamwaffles: I've added a combined non-zero origin and translate test.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -0,0 +1,543 @@
...
+    type PointsIter = Points;
+
+    fn points(&self) -> Self::PointsIter {
+        Points::new(self)
+    }
+}
+
+impl ContainsPoint for Ellipse {
+    fn contains(&self, point: Point) -> bool {
+        // let size_sq = Size::new(self.size.width.pow(2), self.size.height.pow(2));
```
rfuest, line 103:
```suggestion
```

embedded-graphics/src/primitives/ellipse.rs
rfuest: You only changed this in `circles_points` and not in `test_circles`.

jamwaffles:
ralf

jamwaffles:
:heart: :pray:

Issue #314, opened 2020-04-07
Tracking issue for #156 and #182


rfuest:
This is a tracking issue for the ongoing work on issues #156 and #182.
- [x] Change `Rectangle` to use top-left and size (#290)
- [x] Add `Points` iterators (#307)
- [x] Use `Rectangle` for `Dimensions` trait (#312)
- [x] Fix circle `Dimensions` (#317)
- [x] Add `contains(point: Point)` method to primitives (#318)
- [x] Support stroke alignment in `PrimitiveStyle` (#331)
- [x] Add `bottom_right()` getter to `Rectangle` (#333)
- [x] Use `bottom_right` in `Rectangle::contains` (#333)
- [x] `line::Points` should return exactly one `Point` for `start == end` (#353)
- [x] Check duplicate points (https://github.com/jamwaffles/embedded-graphics/pull/292#discussion_r419618612, [this test](https://github.com/jamwaffles/embedded-graphics/pull/292/files#diff-c1953210af5a72934cade95ee9ef2334R447-R455))
- [x] Add `fill_area` and `stroke_area` methods for styled primitives (#400)
- Reduce code duplication for styled and points iterators
  - [x] Circle (#317)
  - [x] Rectangle (#319)
  - [x] Line (#353)
  - [x] Triangle (#363)
- ~[ ] `Dimension` for styled primitive should consider stroke width and alignment~ (migrated to issue #428)
- ~[ ] Add more tests to check that drawing is consistent with #182~
- ~[ ] Rename `ContainsPoint`? (see https://github.com/jamwaffles/embedded-graphics/pull/318#issuecomment-611607535)~
- ~[ ] Investigate why performance of `cargo bench 'unfilled rectangle'` was reduced by 50% in PR #319~ (Rectangles are now drawn without using the iterator, so this should be invalid now)
- ~[ ] Support thick strokes and stroke alignment for triangles (or add note if it doesn't work https://github.com/jamwaffles/embedded-graphics/pull/331#pullrequestreview-400205527) ~ (will be implemented as part of PR #386)

rfuest:
I will close this issue and open individual issues for the remaining points to make it easier to track using the project.

jamwaffles:
Sounds good, thanks

Issue #268, opened 2020-03-11
Variable width font styling


jamwaffles:
Off the back of #267, I think it should be possible to make all built-in fonts variable width by default. Here's my idea:

* As far as I can see, all built in fonts can support variable widths, so set the `VARIABLE_WIDTH` property on `Font` to `true` by default (but retain it so there's an option to enforce monospace behaviour for third party fonts)

* One of the following ways of specifying letter spacing
    1. Two properties, `letter_spacing` for the number of pixels between characters (default 1 px) and `font_family` (inspired by the CSS property of the same name - could use a better one) which is an enum of `Monospace` or `VariableWidth`.

    2. An enum that encapsulates both properties:

        ```rust
        enum LetterSpacing {
            Monospace,
            VariableWidth { spacing: u32 }
        }
        ```

    3. Other suggestions accepted!

* If a font specifies `VARIABLE_WIDTH: false` in its impl of `Font`, the letter spacing properties on `TextStyle` shall have no effect.

    * _Or_ it could only respect the letter spacing but not shrink spacing greater than 1px?

@rfuest any thoughts?

aurelj:
I like the proposition overall.

But there is one catch. Currently the variable width font code only remove the blank columns at the right of the characters but not on the left, and the current `Monospace` fonts have characters such as `i` or `!` smack in the middle. So the variable width font code should first be improved to also remove blank columns at the left of the characters before being able to use every font as variable width.

jamwaffles:
> But there is one catch.

Ah, hm that's a good point. It shouldn't be too difficult to add a left-trimming iterator too though? I'd prefer not to modify the font bitmaps as that would complicate/break existing monospace behaviour.

I think a longer term goal might be to adopt the BDF font format (or a custom derived format for space reasons) which IIRC allows defining a bounding box for a character inside the nominal grid.

#269 also raised the ability to add line spacing now that multiline text is supported from that PR. `TextStyle` should probably also be able to set line spacing (but only _between_ lines).

rfuest:
I'm not a fan of trying to automatically crop glyphs to convert a monospace font into a proportional one. This isn't very flexible and doesn't allow any kind of kerning.
And IMO selecting if a font is monospaced or proportional isn't something that belongs into `TextStyle` but inside the `Font`.

My suggestion:
[...]

Issue #263, opened 2020-02-26
DrawTarget::size is broken


rfuest:
The recent changes in #261 caused some issues I didn't noticed because apparently the `size` function isn't used anywhere. But when I tried to refactor #259 an error occurred:
```
error[E0283]: type annotations needed
  --> simulator/examples/hello-world.rs:23:48
   |
23 |     let bottom_right = Point::zero() + display.size() - Point::new(1, 1);
   |                                                ^^^^ cannot infer type for type parameter `PixelC`
   |
   = note: cannot resolve `_: embedded_graphics::pixelcolor::PixelColor`
   = note: required because of the requirements on the impl of `embedded_graphics::draw_target::DrawTarget<_>` for `embedded_graphics_simulator::display::SimulatorDisplay<embedded_graphics::pixelcolor::binary_color::BinaryColor
```
That makes sense, because `size` is the only function that doesn't have any relation to the color type. I'm not sure what the best way to fix this is. Some options are:
* Remove automatic color conversion
  This would make using different color format images harder to use on color displays. But maybe some kind of conversion function (maybe as an extension trait) could be added to make this simpler. I guess at the moment you would need to use `map` on the iter to convert the colors manually, which isn't comfortable at all.
* Change the `draw_...` functions in the `DrawTarget` trait
   ```rust
   /// Draws a pixel on the display.
   fn draw_pixel(&mut self, item: drawable::Pixel<C>) -> Result<(), Self::Error>;
   ```
   to:
   ```rust
   /// Draws a pixel on the display.
   fn draw_pixel<PixelC>(&mut self, item: drawable::Pixel<PixelC>) -> Result<(), Self::Error>
      where PixelC: PixelColor + Into<C>;
   ```
   But then `C` should be a associated type instead, that represents the native color type of the framebuffer.
* Remove `size` from `DrawTarget`
  This way all functions would depend on `C` and `size` could be moved to something like `DrawTargetDimensions`.

Having to change this now isn't ideal, but I guess we need to do something, because all drivers will depend on this.

jamwaffles:
Isn't that because the pixel color type isn't defined [here](https://github.com/jamwaffles/embedded-graphics/pull/259/files#diff-a8c32b73fb30b81a0aa98efb1a3fdeefR12)? One or more of the SSD1306 examples use `disp.size()` without issue, so I think this might be ok, at least from the display driver side? I think this might be more to do with the simulator being a slightly odd case.

[...]

PR #324, opened 2020-04-15
Update v0.6

jamwaffles:
Update the `v0.6` with all the non-breaking stuff in master. This was a rebase, followed by an interactive rebase that removes some commits and ruins some others to just make it work.

This has been an _absolute nightmare_. Mostly my fault with the squash-n-merges going on. If we want to do parallel releases (and I very much do), we're going to have to be super careful about how we build on previous breaking changes.

There are probably loads of issues with this PR and I'd be more than happy to throw it away in favour of a better attempt/approach, but here we are for now :)

It goes without saying that this should be a rebase onto `v0.6`, at which point I can do a non-breaking release of `0.6.2` and create a PR to update the changelog in `master`.

### DIFF
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- [#307](https://github.com/jamwaffles/embedded-graphics/pull/307) Added `Primitive::points` to get an iterator over all points inside a primitive.
+
 ## [0.6.1] - 2020-04-01
 
 ### Added
@@ -456,8 +460,8 @@ A big release, focussed on ergonomics. There are new macros to make drawing and 
   ```
 
 <!-- next-url -->
-[unreleased]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.1...HEAD
 
+[unreleased]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.1...HEAD
 [0.6.1]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.0...embedded-graphics-v0.6.1
 [0.6.0]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.0-beta.2...embedded-graphics-v0.6.0
 [0.6.0-beta.2]: https://github.com/jamwaffles/embedded-graphics/compare/embedded-graphics-v0.6.0-beta.1...embedded-graphics-v0.6.0-beta.2
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -11,6 +11,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 - [#307](https://github.com/jamwaffles/embedded-graphics/pull/307) Added `Primitive::points` to get an iterator over all points inside a primitive.
 - [#318](https://github.com/jamwaffles/embedded-graphics/pull/317) Added `ContainsPoint` trait to check if a point is inside a closed shape.
 - [#320](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Ellipse` primitive.
+- [#320](https://github.com/jamwaffles/embedded-graphics/pull/320) Added `DrawTarget::draw_ellipse`
 
 ### Fixed
 
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -10,6 +10,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 - [#307](https://github.com/jamwaffles/embedded-graphics/pull/307) Added `Primitive::points` to get an iterator over all points inside a primitive.
 - [#318](https://github.com/jamwaffles/embedded-graphics/pull/317) Added `ContainsPoint` trait to check if a point is inside a closed shape.
+- [#320](https://github.com/jamwaffles/embedded-graphics/pull/320) Added the `Ellipse` primitive.
 
 ### Fixed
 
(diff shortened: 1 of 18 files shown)


rfuest:
The changes to the `Primitive` crate make some of these changes breaking. It's very unlikely that `Primitive` is implemented externally, but because we haven't prevented this, it's a breaking change IMO.

TBH I would prefer to instead keep the `0.7` development cycle short and release it after #314 and the other primitive changes are merged. Other additions and improvements, like sub images or a fixed `DrawTarget` trait, can be moved to later releases.

jamwaffles:
> The changes to the Primitive crate make some of these changes breaking.

Good point. I can remove it from this PR pretty easily I think - it's just the `fn points()` stuff right?

> TBH I would prefer to instead keep the 0.7 development cycle short and release it after #314 and the other primitive changes are merged.

I don't want to release too many breaking changes in quick succession, hence this (hopefully) non-breaking cherry pick PR for 0.6. It sucks for the growing community around this crate to have to repeatedly fix code, wait for drivers, nag authors, etc when we break stuff all the time. We can also make our lives easier in the future by not bundling up so many changes into one huge 0.6 branch update :)

> ... and release it after #314

Wasn't there some breaking font work (#225, #268, etc) that we wanted to get done as well?

> ... sub images ...

That would be a non-breaking change I think, so could be nicely released in 0.6.

> ... or a fixed DrawTarget trait, can be moved to later releases.

What needs fixing? #263 and #181? Anything else?

rfuest:
> I don't want to release too many breaking changes in quick succession, hence this (hopefully) non-breaking cherry pick PR for 0.6. It sucks for the growing community around this crate to have to repeatedly fix code, wait for drivers, nag authors, etc when we break stuff all the time. We can also make our lives easier in the future by not bundling up so many changes into one huge 0.6 branch update :)

I get your point, but I would prefer to have a shorter development cycle than for `0.6`. Releasing the coordinate system fixes early is important IMO, because otherwise users will have to deal with off by one errors later, which is a bigger pain than an API change that causes a compiler error.

Can we limit `0.7` to primitives, coordinate system, fixing `DrawTarget` and `Transform`? IMO this could be a good compromise.

It would be nice if we could decouple the driver interface to make migration between versions less of a pain. But because we use the primitives in `DrawTarget` I don't think there is a good way to achieve this.

> Good point. I can remove it from this PR pretty easily I think - it's just the `fn points()` stuff right?

That should be all at the moment.

Are you sure that adding `Ellipse` is a good idea? This would introduce more inconsistencies in the primitives module by using different parameters than `Circle` and being of by one for `size = radius * 2`.

> Wasn't there some breaking font work (#225, #268, etc) that we wanted to get done as well?

IMO it would be nice if external font renderers could be used in a similar way than the internal font support. To test this integration we would first need to implement an external renderer, for example for BDFs, but this would delay `0.7` further. If we don't test the integration before the `0.7` release we might have to break the API again in the next version which isn't great either.

> That would be a non-breaking change I think, so could be nicely released in 0.6.

But this could, for example, benefit from the `Points` iterators, which is a breaking change. This was the main reason I started working on the primitives before sub images.

> What needs fixing? #263 and #181? Anything else?

Not quite sure yet, I will need to look into this some more.

jamwaffles:
> I get your point, but I would prefer to have a shorter development cycle than for 0.6

Absolutely, I think that was nearly a year long!

> IMO it would be nice if external font renderers could be used in a similar way than the internal font support.

Not sure what you mean really. There are already external crates with custom font impls. What am I missing?

> Can we limit 0.7 to primitives, coordinate system, fixing DrawTarget and Transform?

Sure, that sounds good. However, I'm concerned that we're artificially spacing out breaking changes over multiple versions. Is there anything you have ideas about that's breaking that you didn't want to get into 0.7? I think people can put up with a couple of breaking releases, but probably not much more than that in quick succession. If we want to redesign the API, let's do it all at once. 

Going back to `master`-only development would certainly cause me less pain - this PR was horrendous to put together, and the end result is still messy. To ease migration from 0.6 to 0.7, perhaps we should release 0.7 alpha versions early (i.e. basically around now)? That would allow us to get the improved coordinate system, etc exercised by end users.

rfuest:
> Not sure what you mean really. There are already external crates with custom font impls. What am I missing?

I meant crates that add support for different font formats, like BDFs. IMO a text using a BDF font should also be drawable using the `embedded_graphics::font::Text` struct.

> Going back to `master`-only development would certainly cause me less pain - this PR was horrendous to put together, and the end result is still messy.

Maybe that's the easiest way for now. But releasing easy fixes like #309 in the 0.6 branch might still be a good idea.

> To ease migration from 0.6 to 0.7, perhaps we should release 0.7 alpha versions early (i.e. basically around now)? That would allow us to get the improved coordinate system, etc exercised by end users.

I would suggest to release the first alpha after #314 is done.

jamwaffles:
> I meant crates that add support for different font formats, like BDFs. IMO a text using a BDF font should also be drawable using the embedded_graphics::font::Text struct.

Yes, that makes a lot of sense. Could we create a `SpritemapFont` (or better name) for the existing bitmap fonts and move all common behaviour to a `Font` trait. Both the builtin bitmaps fonts and BDFs (etc, etc) could implement `Font` as well. Shall I open an issue to discuss this? If you have a better idea please open one instead :)

> Maybe that's the easiest way for now. But releasing easy fixes like #309 in the 0.6 branch might still be a good idea.

I'll get a PR together for #309 and any other small changes so far. I think we can close this PR now - it's too messy and far reaching in its changes. Let me know if you think otherwise.

> I would suggest to release the first alpha after #314 is done.

:+1:

rfuest:
> Yes, that makes a lot of sense. Could we create a `SpritemapFont` (or better name) for the existing bitmap fonts and move all common behaviour to a `Font` trait. Both the builtin bitmaps fonts and BDFs (etc, etc) could implement `Font` as well. Shall I open an issue to discuss this? If you have a better idea please open one instead :)

Yes, open an issue to discuss this.

jamwaffles:
I've opened #329 to discuss.

I think this PR can be closed now as it wasn't a good approach to start with. Please reopen if you disagree!

PR #325, opened 2020-04-15
Add Embedded Graphics Web Simulator

rahul-thakoor:
Mention embedded graphics web simulator in the display drivers section


rahul-thakoor:
added it. Thanks :)

PR #331, opened 2020-04-23
Add stroke alignment to PrimitiveStyle

rfuest:

## PR description

This PR adds support for setting the stroke alignment (as described in #182) to `PrimitiveStyle`.


embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -97,15 +97,31 @@ impl Rectangle {
         }
     }
 
+    fn with_center(center: Point, size: Size) -> Self {
```
rfuest, line 100: Should we make this `pub`?

jamwaffles: Yeah I think that's a good idea. Might be worth mentioning the even height behaviour like the `fn center` doc comment.

embedded-graphics/src/style/primitive_style.rs
```diff
@@ -199,12 +223,47 @@ where
...
+impl<C> From<&PrimitiveStyle<C>> for PrimitiveStyleBuilder<C>
+where
+    C: PixelColor,
+{
+    fn from(style: &PrimitiveStyle<C>) -> Self {
+        Self {
+            style: style.clone(),
+        }
+    }
+}
```
jamwaffles, line 248: Nice addition, it makes the simulator example code a lot cleaner :ok_hand:

rfuest:
> Could we add alignment to `Line` as well?

I thought we decided not to: https://github.com/jamwaffles/embedded-graphics/issues/182#issuecomment-600772351

> It would be required for `Polyline`, `Polygon` and `Triangle` to support stroke alignment as they use/will use `Line` under the hood. That said, we/I still need to figure out joint styles so we could have `Line` ignore stroke alignment for now.

I did some initial investigation on a possible implementation for `Triangle` and it could work without using the thick line functions. If `Triangle::expand` and `Triangle::shrink` can be implemented, similar to `Rectangle` and `Circle`, drawing strokes would be easy. To implement `expand` and `shrink` an algorithm similar to the perpendicular line bresenham with thickness accumulator could be used. But there would need to be some special considerations for triangles with very small angles or where the three points are on a line.

jamwaffles:
> I thought we decided not to

:+1: forgot about that

> I did some initial investigation on a possible implementation for Triangle and it could work without using the thick line functions.

How does it handle the joints between thick lines (if at all)? I'd like to add line cap/joint styles on `Line`, `Triangle`, `Polyline` and `Polygon` in the future so a solution would have to take that into account.

rfuest:
> How does it handle the joints between thick lines (if at all)? I'd like to add line cap/joint styles on `Line`, `Triangle`, `Polyline` and `Polygon` in the future so a solution would have to take that into account.

That solution wouldn't properly handle join styles and would only be a temporary solution until something better is implemented. But implementing `shrink` and `expand` would still be useful even if it wouldn't be used by the line drawing anymore.

One advantage of this simple approach is that the stroke and fill area are always triangle shaped and can easily be used for hit testing. If rounded and bevel joins are introduced the `stroke_area` won't be triangular anymore. I guess we need to reconsider if https://github.com/jamwaffles/embedded-graphics/pull/307#issuecomment-609475823 is still a viable idea.

embedded-graphics/src/primitives/rectangle.rs
rfuest: Done. I hope the comment is OK.

embedded-graphics/src/style/primitive_style.rs
rfuest: Thanks for reminding me that I added this, I forgot to add this to the changelog.

jamwaffles:
> That solution wouldn't properly handle join styles and would only be a temporary solution until something better is implemented.

:+1: 

> But implementing shrink and expand would still be useful even if it wouldn't be used by the line drawing anymore.

Yeah definitely :)

> I guess we need to reconsider if #307 (comment) is still a viable idea.

I think it is, at least for now. If there's a desire for pixel-perfect hit testing in the future we can add that in. It might be as simple as `impl ContainsPoint for Styled<Item>` to get a super slow but accurate version of it.

embedded-graphics/src/primitives/rectangle.rs
jamwaffles: Yep! Looks good

rfuest:
I've added a note to `PrimitiveStyle::stroke_alignment` and the example. I don't think we need to add this to every open shape primitive and it should be described once in the styling module.

I'll add `Triangle` to the tracking issue and try to implement thick strokes at some time. If this doesn't work I'll add some notes that triangles aren't supported in another PR.

PR #333, opened 2020-04-24
Add Rectangle::bottom_right()

rfuest:

## PR description

This PR adds a `bottom_right` method to `Rectangle`.

### DIFF
--- a/embedded-graphics/src/primitives/rectangle.rs
+++ b/embedded-graphics/src/primitives/rectangle.rs
@@ -64,10 +64,9 @@ impl Primitive for Rectangle {
 impl ContainsPoint for Rectangle {
     fn contains(&self, point: Point) -> bool {
         if point.x >= self.top_left.x && point.y >= self.top_left.y {
-            // FIXME: use Rectangle::bottom_right
-            let delta = Size::from_bounding_box(self.top_left, point);
-
-            delta.width <= self.size.width && delta.height <= self.size.height
+            self.bottom_right().map_or(false, |bottom_right| {
+                point.x <= bottom_right.x && point.y <= bottom_right.y
+            })
         } else {
             false
         }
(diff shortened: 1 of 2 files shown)


embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -107,6 +106,19 @@ impl Rectangle {
 
         self.top_left + Size::new(dx, dy)
     }
+
+    /// Returns the bottom right corner of this rectangle.
+    ///
+    /// Because the smallest rectangle that can be represented by its corners
+    /// has a size of 1 x 1 pixel this functions returns `None` if the width or
+    /// height of the rectangle is zero.
```
jamwaffles, line 114:
```suggestion
    /// Because the smallest rectangle that can be represented by its corners
    /// has a size of 1 x 1 pixels, this function returns `None` if the width or
    /// height of the rectangle is zero.
```

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -107,6 +106,19 @@ impl Rectangle {
...
+    pub fn bottom_right(&self) -> Option<Point> {
```
jamwaffles, line 115: What's the reasoning behind returning an `Option` and not clamping the `Point` to `self.top_left`? If the rectangle has a size of zero, AIUI from #182 wouldn't `bottom_right == top_left`?

rfuest: This is mentioned in the "Problems" paragraph in #182. `bottom_right == top_left` is a rectangle of `Size::new(1, 1)`.

I would expect that this code works for all rectangles:
```rust
let r = Rectangle::new(...);
assert_eq!(r, Rectangle::with_corners(r.top_left, r.bottom_right.unwrap());
```

If `bottom_right` is clamped to `top_left` it would return the wrong value for zero sized rectangles. This means that the code above would convert a rectangle with `Size::zero()` to a rectangle with `Size::new(1, 1)`.

jamwaffles: I read throught that bullet but didn't quite get the meaning behind it, but I see what you mean now. Makes sense - thanks for explaining 👍

rfuest:
> Got some RSI I'm trying to heal up :(

That sucks, I hope it gets better soon.

PR #334, opened 2020-05-04
Add note about new linkchecker version for 20.04

jamwaffles:
Adds a quick note to the dev env setup to support Ubuntu 20.04 and other systems that only provide Python 3 now.

### DIFF
--- a/embedded-graphics/README.md
+++ b/embedded-graphics/README.md
@@ -255,8 +255,16 @@ rustup update
 # Ensure rustfmt is installed
 rustup component add rustfmt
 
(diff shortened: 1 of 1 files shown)


rfuest:
LGTM and thanks for this note. I wasn't using `linkchecker` locally because I didn't want to install python 2 just to run this.

jamwaffles:
I'd rather be using a pure Rust version so we can remove the Python dependency, but I haven't found a Rust linkchecker that does exactly what the Python version does yet. `$num_sideprojects + 1` ...

PR #336, opened 2020-05-05
Rounded rectangles

### LINKED ISSUE #83: Rounded corners on rectangles
It should be possible to get all fancy and have rounded corners on rectangles.

jamwaffles:

## PR description

Adds the `RoundedRectangle` primitive.

Closes #83 (old school!)


rfuest:
Oh, individual corner radii and even non circular corners, that's fancier than I expected, nice.

Can you start assigning yourself to issues you are working on? I was considering to also add this and that might have caused duplicate work. I'll try to remember to do the same.

jamwaffles:
> that's fancier than I expected, nice

Thanks :grin: go big or go home I suppose!

> Can you start assigning yourself to issues you are working on?

Of course, I'll try to remember to do that from now on.

embedded-graphics/src/geometry/size.rs
```diff
@@ -180,6 +180,11 @@ impl Size {
     pub(crate) fn center_offset(self) -> Self {
         self.saturating_sub(Size::new(1, 1)) / 2
     }
+
+    /// Returns the componentwise minimum of two `Size`s
+    pub(crate) fn component_min(self, other: Self) -> Self {
```
jamwaffles, line 185: Should I add `Point::component_min`, as well as `component_max` for both `Point` and `Size`?

I could also make this `pub`. Thoughts?

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,361 @@
...
+    DrawTarget,
+};
+
+/// Rounded rectangle primitive
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct RoundedRectangle {
+    /// The base rectangle
+    pub rectangle: Rectangle,
+
+    corners: [EllipseQuadrant; 4],
```
jamwaffles, line 23: I decided not to make this `pub` so we don't have to export `EllipseQuadrant`. If you think it's a good idea export it, I can make it pub, but it might be a bit of an odd thing to export as an e-g primitive?

embedded-graphics/src/primitives/ellipse_quadrant.rs
```diff
@@ -0,0 +1,434 @@
...
+            bounding_box: Rectangle::new(top_left, radius),
+        }
+    }
+
+    /// Expand the ellipse quadrant by a given offset
+    ///
+    /// This method will expand the ellipse quadrant in the direction of its outer edge. The center
+    /// point will not be moved.
+    ///
+    /// If the radius of the ellipse quadrant is zero, no expand will be performed.
```
jamwaffles, line 54: Lots of caveats here, but this is the exact behaviour I need for `RoundedRectangle`. Should it be made more general with a `RoundedRectangle`-specific method somewhere else?

embedded-graphics/src/geometry/size.rs
rfuest: > Should I add `Point::component_min`, as well as `component_max` for both `Point` and `Size`?

IMO that's a good idea, I could have used these methods before.

> I could also make this `pub`. Thoughts?

I think I would make them `pub`, but I would like to another look at the `Point` and `Size` API before making a final decision. Can you leave them `pub(crate)` for now and open an issue to discuss this change? There were some other possible additions to the API I considered to make some time ago but can't remember at the moment.

embedded-graphics/src/primitives/ellipse_quadrant.rs
rfuest: If this primitive is only used internally we don't need a generic `expand` impl. But I would change the name to make it clear that this function is different than the other `expand` methods.

embedded-graphics/src/primitives/rounded_rectangle.rs
rfuest: I haven't had time to look at the code in detail, but does this `EllipseQuadrant` primitive offer many advantages? Would splitting the code into `RoundedRectangle` and `EllipseQuadrant` make it harder to implement future optimizations, like drawing a rounded rectangle by using scanlines? (A scanline could start in one `EllipseQuadrant` and end in antoher)

> If you think it's a good idea export it, I can make it pub, but it might be a bit of an odd thing to export as an e-g primitive?

I agree that exporting `EllipseQuadrant` would be odd. If possible I would like to revert to the earlier version that used a `Size` to specify the corner radii. Not being able to access the `corners` field would make this primitive different than all others and we would need to add getters and setters to access and change the radii.

Regardless of whether we use `EllipseQuadrant` or `Size` can we also use a struct instead of an array for the radii?
```rust
pub struct CornerRadii {
    pub top_left: Size,
    pub top_right: Size,
    pub bottom_left: Size,
    pub bottom_right: Size,
}
```
IMO it's easier to use than having to remember which array index corresponds to which corner. And we could also add additional constructors like `CornerRadii::new_equal(size)`, `CornerRadii::new_hv(h_size, v_size)` or `CornerRadii::new_circular(10)`.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,361 @@
...
+        Self::with_corners(
+            rectangle,
+            [corner_radius, corner_radius, corner_radius, corner_radius],
+        )
+    }
+
+    /// Creates a new rounded rectangle with different corner radii.
+    ///
+    /// Corner radii are specified from the top-left corner in a clockwise direction
+    pub fn with_corners(rectangle: Rectangle, corner_radii: [Size; 4]) -> Self {
```
rfuest, line 59: This is a confusing name, because we also have `Rectangle::with_corners` which does something different.

embedded-graphics/src/geometry/size.rs
jamwaffles: I've added all the methods and changed them to `pub(crate)` and ticket opened at #338 .

embedded-graphics/src/primitives/rounded_rectangle.rs
jamwaffles: I've changed to a struct which is much better, thanks for the suggestion.

> does this EllipseQuadrant primitive offer many advantages?

Mostly code separation. The `ellipse_quadrant.rs` file is pretty long and it has its own unit tests, etc. 

> Would splitting the code into RoundedRectangle and EllipseQuadrant make it harder to implement future optimizations, like drawing a rounded rectangle by using scanlines?

If we leave it as an internal implementation detail, we can optimise it out later. 

The other option is ripping out bits of `Ellipse` and reusing them directly in `rounded_rectangle::Points` and `rounded_rectangle::StyledRoundedRectangleIterator`. That way I think we could get rid of `EllipseQuadrant` completely. I might take a look at this approach.

embedded-graphics/src/primitives/rounded_rectangle.rs
jamwaffles: Good point. I've changed it to `with_radii`. Any ideas for a better name?

rfuest: Maybe rename `with_corners` to `new` and rename the current `new` to `with_equal` or `with_equal_radii`?

jamwaffles: I've gone with `with_equal_corners` as `with_equal_radii` might be misunderstood as a rounded rectangle with circular corners (i.e. `equal` meaning equal X and Y radius).

rfuest: SGTM :+1:

jamwaffles:
Might still be some rough edges and there are still things I'm not happy with, but I'm starting to lose the forest for the trees so a quick first review with any obvious suggestions would be much appreciated.

Anything around improving `RoundedRectangle::contains` would be great - both code and perf wise.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    /// Creates a new rounded rectangle with different corner radii.
+    pub fn new(rectangle: Rectangle, corners: CornerRadii) -> Self {
+        let Rectangle { size, .. } = rectangle;
+
+        let corners = CornerRadii {
+            top_left: corners.top_left.component_min(size / 2),
+            top_right: corners.top_right.component_min(size / 2),
+            bottom_right: corners.bottom_right.component_min(size / 2),
+            bottom_left: corners.bottom_left.component_min(size / 2),
```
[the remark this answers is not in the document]
jamwaffles: Ah, good catch. I could limit the sum of two corner radii to their total edge length perhaps? Do you have a better idea in mind?

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,596 @@
...
+            corners.bottom_right,
+            Quadrant::BottomRight,
+        );
+        let bl = EllipseQuadrant::new(
+            top_left + size.y_axis() - corners.bottom_left.y_axis(),
+            corners.bottom_left,
+            Quadrant::BottomLeft,
+        );
+
+        if tl.bounding_box().contains(point) {
```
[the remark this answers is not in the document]
jamwaffles: ~7% performance improvement, hooray

jamwaffles:
Most changes made, thanks for the suggestions.

> The performance of the contains methods isn't great for all complex primitives, because there is no way of caching intermediate values.

It would be super helpful if we could start putting private fields on the primitives. Maybe worth asking in Matrix/Twitter to see if anybody constructs primitives not using the public methods. I think we're leaving quite a bit of performance on the table.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,596 @@
...
+            rectangle, corners, ..
+        } = self;
+
+        let Rectangle { top_left, size, .. } = *rectangle;
+
+        if !rectangle.contains(point) {
+            return false;
+        }
+
+        let tl = EllipseQuadrant::new(top_left, corners.top_left, Quadrant::TopLeft);
```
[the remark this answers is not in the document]
jamwaffles: Very good idea, thanks!

embedded-graphics/src/geometry/point.rs
```diff
@@ -666,4 +694,15 @@ mod tests {
 
         assert_eq!(Point::from(&c), Point::new(20, 20));
     }
+
+    #[test]
+    // TODO: Delete this test if component_min/component_max are made public. It will be redundant
+    // with their respective doc comments.
```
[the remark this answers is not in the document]
jamwaffles: I've removed all the `TODO` comments. The doc/unit tests are mostly dupes of each other now. Maybe some room for improvement but it can't hurt to double check.

embedded-graphics/src/primitives/rounded_rectangle.rs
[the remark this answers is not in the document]
rfuest: No, I don't have a better idea and i think limiting the sum is the right way to handle this. But the exact implementation how the radii are reduced to fit will require some consideration.

For reference: The way CSS handles this is described in: https://www.w3.org/TR/css-backgrounds-3/#corner-overlap. But this might require floating point math.

rfuest:
> It would be super helpful if we could start putting private fields on the primitives. Maybe worth asking in Matrix/Twitter to see if anybody constructs primitives not using the public methods. I think we're leaving quite a bit of performance on the table.

I consider the primitives as minimal descriptions of a geometric shapes that should be as small as possible. Adding additional fields will increase the memory consumption for primitives of that type, even if none of methods that benefit from the additional fields are used.

We could add something similar to `Styled` that wraps a primitive with the additional fields required for fast hit testing or similar operations. This way the user can selectively create the additional fields when they are necessary.

jamwaffles:
> I consider the primitives as minimal descriptions of a geometric shapes that should be as small as possible.

Sure, that makes sense.

> We could add something similar to Styled that wraps a primitive with the additional fields

I had this thought too. Something for a later time :)

embedded-graphics/src/primitives/rounded_rectangle.rs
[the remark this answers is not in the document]
jamwaffles: I'll take a look at the CSS shrinking method later and see if it's doable with integer maths :crossed_fingers:

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,777 @@
...
+            bounding_box.width as i32 - self.top_left.width as i32 - self.top_right.width as i32;
+        let free_space_right = bounding_box.height as i32
+            - self.top_right.height as i32
+            - self.bottom_right.height as i32;
+        let free_space_bottom = bounding_box.width as i32
+            - self.bottom_left.width as i32
+            - self.bottom_right.width as i32;
+        let free_space_left = bounding_box.height as i32
+            - self.top_left.height as i32
+            - self.bottom_left.height as i32;
```
jamwaffles, line 58: Thoughts on how to strip all the `as i32`s? I considered adding `TryFrom<Size> for Point` which would probably make this cleaner, but would either require `Result` or some `unwrap()`s which I'm not keen on.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,777 @@
...
+/// each corner may have a separate radius applied to it.
+///
+/// # Overlapping corners
+///
+/// If one or more corner radii are too large to be contained along an edge of the rectangle, all
+/// corner radii will be reduced by half the largest overlap in pixels (rounded down).
+///
+/// This is similar but not identical to
+/// [how the CSS specification works](https://www.w3.org/TR/css-backgrounds-3/#corner-overlap) as it
+/// relies on floating point calculations.
```
jamwaffles, line 97: I'm happy to remove this if you think it's not necessary. It might be irrelevant, but also might provide some better context for the corner shrinking behaviour.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,779 @@
...
+///
+/// Creates a rectangle with rounded corners. Corners can be circular or elliptical in shape, and
+/// each corner may have a separate radius applied to it.
+///
+/// # Overlapping corners
+///
+/// If one or more corner radii are too large to be contained along an edge of the rectangle, all
+/// corner radii will be reduced by half the largest overlap in pixels (rounded down).
+///
+/// The [example further down the page](#overlapping-corner-radii) demonstrates this behaviour.
```
jamwaffles, line 95: Better descriptions gladly accepted :grin:

embedded-graphics/src/primitives/rounded_rectangle.rs
rfuest: You could use `checked_sub`:
```rust
let overlap_top = (self.top_left.width + self.top_right.width).checked_sub(bounding_box.width);
```
This would return `None` if there is no overlap and `Some(u32)` if it overlaps.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,768 @@
...
+        // Compute free space for each edge. If there is remaining free space, the call to
+        // checked_sub() will return `None`.
+        let free_space_top =
+            (self.top_left.width + self.top_right.width).checked_sub(bounding_box.width);
+        let free_space_right =
+            (self.top_right.height + self.bottom_right.height).checked_sub(bounding_box.height);
+        let free_space_bottom =
+            (self.bottom_left.width + self.bottom_right.width).checked_sub(bounding_box.width);
+        let free_space_left =
+            (self.top_left.height + self.bottom_left.height).checked_sub(bounding_box.height);
```
jamwaffles, line 55: Hm, the original commented got deleted during rebase but wow... Rust provides just the right methods once again! Thanks for the suggestion.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,768 @@
...
+            bottom_left: radius,
+        }
+    }
+
+    /// Confine corner radii that are too large to a given bounding rectangle
+    fn confine(self, bounding_box: Size) -> Self {
+        // Compute free space for each edge. If there is remaining free space, the call to
+        // checked_sub() will return `None`.
+        let free_space_top =
+            (self.top_left.width + self.top_right.width).checked_sub(bounding_box.width);
```
rfuest, line 49: Would `saturating_sub` and using `largest_overlap > 0` as the condition below also work? This way the largest overlap could be calculated without having to deal with the `Option`s.
```suggestion
            (self.top_left.width + self.top_right.width).saturating_sub(bounding_box.width);
```

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,768 @@
...
+            free_space_bottom,
+            free_space_left,
+        ]
+        .iter()
+        .filter_map(|s| *s)
+        .max();
+
+        if let Some(largest_overlap) = largest_overlap {
+            // Reduce each corner radius by (largest overlap / 2)
+            let reduce_by = Size::new_equal(largest_overlap / 2);
```
rfuest, line 70: When the overlap is e.g. 1px `reduce_by` will be `Size::zero()` and the corners would still overlap. This could be fixed by rounding up: `Size::new_equal((largest_overlap + 1) / 2)`

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,768 @@
...
+
+/// Rounded rectangle primitive.
+///
+/// Creates a rectangle with rounded corners. Corners can be circular or elliptical in shape, and
+/// each corner may have a separate radius applied to it.
+///
+/// # Overlapping corners
+///
+/// If one or more corner radii are too large to be contained along an edge of the rectangle, all
+/// corner radii will be reduced by half the largest overlap in pixels (rounded down).
```
rfuest, line 92: See comment above, about rounding of the half overlap.

embedded-graphics/src/primitives/rounded_rectangle.rs
rfuest: I'll need take another look at this in my next review pass, because I'm not sure how relevant this is at the moment and I haven't really looked at the doc comments yet.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,768 @@
...
+    /// corners.
+    pub fn with_equal_corners(rectangle: Rectangle, corner_radius: Size) -> Self {
+        Self::new(rectangle, CornerRadii::new_equal(corner_radius))
+    }
+
+    /// Creates a new rounded rectangle with different corner radii.
+    pub fn new(rectangle: Rectangle, corners: CornerRadii) -> Self {
+        Self {
+            rectangle,
+            corners: corners.confine(rectangle.size),
```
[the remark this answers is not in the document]
jamwaffles: Hm, they don't, and I'm not sure there's a good solution. Where would you suggest confining them? I could confine them in `Points::new()` or `StyledRoundedRectangleIterator::new()`, but that could allow a plain `RoundedRectangle` to be created in an invalid state. It would also require re-confining every time `.contains()` is called. 

What about adding another private struct, `ConfinedCornerRadii`? `CornerRadii::confine()` could produce a `ConfinedCornerRadii` which `RoundedRectangle::new()` accepts.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,768 @@
...
+        }
+    }
+
+    fn get_corner_quadrant(&self, quadrant: Quadrant) -> EllipseQuadrant {
+        let Self {
+            rectangle, corners, ..
+        } = self;
+
+        let Rectangle { top_left, size, .. } = *rectangle;
```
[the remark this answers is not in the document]
jamwaffles: Yes, they're there for future proofing. I can remove them if you think it adds too much clutter?

simulator/src/bin/generate-example-screenshots.rs
```diff
@@ -197,6 +197,40 @@ some display drivers implement accelerated drawing of iterators."#,
...
+                .stroke_width(3)
+                .fill_color(Rgb888::GREEN)
+                .build();
+            {}
+            let corners = CornerRadii {
+                top_left: Size::new(10, 10),
+                top_right: Size::new(10, 10),
+                bottom_right: Size::new(10, 10),
+                bottom_left: Size::new(10, 10),
+            };
```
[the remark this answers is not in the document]
jamwaffles: Ok, that makes sense. I originally went for unequal corners but the generated screenshot never looked like a good example of a rounded rect. I'll change this to `with_equal_corners`.

embedded-graphics/src/primitives/ellipse.rs
```diff
@@ -99,6 +97,12 @@ impl Ellipse {
     }
 }
 
+pub(crate) fn center_2x(top_left: Point, size: Size) -> Point {
```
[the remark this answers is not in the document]
jamwaffles: I cheaped out and copypastad the `Ellipse::center_2x` docs nearly verbatim. Should be ok for now?

jamwaffles:
Thanks for the review. Some changes made, other points commented on.

embedded-graphics/src/primitives/rounded_rectangle.rs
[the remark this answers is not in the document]
rfuest: No, I don't mind leaving them in. I just wanted to make sure I understand the reason they are there.

embedded-graphics/src/primitives/rounded_rectangle.rs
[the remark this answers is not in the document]
rfuest: The first question we need to answer is if it should be allowed to have overlapping radii in `RoundedRectangle`. One advantage of not confining the radii is that you can create a small rectangle with large radii and then scale it up by changing `rrect.rectangle.size` without loosing the original radii. This could, for example, be used in an animation of an expandable UI element.

> What about adding another private struct, `ConfinedCornerRadii`? `CornerRadii::confine()` could produce a `ConfinedCornerRadii` which `RoundedRectangle::new()` accepts.

I don't quite understand how this would work. Is the user supposed to call `confine` before passing the radii into `RoundedRectangle::new`?

embedded-graphics/src/primitives/ellipse.rs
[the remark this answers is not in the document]
rfuest: Yes, that's fine.

rfuest:
Reducing all radii equally to confine them causes some problems.

This can be demonstrated by changing the drawing code in the example to:
```
    RoundedRectangle::new(
        Rectangle::new(top_left, size),
        CornerRadii {
            top_left: Size::new_equal(10),
            top_right: radius,
            bottom_left: Size::new_equal(10),
            bottom_right: radius,
        },
    )
    .into_styled(
        PrimitiveStyleBuilder::new()
            .stroke_width(stroke_width)
            .stroke_color(Rgb888::RED)
            .stroke_alignment(align)
            .fill_color(Rgb888::GREEN)
            .build(),
    )
    .draw(display)
    .unwrap();
```
If you set a large radius and begin to reduce the height of the rectangle the smaller radii disappear as soon as there is some overlap. This happens even if there is more than enough space to draw the small radii and isn't the behavior users expect IMO.

jamwaffles:
I'm not sure what the better behaviour (that's implementable with integer maths) would be tbh. Have you got any suggestions? If we can't think of anything, it might be best to add a note to the confining behaviour description that the behaviour could change in the future.

embedded-graphics/src/primitives/rounded_rectangle.rs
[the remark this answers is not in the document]
jamwaffles: > The first question we need to answer is if it should be allowed to have overlapping radii in RoundedRectangle.

I don't think we should. It breaks the `ContainsPoint` results in unexpected ways. If we call `.confine()` in the `Points` and `StyledRoundedRectangle` iterator constructors, they would  produce a different shape than the theoretical representation in `RoundedRectangle`. To exemplify the `ContainsPoint` issue, `shape.contains()` would return a lot of false positives for the following rounded rect:

image

> This could, for example, be used in an animation of an expandable UI element.

In this specific example with the current behaviour, corner radii will scale with the shape as it animates, up to their configured size which would work well for a scaling animation.

> Is the user supposed to call confine before passing the radii into RoundedRectangle::new?

That was the idea, yeah. Not a great one so open to others :)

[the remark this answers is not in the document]
rfuest: > I don't think we should. It breaks the `ContainsPoint` results in unexpected ways. If we call `.confine()` in the `Points` and `StyledRoundedRectangle` iterator constructors, they would produce a different shape than the theoretical representation in `RoundedRectangle`. To exemplify the `ContainsPoint` issue, `shape.contains()` would return a lot of false positives for the following rounded rect:

I didn't mean that they should use different radii. All methods and iters, including `contains`, would also need to use the confined radii internally. Doing this for every point will be slow, but this could be improved later by adding the wrapper we discussed earlier.

> In this specific example with the current behaviour, corner radii will scale with the shape as it animates, up to their configured size which would work well for a scaling animation.

If you create a new `RoundedRectangle` this works, but not by reusing a rectangle and setting `rectangle.size` to a new value. Or am I missing something?

> That was the idea, yeah. Not a great one so open to others :)

IMO we would need to make the fields private, otherwise we can't guarantee that the radii are correctly confined.

rfuest:
> I'm not sure what the better behaviour (that's implementable with integer maths) would be tbh. Have you got any suggestions? If we can't think of anything, it might be best to add a note to the confining behaviour description that the behaviour could change in the future.

Never mind, I should read the documents I link more carefully before commenting. I wrongly assumed that CSS would only scale the corners that are overlapping and not all corners.

After reading the CSS document again I think it should be possible to implement the same algorithm using integer math:
```
f = min(L_i / S_i)
=> R_scaled = R * f = (R * L_i) / S_i
```

This might preserve the shape of corners better, but personally I don't care too much about non circular corners and would also be OK with keeping the current algorithm.

jamwaffles:
No worries. Let's leave it as is :)

embedded-graphics/src/primitives/rounded_rectangle.rs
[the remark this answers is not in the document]
jamwaffles: > If you create a new RoundedRectangle this works, but not by reusing a rectangle and setting rectangle.size to a new value.

Ah, I see what you mean now. Yes, that wouldn't work which is bad. I'll change the confine to the iters/contains code and just swallow the perf hit for now. This also means we can keep all the `RoundedRectangle` fields public.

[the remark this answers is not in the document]
jamwaffles: Changed. I removed the doc example as it's no longer possible to assert the confined corners, or show them in an example in a meaningful way.

The benchmark only slowed down by ~10% as well which is a pleasant surprise.

simulator/examples/primitives-rounded-rectangle.rs
```diff
@@ -0,0 +1,118 @@
+//! # Example: Rounded rectangle
+//!
+//! This example draws a `RoundedRectangle`. Click and drag to move the lower-left corner around
+//! the screen. The up/down arrow keys adjust stroke width, left/right the corner radius. Space
```
rfuest, line 4: The dragged corner isn't always the lower-left corner.

simulator/examples/primitives-rounded-rectangle.rs
```diff
@@ -0,0 +1,118 @@
...
+use sdl2::keyboard::Keycode;
+
+fn draw(
+    top_left: Point,
+    size: Size,
+    radius: Size,
+    stroke_width: u32,
+    align: StrokeAlignment,
+    display: &mut SimulatorDisplay<Rgb888>,
+) {
```
rfuest, line 25:
```suggestion
fn draw(
    base_rectangle: Rectangle,
    radius: Size,
    stroke_width: u32,
    align: StrokeAlignment,
    display: &mut SimulatorDisplay<Rgb888>,
) {
```

simulator/examples/primitives-rounded-rectangle.rs
```diff
@@ -0,0 +1,118 @@
...
+        .scale(2)
+        .pixel_spacing(1)
+        .build();
+    let mut window = Window::new("Rounded rectangle debugger", &output_settings);
+
+    let top_left = Point::new(20, 20);
+
+    let mut mouse_down = false;
+
+    let mut bounding_rect = Rectangle::with_corners(top_left, Point::new(100, 100));
```
rfuest, line 53: The docs for `RoundedRectangle::new` call this the "base rectangle", so I would suggest to use the same name here.

rfuest:
I just noticed that there is no macro to create this primitive.

Do you also want to add this to the `primitives-fill`, `primitives-fill-macros` and `primitives-stroke` examples?

embedded-graphics/src/primitives/rounded_rectangle.rs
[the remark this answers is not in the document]
rfuest: > The benchmark only slowed down by ~10% as well which is a pleasant surprise.

Nice, that's far less than I expected.

I've opened #343 to discuss future performance improvements like the wrapper we discussed earlier.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,743 @@
...
+    pub top_right: Size,
+
+    /// Bottom right corner radius
+    pub bottom_right: Size,
+
+    /// Bottom left corner radius
+    pub bottom_left: Size,
+}
+
+impl CornerRadii {
```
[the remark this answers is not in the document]
jamwaffles: I think we should keep `CornerRadii::new_equal` as it's a common operation. We could rename it to `new` and let the doc comment explain its behaviour.

I like the idea of a builder so I've added one in. Let me know what you think.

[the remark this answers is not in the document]
rfuest: > I think we should keep `CornerRadii::new_equal` as it's a common operation. We could rename it to `new` and let the doc comment explain its behaviour.

Most of the time the user will use `RoundedRectangle::with_equal_corners` for this case. But we can also keep this, but I would rename it to `new` as you suggested.

> I like the idea of a builder so I've added one in. Let me know what you think.

Looks good. Can you also add a `From<&CornerRadii> for CornerRadiiBuilder` impl? That turned out to be a good idea for `PrimitiveStyle`, if you want to modify an existing object using the same builder methods.

jamwaffles:
> I just noticed that there is no macro to create this primitive.

Well spotted. I wrote this:

```rust
#[macro_export]
macro_rules! egroundedrectangle {
    (base_rectangle = $base_rectangle:expr, radii = $radii:expr $(,)?) => {{
        $crate::egroundedrectangle!(
            base_rectangle = $base_rectangle,
            radii = $radii,
            style = $crate::style::PrimitiveStyle::default()
        )
    }};
    (base_rectangle = $base_rectangle:expr, radii = $radii:expr, style = $style:expr $(,)?) => {{
        $crate::primitives::RoundedRectangle::new($base_rectangle, $radii).into_styled($style)
    }};
}
```

But then wondered if we may want shorter ways of specifying the base rect and/or radii. Thoughts?

> Do you also want to add this to the primitives-fill, primitives-fill-macros and primitives-stroke examples?

Yes, that's a good idea.

rfuest:
> But then wondered if we may want shorter ways of specifying the base rect and/or radii. Thoughts?

Hmm, if you need to build the base rect and radii manually the macro doesn't provide any advantage IMO. The only solution for shortening the base rect definition I see at the moment is copying the `top_left`+`size` and `corners` variants from the `egrectangle` macro. We could also add separate variants for `radii` and a single `radius`, which calls `CornerRadii::new(radius)` internally.

jamwaffles:
I've modified the examples, added a macro and fixed a bug (with added test) with non-filled rounded rects. Ready for another pass :)

embedded-graphics/src/primitives/corner_radii.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Bottom right corner radius
+    pub bottom_right: Size,
+
+    /// Bottom left corner radius
+    pub bottom_left: Size,
+}
+
+impl CornerRadii {
+    /// Create a new set of corner radii with all corners having equal values.
+    pub fn new(radius: Size) -> Self {
```
rfuest, line 23:
```suggestion
    pub const fn new(radius: Size) -> Self {
```

embedded-graphics/src/primitives/corner_radii.rs
```diff
@@ -0,0 +1,356 @@
...
+
+impl CornerRadiiBuilder {
+    /// Create a new corner radii builder.
+    ///
+    /// All radii are defaulted to 0px x 0px.
+    pub fn new() -> Self {
+        Self {
+            corners: CornerRadii::default(),
+        }
+    }
```
rfuest, line 79:
```suggestion
    pub const fn new() -> Self {
        Self {
            corners: CornerRadii::new(Size::zero()),
        }
    }
```

embedded-graphics/src/primitives/corner_radii.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/// [`CornerRadii`] builder.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub struct CornerRadiiBuilder {
+    corners: CornerRadii,
+}
+
+impl CornerRadiiBuilder {
```
rfuest, line 71: All methods can be `const`.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,778 @@
...
+/// )
+/// .into_styled(style)
+/// .draw(&mut display)?;
+/// # Ok::<(), core::convert::Infallible>(())
+/// ```
+///
+/// ## Different corner radii
+///
+/// This example creates a rounded rectangle 50px wide by 60px tall. Each corner is given a distinct
+/// radius in the x and y direction by creating a [`CornerRadii`](../struct.CornerRadiiBuilder.html)
```
rfuest, line 71: The link is wrong.

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -0,0 +1,778 @@
...
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct RoundedRectangle {
+    /// The base rectangle
+    pub rectangle: Rectangle,
+
+    /// The radius of each corner
+    pub corners: CornerRadii,
+}
+
+impl RoundedRectangle {
```
rfuest, line 145: Should we add a `pub fn confine_radii(&self) -> RoundedRectangle` method, that returns a copy of the rectangle with confined radii?

That could be used if a user wants to implement some kind of algorithm that needs to know the "real" radii used by e-g.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -460,6 +469,161 @@ macro_rules! egtriangle {
...
+///     style = primitive_style!(
+///         stroke_color = Rgb565::RED,
+///         stroke_width = 3,
+///         fill_color = Rgb565::GREEN
+///     )
+/// );
+/// ```
+///
+/// Style properties like `stroke_color` map to methods on the [`PrimitiveStyleBuilder`] struct.
+/// For example, the following code makes two identical rounded rectangles:
```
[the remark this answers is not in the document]
jamwaffles: I replicated the structure of the other macros, but that's a good point. Shall I remove it along with the example code below it? I could also do that for the other macro examples if you think it's a good idea. I'll open an issue if so.

embedded-graphics/src/primitives/corner_radii.rs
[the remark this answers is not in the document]
jamwaffles: Do we also want to demonstrate that `CornerRadii` can be constructed using its public fields as well?

embedded-graphics/src/primitives/rounded_rectangle.rs
jamwaffles: Good idea, added!

embedded-graphics/src/primitives/mod.rs
```diff
@@ -460,6 +469,161 @@ macro_rules! egtriangle {
...
+/// use embedded_graphics::{
+///     egroundedrectangle,
+///     pixelcolor::Rgb565,
+///     prelude::*,
+///     primitive_style,
+///     primitives::{CornerRadii, RoundedRectangle},
+///     style::{PrimitiveStyle, Styled},
+/// };
+///
+/// let equal_radius: Styled<RoundedRectangle, PrimitiveStyle<Rgb565>> = egroundedrectangle!(
```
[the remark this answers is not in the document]
jamwaffles: I've added some more text to the description of the example as well as some inline comments. Can't hurt to have both I suppose.

jamwaffles:
All changes addressed, please take a final final look ;)

embedded-graphics/src/primitives/rounded_rectangle.rs
```diff
@@ -162,6 +162,48 @@ impl RoundedRectangle {
         Self::new(rectangle, CornerRadii::new(corner_radius))
     }
 
+    /// Return the rounded rectangle with confined corner radii.
```
rfuest, line 165: I would add a note that calling this method isn't necessary most of the time, because all operations provided by e-g confine the radii internally.

embedded-graphics/src/primitives/mod.rs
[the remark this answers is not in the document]
rfuest: Looks good.

embedded-graphics/src/primitives/mod.rs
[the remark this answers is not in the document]
rfuest: Yes, I think that's a good idea. I would remove this part and replace it by a note that links to the `PrimitiveStyle` doc, which describes how styles can be created.

embedded-graphics/src/primitives/corner_radii.rs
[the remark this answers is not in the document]
rfuest: I would suggest to always use the builder in the examples. Because builders are also used in other instances, where a struct literal isn't possible, like building styles, this pattern is familiar in code that uses e-g.

Being able to create objects by using struct literals is a basic feature of rust that doesn't need to be explained in our examples IMO. And the additional methods in the builder make it more powerful than using plain struct literals.

jamwaffles:
> The radii don't seem to be correctly confined in some cases when drawing a rectangle. 

Managed to reproduce it. I didn't have much luck with a solution though (bit braindead today)... any ideas?

rfuest:
It's caused by a problem with the `confine` algorithm. The overlap is divided by 2 and applied to two adjacent corners using `saturating_sub`. But if one of the corners has a size `< largest_overlap / 2` the corner size cannot be reduced enough to remove the overlap, because it's limited at `0` and the effective reduction in overlap is `< largest_overlap`.

jamwaffles:
Sorry, yeah I figured that out pretty quick. Just struggling with a better algorithm at the moment. I did give your integer version of the CSS spec a go but couldn't get it to behave.

rfuest:
Ugly quick implementation, but I think it works:
```rust
    pub(crate) fn confine(self, bounding_box: Size) -> Self {
        let mut overlap = 0;
        let mut size = 0;
        let mut corner_size = 0;

        let o = (self.top_left.width + self.top_right.width).saturating_sub(bounding_box.width);
        if o > overlap {
            size = bounding_box.width;
            corner_size = self.top_left.width + self.top_right.width;
            overlap = o;
        }

        let o =
            (self.bottom_left.width + self.bottom_right.width).saturating_sub(bounding_box.width);
        if o > overlap {
            size = bounding_box.width;
            corner_size = self.bottom_left.width + self.bottom_right.width;
            overlap = o;
        }

        let o = (self.top_left.width + self.bottom_left.width).saturating_sub(bounding_box.height);
        if o > overlap {
            size = bounding_box.height;
            corner_size = self.top_left.height + self.bottom_left.height;
            overlap = o;
        }

        let o =
            (self.top_right.width + self.bottom_right.width).saturating_sub(bounding_box.height);
        if o > overlap {
            size = bounding_box.height;
            corner_size = self.top_right.height + self.bottom_right.height;
            overlap = o;
        }

        if overlap > 0 && corner_size > 0 {
            Self {
                top_left: (self.top_left * size) / corner_size,
                top_right: (self.top_right * size) / corner_size,
                bottom_right: (self.bottom_right * size) / corner_size,
                bottom_left: (self.bottom_left * size) / corner_size,
            }
        } else {
            self
        }
    }
```

jamwaffles:
Thanks for the kick start, that helped me get a solution over the line. I tried cleaning it up with iterators but that made it significantly slower, so I've left it with an `if` chain for now. We can optimise things much better in #343 than fiddling with things here I think.

rfuest:
Did you read my comment about always using `CornerRadiiBuilder`?

embedded-graphics/src/primitives/corner_radii.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Create a new corner radii builder.
+    ///
+    /// All radii are defaulted to 0px x 0px.
+    pub const fn new() -> Self {
+        Self {
+            corners: CornerRadii::new(Size::zero()),
+        }
+    }
+
+    /// Set all corner radii to the same value.
```
rfuest, line 126: Nit: The examples should all have a `# Examples` title.

jamwaffles:
> Did you read my comment about always using CornerRadiiBuilder?

I did but forgot to change it, thanks for reminding me. There are some occurrences of `RoundedRectangle::with_equal_corners` which I left in, as well as using the struct literal in the `egroundedrectangle` macro. I think it makes sense to leave those as is. Let me know if I've missed any examples you think should use the builder instead.

I've addressed the `# Examples` nit also.

rfuest:
> I did but forgot to change it, thanks for reminding me. There are some occurrences of `RoundedRectangle::with_equal_corners` which I left in,...

I might not have explained this good enough, because using `RoundedRectangle::with_equal_corners` and also `CornerRadii::new` is fine IMO, so that one of the changes wouldn't have been necessary. (But it doesn't hurt either)

> ..as well as using the struct literal in the `egroundedrectangle` macro. I think it makes sense to leave those as is.

IDK, looks weird either way, maybe the proper way would be to add a `corner_radii` macro to make it similar to the other builders. But I don't think that's necessary at the moment and you can leave it anyway you prefer.

jamwaffles:
>  the proper way would be to add a corner_radii macro

It would, but let's do that in a new PR. This one is huge!

jamwaffles:
Thanks for spending the time on this!

PR #337, opened 2020-05-06
Add methods to get points with axes

jamwaffles:

## PR description

Adds `x_axis` and `y_axis` methods to get a `Point` or `Size` with one dimension set to zero. Some places can already use these, but they would be particularly useful in the `RoundedRectangle` PR. Also adds `Point::fill` and `Size::fill` which will be less useful but there are a couple of places they can be used. If these are really useless I don't mind removing them.

### DIFF
--- a/embedded-graphics/src/geometry/point.rs
+++ b/embedded-graphics/src/geometry/point.rs
@@ -78,6 +78,19 @@ impl Point {
         Point { x, y }
     }
 
+    /// Creates a point with X and Y values set to an equal value.
+    ///
+    /// ```rust
+    /// use embedded_graphics::geometry::Point;
+    ///
+    /// let point = Point::new_equal(11);
+    ///
+    /// assert_eq!(point, Point { x: 11, y: 11 });
+    /// ```
+    pub const fn new_equal(value: i32) -> Self {
+        Point { x: value, y: value }
+    }
+
     /// Creates a point with X and Y equal to zero.
     pub const fn zero() -> Self {
         Point { x: 0, y: 0 }
--- a/embedded-graphics/src/geometry/point.rs
+++ b/embedded-graphics/src/geometry/point.rs
@@ -83,6 +83,48 @@ impl Point {
         Point { x: 0, y: 0 }
     }
 
+    /// Returns a point with equal `x` value and `y` set to `0`.
+    ///
+    /// # Examples
+    ///
+    /// ## Move a `Point` along the X axis.
+    ///
+    /// ```rust
+    /// use embedded_graphics::geometry::Point;
+    ///
+    /// let translate = Point::new(20, 30);
+    ///
+    /// let point = Point::new(10, 15);
+    ///
+    /// let moved_x = point + translate.x_axis();
+    ///
+    /// assert_eq!(moved_x, Point::new(30, 15));
+    /// ```
+    pub const fn x_axis(self) -> Self {
+        Self { x: self.x, y: 0 }
+    }
+
+    /// Returns a point with equal `y` value and `x` set to `0`.
+    ///
+    /// # Examples
+    ///
+    /// ## Move a `Point` along the Y axis.
+    ///
+    /// ```rust
+    /// use embedded_graphics::geometry::Point;
+    ///
+    /// let translate = Point::new(20, 30);
+    ///
+    /// let point = Point::new(10, 15);
+    ///
+    /// let moved_y = point + translate.y_axis();
+    ///
+    /// assert_eq!(moved_y, Point::new(10, 45));
+    /// ```
+    pub const fn y_axis(self) -> Self {
+        Self { x: 0, y: self.y }
+    }
+
     /// Remove the sign from a coordinate
     ///
     /// ```
--- a/embedded-graphics/src/geometry/size.rs
+++ b/embedded-graphics/src/geometry/size.rs
@@ -72,6 +72,22 @@ impl Size {
         Size { width, height }
     }
 
+    /// Creates a size with width and height set to an equal value.
+    ///
+    /// ```rust
+    /// use embedded_graphics::geometry::Size;
+    ///
+    /// let size = Size::new_equal(11);
+    ///
+    /// assert_eq!(size, Size { width: 11, height: 11 });
+    /// ```
+    pub const fn new_equal(value: u32) -> Self {
+        Size {
+            width: value,
+            height: value,
+        }
+    }
+
     /// Creates a size with width and height equal to zero.
     pub const fn zero() -> Self {
         Size {
--- a/embedded-graphics/src/geometry/size.rs
+++ b/embedded-graphics/src/geometry/size.rs
@@ -80,6 +80,54 @@ impl Size {
         }
     }
 
+    /// Returns a size with equal `width` value and `height` set to `0`.
+    ///
+    /// # Examples
+    ///
+    /// ## Move a `Point` along the X axis.
+    ///
+    /// ```rust
+    /// use embedded_graphics::geometry::{Point, Size};
+    ///
+    /// let size = Size::new(20, 30);
+    ///
+    /// let point = Point::new(10, 15);
+    ///
+    /// let moved_x = point + size.x_axis();
+    ///
+    /// assert_eq!(moved_x, Point::new(30, 15));
+    /// ```
+    pub const fn x_axis(self) -> Self {
+        Self {
+            width: self.width,
+            height: 0,
+        }
+    }
+
+    /// Returns a size with equal `height` value and `width` set to `0`.
+    ///
+    /// # Examples
+    ///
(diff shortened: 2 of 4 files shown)


rfuest:
> Adds `x_axis` and `y_axis` methods to get a `Point` or `Size` with one dimension set to zero. Some places can already use these, but they would be particularly useful in the `RoundedRectangle` PR.

Good idea!

> Also adds `Point::fill` and `Size::fill` which will be less useful but there are a couple of places they can be used. If these are really useless I don't mind removing them.

I did temporarily add similar methods before, but removed them because they weren't totally necessary for that particular PR and I couldn't come up with a good name. I think we should keep them, but maybe we can come up with a better name, because I wouldn't associate the name `fill` with a constructor. Normally additional constructors should be prefixed by `with_`, but that doesn't work well in this case IMO. Maybe `new_square`?

jamwaffles:
The name was inspired by [nalgebra's `Matrix::fill`](https://docs.rs/nalgebra/0.21.0/nalgebra/base/struct.Matrix.html#method.fill), so might be familiar to some users. An alternative could be `new_equal` or `new_filled` perhaps? I think this would then need a shim adding for the `features = [ "nalgebra" ]` which is easy enough.

I deliberately didn't do a find/replace with the `::fill()` methods as it makes most of the examples and tests where it's used a bit less clear IMO. There are a couple of usages in internal code, but we should probably change those manually going forward.

rfuest:
> The name was inspired by [nalgebra's `Matrix::fill`](https://docs.rs/nalgebra/0.21.0/nalgebra/base/struct.Matrix.html#method.fill), so might be familiar to some users. An alternative could be `new_equal` or `new_filled` perhaps?

Am I right that `Matrix::fill` isn't a constructor? If that's the case I would go with `new_filled`, which still remains similar to the `nalgebra` method.

> I think this would then need a shim adding for the `features = [ "nalgebra" ]` which is easy enough.

Why would this be necessary?

> I deliberately didn't do a find/replace with the `::fill()` methods as it makes most of the examples and tests where it's used a bit less clear IMO. There are a couple of usages in internal code, but we should probably change those manually going forward.

SGTM

jamwaffles:
I've changed `fill` changed to `new_filled`. It's been a while since I used nalgebra - looks like I got some things confused. Ready to review now :)

embedded-graphics/src/geometry/point.rs
```diff
@@ -78,11 +78,66 @@ impl Point {
...
+    /// ## Move a `Point` along the X axis.
+    ///
+    /// ```rust
+    /// use embedded_graphics::geometry::{Point, Size};
+    ///
+    /// let size = Size::new(20, 30);
+    ///
+    /// let point = Point::new(10, 15);
+    ///
+    /// let moved_x = point + size.x_axis();
```
rfuest, line 112: This isn't really an example of `Point::x_axis`, but `Size::x_axis`. Because there are no additional tests for these functions the example should use `Point::x_axis`.

embedded-graphics/src/geometry/point.rs
```diff
@@ -78,11 +78,66 @@ impl Point {
...
+    /// ## Move a `Point` along the Y axis.
+    ///
+    /// ```rust
+    /// use embedded_graphics::geometry::{Point, Size};
+    ///
+    /// let size = Size::new(20, 30);
+    ///
+    /// let point = Point::new(10, 15);
+    ///
+    /// let moved_y = point + size.y_axis();
```
rfuest, line 133: See comment for example above.

embedded-graphics/src/geometry/point.rs
```diff
@@ -78,11 +78,66 @@ impl Point {
...
+    pub const fn new_filled(fill: i32) -> Self {
+        Point { x: fill, y: fill }
+    }
+
     /// Creates a point with X and Y equal to zero.
     pub const fn zero() -> Self {
         Point { x: 0, y: 0 }
     }
 
+    /// Return the `Point` with `y` set to `0`.
```
rfuest, line 99: IMO the comments for the `x_axis` and `y_axis` methods should also include what happens to the other value.
```suggestion
    /// Returns a point with equal `x` value and `y` set to `0`.
```

embedded-graphics/src/geometry/point.rs
```diff
@@ -78,11 +78,66 @@ impl Point {
...
+    ///
+    /// let moved_x = point + size.x_axis();
+    ///
+    /// assert_eq!(moved_x, Point::new(30, 15));
+    /// ```
+    pub const fn x_axis(self) -> Self {
+        Self { x: self.x, y: 0 }
+    }
+
+    /// Return the `Point` with `x` set to `0`.
```
rfuest, line 120:
```suggestion
    /// Returns a point with equal `y` value and `x` set to `0`.
```

embedded-graphics/src/geometry/size.rs
```diff
@@ -80,6 +96,54 @@ impl Size {
         }
     }
 
+    /// Return the `Size` with `height` set to `0`.
```
rfuest, line 99:
```suggestion
    /// Returns a size with equal `width` value and `height` set to `0`.
```

embedded-graphics/src/geometry/size.rs
```diff
@@ -80,6 +96,54 @@ impl Size {
...
+    /// assert_eq!(moved_x, Point::new(30, 15));
+    /// ```
+    pub const fn x_axis(self) -> Self {
+        Self {
+            width: self.width,
+            height: 0,
+        }
+    }
+
+    /// Return the `Size` with `width` set to `0`.
```
rfuest, line 123:
```suggestion
    /// Returns a size with equal `height` value and `width` set to `0`.
```

jamwaffles:
Thanks for the review. Comments addressed

rfuest:
I did give the name of the constructor another thought and I think we should change it.

IMO the word `filled` could easily be mistaken for something related to styling, if someone isn't familiar with e-g:
```rust
let rect = Rectangle::new(Point::new(10, 20), Size::new_filled(10));
// does this create a "filled rectangle"?
```
For this case the `new_square` would work:
```rust
let rect = Rectangle::new(Point::new(10, 20), Size::new_square(10));
```
but `Point::new_square` might not be the greatest name.

Where can I put my :bike:? :wink:

jamwaffles:
Yes that's true, in context `new_filled` isn't very clear. That said, `new_square` doesn't look quite right when used with `Ellipse` as it makes a circle:

```rust
Ellipse::new(Point::new(10, 20), Size::new_square(10));
```

The best I can come up with is `Size::new_equal(u32)` or maybe even just `Size::equal(u32)`. If you're not keen on anything `equal` based, I'll change this PR to `new_square`, let the docs explain what it does for user who might be confused and leave it there.

Can I share your :lock: for my :bike:? :grin:

rfuest:
> The best I can come up with is `Size::new_equal(u32)`

OK, let's just go with that.

jamwaffles:
Changes pushed :+1:

embedded-graphics/src/geometry/point.rs
```diff
@@ -78,11 +78,66 @@ impl Point {
...
+    /// use embedded_graphics::geometry::Point;
+    ///
+    /// let filled = Point::new_equal(11);
+    ///
+    /// assert_eq!(filled, Point { x: 11, y: 11 });
+    /// ```
+    pub const fn new_equal(fill: i32) -> Self {
+        Point { x: fill, y: fill }
+    }
+
```
rfuest, line 93:
```suggestion
    /// Creates a point with X and Y values set to an equal value.
    ///
    /// ```rust
    /// use embedded_graphics::geometry::Point;
    ///
    /// let point = Point::new_equal(11);
    ///
    /// assert_eq!(point, Point { x: 11, y: 11 });
    /// ```
    pub const fn new_equal(value: i32) -> Self {
        Point { x: value, y: value }
    }

```

embedded-graphics/src/geometry/size.rs
```diff
@@ -72,6 +72,22 @@ impl Size {
...
+    ///
+    /// assert_eq!(filled, Size { width: 11, height: 11 });
+    /// ```
+    pub const fn new_equal(fill: u32) -> Self {
+        Size {
+            width: fill,
+            height: fill,
+        }
+    }
+
```
rfuest, line 90:
```suggestion
    /// Creates a size with width and height set to an equal value.
    ///
    /// ```rust
    /// use embedded_graphics::geometry::Size;
    ///
    /// let size = Size::new_equal(11);
    ///
    /// assert_eq!(size, Size { width: 11, height: 11 });
    /// ```
    pub const fn new_equal(value: u32) -> Self {
        Size {
            width: value,
            height: value,
        }
    }

```

jamwaffles:
Those examples do look better, thanks

jamwaffles:
Thanks!

PR #340, opened 2020-05-12
Build speedup

jamwaffles:

## PR description

Optimises the build a bit. The old home dir cache path is incorrect. The new absolute path works in another project I use at work so I'm copying that change here. Builds should speed up by a bit now.

The cache key removal is explained in the commit itself :)


jamwaffles:
Cached builds currently run between 7 and 11 minutes. This branch reduces that to ~5m 30s. Still quite long but a decent improvement nonetheless.

PR #342, opened 2020-05-14
Draw target improvements

### LINKED ISSUE #330: DrawTarget improvements
As previously discussed in #326, the following changes to the `DrawTarget` trait were suggested:
- Replace `draw_image` by `draw_rectangle_iter(&mut self, rect: &Rectangle, iter: Iterator<Item = C>)`
  `draw_rectangle_iter` can be used to fill a rectangular region of the target with the colors provided by the `iter`. This can e.g. be used for images and font rendering.
-  Add a `fill_rectangle(&mut self, rect: &Rectangle, color: C)` method
  This fills the given rectangle with a single color, which might be more efficient that using `draw_rectangle_iter` if the display supports filling of memory regions.
- Change `draw_rectangle` to use `fill_rectangle` or `draw_rectangle_iter`
  For opaque styled rectangles without a stroke the default impl of `draw_rectangle` can call directly `fill_rectangle` with `styled.primitive` as the rectangle.
  Styled rectangles without a fill can be drawn using 4 individual `fill_rectangle` calls.
  Drawing rectangles with fill and stroke can be implemented using 2 overlapping `fill_rectangle` calls, 5 non overlapping `fill_rectangle` calls or a single `draw_rectangle_iter` call.
- Add docs for driver authors that suggest to implement `fill_rectangle` and `draw_rectangle_iter` instead of the other styled primitve `draw_...` methods.

Open questions:
- [ ] `draw_rectangle_iter` should be able to handle iterators that are shorter than `rect.size.width * rect.size.height` without leaving the display in an invalid state.
   Should we provide any method to make this easier for driver implementations or simply state this in the docs?
- [ ] `draw_rectangle_iter` needs a better name, that can't be easily confused with `draw_rectangle`
- [ ] Decide if `draw_rectangle` for styled rectangles with fill and stroke should be implemented using `fill_rectangle` or `draw_rectangle_iter`
  Using 2 overlapping `fill_rectangle` calls might not be a good idea if we plan to support transparency in the future.
- [ ] Would something like `draw_bounded_iter(&mut self, bound: Rectangle, iter: Iterator<Item=Pixel<C>>)` be useful?
  This method would be used for iterators that don't change every pixel in the given rectangle, but are guaranteed to only change pixels inside the given bounding box. This could e.g. be used for lines or transparent text.
  I'm not sure if this would provide any benefits for drivers, but maybe it could be used to implement some kind of caching.
- [ ] DrawTarget::size -> #263

jamwaffles:

## PR description

Closes #330


embedded-graphics/src/draw_target.rs
```diff
@@ -226,161 +221,66 @@ where
...
-    fn draw_line(
-        &mut self,
-        item: &Styled<primitives::Line, PrimitiveStyle<C>>,
-    ) -> Result<(), Self::Error> {
-        self.draw_iter(item)
+        self.draw_iter(
+            area.points()
+                .zip(colors)
+                .filter_map(|(pos, color)| color.map(|c| Pixel(pos, c))),
+        )
```
jamwaffles, line 241: @rfuest 

This is obviously the incorrect behaviour, but how should we fix this for sparse primitives like `Circle`, etc? Should we add a wrapper iterator over styled primitive iterators to produce a bounding box iterator yielding `Option<C>`?

rfuest: > This is obviously the incorrect behaviour,

I think I'm a little slow today, how exactly is this the incorrect behavior?

I haven't really thought about how we should handle iterators in the future. If a drawable can be implemented by only calling functions in `DrawTarget`, like it's possible for rectangles, the iterator impl becomes kind of redundant. To keep the iterator support for primitives, which we should, we will have to maintain both variants in the future and make sure they are resulting in the same output.

> Should we add a wrapper iterator over styled primitive iterators to produce a bounding box iterator yielding `Option<C>`?

I think the other way around makes more sense in most cases. When a iterator, like for a styled circle, already iterates over the bounding box internally it would be easy to return a `Option<C>` instead of skipping these pixels.

jamwaffles: > I think I'm a little slow today, how exactly is this the incorrect behavior?

`area.points()` returns a contiguous iterator of x * y pixels, whereas `colors` will return fewer items as it's a sparse iterator.

> it would be easy to return a Option<C> instead of skipping these pixels.

It would indeed and I did consider this approach. Are we ok doing breaking changes to a primitive's styled iterator? They're pretty internal (albeit `pub`) so I don't think it'll break much/any user code.

rfuest: > `area.points()` returns a contiguous iterator of x * y pixels, whereas `colors` will return fewer items as it's a sparse iterator.

But `fill_sparse` takes an `Iterator<Item = Option<C>>` which is required to return x * y pixels. If `colors` would return fewer pixels there would be no way of knowing where to put the pixels, because the item only contains the color and not the position.

> It would indeed and I did consider this approach. Are we ok doing breaking changes to a primitive's styled iterator?

Yes, I expected that we would need to break the iterators during this refactoring.

jamwaffles: > But fill_sparse takes an Iterator<Item = Option<C>> which is required to return x * y pixels.

Sorry, I can see why my comment would be confusing now: when I made it, the current iterators don't do this, hence me mentioning the incorrect behaviour. At any rate, I'll change them to produce `Option<C>` :)

> Yes, I expected that we would need to break the iterators during this refactoring.

Sweet, I'll get that done asap :)

jamwaffles:
This code:

```rust
Rectangle::new(Point::new(0, 0), Size::new(1, 1))
    .into_styled(PrimitiveStyle::default())
    .into_iter()
    .chain(
        Circle::new(Point::new(1, 1), 3)
            .into_styled(PrimitiveStyle::default())
            .into_iter(),
    )
```

fails to compile. 

`StyledRectangleIterator` produces `Pixel`s, whereas I've just changed `StyledCircleIterator` to produce `Option<C>`. 

I would very much like to keep chaining as a feature. Should we change all `Styled*Iterators` to produce `Option<C>`?

rfuest:
> I would very much like to keep chaining as a feature. Should we change all `Styled*Iterators` to produce `Option<C>`?

I agree that we should keep this feature in some form, but we also need to consider that it makes using accelerated drawing routines impossible.

Can we remove the `fill_sparse` method in this PR? This way we could merge the first part of the refactor and discuss possible solutions without resulting in a huge PR.

jamwaffles:
> Can we remove the fill_sparse method in this PR?

Sure thing, I'll leave it out for now.

---

To support `.into()`, I've ended up adding some extra bounds to the `DrawTarget` methods. For example:

```rust
struct TestDisplay;

impl DrawTarget for TestDisplay {
    type Color = Rgb888;

    fn size(&self) -> Size {
        Size::zero()
    }

    fn draw_iter<I, C>(&mut self, pixels: I) -> Result<(), Self::Error>
    where
        C: PixelColor + Into<Self::Color>,
        I: IntoIterator<Item = Pixel<C>>,
    {
        for Pixel(pos, col) in pixels.into_iter() {
            let color: Rgb888 = col.into();

            //
        }

        Ok(())
    }
}
```

I'm pretty alright changing the trait definition to add more bounds, but is there perhaps a cleaner way? I've tried changing the `type Color` bounds to `Into<...>` but that spews a whole pile of errors.

rfuest:
> I'm pretty alright changing the trait definition to add more bounds, but is there perhaps a cleaner way? I've tried changing the `type Color` bounds to `Into<...>` but that spews a whole pile of errors.

We discussed this before (I think in #330). I would like to not implement automatic color conversion this way and instead use a wrapper around a `DrawTarget` to implement the software color conversion. Something like:
```rust
struct ColorConversionDrawTarget<'a, T, FROM_C, TO_C>
where
    T: DrawTarget<TO_C>,
    FROM_C: PixelColor + Into<TO_C>,
{
   draw_target: &'a mut T,
   color: PhantomData<TO_C>,
}

impl <T, FROM_C, TO_C> DrawTarget for ColorConversionDrawTarget<'_, T, FROM_C, TO_C> {
   type Color = TO_C;
   type Error = T::Error;

   ... (draw methods that use `into`) ...
}
```
To make this easier to use an extension trait for `DrawTarget` can provide a method that wraps the draw target:
```
image_rgb565.draw(display_rgb888.convert_colors())?;
```

jamwaffles:
This PR is already pretty large so I'm going to tentatively put it up for initial review. I think the docs and tests could use some improvement, but not sure what at this point. Guidance gladly accepted.

In terms of scope, I think it's a good idea to keep this PR to `DrawTarget` changes plus a changed impl of `Rectangle`'s `draw` method using `fill_solid` to exercise the new code. 

Also happy to continue bikeshedding the new `DrawTarget` method names.

jamwaffles:
I've rebased this after some conflicts from #353. @rfuest if you've got time for an initial review that would be most appreciated :)

rfuest:
> @rfuest if you've got time for an initial review that would be most appreciated :)

It's on my TODO list, but I wanted to finish #353 before doing anything else. I'll try to review it tomorrow.

jamwaffles:
Makes sense :) thanks for taking the time.

embedded-graphics/src/draw_target.rs
```diff
@@ -99,118 +95,149 @@ use crate::{
...
+///             bottom_right
+///         } else {
+///             return Ok(())
+///         };
+///
+///         self.send_commands(&[
+///             // Draw rectangle command
+///             0x22,
+///             // Top left X coordinate
+///             area.top_left.x as u8,
```
rfuest, line 185: By using a simple cast here this won't have the correct behavior for coordinates outside the display area.

I just noticed that the drivers will also need to handle rectangles that are partially outside the display area, which makes this more complex then expected. Maybe we should add a clipping function for rectangles to make this easier to implement correctly:
```rust
 fn fill_solid(&mut self, area: &Rectangle, color: Self::Color) -> Result<(), Self::Error> {
    let display_area = Rectangle::new(Point::zero(), display.size());
    let actual_area = area.clip(&display_area);  //clip calculates the intersection of area and display_area
    ...
}
```

embedded-graphics/src/draw_target.rs
```diff
@@ -99,118 +95,149 @@ use crate::{
...
+/// let mut display = ExampleDisplay {
+///     framebuffer: [0; 4096],
+///     iface: SPI1,
+/// };
 ///
-/// // Draw a rectangle from (10, 20) to (30, 40) with a white stroke
-/// Rectangle::with_corners(Point::new(10, 20), Point::new(30, 40))
-///     .into_styled(PrimitiveStyle::with_stroke(Gray8::WHITE, 1))
+/// // Draw a rectangle with 5px green stroke and red fill. The stroke and fill can be broken down
+/// // into multiple individual rectangles, so uses `fill_solid` internally.
```
rfuest, line 208: The fill is green and the stroke is red.

embedded-graphics/src/draw_target.rs
```diff
@@ -99,118 +95,149 @@ use crate::{
...
+///             .stroke_width(5)
+///             .fill_color(Rgb565::GREEN)
+///             .build(),
+///     )
 ///     .draw(&mut display)?;
 ///
-/// // Draw a rectangle on the display using accelerated `draw_rectangle()` function
-/// # Ok::<(), CommError>(())
-/// ```
+/// // Draw a circle with top-left at `(5, 5)` with a diameter of `10` and a megnta stroke with cyan
```
rfuest, line 219:
```suggestion
/// // Draw a circle with top-left at `(5, 5)` with a diameter of `10` and a magenta stroke with cyan
```

embedded-graphics/src/draw_target.rs
```diff
@@ -222,161 +249,78 @@ where
...
-    where
-        Self: Sized,
-    {
-        primitives::Rectangle::new(Point::zero(), self.size())
-            .into_styled(PrimitiveStyle::with_fill(color))
-            .draw(self)
-    }
+    /// This should return the size of the visible area of the display, starting at `(0, 0)`. If a
+    /// display supports drawing to pixels outside this area, that area should not be reported in
+    /// calls to `size`.
```
rfuest, line 256: I think this needs some clarification.

Does this mean that it should also only return the visible area if I use it in conjunction with a larger framebuffer in the controller memory to support scrolling?

jamwaffles: Exactly that, yes. What would you add  to the comment to explain this better?

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -310,8 +310,52 @@ impl<C> Drawable<C> for &Styled<Rectangle, PrimitiveStyle<C>>
...
+        let fill_area = self.primitive.shrink(self.style.inside_stroke_width());
+
+        // Fill rectangle
+        if let Some(fill_color) = self.style.fill_color {
+            display.fill_solid(&fill_area, fill_color)?;
+        }
+
+        // Draw stroke
+        if let Some(stroke_color) = self.style.effective_stroke_color() {
+            let stroke_width = self.style.stroke_width;
```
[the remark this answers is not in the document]
jamwaffles: Hm that's a good point. `stroke_area.size` might also become an issue. There are a couple of places where `saturating_add` would give some safety, but the `as i32` casts and some of the other maths will still cause problems. I'm not really sure what the behaviour should be here. 

We could change the drawtargets to return an error enum:

```rust
enum DisplayError<E> {
    /// Display- or hardware-specific error
    Error(E),

    /// Arithmetic overflow/underflow
    Overflow,

    /// Internal conversion between types failed
    Conversion(TryFromIntError)
}

trait DrawTarget {
    type Error;

    // ...

    fn draw_the_thing(&mut self, thing: Rectangle) -> Result<(), DisplayError<Self::Error>> {
        // ...

        Ok(())
    }
}
```

This would allow usage of `try_into` and/or the `checked_*` methods. Open to better ideas!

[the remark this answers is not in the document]
rfuest: I wasn't fully awake anymore while writing the comments yesterday, so I didn't explain this very well. Handling of integer overflows in general could need some more considerations, but that's not what I meant in this case.

If `stroke_with` is larger than the size of rectangle the border will be drawn in the wrong location, because the unlimited `stroke_with` is used to calculate the size and position of the border rectangles.

Introducing an error type to catch integer errors seems to be cumbersome to use and I don't think there is much value to the user to catch these errors, because there isn't really a good way to handle them. I would prefer that we try to minimize the effects these overflows can have, e.g. by using `saturating_...`, but I don't think it's possible to prevent them entirely. I wouldn't mind if these out of range values would cause graphical corruption, but they shouldn't cause panics or endless loops.

embedded-graphics/src/draw_target.rs
rfuest: I'm not sure. There might be situations where a target is larger than the actual display and where it would make sense for `size` to return a size larger than the display. If, for example, a larger framebuffer (independend of the one used by the display/driver) is used to support scrolling an UI layout algorithm might need to know the available space in that buffer.

jamwaffles: The main purpose I have in mind for the return value of this method is for doing layout in the visible area of the screen, but we do need to be able to get the renderable area for proper clipping and scrolling as you mention. How about having two methods, `visible_size` and `render_size` for visible and total pixel area respectively where `render_size`' default impl simply calls visible size?

rfuest: I'm not sure this solves the problem, if `visible_size` is `< render_size` the layout algorithm still wouldn't know where the top left coordinate of the visible window is.

jamwaffles: If the visible area top left is `(0, 0)`, the off-screen area could be a `Rectangle` with a negative `top_left` value. Would that work?. This way, it keeps the drawable area positioned within the render area at a known offset whilst keeping the origin at `(0, 0)` which is what I think most end users will want. 

Also, AFAIK it's more common to have a display with `visible_size == render_size` rather than `visible_size < render_size`, so to me it makes sense to use `(0, 0)` as the _visible_ origin.

rfuest: > If the visible area top left is `(0, 0)`, the off-screen area could be a `Rectangle` with a negative `top_left` value. Would that work?. This way, it keeps the drawable area positioned within the render area at a known offset whilst keeping the origin at `(0, 0)` which is what I think most end users will want.

Yes, this is what most people want A better way to implement this might be to add a `TranslatedDrawTarget` wrapper (which I suggested some time ago, but I have no idea where). The display driver would implement a basic draw target which covers the entire framebuffer and reports it's size in the `size` function. This draw target can then be wrapped, which causes all draw operations to work as expected by using coordinates starting from `(0, 0)`.

jamwaffles: That sounds like a good idea. For this PR, I think the best approach is to change the `size` docs to mention it should return the entire drawable area and add `TranslatedDrawTarget` in a later PR. Shall I open a ticket to add it once this PR is merged?

rfuest: Yes, let's try this approach and see if it works out in practice.

jamwaffles: Ok, sounds good. #355 opened to track. I'll change the docs in this PR momentarily.

embedded-graphics/src/draw_target.rs
```diff
@@ -307,16 +307,77 @@ pub trait DrawTarget {
...
-        )
+        if let Some(area) = Rectangle::new(Point::zero(), self.size()).intersection(&area) {
+            self.draw_iter(
+                area.points()
+                    .zip(colors)
+                    .map(|(pos, color)| Pixel(pos, color)),
+            )
+        } else {
+            Ok(())
+        }
```
rfuest, line 380: IMO the default implementation shouldn't do any clipping and this should be handled by the target impl. If a display driver uses a larger framebuffer than the display size the overlap could be visible later if the entire screen is scrolled in that direction.

jamwaffles: Ok, that's a fair point. I've removed the code from the default impl.

embedded-graphics/src/draw_target.rs
jamwaffles: > By using a simple cast here this won't have the correct behavior for coordinates outside the display area.

I'm struggling to come up with a clean-ish solution to show in this demo code. We could go for `var.max(0).min(63) as u8`, or something similar to the current `if let Ok((x @ 0..=63, y @ 0..=63)) = coord.try_into() { ... }`, but both feel quite messy.

rfuest: If you intersect the `area` rectangle with a rectangle that spans the visible display area the result is always in the valid range and you can just cast the coordinates, if the intersection is not `None`.

jamwaffles: Thanks for the hint. I've pushed a change to that example. Should be safe but still reasonably easy to understand.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -138,6 +138,78 @@ impl Rectangle {
...
+    ///         ". #   #    .",
+    ///         ". #   #    .",
+    ///         "..#####    .",
+    ///         "  .        .",
+    ///         "  ..........",
+    ///     ])
+    /// );
+    /// # Ok::<(), core::convert::Infallible>(())
+    /// ```
+    pub fn intersection(&self, other: &Rectangle) -> Option<Rectangle> {
```
rfuest, line 194: Should we change this to simply return a `Rectangle`? Because a `Rectangle`can already be zero sized adding an `Option` should be unnecessary. The `fill_solid` example demonstrates this problem IMO, because it needs to use two if statements.

jamwaffles: It could indeed, with checking for an intersection changing to `intersection.size == Size::zero()` which is pretty clean. In the case of no intersection at all, what should the `top_left` component of the zero-sized rectangle be? `Point::zero()`?

rfuest: >  In the case of no intersection at all, what should the `top_left` component of the zero-sized rectangle be? `Point::zero()`?

Yes, I think that's the best option.

jamwaffles: Changes pushed :ok_hand:

embedded-graphics/src/draw_target.rs
```diff
@@ -222,161 +268,136 @@ where
...
     ///
-    /// This method should not be called directly from application code. It is used to define the
-    /// internals of the [`draw`] method used for the [`Styled`] [`Circle`] primitive. To draw a
-    /// circle, call [`draw`] on a `Styled<Circle>` object.
+    ///         self.draw_iter(
+    ///             area.points()
+    ///                 .zip(colors)
+    ///                 .map(|(pos, color)| Pixel(pos, color)),
+    ///         )
+    ///     }
```
rfuest, line 366: I just noticed that this won't work. The intersection might, for example, be narrower than `area`, which means that each row of the points iter has fewer pixels than the colors iter which will mess up the display.

jamwaffles: Oh nuts, very true :facepalm:. We could do this:

```rust
// Clamp area to drawable part of the display target
let drawable_area = area.intersection(&Rectangle::new(Point::zero(), self.size()));

self.draw_iter(
    area.points()
        .zip(colors)
        .filter(|(pos, _color)| drawable_area.contains(pos))
        .map(|(pos, color)| Pixel(pos, color)),
)
```

Although it's pretty inefficient for both the demo code and the default impl. Not sure there's another way round it tbh.

rfuest: > Although it's pretty inefficient for both the demo code and the default impl.

The default impl doesn't need to clip, so it's fine the way it is.

For the demo code we could go with the example you provided. It's easy to understand and shouldn't be too bad performance wise. If we want to improve this in the future we could implement some kind of adapter around the iterator that skips the appropriate amount of pixels. But that seems more complicated then necessary at the moment.

jamwaffles: > For the demo code we could go with the example you provided.

Changes pushed :ok_hand:

embedded-graphics/src/primitives/rectangle.rs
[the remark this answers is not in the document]
jamwaffles: I've refactored the impl and added some tests for these overdraw cases. It should now work properly.

embedded-graphics/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - **(breaking)** [#312](https://github.com/jamwaffles/embedded-graphics/pull/312) The methods in the `Dimension` trait are replaced by a single `bounding_box` method that returns a `Rectangle`.
 - **(breaking)** [#353](https://github.com/jamwaffles/embedded-graphics/pull/353) Lines with equal start and end points are now drawn by assuming they are oriented horizontally. This means that they are now drawn as a `stroke_width` high and 1px wide rectangle, instead of not being drawn at all.
 - **(breaking)** [#353](https://github.com/jamwaffles/embedded-graphics/pull/353) `primitives::line::StyledLineIterator` was renamed to `primitives::line::StyledIterator`.
+- **(breaking)** [#342](https://github.com/jamwaffles/embedded-graphics/pull/342) Refactored the `DrawTarget` trait to better support common hardware capabilities.
```
jamwaffles, line 33: This will eventually need a migration guide item describing how to migrate in detail.

jamwaffles:
Ready for another look :ok_hand:

embedded-graphics/src/draw_target.rs
```diff
@@ -222,161 +273,139 @@ where
...
     }
 
-    /// Draws an image with known size
+    /// Fill the entire display with a solid color.
     ///
-    /// This default trait method can be overridden if a display provides hardware-accelerated
-    /// methods for drawing an image with known size.
+    /// The default implementation of this method delegates to [`fill_solid`] where the fill area
+    /// is specified as `(0, 0)` to `(width, height)`. If the target hardware supports a more
+    /// optimised way of filling the entire display with a solid color, this method should be
```
rfuest, line 404:
```suggestion
    /// optimized way of filling the entire display with a solid color, this method should be
```

embedded-graphics/CHANGELOG.md
jamwaffles: Added :+1:

embedded-graphics/src/draw_target.rs
```diff
@@ -222,161 +273,139 @@ where
...
-        }
-
-        Ok(())
-    }
-
     /// Returns the dimensions of the `DrawTarget` in pixels.
+    ///
+    /// This should return the size of the entire drawable area of the display. If a display
+    /// supports drawing to pixels outside the visible area, that area should also be reported in
+    /// the dimensions returned by `size`.
```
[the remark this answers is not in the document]
jamwaffles: Yeah, I think that's a good idea. It should be easier to document clearly by linking to `SubDrawTarget` and its associated example code. I'll add a link to this comment to #355 as a reminder.

jamwaffles:
I ran into some frustrating problems whilst adding some links to the `DrawTarget` docs. Because `DrawTarget` was exported both at the top level and in the prelude, the relative doc links like `../primitives/rectangle/struct.Rectangle.html#method.intersection` works in some places but not others. Changing this to `./primitives/rectangle/struct.Rectangle.html#method.intersection` (`./` vs `../`) swaps that around but still not all links work.

I also experimented with intra doc links using `cargo +nightly doc` which did work but I'm not sure I want to add a requirement for nightly right now so I left that alone.

After a bit of yak shaving, I've removed the top-level export of `DrawTarget`. This fixes the doc links and also feels a bit cleaner generally as everything is now exported out of sub-modules. What are your thoughts on this?

rfuest:
> I also experimented with intra doc links using `cargo +nightly doc` which did work but I'm not sure I want to add a requirement for nightly right now so I left that alone.

I hope they will be stabilized soon, this would make writing docs with links a lot easier. But until then...

> After a bit of yak shaving, I've removed the top-level export of `DrawTarget`. This fixes the doc links and also feels a bit cleaner generally as everything is now exported out of sub-modules. What are your thoughts on this?

Do you understand why some structs and traits are inlined into the prelude and some are listed as re-exports? IMO they should all be re-exports, that way they don't cause additional problems with links.

I don't like having modules that only contain a single type and IMO the fundamental types `DrawTarget` and `Drawable` could both be contained in the root, without a module.

jamwaffles:
> Do you understand why some structs and traits are inlined into the prelude and some are listed as re-exports?

I haven't looked into it too deeply but I have absolutely no idea :(

> IMO they should all be re-exports, that way they don't cause additional problems with links.

That's what I wanted as well, but something shifty is going on which makes this not happen. I'll take a look later and see if I can figure out what's going on.

> I don't like having modules that only contain a single type and IMO the fundamental types DrawTarget and Drawable could both be contained in the root, without a module.

Do you mean move them into `lib.rs`? That would make it over 1000 lines long, mostly comments. `draw_target.rs` is already pretty difficult to scroll around so throwing it all in `lib.rs` would make matters worse, no?

rfuest:
> Do you mean move them into `lib.rs`? That would make it over 1000 lines long, mostly comments. `draw_target.rs` is already pretty difficult to scroll around so throwing it all in `lib.rs` would make matters worse, no?

No, not I didn't mean moving the code physically. The types in the `draw_target` and `drawable` modules should both be exported without having the internal modules in the path: `embedded_graphics::DrawTarget` and `embedded_graphics::Drawable`. Having to click through the empty modules makes finding these fundamental types in the docs harder than necessary IMO.

But this would depend on being able to fix the re-export issue.

jamwaffles:
Found a (reasonably elegant) fix! Adding [`#[doc(no_inline)]`](https://doc.rust-lang.org/rustdoc/the-doc-attribute.html#docno_inlinedocinline) puts the `DrawTarget` docs in the right place. It also changes it to be a re-export from `prelude` now which makes a lot more sense. I'm still unsure why there are some items that don't show up as re-exports, but anyway...

I didn't change the other exports from that module, or change `Drawable` to be a top-level export as suggested. I'd like to keep those changes for a separate PR.

rfuest:
> Found a (reasonably elegant) fix! Adding [`#[doc(no_inline)]`](https://doc.rust-lang.org/rustdoc/the-doc-attribute.html#docno_inlinedocinline) puts the `DrawTarget` docs in the right place.

Nice.

> I didn't change the other exports from that module, or change `Drawable` to be a top-level export as suggested. I'd like to keep those changes for a separate PR.

OK, using the `no_inline` attribute for the entire `pub use` statement in the prelude seems to be a good idea IMO. Can you open issues for these?

embedded-graphics/src/draw_target.rs
```diff
@@ -1,34 +1,41 @@
...
-/// Defines a display that can be used to render [`Drawable`] objects.
+/// A target for embedded-graphics drawing operations.
 ///
-/// To to add embedded-graphics support to a display driver, `DrawTarget` must be implemented. Once
-/// a `DrawTarget` is defined, it can be used to render [`Drawable`]s. Note that any iterator over
-/// [`Pixel`]s can be drawn as [`Drawable`] is implemented for `Iterator<Item = Pixel<C:
-/// PixelColor>>`. See the [`Drawable`] trait documentation for more details.
+/// The `DrawTarget` trait is used to add embedded-graphics support to a display
+/// driver or similar targets like framebuffers or image files.
+/// Targets are required to at least implement the `size` and `draw_iter` methods. All other
```
rfuest, line 12:
```suggestion
/// Targets are required to at least implement the [`size`] and [`draw_iter`] methods. All other
```

jamwaffles:
Agreed. #358 opened.

embedded-graphics/src/mock_display.rs
```diff
@@ -193,6 +194,10 @@ where
         self.allow_overdraw = value;
     }
 
+    fn size(&self) -> Size {
+        DISPLAY_AREA.size
+    }
+
```
rfuest, line 200: This isn't required, because `MockDisplay` also implements `DrawTarget::size`.

embedded-graphics/src/mock_display.rs
```diff
@@ -201,10 +206,24 @@ where
     }
 
     /// Changes the color of a pixel.
-    pub fn set_pixel(&mut self, p: Point, color: Option<C>) {
-        let Point { x, y } = p;
+    pub fn set_pixel(&mut self, point: Point, color: Option<C>) {
```
rfuest, line 209: `set_pixel` did intentionally not check for overdraw. It's supposed to be a lower level access than using the `DrawTarget`.

embedded-graphics/src/mock_display.rs
```diff
@@ -400,32 +419,21 @@ where
     }
 }
 
-impl<C> DrawTarget<C> for MockDisplay<C>
+impl<C> DrawTarget for MockDisplay<C>
 where
-    C: PixelColor,
+    C: PixelColor + ColorMapping,
```
rfuest, line 424: `ColorMapping` isn't required to implement `DrawTarget`.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -139,6 +139,104 @@ impl Rectangle {
...
+    /// This example creates two rectangles with no intersection between them. In this case,
+    /// `intersection` returns a zero-sized rectangle with its origin at `(0, 0)`.
+    ///
+    /// ```rust
+    /// use embedded_graphics::{
+    ///     mock_display::MockDisplay, pixelcolor::BinaryColor, prelude::*, primitives::Rectangle,
+    ///     style::PrimitiveStyle,
+    /// };
+    ///
+    /// let mut display: MockDisplay<BinaryColor> = MockDisplay::new();
```
rfuest, line 210: The display isn't used in the example.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -139,6 +139,104 @@ impl Rectangle {
             None
         }
     }
+
+    /// Returns a new `Rectangle` containing the intersection of `self` and `other`.
+    ///
+    /// If no intersection is present, this method will return a zero sized rectangle with its top
+    /// left corner set to `(0, 0)`.
```
rfuest, line 146: Do we need to document that `top_left` is (0, 0)? IMO the user shouldn't care about the `top_left` corner if a rectangle is zero sized.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -139,6 +139,104 @@ impl Rectangle {
...
+    ///
+    /// let mut display: MockDisplay<BinaryColor> = MockDisplay::new();
+    ///
+    /// let rect1 = Rectangle::new(Point::zero(), Size::new(7, 8));
+    /// let rect2 = Rectangle::new(Point::new(10, 15), Size::new(10, 7));
+    ///
+    /// let intersection = rect1.intersection(&rect2);
+    ///
+    /// assert_eq!(intersection.size, Size::zero());
+    /// assert_eq!(intersection.top_left, Point::zero());
```
rfuest, line 218: Related to the note above, I would remove this check.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -139,6 +139,104 @@ impl Rectangle {
...
+    ///         "  ..........",
+    ///     ])
+    /// );
+    /// # Ok::<(), core::convert::Infallible>(())
+    /// ```
+    ///
+    /// ## No intersection
+    ///
+    /// This example creates two rectangles with no intersection between them. In this case,
+    /// `intersection` returns a zero-sized rectangle with its origin at `(0, 0)`.
```
rfuest, line 202: See note above.

embedded-graphics/src/primitives/rectangle.rs
```diff
@@ -139,6 +139,104 @@ impl Rectangle {
...
+    ///
+    /// let intersection = rect1.intersection(&rect2);
+    ///
+    /// assert_eq!(intersection.size, Size::zero());
+    /// assert_eq!(intersection.top_left, Point::zero());
+    /// # Ok::<(), core::convert::Infallible>(())
+    /// ```
+    pub fn intersection(&self, other: &Rectangle) -> Rectangle {
+        let other_bottom_right = other.bottom_right().unwrap_or_else(Point::zero);
+        let self_bottom_right = self.bottom_right().unwrap_or_else(Point::zero);
```
rfuest, line 223: I think this contains an error and this test should show the problem (haven't tried to run it):
```rust
let rect1 = Rectangle::new(Point::new(0, 0), Size::new(0, 0));
let rect2 = Rectangle::new(Point::new(-10, -10), Size::new(20, 20));

assert_eq!(rect1.intersection(&rect2).size, Size::zero());
```
If any `bottom_right` is `None` the intersection must always be `Size::zero()`.

jamwaffles: That is indeed broken, well spotted. Should we add the size check to `Rectangle`'s `ContainsPoint::contains` impl as well?

embedded-graphics/src/mock_display.rs
jamwaffles: This and some of the other `MockDisplay` weirdness was a bad rebase on my part I think. Anyway, the bounds checking has been moved into `MockDisplay`'s `DrawTarget` impl which I think is the correct place?

embedded-graphics/src/primitives/rectangle.rs
rfuest: I don't think that `contains` needs an additional size check because it checks the  result of `bottom_right`, that is `None` if the rectangle is zero sized. But we could use the same check for `intersection`:
```rust
    pub fn intersection(&self, other: &Rectangle) -> Rectangle {
        if let (Some(other_bottom_right), Some(self_bottom_right)) =
            (other.bottom_right(), self.bottom_right())
        {
            // Check for overlap
            if self.contains(other.top_left)
                || self.contains(other_bottom_right)
                || other.contains(self.top_left)
                || other.contains(self_bottom_right)
            {
                return Rectangle::with_corners(
                    self.top_left.component_max(other.top_left),
                    self_bottom_right.component_min(other_bottom_right),
                )
            }
        }

        // No overlap present
        Rectangle::new(Point::zero(), Size::zero())
    }
```

embedded-graphics/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - **(breaking)** [#353](https://github.com/jamwaffles/embedded-graphics/pull/353) Lines with equal start and end points are now drawn by assuming they are oriented horizontally. This means that they are now drawn as a `stroke_width` high and 1px wide rectangle, instead of not being drawn at all.
 - **(breaking)** [#353](https://github.com/jamwaffles/embedded-graphics/pull/353) `primitives::line::StyledLineIterator` was renamed to `primitives::line::StyledIterator`.
 - **(breaking)** [#357](https://github.com/jamwaffles/embedded-graphics/pull/357) Additional checks for overdraw and out of bounds drawing were added to `MockDisplay` which can cause tests to panic. See the `mock_display` module docs for more information.
+- **(breaking)** [#342](https://github.com/jamwaffles/embedded-graphics/pull/342) Refactored the `DrawTarget` trait to better support common hardware capabilities.
```
rfuest, line 36: `Rectangle::intesection` needs a changelog entry.

embedded-graphics/src/mock_display.rs
rfuest: Yes, looks good now.

jamwaffles:
Changelog entry added, intersection impl changed for your improved version :+1:

rfuest:
LGTM :+1:

jamwaffles:
Thanks again for the detailed review!

image

I'll buy you a pint/coffee/drink of choice when we get one with over 200 ;)

PR #349, opened 2020-05-21
Add egpolyline macro

### LINKED ISSUE #344: Add egpolyline macro
I forgot to add a macro for the `Polyline` primitive added in #292 :facepalm:

jamwaffles:

## PR description

Adds the `egpolyline!` macro.

Closes #344.

The reference `&` is left at the call site of the macro, i.e. `egpolyline!(points = &points)` as opposed to letting the macro add the reference internally with `egpolyline!(points = points)`. The latter is a little cleaner but hides some of the behaviour which isn't great.

I also wanted to add a variant where the following is possible:

```rust
let stroke_line: Styled<Polyline, PrimitiveStyle<Rgb888>> = egpolyline!(
    points = [(10, 20), (30, 40)],
    style = primitive_style!(stroke_color = Rgb888::BLUE)
);
```

where `points` would be converted into a list of `Point` by the macro, however `Polyline::new` takes a reference which means this isn't possible in the macro due to lifetime issues.

### DIFF
--- a/embedded-graphics/src/primitives/mod.rs
+++ b/embedded-graphics/src/primitives/mod.rs
@@ -308,6 +308,51 @@ macro_rules! egline {
     }};
 }
 
+/// Create a [`Polyline`](./primitives/polyline/struct.Polyline.html) with optional styling using a
+/// convenient macro.
+///
+/// Note that styled polylines currently only support the `stroke_color` and `stroke_width` style
+/// properties. The `stroke_width` property is currently clamped at 1px. All other style properties
+/// are ignored.
+///
+/// # Examples
+///
+/// ```rust
+/// use embedded_graphics::{
+///     egpolyline,
+///     pixelcolor::Rgb565,
+///     prelude::*,
+///     primitive_style,
+///     primitives::Polyline,
+///     style::{PrimitiveStyle, Styled},
+/// };
+///
+/// let points: [Point; 5] = [
+///     Point::new(8, 8),
+///     Point::new(48, 16),
+///     Point::new(32, 48),
+///     Point::new(16, 32),
+///     Point::new(32, 32),
+/// ];
+///
+/// let line: Styled<Polyline, PrimitiveStyle<Rgb565>> = egpolyline!(points = &points);
+///
+/// let stroke_line: Styled<Polyline, PrimitiveStyle<Rgb565>> = egpolyline!(
+///     points = &points,
+///     style = primitive_style!(stroke_width = 1, stroke_color = Rgb565::BLUE)
+/// );
+/// ```
+#[macro_export]
+macro_rules! egpolyline {
+    (points = $points:expr $(,)?) => {
+        $crate::primitives::Polyline::new($points)
+            .into_styled($crate::style::PrimitiveStyle::default())
+    };
+    (points = $points:expr, style = $style:expr $(,)?) => {
+        $crate::primitives::Polyline::new($points).into_styled($style)
+    };
+}
+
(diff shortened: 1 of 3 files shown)


embedded-graphics/src/primitives/mod.rs
```diff
@@ -308,6 +308,51 @@ macro_rules! egline {
     }};
 }
 
+/// Create a [`Polyline`](./primitives/polyline/struct.Polyline.html) with optional styling using a
+/// convenient macro.
+///
+/// Note that styled polylines currently only support the `stroke_color` and `stroke_width` style
+/// properties. The `stroke_width` property is currently clamped at 1px. All other style properties
+/// are ignored.
```
rfuest, line 316: IMO this should be in the `Polyline` docs and not here, because it's easy to miss if this changes in the future.  And the macro doesn't use the mentioned properties directly, which could be confusing.

I would suggest to replace it by a note that that styling options are limited for polylines and that the `Polyline` docs provide more info.

embedded-graphics/src/primitives/mod.rs
```diff
@@ -308,6 +308,51 @@ macro_rules! egline {
     }};
 }
 
+/// Create a [`Polyline`](./primitives/polyline/struct.Polyline.html) with optional styling using a
+/// convenient macro.
```
rfuest, line 312: I know that this is the same as the other macros, but I think we should change this. The macros never create a raw primitive and always apply styling.

Now that I'm thinking about this, does having the macro variant without `style = ...` make sense at all? Why would you want to create a styled primitive that doesn't draw anything? Maybe it would be more useful to return a raw primitive in that case.

rfuest:
> I also wanted to add a variant where the following is possible:
> 
> ```rust
> let stroke_line: Styled<Polyline, PrimitiveStyle<Rgb888>> = egpolyline!(
>     points = [(10, 20), (30, 40)],
>     style = primitive_style!(stroke_color = Rgb888::BLUE)
> );
> ```
> 
> where `points` would be converted into a list of `Point` by the macro, however `Polyline::new` takes a reference which means this isn't possible in the macro due to lifetime issues.

I think this could work for a limited subset, like the example you provided, but not every `T: Into<Point>`. For this case the points could be stored in an internal `const`. Example macro output:
```rust
{
    const POINTS: &[Point] = [Point::new(10, 20), ...];
    egpolyline!(points = POINTS, ...)
}
```

embedded-graphics/src/primitives/mod.rs
jamwaffles: Yeah, the macro docs need an overhaul. There's lots of semi-incorrect copypasta. 

> Now that I'm thinking about this, does having the macro variant without style = ... make sense at all?

Not in their current form. We could change the variants without `style =` return a plain primitive, but that might get confusing. We could also remove the macros completely (although worth consulting with the e-g matrix chat if we do this).

Do you mind if we move this specific change and the other macro stuff to another PR? I'll open an issue to track.

jamwaffles: See #350

rfuest: > We could also remove the macros completely (although worth consulting with the e-g matrix chat if we do this).

TBH, I never use the macros and don't see a huge benefit in having them, so I would support this idea.

rfuest: > Do you mind if we move this specific change and the other macro stuff to another PR? I'll open an issue to track.

No, I just noticed this while doing the review and it should be handled in another PR.

PR #357, opened 2020-06-02
Add MockDisplay features and cleanup font tests

### LINKED ISSUE #356: Add something to MockDisplay to check overdraw/diffs
As discussed in https://github.com/jamwaffles/embedded-graphics/pull/353#discussion_r432920157

In writing #342 I've found a need for checking for overdraw, so it would be useful to make the function (or functionality) in that comment thread part of the e-g test harness.

Another feature that would be very useful is some kind of diff. As of #342, draw targets won't necessarily use the primitive's iterator to output pixels, so we need to add a check to ensure their outputs are equal. A method of diffing the mock display will help ensure we keep the outputs the same.

rfuest:

## PR description

This PR adds the additional checks, we discussed in #356, to `MockDisplay`. I've adjusted some dimensions in the existing tests and examples to make them fit in the 64 px x 64 px display area. IMO this is better than allowing out of bounds drawing for many examples.

In addition to the `MockDisplay` changes the PR also cleans up the font test a bit. Some tests in the fonts were actually tests of the styled text drawable, so I moved them there. The fonts should now only contain tests that are relevant to the individual font.

There might be other tests, for example in the `primitives` module, that could benefit from the new features, but this can be implemented in another PR. There is also some overlap with the changes in #342, so we need to decide which to merge first.

Closes #356.

### DIFF
--- a/embedded-graphics/src/lib.rs
+++ b/embedded-graphics/src/lib.rs
@@ -103,23 +103,23 @@
 //! ```rust
 //! use embedded_graphics::{
 //!     fonts::{Font6x8, Text},
-//!     mock_display::MockDisplay,
+//! #   mock_display::MockDisplay,
 //!     pixelcolor::Rgb565,
 //!     prelude::*,
 //!     primitives::Circle,
 //!     style::{PrimitiveStyle, TextStyle},
 //! };
 //!
-//! // Create a draw target using the builtin MockDisplay. In real applications this would be
-//! // replaced by a draw target that is provided by a display driver crate.
-//! let mut display = MockDisplay::default();
-//! display.set_allow_overdraw(true);
-//! display.set_allow_out_of_bounds_drawing(true);
+//! # let mut display = MockDisplay::default();
+//! # display.set_allow_overdraw(true);
+//! # display.set_allow_out_of_bounds_drawing(true);
 //!
 //! let c = Circle::new(Point::new(12, 12), 17).into_styled(PrimitiveStyle::with_fill(Rgb565::RED));
 //! let t = Text::new("Hello Rust!", Point::new(20, 16))
 //!     .into_styled(TextStyle::new(Font6x8, Rgb565::GREEN));
 //!
+//! // The `display` variable contains a `DrawTarget` implementation provided by the display driver
+//! // crate. See the driver crate documentation for more information about how it is constructed.
 //! c.draw(&mut display)?;
 //! t.draw(&mut display)?;
 //! # Ok::<(), core::convert::Infallible>(())
--- a/embedded-graphics/src/lib.rs
+++ b/embedded-graphics/src/lib.rs
@@ -113,6 +113,8 @@
 //! // Create a draw target using the builtin MockDisplay. In real applications this would be
 //! // replaced by a draw target that is provided by a display driver crate.
 //! let mut display = MockDisplay::default();
+//! display.set_allow_overdraw(true);
+//! display.set_allow_out_of_bounds_drawing(true);
 //!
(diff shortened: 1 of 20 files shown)


jamwaffles:
It would be great to get #342 over the line first, but of course I'm biased. Hopefully there's not much left on that PR so hopefully isn't too much of a holdup.

embedded-graphics/src/lib.rs
```diff
@@ -113,6 +113,8 @@
 //! // Create a draw target using the builtin MockDisplay. In real applications this would be
 //! // replaced by a draw target that is provided by a display driver crate.
 //! let mut display = MockDisplay::default();
+//! display.set_allow_overdraw(true);
+//! display.set_allow_out_of_bounds_drawing(true);
```
jamwaffles, line 117: Should these be hidden in this example? If not, a oneline comment explaining why they're present would be useful.

embedded-graphics/src/mock_display.rs
```diff
@@ -324,19 +403,32 @@ where
...
 
-        let i = x + y * SIZE as i32;
-        self.0[i as usize] = Some(color);
+        if !self.allow_overdraw {
+            if self.get_pixel(point).is_some() {
+                panic!("tried to draw pixel twice (x: {}, y: {})", point.x, point.y);
+            }
+        }
+
+        self.set_pixel(point, Some(color));
```
jamwaffles, line 425: Clippy is complaining that this could be the following, but I understand if you want to keep it as is for consistency with the above `if { if { ... } }`

```suggestion
        if !self.allow_overdraw && self.get_pixel(point).is_some() {
            panic!("tried to draw pixel twice (x: {}, y: {})", point.x, point.y);
        }
```

embedded-graphics/src/mock_display.rs
```diff
@@ -207,11 +247,46 @@ where
 
         mirrored
     }
+
+    /// Maps a `MockDisplay<C>' to a `MockDisplay<CT>` by applying a function
+    /// to each pixel.
+    ///
+    /// # Examples
```
jamwaffles, line 254: It took me a bit long than I'd like to admit to figure out what this example does, so a heading would help clarify it at a glance IMO.

```suggestion
    /// # Examples
    ///
    /// ## Invert the colors in a `BinaryColor` pattern
```

embedded-graphics/src/lib.rs
rfuest: The `MockDisplay` creation was unhidden, when `lib.rs` and `README.md` shared the same example. But if this example is only used in `lib.rs` I would suggest to also hide the `MockDisplay::default()` line and replace that by a comment that `display` is the `DrawTarget` impl from a driver crate. OK?

jamwaffles: Yep, that sounds good to me

embedded-graphics/src/mock_display.rs
rfuest: Good point! I've added a comment.

embedded-graphics/src/mock_display.rs
rfuest: I've changed it to make clippy happy.

embedded-graphics/src/lib.rs
rfuest: Is the comment I've added OK?

jamwaffles: Yep that looks good, thanks

rfuest:
> I'll handle any rebase conflicts over in that PR.

Thanks for doing the extra work!

PR #363, opened 2020-06-05
Split primitives files into multi-file modules

### LINKED ISSUE #361: Split primitives into multiple files
I think it's about time we did this. I was doing some work towards #343 which would likely make the primitives files even longer, spurring me to create this issue.

I'm going to hopefully create a PR addressing this asap unless @rfuest is working on this already.

jamwaffles:

## PR description

Closes #361


jamwaffles:
This should be ready for an initial review. Some points:

* This should be reviewable commit-by-commit. I tried to make them self contained so it's easier to go through. 
* The last commit ("Reduce code duplication in Triangle iterators") is my attempt at deduping the triangle iter code for #314. You might've had a different idea though so I'm happy to change my approach.
* I wasn't entirely sure how to split the tests as some of them apply to multiple files in a module. If you find tests that should obviously be in a different file please let me know.
* `CornerRadii` and `EllipseQuadrant` are moved into `rounded_rectangle` as they're only used in there for now. We could move `EllipseQuadrant` out later if we expose it as a primitive.

rfuest:
> The last commit ("Reduce code duplication in Triangle iterators") is my attempt at deduping the triangle iter code for #314. You might've had a different idea though so I'm happy to change my approach.

I have never looked at the triangle code in detail before and didn't have a plan how to change it yet. But on first glance your version look much nicer than the old code.

> `CornerRadii` and `EllipseQuadrant` are moved into `rounded_rectangle` as they're only used in there for now. We could move `EllipseQuadrant` out later if we expose it as a primitive.

Sounds good.

embedded-graphics/src/primitives/circle/mod.rs
```diff
@@ -290,50 +189,19 @@ where
...
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{
-        mock_display::MockDisplay,
-        pixelcolor::BinaryColor,
-        style::{PrimitiveStyleBuilder, StrokeAlignment},
-    };
+    use crate::{mock_display::MockDisplay, pixelcolor::BinaryColor, prelude::*};
```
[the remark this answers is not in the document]
jamwaffles: >  If the code is compiled as an separate crate and uses use embedded_graphics::... it should use the prelude and if it uses use crate::... it shouldn't.

That's a good rule of thumb, I'll try and remember it from now on.

> But I would prefer to use the same style for all tests modules.

Agreed. I've removed the prelude from all e-g primitives internals now.

jamwaffles:
> One suggestion to solve this: Rename all styled_iterator.rs filed into styled.rs and move all Styled related code into that file, not only the iterator but also Drawable and IntoIterator. Then all styling depended test could also go into that file.

I briefly considered this idea but decided not to go in that direction for some reason. After the initial split, I think this is indeed the best way to go, so I'll refactor the primitives to follow this pattern.

embedded-graphics/src/primitives/circle/styled_iterator.rs
```diff
@@ -0,0 +1,154 @@
+use crate::{
+    drawable::Pixel,
+    pixelcolor::PixelColor,
+    primitives::circle::{diameter_to_threshold, distance_iterator::DistanceIterator, Circle},
+    style::{PrimitiveStyle, Styled},
+};
+
+/// Pixel iterator for each pixel in the circle border
```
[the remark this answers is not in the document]
jamwaffles: > All Styled... iterators need a new comment.

Not entirely sure what you mean by this. Do they need tweaking or expanding or something else?

> We don't use "border" elsewhere IIRC and it doesn't describe the purpose very well IMO.

True, I'll fix this.

jamwaffles:
_wheeze_

5 million <kbd>Ctrl</kbd> + <kbd>X</kbd>, <kbd>Ctrl</kbd> + <kbd>V</kbd> later and I think this is in a much better place now. The new commits should be reasonably self-contained so might help to go through them one at a time to keep the file moving noise down.

embedded-graphics/src/primitives/circle/styled_iterator.rs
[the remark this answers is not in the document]
rfuest: > Not entirely sure what you mean by this. Do they need tweaking or expanding or something else?

The main change was described in the following sentence, but IMO they could also be expanded a bit. IMO the comment should describe what a `StyledIterator` is and how it can be created.

embedded-graphics/src/mock_display.rs
```diff
@@ -288,6 +288,22 @@ where
     }
 }
 
+// impl MockDisplay<BinaryColor> {
```
rfuest, line 291: I think this shouldn't be here.

embedded-graphics/src/mock_display.rs
```diff
@@ -349,6 +365,21 @@ where
...
+
+    /// TODO: Docs
+    pub fn from_points<I>(points: I) -> Self
+    where
+        I: Iterator<Item = Point>,
+    {
+        let mut display = MockDisplay::new();
+
+        points
+            .map(|p| Pixel(p, C::on()))
```
rfuest, line 377: I like the general idea of having `ColorMapping::on`, but I'm not sure if we need this here. My idea for `from_points` was only to use it for testing points iterators against patterns, which doesn't require other colors. And if you need that feature you could also use `map` to get any color: `MockDisplay::from_points(p).map(|_| Rgb888::MAGENTA)`.

Or we could implement `MockDisplay::draw_points(&mut self, points: I, color: C)` instead. This would allow us to combine multiple points iters into a single mock display pattern. This would also have the advantage that you could disable the overdraw and out of bounds checks before drawing.

jamwaffles:
> (or organised if you prefer that :wink:)

image

embedded-graphics/src/mock_display.rs
jamwaffles: I uncommented this and removed the other `from_points` method you commented on below to get `from_points` closer to what you were thinking. Should we try and merge this impl block with the `impl<C> MockDisplay<C>` above it? I tried before but ran into generics issues.

embedded-graphics/src/mock_display.rs
jamwaffles: See other comment

embedded-graphics/src/mock_display.rs
rfuest: > Should we try and merge this impl block with the impl<C> MockDisplay<C> above it? I tried before but ran into generics issues.

This would only work if we had some `C` parameter, like: `fn from_points(iter: I, color: C) -> Self`. But maybe replacing `from_points` by `draw_points` (suggested below) would be the most flexible option. Or keep both, if you like, where `from_points` only works with `BinaryColor` and `draw_points` with any `C`.

One thing I would like to have changed is the use of `Pixel::draw` in `from_points`. IMO this code should be independent of `Drawable`, so we don't use the same code path if something breaks, which could lead to false negatives.

To implement this I would suggest to revert what I sad about `set_pixel` in a previous PR (sorry for the back and forth) and implement the bounds and overdraw checks in that method. If the unchecked behavior is required the user can always disable the checks temporarily. But I would suggest to rename it to `draw_pixel` to make it clearer that it is not a simple setter, which should also be mentioned in the docs.

jamwaffles: I've pushed a new commit with some changes in it. I've moved the bounds checking to `set_pixel` (now renamed to `draw_pixel`) and added some docs.

In terms of `draw_points` and `from_points`, I've gone with just having `from_points` as that's all we need currently.

PR #365, opened 2020-06-09
Initial SparseIterator experiment

### LINKED ISSUE #330: DrawTarget improvements

jamwaffles:

## PR description

Experiments with adding a `SparseIterator` struct which wraps a non-contiguous iterator into something that produces an `Option<C>` for each point in a rectangular area. This is the first idea that came to mind and it might be completely the wrong approach, but either way it would be great to get some feedback.

Closes #330

### DIFF
--- a/embedded-graphics/src/draw_target/mod.rs
+++ b/embedded-graphics/src/draw_target/mod.rs
@@ -302,23 +302,22 @@ pub trait DrawTarget: Dimensions {
     /// [`core::convert::Infallible`]: https://doc.rust-lang.org/stable/core/convert/enum.Infallible.html
     type Error;
 
-    // TODO: Reenable this in a new PR with modified primitives iterators
-    // // TODO: Mention performance
-    // // TODO: Mention default impl behaviour
-    // /// Fill a given area with a transparent pixel iterator.
-    // ///
-    // /// Iteration order is guaranteed from top left to bottom right, however some pixels may be
-    // /// transparent which are represented as `None`.
-    // fn fill_sparse<I>(&mut self, area: &Rectangle, colors: I) -> Result<(), Self::Error>
-    // where
-    //     I: IntoIterator<Item = Option<Self::Color>>,
-    // {
-    //     self.draw_iter(
-    //         area.points()
-    //             .zip(colors)
-    //             .filter_map(|(pos, color)| color.map(|c| Pixel(pos, c))),
-    //     )
-    // }
+    // TODO: Mention performance
+    // TODO: Mention default impl behaviour
+    /// Fill a given area with a transparent pixel iterator.
+    ///
+    /// Iteration order is guaranteed from top left to bottom right, however some pixels may be
+    /// transparent which are represented as `None`.
+    fn fill_sparse<I>(&mut self, area: &Rectangle, colors: I) -> Result<(), Self::Error>
+    where
+        I: IntoIterator<Item = Option<Self::Color>>,
+    {
+        self.draw_iter(
+            area.points()
+                .zip(colors)
+                .filter_map(|(pos, color)| color.map(|c| Pixel(pos, c))),
+        )
+    }
 
     /// Draw individual pixels to the display without a defined order.
     ///
--- a/embedded-graphics/src/iterator/mod.rs
+++ b/embedded-graphics/src/iterator/mod.rs
@@ -58,7 +58,7 @@ where
     I::Item: PixelColor,
 {
     fn into_pixels(self, bounding_box: &Rectangle) -> contiguous::IntoPixels<Self> {
-        contiguous::IntoPixels::new(self, bounding_box.clone())
+        contiguous::IntoPixels::new(self, *bounding_box)
     }
 }
 
--- a/embedded-graphics/src/iterator/mod.rs
+++ b/embedded-graphics/src/iterator/mod.rs
@@ -1,7 +1,11 @@
 //! Iterators.
 
 use crate::{
-    draw_target::DrawTarget, geometry::Point, pixelcolor::PixelColor, primitives::Rectangle, Pixel,
+    draw_target::DrawTarget,
+    geometry::{Dimensions, Point},
+    pixelcolor::PixelColor,
+    primitives::Rectangle,
+    Pixel,
 };
 
 pub mod contiguous;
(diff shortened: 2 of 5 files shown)


rfuest:
> Experiments with adding a `SparseIterator` struct which wraps a non-contiguous iterator into something that produces an `Option<C>` for each point in a rectangular area. This is the first idea that came to mind and it might be completely the wrong approach, but either way it would be great to get some feedback.

I was thinking that it should be implemented the other way around. Take the styled circle iterator for example:
```rust
    for (point, distance) in &mut self.iter {
        let color = if distance < self.inner_threshold {
            self.inner_color
        } else if distance < self.outer_threshold {
            self.outer_color
        } else {
            None
        };

        if let Some(color) = color {
            return Some(Pixel(point, color));
        }
    }

    None
}
```
Only using the inner part of the loop is a sparse iterator:
```rust
    self.iter.map(|(point, distance)| {
        if distance < self.inner_threshold {
            self.inner_color
        } else if distance < self.outer_threshold {
            self.outer_color
        } else {
            None
        }
    })
}
```
And it would be wasteful to convert a spare iterator in a non-contiguous iterator just to convert it back again.

jamwaffles:
That makes a lot more sense. I've changed `SparseIterator` around and converted the circle rendering to use `fill_sparse` as a first test. I've left the styled circle iterator to continue returning sparse `Pixel`s as a deliberate choice. Do you think we should keep the styled iterator behaviour the same as master (i.e. sparse if necessary) and only add non-sparse code to support `fill_sparse` internally?

rfuest:
> I've left the styled circle iterator to continue returning sparse `Pixel`s as a deliberate choice. Do you think we should keep the styled iterator behaviour the same as master (i.e. sparse if necessary) and only add non-sparse code to support `fill_sparse` internally?

The name `StyledPixels` implies that it returns pixels, so if we decide to keep the name it also should return pixels IMO.

The `IntoIterator` trait is very generic and the name `into_iter` doesn't provide any information about what iterator is created. Can we replace it by custom traits to support different iterator types, like we did for the `.points()` method? `styled_primitive.pixels()` is much easier to understand than `styled_primitives.into_iter()` IMO. And we could also add additional methods to create the `Item = Option<C>` iterator variant.

embedded-graphics/src/draw_target.rs
```diff
@@ -297,23 +297,22 @@ pub trait DrawTarget {
...
+    /// transparent which are represented as `None`.
+    fn fill_sparse<I>(&mut self, area: &Rectangle, colors: I) -> Result<(), Self::Error>
+    where
+        I: IntoIterator<Item = Option<Self::Color>>,
+    {
+        self.draw_iter(
+            area.points()
+                .zip(colors)
+                .filter_map(|(pos, color)| color.map(|c| Pixel(pos, c))),
+        )
```
rfuest, line 314: Can't this be replaced by using `SparseIterator`?

jamwaffles:
The original code was getting pretty stale, so I reset this branch and reimplemented sparse/non-sparse for `Circle`. Can you take a quick look to see if I'm on the right track again?

`FilteredIterator` could _really_ use a better name. It takes a sparse iterator and converts it to a "normal" iterator of `Pixel<C>`. It should be resuable across all primitives, reducing a bit of boilerplate.

jamwaffles:
It just occurred to me that we don't necessarily need `FilteredIterator` for our purposes, but it might be useful for others if they want all styled pixels in a primitive without having to filter it themselves. Should we keep it in, along with the `IntoPixels` impl for each primitive?

rfuest:
> Can you take a quick look to see if I'm on the right track again?

It's the track we discussed, but maybe not the right one.

I think that `Circle` (and `Ellipse`) could be implemented more efficiently by using a scanline approach using `fill_solid`. This could leverage the symmetry inherit in these shapes to reduce the number of required calculations. And it wouldn't require any caching on the display driver side.

This would only leave text rendering which uses this feature. Perhaps we should just remove the sparse iter entirely and use `draw_iter` instead. The display driver could still try to combine multiple pixels into larger blocks and I think this might not even be much more efficient when using a sparse iterator. With a sparse iter we discard the position information that the display driver needs to recreate and this isn't required when using `draw_iter`. The algorithm for creating larger blocks might be a little more complicated without knowing the iteration order, but this would only needed to be implemented once (in e-g or another crate) and can be shared between multiple display drivers.

jamwaffles:
> I think that Circle (and Ellipse) could be implemented more efficiently

Certainly, although for this PR, that code was a precursor to implementing fill_sparse for any other primitives that needed it. A scanline iterator would work a lot better though. It could even use `fill_rect` for each line for a tidy speed boost.

> Perhaps we should just remove the sparse iter entirely and use draw_iter instead.

Ok, so if I understand correctly (sorry, it's very warm here...), you're suggesting closing this PR and forgetting the idea of `fill_sparse` entirely? This would also allow closure of #330 which I think is only missing this `fill_sparse` behaviour.

rfuest:
> It could even use `fill_rect` for each line for a tidy speed boost.

I think `fill_rect` is called `fill_solid`. But maybe I'm just imagining things because we have similar temperatures.

> Ok, so if I understand correctly (sorry, it's very warm here...), you're suggesting closing this PR and forgetting the idea of `fill_sparse` entirely? This would also allow closure of #330 which I think is only missing this `fill_sparse` behaviour.

Yes, that's what I'm suggesting. If we change our mind later it wouldn't even be a breaking change because we would provide a default implementation.

jamwaffles:
> I think fill_rect is called fill_solid.

Oops yes it is, that's what I meant.

> Yes, that's what I'm suggesting. If we change our mind later it wouldn't even be a breaking change because we would provide a default implementation.

Sweet. I'll close this issue and yeah I agree - adding stuff in later is a lot easier and less breaky than doing it poorly now.

rfuest:
@jamwaffles We still need to remove the commented out code in `DrawTarget`. Will you create a PR for that?

jamwaffles:
Yep, on it now.

PR #369, opened 2020-06-12
Split arc sector files

### LINKED ISSUE #368: Split arc and sector primitives into multiple files
The arc and sector primitives should use the same layout that was implemented for the other primitives in #363.

jamwaffles:

## PR description

Closes #368.


embedded-graphics/src/primitives/arc/linear_equation.rs
```diff
@@ -0,0 +1,48 @@
...
+        let (a, b) = match angle.tan() {
+            None => (Real::from(1.0), Real::from(0.0)),
+            Some(a) => (-a, Real::from(-1.0)),
+        };
+        let c = -(a * point.x.into() + b * point.y.into());
+        LinearEquation { a, b, c }
+    }
+
+    /// Create a flat line equation
+    pub(in crate::primitives) fn flat() -> Self {
```
rfuest, line 32:
```suggestion
    /// Create a horizontal line equation
    pub(in crate::primitives) fn new_horizontal() -> Self {
```

embedded-graphics/src/primitives/arc/linear_equation.rs
```diff
@@ -0,0 +1,48 @@
+use crate::geometry::{Angle, Point, Real, Trigonometry};
+
+/// Define one side of a line
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub(in crate::primitives) enum LineSide {
```
rfuest, line 5: If this type is never exported in another public module, this would be inaccessible from the outside, even if it were declared as `pub` here. Should we stick to `pub` for these cases (not only for this type)? I'm not sure if there is any preferred way of handling this.

embedded-graphics/src/primitives/arc/linear_equation.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+pub(in crate::primitives) struct LinearEquation {
+    a: Real,
+    b: Real,
+    c: Real,
+}
+
+impl LinearEquation {
+    /// Create a new linear equation based on one point and one angle
+    pub(in crate::primitives) fn from_point_angle(point: Point, angle: Angle) -> Self {
```
rfuest, line 22: Related to the comment above: If the whole struct is only `pub(in crate::primitives)`, do we also need to limit the scope of each constructor? This way we would need to change multiple locations, if we ever decide to widen the scope of, in this example, `LinearEquation`.

embedded-graphics/src/primitives/arc/plane_sector.rs
```diff
@@ -0,0 +1,150 @@
+use crate::{
+    geometry::{angle_consts::*, Angle, Dimensions, Point, Size},
+    primitives::{
+        arc::{linear_equation::LineSide, LinearEquation},
+        rectangle, Primitive, Rectangle,
+    },
+};
+
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
```
rfuest, line 9:
```suggestion
/// Sector shaped part of a plane.
///
/// The shape is described by two half-planes that divide the XY plane along the two
/// lines from the center point to the arc's end points. For sweep angles < 180° the
/// intersection of both half-planes is used and for angles >= 180° the union of both
/// half-planes.
#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
```

embedded-graphics/src/primitives/arc/plane_sector.rs
```diff
@@ -0,0 +1,150 @@
...
+            correct_a_side && correct_b_side
+        } else if self.sweep < ANGLE_360DEG {
+            correct_a_side || correct_b_side
+        } else {
+            true
+        }
+    }
+}
+
+/// Iterator that returns only the points which are inside a plane sector defined by two lines.
```
rfuest, line 66:
```suggestion
/// Iterator over the points in the intersection of a plane sector and the bounding box of a primitive.
```

embedded-graphics/src/primitives/arc/styled.rs
```diff
@@ -0,0 +1,171 @@
...
+                PlaneSectorIterator::new(
+                    &stroke_area,
+                    stroke_area.center(),
+                    stroke_area.angle_start,
+                    stroke_area.angle_sweep,
+                ),
+            )
+        } else {
+            DistanceIterator::new(Point::zero(), PlaneSectorIterator::empty())
+        };
```
rfuest, line 54:
```suggestion
        let points = if !styled.style.is_transparent() {
            PlaneSectorIterator::new(
                &stroke_area,
                stroke_area.center(),
                stroke_area.angle_start,
                stroke_area.angle_sweep,
            )
        } else {
            PlaneSectorIterator::empty()
        };
        
        let iter = DistanceIterator::new(stroke_area.center_2x(), points);
```

embedded-graphics/src/primitives/arc/plane_sector.rs
```diff
@@ -0,0 +1,150 @@
...
+        assert_eq!(iter.next(), Some((Point::new(1, 1), 0)));
+        assert_eq!(iter.next(), Some((Point::new(2, 1), 4)));
+        assert_eq!(iter.next(), None);
+    }
+
+    #[test]
+    fn distance_iter_empty() {
+        let mut iter = DistanceIterator::new(Point::zero(), PlaneSectorIterator::empty());
+        assert_eq!(iter.next(), None);
+    }
```
rfuest, line 149: IMO we don't need these anymore, because the they test `DistanceIter`, which is in another module.

embedded-graphics/src/primitives/arc/plane_sector.rs
```diff
@@ -0,0 +1,150 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        geometry::AngleUnit,
+        primitives::{circle::DistanceIterator, Arc},
+    };
+
+    #[test]
+    fn plane_arc_iter() {
```
rfuest, line 112:
```suggestion
    fn plane_sector_iter() {
```

embedded-graphics/src/primitives/arc/linear_equation.rs
jamwaffles: I'm not entirely sure what you mean. Ideally, everything that doesn't need to be `pub` isn't, so there might be some tightening we can do around that.

In this specific case, `LineSide` is also used in the adjacent `plane_sector` module so can't remain private. I could move it into `arc/mod.rs` which would allow it to be private but also used in both modules which might be a better solution.

embedded-graphics/src/primitives/arc/linear_equation.rs
jamwaffles: Ah, I see what you mean now. That's a good point. I think I can just make these `pub` and rely on the struct visibility as you suggest. I didn't think of that. I'll change the arc/sector methods accordingly.

embedded-graphics/src/primitives/arc/plane_sector.rs
jamwaffles: Good point. I thought it would be useful to test the combination of `DistanceIterator` and `PlaneSectorIterator` but they're pretty redundant with the two tests before these. I'll remove them.

embedded-graphics/src/primitives/arc/linear_equation.rs
rfuest: > I'm not entirely sure what you mean. Ideally, everything that doesn't need to be `pub` isn't, so there might be some tightening we can do around that.

Example:
```rust
// in arc/linear_equation.rs
pub(in crate::primitives) struct LinearEquation { ... }

// in arc/mod.rs
pub(in crate::primitives) use linear_equation::LinearEquation;
```
Do we need to specify `in crate::primitives` in both cases? Because if it isn't exported in `mod.rs` it wouldn't be visible outside `primitives::arc` anyway. Reducing the number of specific  `in ...` scopes should make it easier to refactor code.

jamwaffles: Thanks for explaining, yeah that's not very clean. Shall I keep the `pub(in crate::primitives)` on the item, or on the `use`? I'm leaning towards the item but you might have a good reason to put it on the `use`.

rfuest: I think it needs to be on the `use`, because a simple `pub use ...` would expand the scope and probably cause a compiler error, because the exported type is only visible in the `primitives` module.

jamwaffles:
Comments addressed as far as I know :)

jamwaffles:
Thanks!

> And also good catch with the unrelated changes.

Yeah, not sure why I didn't catch them as part of #361 but hey, at least we caught them before public release.

PR #370, opened 2020-06-12
Remove redundant PixelColor trait bounds

rfuest:

## PR description

This PR removes some redundant `PixelColor` trait bounds and makes all of them use the `where` syntax.

### DIFF
--- a/embedded-graphics/src/drawable.rs
+++ b/embedded-graphics/src/drawable.rs
@@ -97,7 +97,9 @@ where
 /// [`Drawable`]: trait.Drawable.html
 /// [`DrawTarget`]: ./trait.DrawTarget.html
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
-pub struct Pixel<C: PixelColor>(pub Point, pub C);
+pub struct Pixel<C>(pub Point, pub C)
+where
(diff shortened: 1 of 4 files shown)


rfuest:
@jamwaffles Should we use the `where` syntax for all trait bounds? The `Drawable` impls currently use the abbreviated syntax.

rfuest:
> I’m pretty ambivalent but it’s less effort to convert to the where syntax than the other way round. I wouldn’t reject a PR that changes it all for the abbreviated syntax though - I do realise it’s shorter.

I don't think that all trait bounds can be expressed using the shorter syntax, so the `where` syntax would be the only way to use the same for all trait bounds.

If we decide to also use the shorter syntax we need to define a point at which it's better to start using the `where` syntax. I wouldn't mind `<C: PixelColor>`, but I prefer the `where` syntax for `Drawable::draw`:
```rust
fn draw<D: DrawTarget<Color = C>>(self, display: &mut D) -> Result<(), D::Error> {
    display.fill_contiguous(&self.bounding_box(), self.into_iter().map(|p| p.1))
}

fn draw<D>(self, display: &mut D) -> Result<(), D::Error> 
where
    D: DrawTarget<Color = C>,
{
    display.fill_contiguous(&self.bounding_box(), self.into_iter().map(|p| p.1))
}
```

PR #378, opened 2020-06-27
0.6 maintenance

### LINKED ISSUE #377: Possibility of a new 0.6.x bugfix release
Latest stable currently calculates `Size` for bounding boxes incorrectly, and has possibly other mistakes. I would like to take some time and gather bugfixes that were made to the codebase since release, and backport those to a maintenance branch. @jamwaffles  would you be interested in making a new bugfix release from that, while 0.7 is in the works?

bugadani:

## PR description

I wasn't sure if this branch is the best to open the PR against, but this is my best guess. I've seen the discussion in #288 and I strongly believe a maintenance branch and manual releasing instead of cargo-release (although I'm not familiar with that but I see no reason why that shouldn't work either) should be the simplest solution for a bugfix.

I'm still looking for code that could be backported, but I'm also trying to do the least amount of work because there was a significant amount of refactoring done since 0.6.2.

My reasoning behind this PR is that embedded-layout kinda-sorta relies a LOT on bounding box sizes currently and so a point release would make my life much easier while coding against the latest stable.

Closes #377

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,8 +7,3 @@ members = [
     "tinytga",
 ]
 
-[patch.crates-io]
-embedded-graphics = { path = "embedded-graphics" }
-embedded-graphics-simulator = { path = "simulator" }
-tinytga = { path = "tinytga" }
-tinybmp = { path = "tinybmp" }
--- a/embedded-graphics/CHANGELOG.md
+++ b/embedded-graphics/CHANGELOG.md
@@ -6,6 +6,8 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+## [0.6.2] - 2020-04-20
+
 ### Fixed
 
 - [#309](https://github.com/jamwaffles/embedded-graphics/pull/309) Prevent triangles with off-screen vertices from infinitely looping.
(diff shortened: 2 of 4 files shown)


jamwaffles:
This PR is a bit confusing as there are two commits in it that probably shouldn't be here. Can you rebase them out or something? I think the only commit that should be present is 48513c4, right? The CI build also failed with some broken tests from the cherrypicked fix.

The previous 0.6 backport/release process was pretty painful which is likely why you're seeing lots of weird commits as you mention in #377.

I'm certainly not opposed to getting this fix into 0.6 unless @rfuest has any objections, but it does raise the question of whether to forego a 0.6 point release and release 0.7.0-alpha.1.

bugadani:
Certainly, there was no perfect point to branch off, and to open the pr against. I'm not sure about the timeline of 0.7 and I don't want to rush that at all, but since that may be still some tume away until stable, I really hope a bugfix release could be possible.

This PR was not yet about the code, I'm planning on collectung as many fixes as possibe, so don't worry about the specifics yet.

rfuest:
I'm not sure if we should fix this in 0.6. There are some crates that expect this wrong behavior and changing it should be considered a breaking change IMO. That was the reason I decided to postpone these changes to 0.7 and not include any geometry changes in minor 0.6 releases.

jamwaffles:
I did consider that but decided it might be minor enough to release in a 0.6 point version, but on second thoughts that would be unwise.

I've also opened #380 to track what can go in a next alpha release FWIW.

bugadani:
All right, sorry for wasting your time on this, then. Looking fwd for 0.7 ;)

jamwaffles:
Not a waste at all! I'm sorry you didn't get the resolution you were looking for. Hopefully future versions of e-g will have fewer breaking changes so releases don't take so long, or are at least easier to break into patches.

PR #379, opened 2020-06-28
Clearer feature flags

jamwaffles:
Hi! Thank you for helping out with Embedded Graphics development! Please:

- [x] Check that you've added passing tests and documentation
- [x] Add a simulator example(s) where applicable
- [x] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) and appropriate crate (`embedded-graphics`, `simulator`, `tinytga`, `tinybmp`) if your changes affect the **public API**
- [x] Run `rustfmt` on the project
- [x] Run `./build.sh` (Linux/macOS only) and make sure it passes. If you use Windows, check that CI passes once you've opened the PR.

## PR description

Adds notes and comments about the `graphics` feature of tinytga and tinybmp.

This also adds in the sector and arc example generators - it looks like the example was added manually in #310 but not to the generator.

Closes #374 
Closes #375

PR #381, opened 2020-06-28
Cargo readme

### LINKED ISSUE #332: hello-world example in README.md is broken
The copy of the hello-world example in the readme is currently broken again.

Should we leave this issue open and fix this once before the release of the new stable version? Or should we reenable the readme check in master and comment the check out again before the release? Or is there a better permanent solution?

jamwaffles:

## PR description

Use `cargo-readme` to generate each crate's `README.md` from its `lib.rs`.

~Currently based on #379 so will need rebasing to master once that's merged.~

This is a somewhat exploratory PR into using `cargo-readme` to generate all crate readmes. This makes `lib.rs` a single source of truth for each crate and allows for easier testing of the examples contained within it. I've also reorganised `embedded-graphics/lib.rs` to hopefully make it a bit more approachable, and more relevant to both the readme ordering and docs ordering.

Closes #332

I also believe this closes #387


jamwaffles:
Installing cargo-readme in CI adds about 30 seconds to build time. I'm sort of ok with this, but if it gets much slower I can look into creating a custom docker image with some dependencies preinstalled.

jamwaffles:
Well that's frustrating... the links aren't removed. There's an open issue about that here: https://github.com/livioribeiro/cargo-readme/issues/18

Someone in that thread links to this regex solution which strips the links: https://github.com/porglezomp/pixel-canvas/blob/develop/gen-readme.sh not that clean, but generated readmes might outweigh the hax

rfuest:
GitHub also seems to remove the base64 encoded images for the examples when rendering the markdown.

jamwaffles:
Yeah, I noticed that earlier. I fixed the links by adding a script to strip them, but I don't think I can do anything about the inline images. I could upload them into the repo, but that would make testing them locally a pain.

rfuest:
Can we move the examples for the individual drawables somewhere else and just link to them from the README? IMO the old example in the README gave a good overview of e-g's functions and having all examples included in the README makes it really long.

jamwaffles:
This has turned out to be a lot harder than expected. What I tried:

* Outputting the generated examples to an `examples.rs` module and including it in the docs. This is quite ugly, but works with minimal changes.
* Including the generated example docs in the top of `primitives/mod.rs` with various macro/`include_str!()`/`include!()` hacks with the `#![doc]` attribute. None of these worked, spewing various errors caused by the top-level (`#!`) requirement.
* Rendering the generated examples, feeding the output through `rustdoc` and uploading the file to Github. The screenshots would be output separately and tracked in the repo. This won't work as Github renders the HTML code, not the page itself.
* Change the generator to output plain Markdown and separate images and upload both to Github. This would work as far as I can see, but we loose the ability to `rustfmt` the examples beyond their formatting in the `generate-example-screenshots.rs` file.

Do you have a preferred/ideal way this should be done? I'm at a bit of a loss to be honest.

I did at least put the README example back and removed the list of generated examples, so I've got that going for me :grin:

jamwaffles:
Oh come on! lol

The simulator is `std`, so can't be included in the dev dependencies in e-g to build the simulator-using example in `lib.rs`, hence the failing target builds. sigh.

jamwaffles:
Getting closer! I'm now stuck on trying to pull in e-g-simulator as a dev dependency of e-g to build the simulator example in e-g's `lib.rs`. I added this to `embedded-graphics/Cargo.toml`:

```toml
[target.'cfg(target_arch = "x86_64")'.dev-dependencies]
embedded-graphics-simulator = "0.2.0"
```

however got bit by rust-lang/rust#7915.

We could switch to using `MockDisplay` in the example and keep the simulator screenshot and the instructions on how to run the sim examples. The simulator's `lib.rs` could then contain the complete, runnable example that uses the simulator.

jamwaffles:
Alright mint, this is ready for review. I decided to use `MockDisplay` in the e-g readme as it's the least faffy way of making this whole thing work. I could've kept the simulator example and done `rust,ignore` but that defeats the purpose of this PR I think.

embedded-graphics/README.tpl
```diff
@@ -0,0 +1,82 @@
...
+</p>
+
+# Embedded graphics
+
+{{readme}}
+
+## Generating readmes
+
+The various `README.md` files in this project are generated from each crate's `lib.rs` comment. To
+regenerate a readme, ensure [`cargo-readme`](https://crates.io/crates/cargo-readme) is installed
```
bugadani, line 18: Isn't it possible to include cargo-readme in the dev-dependencies instead?

embedded-graphics/README.md
```diff
@@ -11,188 +11,250 @@
...
-- [Simple layout/alignment functions - `embedded-layout`](https://crates.io/crates/embedded-layout)
+[simulator]: https://github.com/jamwaffles/embedded-graphics/tree/c4f74c12dae9f0a0193fa48192f905a002bf8c9d/simulator
+[simulator examples]: https://github.com/jamwaffles/embedded-graphics/tree/c4f74c12dae9f0a0193fa48192f905a002bf8c9d/simulator/examples
 
-Note that some of these crates may not support the latest version of embedded-graphics.
+## Crate features
 
-If you know of a crate that is not in this list, please [open an
-issue](https://github.com/jamwaffles/embedded-graphics/issues/new).
+Add these to your `Cargo.toml` to turn on extra bits of functionality.
```
rfuest, line 109: IMO this should be more generic, because I wouldn't consider `fixed_point` an "extra bit of functionality".

embedded-graphics/README.md
```diff
@@ -11,188 +11,250 @@
...
+* `fixed_point` - use fixed point arithmetic instead of floating point for all trigonometric
+calculation.
 
-Note that some drivers may not support the latest version of embedded-graphics.
+## Migrating from 0.5 to 0.6
+
+Please read [the migration guide](https://github.com/jamwaffles/embedded-graphics/blob/master/embedded-graphics/MIGRATING-0.5-0.6.md).
 
-There may be other drivers out there we don't know about yet. If you know of a driver to add to this list, please open [an issue](https://github.com/jamwaffles/embedded-graphics/issues/new)!
+## Implementing `embedded_graphics` in a driver
```
rfuest, line 121: Is "Implementing `embedded_graphics`..." right in this context? Because you implement the driver API and not e-g itself.

embedded-graphics/README.tpl
jamwaffles: Possibly, but then how would it run from a bash script? Can you install crates with binaries into `target/` or something?

bugadani: Right, I got confused for a bit, sorry about that ^^

embedded-graphics/README.md
```diff
@@ -11,188 +11,250 @@
...
+* [ssd1675](https://crates.io/crates/ssd1675): Rust driver for the Solomon Systech SSD1675 e-Paper display (EPD) controller
+* [st7735-lcd](https://crates.io/crates/st7735-lcd): Rust library for displays using the ST7735 driver
+* [st7789](https://crates.io/crates/st7789): A Rust driver library for ST7789 displays
+* [st7920](https://crates.io/crates/st7920): ST7920 LCD driver in Rust
 
-[![Collage of drawing examples](https://raw.githubusercontent.com/jamwaffles/embedded-graphics/master/doc/assets/all_drawing_ops.png)](https://github.com/jamwaffles/embedded-graphics/blob/master/doc/drawing-examples.md)
+## Simulator
 
-Example usage of drawing primitives, text and images with embedded-graphics can be found [here](https://github.com/jamwaffles/embedded-graphics/blob/master/doc/drawing-examples.md).
+Embedded graphics comes with a [simulator]!
```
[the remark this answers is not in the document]
jamwaffles: I copied the first sentence as suggested, but reworded it to be a bit more professional. Once you're happy with it, I can backport it to the simulator description if you think that's a good idea.

embedded-graphics/src/lib.rs
```diff
@@ -1,5 +1,12 @@
-//! This crate aims to make drawing 2D graphics primitives super easy. It currently supports the
-//! following built in items:
+//! Embedded-graphics is a 2D graphics library that is focused on memory constrained embedded devices.
+//!
+//! A core goal of embedded-graphics is to draw graphics without using any buffers; the crate is
+//! `no_std` compatible and works without a dynamic memory allocator, and without pre-allocating
+//! large chunks of memory. To achieve this, it takes an `Iterator` based approach, where pixel
+//! values and positions are calculated on the fly, with the minimum of saved state. This allows the
```
rfuest, line 6: IMO we should replace "pixel values" by "pixel colors". You normally work with specific types that represent colors and not a raw value.

embedded-graphics/README.tpl
```diff
@@ -0,0 +1,82 @@
...
+    <a href="https://crates.io/crates/embedded-graphics"><img src="https://img.shields.io/crates/v/embedded-graphics.svg" alt="Crates.io"></a>
+    <a href="https://docs.rs/embedded-graphics"><img src="https://docs.rs/embedded-graphics/badge.svg" alt="Docs.rs"></a>
+    <a href="https://matrix.to/#/#rust-embedded-graphics:matrix.org"><img src="https://img.shields.io/matrix/rust-embedded-graphics:matrix.org" alt="embedded-graphics on Matrix"></a>
+</p>
+
+# Embedded graphics
+
+{{readme}}
+
+## Generating readmes
```
rfuest, line 15: Do we need to include this in the user facing readme file? I would suggest to move everything that isn't directly related to using the library to a new file in the `doc` directory and only add a link to that file to the readme.

jamwaffles:
Apologies for the force push. I've pushed some fixes in response to your comments. I'm hands-off this branch until you're done with your review :raised_hands:

embedded-graphics/README.tpl
jamwaffles: No problem! The build scripts/environment is pretty involved for e-g :grimacing:

embedded-graphics/src/lib.rs
```diff
@@ -66,12 +74,14 @@
 //!
 //! # Simulator
 //!
-//! Embedded graphics comes with a [simulator]!
+//! Embedded graphics comes with a [simulator]! The simulator can be used to test and debug
+//! embedded graphics code, or produce examples and interactive demos to show of embedded graphics
+//! features.
```
rfuest, line 79: Much better than the version in the simulator readme. It sounds like the primary purpose of the simulator is debugging embedded-graphics itself, but I think we should emphasize that it can also be used to do the same for external libraries or applications.

embedded-graphics/src/lib.rs
```diff
@@ -81,23 +91,27 @@
...
 //! calculation.
 //!
-//! # Implementing `embedded_graphics` in a driver
+//! # Migrating from 0.5 to 0.6
+//!
+//! Please read [the migration guide](https://github.com/jamwaffles/embedded-graphics/blob/master/embedded-graphics/MIGRATING-0.5-0.6.md).
 //!
-//! To add support for embedded_graphics to a display driver, [`DrawTarget`] should be implemented.
-//! This allows all embedded_graphics objects to be rendered by the display. See the [`DrawTarget`]
+//! # Implementing `embedded_graphics` supprot for a driver
```
rfuest, line 111:
```suggestion
//! # Implementing `embedded_graphics` support for a display driver
```

embedded-graphics/src/lib.rs
```diff
@@ -81,23 +91,27 @@
...
+//!
+//! To add support for embedded-graphics to a display driver, [`DrawTarget`] should be implemented.
```
rfuest, line 113:
```suggestion
//! To add support for embedded-graphics to a display driver, [`DrawTarget`] needs to be implemented.
```

embedded-graphics/src/lib.rs
```diff
@@ -81,23 +91,27 @@
...
+//! This allows all embedded-graphics objects to be rendered by the display. See the [`DrawTarget`]
```
rfuest, line 114: Should we use a consistent way to refer to "embedded-graphics objects" in the docs? The list on the top of the readme uses "items" instead. Or we could use "drawables", because of the `Drawable` trait.

simulator/README.tpl
```diff
@@ -0,0 +1,24 @@
+# Embedded graphics simulator
+
+[![Build Status](https://circleci.com/gh/jamwaffles/embedded-graphics/tree/master.svg?style=shield)](https://circleci.com/gh/jamwaffles/embedded-graphics/tree/master)
+[![Crates.io](https://img.shields.io/crates/v/embedded-graphics-simulator.svg)](https://crates.io/crates/embedded-graphics-simulator)
+[![Docs.rs](https://docs.rs/embedded-graphics-simulator/badge.svg)](https://docs.rs/embedded-graphics-simulator)
```
rfuest, line 5: Should we also link the matrix channel in the simulator and the tiny crates?

embedded-graphics/src/lib.rs
jamwaffles: I've backported this text to the simulator as well :+1:

embedded-graphics/README.tpl
jamwaffles: Good point. I've moved it into `doc/`, along with the development setup section.

embedded-graphics/src/lib.rs
jamwaffles: I've kind of standardised on `items`, but I'd prefer not to go through all the readmes/docs in this PR to standardise that wording. I've changed this case in particular from `objects` to `items` though.

doc/development-setup.md
```diff
@@ -0,0 +1,48 @@
+# Development setup
+
+## Minimum supported Rust version
+
+The minimum supported Rust version for embedded-graphics is `1.40.0` or greater.
```
rfuest, line 5: The MSRV is also important for user of the library, so it should be included in the main readme.

rfuest:
The new GitHub layout has messed with the display of the simulator screenshot in the main readme. It's not scaled 1:1 anymore and the clock shows a moiré pattern (but only in the main readme not the simulator readme). IMO we should reduce the width of the image to avoid scaling and also remove the images of real displays from the screenshot collage.

Do you want to include this in this PR? Or should I create a new PR after this is merged?

jamwaffles:
Wel spotted. I'll get it fixed in this PR.

jamwaffles:
I fixed the screenshot. Looks like this (view in a new tab):

image

No Moire no moire ;)

rfuest:
> No Moire no moire ;)

Great, that looks much nicer! Did you consider to remove the photos, like I suggested in my previous comment? IMO they don't belong in that screenshot and are already included in the main readme anyway.

jamwaffles:
Good idea, but we need some extra screenshots to fill the whitespace and balance the image then - the current images won't look that nice. I'd suggest we do this in a new PR as we need to decide on what screenshots to use and how to lay them out.

jamwaffles:
Thanks for the review!

I opened #402 to track the simulator image.

PR #383, opened 2020-07-02
Rename intoiterator

jamwaffles:

## PR description

Inspired by https://github.com/jamwaffles/embedded-graphics/pull/365#issuecomment-644170782, this is a separate PR to change the use of `IntoIterator` trait impls to use a custom trait.

To that end, I've added the `PixelIterator` and `SparsePixelIterator` traits. The latter would be used in #365 and not this PR, but I think it's worth defining what both these traits should look like in the same PR.

Only `Rectangle` is ported so far to keep the diff clean at first so we can get the design sorted before I port all the primitives, fonts, image, etc.

It would also be good for backwards compatibility if we could `impl IntoIterator for T where T: PixelIterator`. I gave this a very quick try but the compiler doesn't like it. Before I spend more time on it, is it worth pursuing?

The current state as of opening is just something to get the conversation started.

### DIFF
--- a/embedded-graphics/src/image/mod.rs
+++ b/embedded-graphics/src/image/mod.rs
@@ -220,7 +220,7 @@ where
 impl<'a, 'b: 'a, I, C> IntoPixels<C> for &'a Image<'b, I, C>
 where
     &'b I: IntoPixelIter<C>,
-    C: PixelColor + From<<C as PixelColor>::Raw> + 'a,
+    C: PixelColor + From<<C as PixelColor>::Raw>,
 {
     type Iter = ImageIterator<'a, 'b, I, C>;
 
--- a/embedded-graphics/src/image/mod.rs
+++ b/embedded-graphics/src/image/mod.rs
@@ -297,7 +297,7 @@ mod tests {
             Pixel(Point::new(1, 1), Gray8::new(0xaa)),
         ];
 
-        assert!(image.into_pixels().eq(expected.iter().copied()));
+        assert!(image.pixels().eq(expected.iter().copied()));
     }
 
     #[test]
--- a/embedded-graphics/src/image/mod.rs
+++ b/embedded-graphics/src/image/mod.rs
@@ -51,6 +51,7 @@ use crate::{
     draw_target::DrawTarget,
     drawable::{Drawable, Pixel},
     geometry::{Dimensions, Point, Size},
+    pixel_iterator::Pixels,
     pixelcolor::PixelColor,
     primitives::Rectangle,
     transform::Transform,
@@ -200,7 +201,7 @@ where
     where
         D: DrawTarget<Color = C>,
     {
-        display.fill_contiguous(&self.bounding_box(), self.into_iter().map(|p| p.1))
+        display.fill_contiguous(&self.bounding_box(), self.pixels().map(|p| p.1))
     }
 }
 
@@ -216,15 +217,14 @@ where
     }
 }
 
-impl<'a, 'b, I, C> IntoIterator for &'a Image<'b, I, C>
+impl<'a, 'b: 'a, I, C> Pixels<'a, C> for Image<'b, I, C>
 where
     &'b I: IntoPixelIter<C>,
-    C: PixelColor + From<<C as PixelColor>::Raw>,
+    C: PixelColor + From<<C as PixelColor>::Raw> + 'a,
 {
-    type Item = Pixel<C>;
-    type IntoIter = ImageIterator<'a, 'b, I, C>;
+    type Iter = ImageIterator<'a, 'b, I, C>;
 
-    fn into_iter(self) -> Self::IntoIter {
+    fn pixels(&'a self) -> Self::Iter {
         ImageIterator {
             it: self.image_data.pixel_iter(),
             image: self,
@@ -296,7 +296,7 @@ mod tests {
             Pixel(Point::new(1, 1), Gray8::new(0xaa)),
         ];
 
-        assert!(image.into_iter().eq(expected.iter().copied()));
+        assert!(image.pixels().eq(expected.iter().copied()));
     }
 
     #[test]
--- a/embedded-graphics/src/image/mod.rs
+++ b/embedded-graphics/src/image/mod.rs
@@ -51,7 +51,7 @@ use crate::{
     draw_target::DrawTarget,
     drawable::{Drawable, Pixel},
     geometry::{Dimensions, Point, Size},
-    pixel_iterator::{IntoPixels, Pixels},
+    pixel_iterator::IntoPixels,
     pixelcolor::PixelColor,
     primitives::Rectangle,
     transform::Transform,
@@ -201,7 +201,7 @@ where
     where
         D: DrawTarget<Color = C>,
     {
-        display.fill_contiguous(&self.bounding_box(), self.pixels().map(|p| p.1))
+        display.fill_contiguous(&self.bounding_box(), self.into_pixels().map(|p| p.1))
     }
 }
 
@@ -272,7 +272,6 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::pixel_iterator::Pixels;
     use crate::pixelcolor::{BinaryColor, Gray8, GrayColor};
 
     #[test]
@@ -297,7 +296,7 @@ mod tests {
             Pixel(Point::new(1, 1), Gray8::new(0xaa)),
         ];
 
-        assert!(image.pixels().eq(expected.iter().copied()));
+        assert!(image.into_pixels().eq(expected.iter().copied()));
     }
 
     #[test]
--- a/embedded-graphics/src/image/mod.rs
+++ b/embedded-graphics/src/image/mod.rs
@@ -51,7 +51,7 @@ use crate::{
     draw_target::DrawTarget,
     drawable::{Drawable, Pixel},
     geometry::{Dimensions, Point, Size},
-    pixel_iterator::Pixels,
+    pixel_iterator::{IntoPixels, Pixels},
     pixelcolor::PixelColor,
     primitives::Rectangle,
     transform::Transform,
@@ -217,14 +217,14 @@ where
     }
 }
 
-impl<'a, 'b: 'a, I, C> Pixels<'a, C> for Image<'b, I, C>
+impl<'a, 'b: 'a, I, C> IntoPixels<C> for &'a Image<'b, I, C>
 where
     &'b I: IntoPixelIter<C>,
     C: PixelColor + From<<C as PixelColor>::Raw> + 'a,
 {
     type Iter = ImageIterator<'a, 'b, I, C>;
 
-    fn pixels(&'a self) -> Self::Iter {
+    fn into_pixels(self) -> Self::Iter {
         ImageIterator {
             it: self.image_data.pixel_iter(),
             image: self,
@@ -272,6 +272,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::pixel_iterator::Pixels;
     use crate::pixelcolor::{BinaryColor, Gray8, GrayColor};
 
     #[test]
@@ -296,7 +297,7 @@ mod tests {
             Pixel(Point::new(1, 1), Gray8::new(0xaa)),
         ];
 
-        assert!(image.pixels().eq(expected.iter().copied()));
+        assert!(image.into_pixels().eq(expected.iter().copied()));
     }
 
     #[test]
--- a/embedded-graphics/src/image/mod.rs
+++ b/embedded-graphics/src/image/mod.rs
@@ -217,7 +217,7 @@ where
     }
 }
 
-impl<'a, 'b: 'a, I, C> IntoPixels for &'a Image<'b, I, C>
+impl<'a, 'b, I, C> IntoPixels for &'a Image<'b, I, C>
 where
     &'b I: IntoPixelIter<C>,
     C: PixelColor + From<<C as PixelColor>::Raw>,
--- a/embedded-graphics/src/image/mod.rs
+++ b/embedded-graphics/src/image/mod.rs
@@ -217,11 +217,13 @@ where
     }
 }
 
-impl<'a, 'b: 'a, I, C> IntoPixels<C> for &'a Image<'b, I, C>
+impl<'a, 'b: 'a, I, C> IntoPixels for &'a Image<'b, I, C>
 where
     &'b I: IntoPixelIter<C>,
     C: PixelColor + From<<C as PixelColor>::Raw>,
 {
+    type Color = C;
+
     type Iter = ImageIterator<'a, 'b, I, C>;
 
     fn into_pixels(self) -> Self::Iter {
--- a/embedded-graphics/src/pixel_iterator.rs
+++ b/embedded-graphics/src/pixel_iterator.rs
@@ -29,25 +29,25 @@ where
 }
 
 /// Pixel iterator trait
-pub trait IntoPixelIterator<C>
+pub trait Pixels<'a, C>
 where
     C: PixelColor,
 {
     /// TODO: Doc
-    type Iter: Iterator<Item = Pixel<C>>;
+    type Iter: Iterator<Item = Pixel<C>> + 'a;
 
     /// TODO: Doc
-    fn pixels(self) -> Self::Iter;
+    fn pixels(&'a self) -> Self::Iter;
 }
 
 /// Sparse pixel iterator
-pub trait IntoSparsePixelIterator<C>
+pub trait SparsePixels<'a, C>
 where
     C: PixelColor,
 {
     ///  TODO: Doc
-    type Iter: Iterator<Item = Option<C>> + Dimensions;
+    type Iter: Iterator<Item = Option<C>> + Dimensions + 'a;
 
     /// TODO: Doc
-    fn sparse_pixels(self) -> Self::Iter;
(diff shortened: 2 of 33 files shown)


rfuest:
> It would also be good for backwards compatibility if we could `impl IntoIterator for T where T: PixelIterator`. I gave this a very quick try but the compiler doesn't like it. Before I spend more time on it, is it worth pursuing?

I don't think it's possible to implement this until specialization in rust is stabilized. And TBH I don't think we should implement this anyway. It's redundant to have both and it wouldn't allow users of the library, who implement a drawable, to use `IntoIterator` for other purposes.

jamwaffles:
Yeah I got the feeling it wasn't possible either. Thought I'd ask in case there's something I missed. Ignore the failing CI for now... not sure what that doc link check is on about...

rfuest:
> Ignore the failing CI for now...

Unrelated question regarding CI: Is there a setting in CircleCI to allow anonymous viewing of the build results? This used to be possible, but now you need to connect your GitHub account in order to view the results.

jamwaffles:
I had a look and couldn't find anything. Maybe they changed it recently? That's a bit disappointing if they have :(

rfuest:
It has been like this for some time now. Their default "Log In with GitHub" button requests access to all repos, including private ones. There is a only public repos option, but having to grant them access to my account just to view the results from another repo still shouldn't be necessary.

jamwaffles:
Yeah that's not good. I used up my last superpower and sent them a tweet which will hopefully just tell me I'm missing a setting somewhere.

I don't really want to change CI providers, but if it's stopping you from contributing we can look at GH Actions or migrate back to Travis.

rfuest:
> I don't really want to change CI providers, but if it's stopping you from contributing we can look at GH Actions or migrate back to Travis.

No, it won't stop me from contributing and I will log in if necessary. I just wanted to mention it because it's also a problem for one time contributors who can no longer easily check why their builds are failing.

jamwaffles:
Yeah, not a great user experience I agree. Anyway, let me know what you think of the PR itself when you've got some time :)

rfuest:
IMO `PixelIterator` isn't the right name for the trait. The object it's implemented for isn't the iterator itself, but can be converted into a pixel iterator, which should be reflected in the name. We could use `IntoPixelIterator` or something completely different.

> ...when you've got some time :)

That's a problem at the moment. I don't have much time to work on e-g at the moment and will also be on holiday in the next week. But I hope to find some time to work through the issues and PRs that need feedback after I'm back.

embedded-graphics/src/pixel_iterator.rs
```diff
@@ -0,0 +1,29 @@
...
+pub trait SparsePixelIterator<C>
+where
+    C: PixelColor,
+{
+    ///  TODO: Doc
+    type Iter: Iterator<Item = Option<C>>;
+
+    /// TODO: Doc
+    fn sparse_pixels(self) -> (Rectangle, Self::Iter);
+}
```
rfuest, line 29: Would this work?

```suggestion
pub trait SparsePixelIterator<C>
where
    C: PixelColor,
{
    ///  TODO: Doc
    type Iter: Iterator<Item = Option<C>> + Dimensions;

    /// TODO: Doc
    fn sparse_pixels(self) -> Self::Iter;
}
```

jamwaffles: Ah yeah that's a much better idea. We'd have to `impl Dimensions` for all the styled iterators which isn't difficult, just a bit of boilerplate.

jamwaffles:
> We could use IntoPixelIterator or something completely different.

`IntoPixelIterator` makes sense to me so I'll use that for now. Should we change the method names to have a `into_` prefix like the `IntoIter` trait? e.g. `into_pixels()`, etc.

> I don't have much time to work on e-g at the moment

No problem! I hope you have a good week off :)

rfuest:
> `IntoPixelIterator` makes sense to me so I'll use that for now. Should we change the method names to have a `into_` prefix like the `IntoIter` trait? e.g. `into_pixels()`, etc.

`into_pixels` would be the "correct" way of naming the method inside `IntoPixelIterator`. But we only implement `IntoPixelIterator` for references. This means that code like `styled.into_pixels()` doesn't actually convert the styled into an iterator and the styled can be reused later, which is unexpected when only considering the names IMO.

I previously considered to change the traits to always work on references instead of having to implement them for references:
```rust
pub trait Pixels<C> //needs a name that doesn't conflict with the `Pixels`  structs.
where
    C: PixelColor,
{
    type Iter: Iterator<Item = Pixel<C>>;
    fn pixels(&self) -> Self::Iter;
}
```
This would solve the naming problem, but would break iterators that borrow from `self` instead of copying all necessary data.

But this might work:
```rust
pub trait Pixels<'a, C>
where
    C: PixelColor,
{
    type Iter: Iterator<Item = Pixel<C>> + 'a;
    fn pixels(&'a self) -> Self::Iter;
}
```
```rust
impl<C> IntoPixelIterator<C> for &Styled<Rectangle, PrimitiveStyle<C>>
where
    C: PixelColor,
// gets replaced by
impl<C> Pixels<'_, C> for Styled<Rectangle, PrimitiveStyle<C>>
where
    C: PixelColor,
```
```rust
impl<'a, 'b, I, C> IntoIterator for &'a Image<'b, I, C>
where
    &'b I: IntoPixelIter<C>,
    C: PixelColor + From<<C as PixelColor>::Raw>,
//gets replaced by
impl<'a, 'b, I, C> Pixels<'a, C> for Image<'b, I, C>
where
    &'b I: IntoPixelIter<C>,
    C: PixelColor + From<<C as PixelColor>::Raw>,
```

> No problem! I hope you have a good week off :)

Thanks!

jamwaffles:
I applied those suggestions to `Rectangle` and `Image`. It's a bit of a lifetime soup now but doesn't look too bad.

jamwaffles:
Hey @rfuest could you give this a quick once over before I make the mechanical changes to all the other primitives please? I'd prefer not to go through the whole project more than once :grin:

rfuest:
I did hope that this change would make it easier to implement `pixels`, especially when the iterator doesn't need a reference to the original struct. But the code new code is pretty much the same complexity IMO and I'm ambivalent which one I prefer. Do you have any preference?

rfuest:
One solution might be to provide both `pixels` for `T` and `into_pixels` for `&T`. Similar to `vec.iter()` and `(&vec).into_iter()`.

rfuest:
On a related topic, which could perhaps have some impact on this PR: What do you think about changing `Drawable::draw` to use a reference to `self`? I consider a drawable to be something that can be drawn to a display and shouldn't be changed in the process, which means a reference is more appropriate. But this means that a pixel iterator can no longer be a `Drawable`, but IMO that's OK because it can still be drawn by `display.draw_iter` or an extension trait.

The description might not be very clear, but I've implemented a first draft of the changes which should make it easier to follow: https://github.com/rfuest/embedded-graphics/commit/f7a1af832a06427c414f8b9b83fdc8d1dc333a47

jamwaffles:
> What do you think about changing Drawable::draw to use a reference to self? I consider a drawable to be something that can be drawn to a display and shouldn't be changed in the process, which means a reference is more appropriate.

I agree completely. I'm unsure why it currently takes `self`. 

> But this means that a pixel iterator can no longer be a `Drawable`, but IMO that's OK because it can still be drawn by `display.draw_iter` or an extension trait.

Hm. I think that's probably ok. I saw the `PixelIteratorExt` trait in your linked commit which should work ok. 

Can you prepare a PR for the `Drawable` changes? I think that can happen separately from the iterator discussion here.

jamwaffles:
So if I understand correctly, your proposal looks about like this (not compiled or tested but you get the gist):

```rust
pub trait Pixels<C>
where
    C: PixelColor,
{
    type Iter: Iterator<Item = Pixel<C>>;

    fn pixels(&self) -> Self::Iter;
}

pub trait IntoPixels<C>
where
    C: PixelColor,
{
    type Iter: Iterator<Item = Pixel<C>>;

    fn into_pixels(self) -> Self::Iter;
}

struct Thing {}

impl<C> Pixels<C> for Thing {
    type Iter = ThingIter;

    fn pixels(&self) -> Self::Iter {
        // ...
    }
}

impl<C> IntoPixels<C> for &Thing {
    type Iter = ThingIter;

    fn into_pixels(self) -> Self::Iter {
        // ...
    }
}
```

rfuest:
> I agree completely. I'm unsure why it currently takes `self`.

It takes `self` to be able to implement it for `&T` and `&mut T`, which is only required for iterators.

> Can you prepare a PR for the `Drawable` changes? I think that can happen separately from the iterator discussion here.

It's currently based on this PR, but it shouldn't be too hard to rebase onto master. Do you want to merge this PR first or should I prepare the other PR first?

rfuest:
> So if I understand correctly, your proposal looks about like this (not compiled or tested but you get the gist):

It wasn't a well thought out proposal yet, but I think `Pixels` could be implemented for all `T where &T: IntoPixels`. Which would mean that `Pixels` would mainly exist to rename the method `into_pixels` into `pixels`, but this would make it consistent with `Vec` and other examples in the standard library.

jamwaffles:
> Do you want to merge this PR first or should I prepare the other PR first? 

Can you prepare yours first? I think it might be a little less work, no?

> Which would mean that Pixels would mainly exist to rename the method into_pixels into pixels, but this would make it consistent with Vec and other examples in the standard library.

Makes sense to me

rfuest:
> Can you prepare yours first? I think it might be a little less work, no?

Not sure if it's less or more work, but I've created #393.

jamwaffles:
I feel like I've misunderstood your comments, but it's pretty late here so I don't know if I really am or not. Can you take a look at the latest commit (81a4affaa1eb3c5da7aa4e6fb504eb2e0c2acb3c) and see if that's going in the right direction? I had to rename calls to `.pixels()` to `.into_pixels()` which doesn't feel right.

jamwaffles:
Oop, I had a sleep and now I think I've got a much better solution in 6b4832c2b45b64b78ade1060b62d95e03eaf2e50. I still haven't ported the other primitives yet - only `Rectangle` and `Image` - but they should be pretty easy if this is the way to go.

I also added a sanity check [here](https://github.com/jamwaffles/embedded-graphics/pull/383/commits/6b4832c2b45b64b78ade1060b62d95e03eaf2e50#diff-72ec886f963e59e83ae167047455429dR360-R372) which is a bit rubbish so suggestions on a better regression check for this would be appreciated.

rfuest:
After the last commit `Pixels` and `IntoPixels` are exactly the same, apart from the method name. I was thinking about something like this:
```rust
pub trait Pixels<'a, C>
where
    C: PixelColor + 'a,
{
    type Iter: Iterator<Item = Pixel<C>> + 'a;
    fn pixels(&'a self) -> Self::Iter;
}

impl<'a, C, T> Pixels<'a, C> for T
where
    C: PixelColor + 'a,
    &'a T: IntoPixels<C> + 'a,
    <&'a T as IntoPixels<C>>::Iter: Iterator<Item = Pixel<C>> + 'a,
{
    type Iter = <&'a T as IntoPixels<C>>::Iter;

    fn pixels(&'a self) -> Self::Iter {
        self.into_pixels()
    }
}
```

This would work better once GATs are stabilized. Maybe we should consider if we should add this at all. I thought that it would be a good idea to provide the `pixels` method, but I'm not sure anymore if we gain much by adding it. `into_pixels` will probably not be called very often in user code and if it's documented that `IntoPixels` is implemented for references it shouldn't be too confusing. I would suggest to only add `IntoPixels` now and see how it turns out. `Pixels` can be added later if we this it's necessary.

jamwaffles:
> I would suggest to only add IntoPixels now and see how it turns out. 

Just to make sure, this would replace `into_iter()` calls with `into_pixels()` and will be implemented on `&'a Styled<...>`, not `Styled<...>` right? The goal being not to load `into_iter()` with too much behaviour, right?

rfuest:
> Just to make sure, this would replace `into_iter()` calls with `into_pixels()` and will be implemented on `&'a Styled<...>`, not `Styled<...>` right? The goal being not to load `into_iter()` with too much behaviour, right?

Yes. Sorry if I did make this more confusing by the whole `pixels` vs `into_pixels` stuff. I should have thought this through before commenting on it.

jamwaffles:
Alright, another attempt pushed. Hopefully this is what you meant in your comments :)

embedded-graphics/src/image/mod.rs
```diff
@@ -216,15 +217,14 @@ where
     }
 }
 
-impl<'a, 'b, I, C> IntoIterator for &'a Image<'b, I, C>
+impl<'a, 'b: 'a, I, C> IntoPixels<C> for &'a Image<'b, I, C>
 where
     &'b I: IntoPixelIter<C>,
-    C: PixelColor + From<<C as PixelColor>::Raw>,
+    C: PixelColor + From<<C as PixelColor>::Raw> + 'a,
```
rfuest, line 223: The additional lifetime bound shouldn't be required anymore.

embedded-graphics/src/primitives/rectangle/styled.rs
```diff
@@ -73,15 +74,14 @@ where
     }
 }
 
-impl<C> IntoIterator for &Styled<Rectangle, PrimitiveStyle<C>>
+impl<'a, C> IntoPixels<C> for &'a Styled<Rectangle, PrimitiveStyle<C>>
 where
-    C: PixelColor,
+    C: PixelColor + 'a,
```
rfuest, line 79: It shouldn't be necessary anymore to explicitly specify the lifetime.

jamwaffles:
Good point on the lifetimes - fixed.

If you're good with this approach, I'll make all the changes required throughout the codebase tomorrow.

Besides being exported in the prelude, `IntoPixels` should rarely, if ever, be used outside of e-g, right? Should we make it a sealed trait? Are there any things you think should be put in the documentation for it before I start writing some?

embedded-graphics/src/pixel_iterator.rs
```diff
@@ -26,3 +26,28 @@ where
         target.draw_iter(self)
     }
 }
+
+///  TODO: Doc
+pub trait IntoPixels<C>
```
rfuest, line 31: I just realized that if would probably better to use an associated type for `C` similar to `IntoIterator::Item`.

rfuest:
> Besides being exported in the prelude, `IntoPixels` should rarely, if ever, be used outside of e-g, right?

Should we also change `DrawTarget::draw_iter` to use `IntoPixels` instead of `IntoIterator`? This would require external implementations of `IntoPixels`.
IMO we shouldn't and I would even prefer to implement the methods in `DrawTarget` for `&mut I : Iterator` instead of `IntoIterator`. This will require an additional `into_pixels()` or `into_iter()` when calling `draw_iter`, but IMO it's better to create the iterator explicitly in this case. And because `draw_iter` shouldn't normally called by a user anyway the API won't get more complicated.

> Should we make it a sealed trait?

I don't think that's necessary. Are there any negative effects of an external implementation? If we decide to change the trait later it will be a breaking change anyway, not only for implementers of the trait but also for users.

> Are there any things you think should be put in the documentation for it before I start writing some?

I can't think of anything we didn't discuss already at the moment. It should be mentioned that e.g. `styled.into_pixel()` doesn't "consume" `styled` because `IntoPixels` is implemented for references and that `IntoPixels` is similar to `IntoIterator`.

jamwaffles:
> IMO we shouldn't

I agree. Leaving it as-is allows end users to pass in their own iterators instead of having to `impl IntoPixels`.

> I would even prefer to implement the methods in DrawTarget for &mut I : Iterator instead of IntoIterator

IMO Accepting `IntoIterator` is the better choice as it allows the programmer to not have to be explicit, mirroring the `for` loop's behaviour of transforming its argument into an iterator. I can see the benefits of being explicit about this though, but will it cause us problems down the line if we leave it as `IntoIterator`? Are there other strong reasons to change the interface beyond it being more explicit?

> I don't think that's necessary. Are there any negative effects of an external implementation?

No, I was thinking that if it's sealed, it's less API surface to support but you're right, it's not really an issue.

rfuest:
> I can see the benefits of being explicit about this though, but will it cause us problems down the line if we leave it as IntoIterator? Are there other strong reasons to change the interface beyond it being more explicit?

OK, you convinced my to leave it as is. It just felt like it might be worth considering because we no longer implement `IntoIterator` out self and won't have many benefits of being more generic.

> No, I was thinking that if it's sealed, it's less API surface to support but you're right, it's not really an issue.

I also considered to replace `image::IntoPixelIter` by the new `IntoPixels` which are essentially the same trait now. And this would need to be implemented externally, e.g. by `tinybmp` and `tinytga`.

jamwaffles:
> I also considered to replace `image::IntoPixelIter` by the new `IntoPixels` which are essentially the same trait now. And this would need to be implemented externally, e.g. by `tinybmp` and `tinytga`.

That's a really good idea. I tried an impl, but ran into type recursion issues as `Image<I>` ends up being implemented for any `I` that impls `IntoPixels` at that point... which also includes `Image`. I could add an extra trait that breaks the recursion cycle, or leave `Image` alone with its `IntoPixelIter` for now.

rfuest:
> ...or leave `Image` alone with its `IntoPixelIter` for now.

OK, I think that's the best idea for this PR. I can look into this when I implement `SubImage` and change the `ImageDimension` trait.

jamwaffles:
Makes sense to me :). In that case, I think I'm done with my changes so this is ready for a review.

PR #386, opened 2020-07-09
Thick polylines

jamwaffles:

## PR description

Adds thick polyline support by computing line joints for each node of the polyline. These are mitred if the mitre isn't too long, in which case they are converted to bevelled corners.

The rasterisation is done using a scanline algorithm, with each line being drawn with a call to `DrawTarget::fill_solid` for a potential performance boost, although each line will usually be pretty short for small stroke widths.

Stroke alignment is currently unsupported - but do we want to support it at all? I'm expecting the answer to be `yes` for polygons (closed shapes as opposed to the open shapes made with `Polyline`s) but I'll tackle that in a future PR. 

This PR does introduce support for stroke alignment where `Inside` is right-aligned an `Outside` is left-aligned to work with clockwise polygones, however it is unused for this PR. I've hardcoded it to `StrokeAlignment::Center` in a couple of places. I'd like to leave this code in as we'll very likely need to use it when supporting thick triangles and thick polygons which will support stroke alignment.


embedded-graphics/src/lib.rs
```diff
@@ -242,7 +242,7 @@
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/jamwaffles/embedded-graphics/191fe7f8a0fedc713f9722b9dc59208dacadee7e/assets/logo.svg?sanitize=true"
 )]
-#![no_std]
+// #![no_std]
```
bugadani, line 245: Just a friendly reminder to uncomment this

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -62,6 +70,47 @@ impl Dimensions for Line {
     }
 }
 
+/// Check signs of two signed numbers
+///
+/// Fastest ASM output compared to other methods. See: https://godbolt.org/z/zVx9cD
+fn same_signs(a: i32, b: i32) -> bool {
+    a ^ b >= 0
```
bugadani, line 77: While longer, maybe (a >= 0) == (b >= 0) is easier to understand?

https://godbolt.org/z/hsEqPv

embedded-graphics/src/lib.rs
jamwaffles: Thanks! Even with this comment I'm 99% guaranteed to forget lol

embedded-graphics/src/primitives/line/mod.rs
jamwaffles: Good shout. It never ceases to amaze me what compilers can optimise down to.

jamwaffles:
A recurring issue/thought I have with my triangulation approach is overdraw along the edges of the triangles that comprise line features. Currently, due to the edge of the triangle being drawn using the Bresenham algorithm, point centers can appear both inside and outside the mathematical true triangle. This means that edges from two triangles that share the same points will contain some pixels that are drawn twice.

There are three options:

1. Accept some overdraw on triangle edges internal to the thick line, but keep the left and right edges of the thick line true to the Bresenham rendering
2. Eliminate overdraw by rendering the mathematically correct triangle, i.e. only rendering points whose centers lie on or inside the triangle edges. This would however make the line slightly thinner as the edges wouldn't be drawn using Bresenham.
3. Go with point (2.) but also draw Bresenham lines on the edges of the thick line to get them to be the right thickness. This would also result in some overdraw.

@rfuest if you can understand this waffle, what would you suggest? I'm leaning towards point 1. at the moment.

rfuest:
I would need to look into this some more to give a good suggestion. But it would be best if we could prevent overdraw. If we start to support alpha blending it would make a difference if a pixel is drawn once or multiple times with a semi transparent color.

jamwaffles:
That's precisely the issue I had in mind. I could go with the non-overdrawing solution for now and fix the outlines in the future. They're not very different from the Bresenham result.

bugadani:
Overdraw is also an issue when somebody uses a custom color that inverts what's on the display.

jamwaffles:
Wouldn't that happen at the framebuffer level where there's only one piece of data per pixel anyway though?

bugadani:
Not necessarily. I've implemented that in a way for the SSD1306 that uses XOR to set the bit in the framebuffer, so if a pixel is drawn multiple times, it flips the bit multiple times.

rfuest:
> I could go with the non-overdrawing solution for now and fix the outlines in the future. They're not very different from the Bresenham result.

Sounds good to me.

jamwaffles:
> Sounds good to me.

Cool, I'll get that impled.

> Not necessarily. I've implemented that in a way for the SSD1306 that uses XOR to set the bit in the framebuffer, so if a pixel is drawn multiple times, it flips the bit multiple times.

Good point, thanks for explaining

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -389,18 +389,25 @@ mod tests {
...
+    //     let triangles = [
+    //         Triangle::new(Point::new(5, 10), Point::new(15, 20), Point::new(10, 15)),
+    //         Triangle::new(Point::new(2, 2), Point::new(2, 4), Point::new(2, 4)),
+    //         Triangle::new(Point::new(2, 2), Point::new(4, 2), Point::new(4, 2)),
+    //     ];
+
+    //     for (idx, triangle) in triangles.iter().enumerate() {
+    //         for point in Rectangle::new(Point::new(-5, -5), Size::new(25, 25)).points() {
+    //             assert_eq!(
+    //                 triangle.contains(point),
```
bugadani, line 404: If I can believe myself at 1:20am, in Triangle::contains() you walk through the edge points of the triangles, and since the triangles are inside the rectangle, basically for every point on their edges you'll get a match. Or I am wrong, which is possible.

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -178,13 +148,119 @@ impl Triangle {
...
+
+    /// Find the center of gravity/centroid of the triangle
+    pub fn centroid(&self) -> Point {
+        (self.p1 + self.p2 + self.p3) / 3
+    }
+
+    /// Point inside triangle, ignoring pixels that partially lie outside triangle lines.
+    pub(self) fn mathematical_contains(&self, point: &Point) -> bool {
+        // Skip expensive calculations below if point is outside the bounding box
+        if !self.bounding_box().contains(*point) {
```
bugadani, line 186: You are skipping the bulk of this function, but not skipping the line walking in Triangle::contains(). Maybe it would be worth to move the bounding box check out so contains don't walk the edges when it doesn't have to?

simulator/examples/polyline-joints.rs
```diff
@@ -40,23 +40,18 @@ fn empty_crosshair(point: Point, color: Rgb888, display: &mut SimulatorDisplay<R
...
-
-    /// Second triangle that forms edge rectangle
-    t2: MathematicalPoints,
-
-    /// Filler triangle (if the current joint style requires it)
-    filler: MathematicalPoints,
     start_idx: usize,
+    t1: Option<Triangle>,
+    t2: Option<Triangle>,
+    filler: Option<Triangle>,
```
bugadani, line 48: Just making a note that this is 84 bytes of memory.

jamwaffles:
Unfortunately there's a little bit of overdraw (white lines/dots):

image

jamwaffles:
image

Triangles aren't immune either :(

A possible solution would be a scanline iterator that treats all the generated triangles as a single polygon instead of filling them individually. I'll do some investigation...

jamwaffles:
Polylines are _so close_ to being good. We're now down to two issues:

1. 2px wide lines are drawn like crap:

polyline-2px

2. There's some minor overdraw (white pixels) for _some_ joints:

overdraw-case1

I'm kind of losing my mind with this now so if someone can suggest a new/better direction to go in that's more robust, I'd be most appreciative.

jamwaffles:
@bugadani I'm pressed for time at the moment so did a very quick/shoddy/broken rebase - please disregard any mistakes you see in this branch. I'll fix those when I have a moment and keep working from there.

bugadani:
Good thing I checked out the branch before the rebase, then 😅

jamwaffles:
So, I'm putting this up for a tentative initial review. It's also pretty late so please forgive any stupid mistakes. Please also see the edited description at the top of the thread.

As far as my unit tests are telling me, there is _no_ overdraw! 

I have some outstanding points:

* Where should `primitives/{line_joint,thick_segment_iter,thick_segment}.rs` really be moved to? Take into account that the thick triangle and `Polygon` will reuse a lot of what's in those files.
* A better example demonstrating polylines and the thickness/joint properties would be great - the one included in this PR right now is more for debugging than showing off.
* What extra docs need to be added?
* What extra tests need to be added to cater for conditions I haven't thought of?
* We've lost ~10% on thick lines, mostly due to the (dormant, but IMO necessary for future work) stroke alignment support added to the Bresenham stuff. I'm ok with this, but if someone wants to benchmark and claw that back, please do. Either way, I'm aware of this perf drop.

rfuest:
Great! I'll try to review it soon. :tada: :champagne:

embedded-graphics/src/primitives/polyline/styled.rs
```diff
@@ -72,19 +116,62 @@ where
...
     }
 }
 
 impl<C> Dimensions for Styled<Polyline<'_>, PrimitiveStyle<C>>
 where
     C: PixelColor,
 {
-    // NOTE: Polyline currently ignores stroke width, so this delegates to the un-styled bounding
-    // box impl.
     fn bounding_box(&self) -> Rectangle {
```
rfuest, line 152: I think this will need to explicitly handle polylines with fewer than two vertices. Otherwise it will return a maximum size rectangle instead of a zero size one.

embedded-graphics/src/primitives/polyline/styled.rs
```diff
@@ -94,51 +181,235 @@ where
...
+            );
+        }
+    }
+
+    #[test]
+    fn degenerate_joint() {
+        let mut display: MockDisplay<BinaryColor> = MockDisplay::new();
+
+        display.set_allow_overdraw(true);
+
```
rfuest, line 346: This isn't required anymore :tada:
```suggestion
```

simulator/examples/polyline-joints.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        for event in window.events() {
+            match event {
+                SimulatorEvent::Quit => break 'running,
+                SimulatorEvent::MouseButtonDown { point, .. } => {
+                    mouse_down = true;
+
+                    end_point = point;
+                }
+                SimulatorEvent::KeyDown { keycode, .. } => match keycode {
```
rfuest, line 73: There seems to be a problem with determining when a bevel join should be used. For some small stroke widths the top vertex is drawn using a bevel join:
grafik

But if the stroke width is increased by one 1px the miter reappears:
grafik

This pattern continues for the next few stroke widths until the vertex stays beveled.

rfuest (changes requested):
A full review of this PR will take a while, but here is some initial feedback.

> Where should `primitives/{line_joint,thick_segment_iter,thick_segment}.rs` really be moved to? Take into account that the thick triangle and `Polygon` will reuse a lot of what's in those files.

Maybe a `common` submodule inside the `primitives` module? I haven't looked at these in detail yet, do they need to be public? Otherwise we could also move them later.

>  We've lost ~10% on thick lines, [...] I'm ok with this, [...]

I agree, we can look into this later. And I'm not sure how accurate the current benchmarks are, because I got some interesting results in the past. Perhaps collecting `Points` into a vector isn't a good representation of actually drawing something, especially now that we've started to use other drawing methods that aren't all iterator based.

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -86,49 +149,216 @@ impl Line {
...
+        if denom == 0 {
+            return Intersection::Colinear;
+        }
+
+        // If we got here, line segments intersect. Compute intersection point using method similar
+        // to that described here: http://paulbourke.net/geometry/pointlineplane/#i2l
+
+        // The denom/2 is to get rounding instead of truncating. It is added or subtracted to the
+        // numerator, depending upon the sign of the numerator.
+        let offset = if denom < 0 { -denom / 2 } else { denom / 2 };
```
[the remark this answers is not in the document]
bugadani: Or a copy-paste error indicated by the comment?

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -86,49 +149,216 @@ impl Line {
...
+        let (start, end) = if self.start.x > self.end.x {
+            (self.end, self.start)
+        } else {
+            (self.start, self.end)
+        };
+
+        Self::new(start, end)
+    }
+
+    /// Get the midpoint of the line.
```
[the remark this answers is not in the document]
bugadani: I'd suggest "Computes".

embedded-graphics/src/primitives/line_joint.rs
```diff
@@ -0,0 +1,305 @@
...
+        mid: Point,
+        end: Point,
+        width: u32,
+        alignment: StrokeAlignment,
+    ) -> Self {
+        let first_line = Line::new(start, mid);
+        let second_line = Line::new(mid, end);
+
+        // Miter length limit is double the line width (but squared to avoid sqrt() costs)
+        let miter_limit = (width * 2).pow(2);
```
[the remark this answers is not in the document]
jamwaffles: Let's add it later :)

embedded-graphics/src/primitives/line_joint.rs
```diff
@@ -0,0 +1,305 @@
+//! Thick line joint.
```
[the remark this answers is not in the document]
jamwaffles: AFAIK I've called it "joint" all the way through, so I'd like to keep it like that if you don't have a strong opinion either way.

embedded-graphics/src/primitives/line/thick_points.rs
```diff
@@ -84,11 +88,14 @@ pub(in crate::primitives::line) struct ParallelsIterator {
 
     /// The next side which will be drawn.
     next_side: Side,
+
+    /// Stroke alignment.
+    alignment: StrokeAlignment,
```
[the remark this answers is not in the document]
jamwaffles: Alignment isn't supported in this PR. I left some comments in the description about this - it'll be required for e.g. thick `Triangle`s and probably `Polygon`s which are closed shapes and, for consistency with the existing closed shapes, should really support stroke alignment.

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -86,49 +149,216 @@ impl Line {
...
+    pub fn midpoint(&self) -> Point {
+        self.start + (self.end - self.start) / 2
+    }
+
+    const fn coefficients(&self, other: &Self) -> (i32, i32, i32, i32, i32, i32, i32) {
```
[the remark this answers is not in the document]
jamwaffles: That would be great. I don't understand the `LinearEquation` code so probably can't do it myself, but it would be a lot cleaner to merge them if possible.

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -66,6 +67,68 @@ impl Dimensions for Line {
...
+    /// Multiple intersection points.
+    ///
+    /// This may be caused by e.g. a shallow line where multiple points lie on the same Y
+    /// coordinate.
+    Line(Line),
+}
+
+impl BresenhamIntersection {
+    /// As a line
+    pub fn as_line(&self) -> Line {
```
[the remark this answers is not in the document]
jamwaffles: Good shout. This might be a bit subjective, but I've gone with `into_line()`. Clippy also suggested making it take `self` not `&self` so I changed that too.

embedded-graphics/src/primitives/line/mod.rs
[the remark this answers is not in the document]
jamwaffles: Nono, this is just me failing a basic programming interview. I've changed to `denom.abs()` like it should be...

embedded-graphics/src/primitives/polyline/styled.rs
jamwaffles: Well spotted. I've added a test case for two and one point with some fixes.

embedded-graphics/src/primitives/polyline/styled.rs
jamwaffles: feelsgoodman.jpeg

embedded-graphics/src/geometry/size.rs
```diff
@@ -187,6 +187,11 @@ impl Size {
         self.saturating_sub(Size::new(1, 1)) / 2
     }
 
+    /// Get the area of this size
+    pub fn area(self) -> u32 {
```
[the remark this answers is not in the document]
jamwaffles: Turns out this method isn't used at all anymore, so I've deleted it completely.

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -86,49 +149,216 @@ impl Line {
...
+                BresenhamIntersection::Point(first)
+            }
+        } else {
+            BresenhamIntersection::Point(first)
+        };
+
+        Some(intersection)
+    }
+
+    /// Get the squared length of the line
```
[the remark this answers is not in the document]
jamwaffles: That's a good idea. I've added it in as a public method.

embedded-graphics/src/primitives/line/thick_points.rs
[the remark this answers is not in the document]
jamwaffles: Internally, yes this should be something else. What would you suggest for the public API though? We could simply never support alignment for lines/polylines and only deal with closed shapes where the inside/outside wording makes sense.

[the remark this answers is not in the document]
jamwaffles: p.s. I also added a test that ensures thick polylines are always drawn centered, regardless of alignment.

simulator/examples/polyline-joints.rs
jamwaffles: I've seen this as well. I think it's an unlucky combination of the joint angle in the demo and its effect on the bevel heuristic. Tbh I'm ok leaving it as is, but do you have a suggestion on how to fix it better?

jamwaffles:
Thanks for the initial pass. Some of the comments were just missed cleanup in terms of `pub` items.

embedded-graphics/src/primitives/line_joint.rs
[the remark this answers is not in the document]
rfuest: OK.

embedded-graphics/src/primitives/line/thick_points.rs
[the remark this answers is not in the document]
rfuest: > What would you suggest for the public API though? We could simply never support alignment for lines/polylines and only deal with closed shapes where the inside/outside wording makes sense.

I think that's what we had decided some time ago.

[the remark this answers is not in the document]
rfuest: > Alignment isn't supported in this PR.

Not for drawing any user visible objects, but if I understand the code correctly it is already supported by `ParallelsIterator`. And that file should have unit tests for the different alignments. But because it isn't used at the moment a simple `TODO: add tests for different alignments` would also be enough IMO.

embedded-graphics/src/primitives/line_joint.rs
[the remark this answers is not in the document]
rfuest: It's your decision, but if you do a google search for "line joint" and "line join" and compare the results the second one gives lots more relevant results. Cairo, HTML5 canvas, OpenGL, System.Drawing (.NET), Apple CoreAnimation all use "line join".

rfuest:
Can we move the intersection code out of `line/mod.rs` into another file? The file is getting long and it makes it harder to understand the core functionality. `Line::bresenham_scanline_intersection` could for instance be moved to `BresenhamIntersection::new`. For `segment_intersection` and `line_intersection` you could create a new struct `LineCoefficient` which contains the 7 coefficients and move most of the code there.

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -66,6 +67,68 @@ impl Dimensions for Line {
...
+impl BresenhamIntersection {
+    /// As a line
+    pub fn into_line(self) -> Line {
+        match self {
+            Self::Line(l) => l,
+            Self::Point(p) => Line::new(p, p),
+        }
+        .sorted_x()
+    }
+}
```
rfuest, line 130: Small optimization, because a single point line doesn't need to be sorted:
```suggestion
    pub fn into_line(self) -> Line {
        match self {
            Self::Line(l) => l.sorted_x(),
            Self::Point(p) => Line::new(p, p),
        }
    }
}
```

rfuest (changes requested):

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,105 @@
+//! A line segment constructed from two line joints.
+//!
+//! Segment always has one open side, so scanline intersections may return one or two points, but
+//! no more.
```
rfuest, line 4: IMO this comment belongs to the `ThickSegment` struct and not the module.

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,105 @@
...
+                self.start_joint.second_edge_start.left,
+            ),
+        )
+    }
+
+    /// Get the bounding box containing the left/right edges of the segment.
+    ///
+    /// Note that this does not include any bevel/cap lines as returned by `perimiter` which is why
+    /// this is not `impl Dimensions`. These lines don't need to be included as  other segments
+    /// in the polyline will expand the bounding box to the right place anyway.
```
rfuest, line 44:
```suggestion
    /// Note that this does not include any bevel/cap lines as returned by `perimeter` which is why
    /// this is not `impl Dimensions`. These lines don't need to be included as other segments
    /// in the polyline will expand the bounding box to the right place anyway.
```

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        let left_br = left.bottom_right().unwrap_or(tl);
+        let right_br = right.bottom_right().unwrap_or(tl);
+
+        let br = left_br.component_max(right_br);
+
+        Rectangle::with_corners(tl, br)
+    }
+
+    /// Get up to 6 lines comprising the perimiter of this segment.
```
rfuest, line 61:
```suggestion
    /// Get up to 6 lines comprising the perimeter of this segment.
```

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,105 @@
...
+    ///
+    /// Note: This array could be stored as a `ThickSegment` property, but it's calculated on the fly
+    /// for memory reasons.
```
rfuest, line 64: But the array will probably still cause a high peak memory consumption (if there is no compiler voodoo involved) because it is > 100 bytes large, which is a significant size on a small embedded device. Rewriting the code to only prepare one line at a timer and merging that into the scanline before moving on to the next one might be a better approach. Although it would probably make the code a bit messier.

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,105 @@
...
+            start_cap[1],
+            end_cap[0],
+            end_cap[1],
+            edges.0.into(),
+            edges.1.into(),
+        ]
+    }
+
+    pub(in crate::primitives) fn intersection(&self, scanline_y: i32) -> Option<Line> {
+        let perimiter = self.perimiter();
```
rfuest, line 81:
```suggestion
        let perimeter = self.perimeter();
```

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        // Loop through perimiter and get any intersections
```
rfuest, line 83:
```suggestion
        // Loop through perimeter and get any intersections
```

rfuest (changes requested):

rfuest:
Could you add changes as new commits instead of force pushing the branch and clean the commits up later? That would make it easier to follow the changes.

embedded-graphics/src/primitives/polyline/scanline_iterator.rs
```diff
@@ -0,0 +1,70 @@
...
+pub struct ScanlineIterator<'a> {
+    scanline_y: i32,
+    scanline_limit: i32,
+    intersections: ScanlineIntersections<'a>,
+    stop: bool,
+}
+
+impl<'a> ScanlineIterator<'a> {
+    /// New.
+    pub fn new<C>(styled: &Styled<Polyline<'a>, PrimitiveStyle<C>>) -> Self
```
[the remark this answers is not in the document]
jamwaffles: I've added a debug assert (to a commit not yet pushed at time of writing). For `stroke_width == 1`, it's far more efficient to use a normal Bresenham line iterator so this iterator doesn't need to support that case.

embedded-graphics/src/primitives/thick_segment.rs
jamwaffles: Both my parents are English teachers...

embedded-graphics/src/primitives/thick_segment.rs
jamwaffles: This comment isn't even correct anymore - the thick segment is now always generated as a closed shape, guaranteeing a line intersection.

jamwaffles:
Making the iterators non-recursive was a little bit off-by-one mind-bendy so if you can see any obvious ways to improve the code, particularly removing the `self.stop` flags, that would be great. 

Vielen Dank for the thorough review! There's so much I didn't test or clean up :grimacing:

embedded-graphics/CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ## [Unreleased] - ReleaseDate
 
+## [0.7.0-alpha.1] - 2020-09-19
```
rfuest, line 9: I think this header should be removed.

embedded-graphics/CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
...
+
+### Added
+
+- [#386](https://github.com/jamwaffles/embedded-graphics/pull/386) Added `Line::midpoint` to get the midpoint of a line.
+- [#386](https://github.com/jamwaffles/embedded-graphics/pull/386) `Polyline` now supports stroke widths greater than 1. Stroke alignment is not supported.
```
rfuest, line 14: Should we mention the stroke alignment in the changelog? This could be interpreted as a limitation of the current version which will be fixed later, but not supporting this was a deliberate choice.

embedded-graphics/src/primitives/polyline/mod.rs
```diff
@@ -20,7 +22,7 @@ pub use styled::StyledPixels;
 ///
```
rfuest, line 22: The doc comment for polyline still mentions the 1px line width limitation.

rfuest (changes requested):
Some small nits.

There are also some unresolved conversations left:
https://github.com/jamwaffles/embedded-graphics/pull/386#discussion_r515173066
https://github.com/jamwaffles/embedded-graphics/pull/386#discussion_r515217445
https://github.com/jamwaffles/embedded-graphics/pull/386#discussion_r515177633
They aren't super important and could be handled in future PRs, but I want to make sure you didn't miss them in the flood of comments.

rfuest:
This commit removes the `stop` flag from `ScanlineIterator`: https://github.com/rfuest/embedded-graphics/commit/c903392460c058ea463e363ba9fb9a4e082dd1c1

I had to change one test to make it pass. But IMO the result looks better now, so there might have been a problem with the old code, but you should probably take another look.

The change should also fix the limitation to `stroke_width > 1`, but I forgot to remove the `debug_assert`.

jamwaffles:
I made a few attempts and got the same result in that test too which worried me a bit. I tested with the `polyline-joints` example and didn't find any difference in how that line renders, so I'm not sure what's going on. I'll cherry pick the change in but I still think it's a good idea to keep the `debug_assert()`, just as a safety net.

rfuest:
> I made a few attempts and got the same result in that test too which worried me a bit.

I believe that the previous version had a bug and was only drawing one intersection on the last scanline, because the `stop` flag was set too early.

jamwaffles:
In a rush atm so didn't check the commit but I've just pushed a new commit to add a custom internal type for stroke offset. It's currently all hard coded to `StrokeOffset::None`, but it will come in useful when doing triangles/polygons. Let me know if it's going in the right direction or not :)

I'll pick up https://github.com/jamwaffles/embedded-graphics/pull/386#discussion_r515217445 later/tomorrow when I have a moment - that's the only other comment that needs addressing from the ones you raised. Thanks for posting them again - yes this PR is getting to be a typical length for e-g features :grin:

jamwaffles:
> I believe that the previous version had a bug and was only drawing one intersection on the last scanline, because the stop flag was set too early.

IIRC that test outputs two zero length lines for the last line, which the change to the iterator removes the last one. But the fixes both I and you have done to the iterator feel correct so I'm at a loss as to what's going on.

rfuest:
> IIRC that test outputs two zero length lines for the last line, which the change to the iterator removes the last one. But the fixes both I and you have done to the iterator feel correct so I'm at a loss as to what's going on.

Do you think the current behavior is wrong? I thought that it is correct now, because it does now draw two pixels in the last scanline. Previously only one pixel was drawn and the old test pattern assumed that this was correct.

rfuest:
This is my attempt of removing the `stop` flag from `ScanlineIntersections`: https://github.com/rfuest/embedded-graphics/commit/8ee0f88e6a1d3f910bb1cf19570648afc3e6958c
`next_segment` could use some comments, but IMO the overall logic is a little clearer than before.

jamwaffles:
> I thought that it is correct now, because it does now draw two pixels in the last scanline. 

I agree, it was just a strange difference to see from the test.

> This is my attempt of removing the stop flag from ScanlineIntersections

Looks good, but doesn't work on 1.40.0 with an error saying "subslice patterns are not stable" which is a shame. I cherry picked it anyway, and applied fixes on top to use an uglier, but compatible approach. 

Should we think about bumping MSRV to 1.47? There are a lot of `const` improvements from recent releases we could make use of as well as cleaning up the subslice stuff.

rfuest:
> Looks good, but doesn't work on 1.40.0 with an error saying "subslice patterns are not stable" which is a shame. I cherry picked it anyway, and applied fixes on top to use an uglier, but compatible approach.

I thought that it might not work but forgot to check. Do you think that we should extend the justfile to default to building with the MSRV by using `cargo +1.40.0 ...`? We would also need to provide some way to override this to another toolchain, e.g. by using an environment variable.

> Should we think about bumping MSRV to 1.47?

I wouldn't mind that. The biggest missing const feature (generics) still isn't included in the latest stable version, but all the other features, like flow control expressions in const fns and subslice patterns could certainly be useful. And it would also allow us to use the latest `just` versions with the same toolchain.

jamwaffles:
> Do you think that we should extend the justfile to default to building with the MSRV by using cargo +1.40.0 ...?

I don't think so. I think it's enough to rely on CI to catch any incompatibilities, as well as the ability to do `rustup override set 1.40.0` in the e-g repo folder.

> I wouldn't mind that.

Cool. I'll upgrade in a future PR.

embedded-graphics/src/primitives/line/thick_points.rs
```diff
@@ -122,11 +134,12 @@ impl ParallelsIterator {
...
             right: Bresenham::new(start_point),
             right_error: 0,
-            next_side: Side::Right,
+            next_side,
+            stroke_offset,
         };
 
         // Skip center line on left side iterator
-        self_.next_parallel(Side::Left);
+        self_.next_parallel(skip_side);
```
rfuest, line 142: Is `skip_side` required? The `match` expression above assigns `skip_side` to be `next_side.swap()`.

embedded-graphics/src/primitives/line/thick_points.rs
```diff
@@ -122,11 +134,12 @@ impl ParallelsIterator {
             left_error: 0,
             right: Bresenham::new(start_point),
             right_error: 0,
-            next_side: Side::Right,
+            next_side,
+            stroke_offset,
         };
 
         // Skip center line on left side iterator
```
rfuest, line 141: The comment doesn't apply to all cases anymore.

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,154 @@
...
+        let right_br = right.bottom_right().unwrap_or(tl);
+
+        let br = left_br.component_max(right_br);
+
+        Rectangle::with_corners(tl, br)
+    }
+
+    pub(in crate::primitives) fn intersection(&self, scanline_y: i32) -> Option<Line> {
+        // let perimeter = self.perimeter();
+
```
rfuest, line 60:
```suggestion
```

embedded-graphics/src/primitives/thick_segment.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+enum State {
+    StartCap,
+    EndCap,
+    Edges,
+    Done,
+}
+
+struct PerimiterLineIterator<'a> {
```
rfuest, line 108: Implementing an iterator for this seems more complicated than necessary, but it works so we can leave it this way. But we should correct the name:
```suggestion
struct PerimeterLineIterator<'a> {
```

simulator/examples/polyline-joints.rs
```diff
@@ -0,0 +1,97 @@
+use embedded_graphics::{
```
rfuest, line 1: You said that this example was mainly intended to debug the code, so should we remove this example in the PR and maybe add add a better one in a future PR? We still have `primitives-polyline` to demonstrate the basics.

simulator/examples/polyline-joints.rs
rfuest: I don't think we need to fix this in this PR, but IMO we should take another look at this later. Can you open an issue for this?

rfuest (changes requested):
Some more nits.

jamwaffles:
> I don't think we need to fix this in this PR, but IMO we should take another look at this later. Can you open an issue for this?

The file is gone now but I'm not sure what you mean by "this" here :thinking:

rfuest:
This was supposed to be a part of the previous conversation about the bevel issue, but somehow it got moved to the end of the thread. I meant that we should take another look at the algorithm that determines if a corner should be beveled, which was a problem that the removed example showed.

rfuest (approved):
Looks good :+1:

jamwaffles:
Ah, I see, yeah that would be a good idea so the line is more stable. Maybe based on the angle instead of how far away the intersection point is or something...

Thanks again for the review :). This has been a long time coming!

PR #387, opened 2020-07-09
Replace example in simulator README

### LINKED ISSUE #385: Simulator example in README uses different size than stated
The simulator README [here](https://github.com/jamwaffles/embedded-graphics/blame/master/simulator/README.md#L13) reads "Simulate a 128x64 SSD1306 OLED", but then the example code uses `Size::new(129, 129)` [here](https://github.com/jamwaffles/embedded-graphics/blame/master/simulator/README.md#L26)

bugadani:

## PR description

This PR closes #385

I've tried to keep true to the original example, with some artistic license. The example renders this image:

image


jamwaffles:
That looks pretty good, however I'm not sure it adds much on top of the existing demo. One regression that sticks out to me is that there's no demonstration of stroke/fill behaviour in the new demo which I think is important to show.

bugadani:
There wasn't any originally ;) The whole point of this PR is to fix the 128x64 vs 129x129px thing. The original readme-example was a simple clockface looking thing with Hello World written inside it, nothing more. I would have copied the more complicated hello world example, but I did not want to make it any more longer and that is already in the e-g readme.

bugadani:
I could, but I think the Windows window style is kinda bad for consistency here and I don't have anything else.

jamwaffles:
I wouldn't worry about it tbh. The simulator docs in `lib.rs` mention multiple platforms (although could use some clarification about supported platforms and copying into the readme), so a Windows screenshot is ok IMO. The e-g example screenshot is from what looks like a GNOME 3 Linux machine but I wouldn't go out of your way to match that window style.

bugadani:
Oh hm, there's an example in there as well. This is getting more complicated than what I intended 😅

jamwaffles:
Quick poke @bugadani is this ready for another review? I'm unsure what's in this that isn't covered by #389. It also needs a rebase :)

bugadani:
Hm, no, it isn't.

> I'm unsure what's in this that isn't covered by #389

#389 updated the hello world of embedded-graphics. This PR updated the simulator hello world. For this reason (having 2 hello worlds) I'm not sure how to proceed.

We can keep two separate examples and that would mean I could update this with some interactivity example (like close on click, that was lost somewhere).

On the other hand, the embedded-graphics hello world also built on the simulator and is a much nicer example, but a little more complex on the graphics side.

I'd leave the decision up to you, and progress from that point.

PR #388, opened 2020-07-10
Update simulator setup instructions

bugadani:

## PR description

A little bit of improvement to the simulator readme. I think the newly linked SDL2 Rust readme gives much simpler instructions compared to the SDL2 wiki on how to set up SDL2 for Rust in Windows, but that may just be subjective.

### DIFF
--- a/simulator/src/lib.rs
+++ b/simulator/src/lib.rs
@@ -6,7 +6,7 @@
 //!
 //! # Setup
 //!
-//! The simulator uses SDL and its development libraries which must be installed to build and run
+//! The simulator uses SDL2 and its development libraries which must be installed to build and run
 //! it.
 //!
 //! ## Linux (`apt`)
@@ -23,8 +23,13 @@
 //!
 //! ## Windows
 //!
-//! The Windows install process is a bit more involved, but it _does_ work. See [the SDL2
-//! wiki](https://wiki.libsdl.org/Installation#WinRT.2FWindows_8.2FWinPhone) for instructions.
+//! The Windows install process is a bit more involved, but it _does_ work. See [the Rust-SDL2
+//! crate's README](https://github.com/Rust-SDL2/rust-sdl2) for instructions. There are multiple
+//! ways to get it working, but probably the simplest method is copying the binaries as shown
+//! [here](https://github.com/Rust-SDL2/rust-sdl2#windows-msvc).
+//!
+//! **Note:** The precompiled binaries need to be added to _all_ Rust versions you use. For example,
+//! this means that you will need to copy the files after every update to a new stable Rust version.
 //!
 //! # Examples
 //!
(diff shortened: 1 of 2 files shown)


rfuest:
@bugadani Do you think that it would be easier for users of e-g on Windows to support the `bundled` flag of the SDL2 crate? According to their docs this should make it easier to use, but I haven't done any Rust development on Windows and can't judge the developer experience on that platform.

bugadani:
Honestly, for someone who knows how to set that up, sure, adding a line to `Cargo.toml` is the simplest there is. But I'm quite sure it's simpler for a _lot_ of people to just copy some files around.

My main issue with the bundled feature is that it's not well documented what _exactly_ needs to be done. I don't think `you will need a C compiler` is enough. Especially, since I thought I had everything set up and was greeted with a "is `cmake` not installed?" error. It doesn't seem as simple as going through the Visual Studio Installer and selecting the necessary package, so I think a lot of people would just give up.

jamwaffles:
I've set up SDL2 on Windows (10) before and from that experience, I'd rather copy a DLL or two around than set up a whole compilation toolchain. My two pence :)

bugadani:
I've converted this to draft - I'm not at all happy with my wording and maybe there are other improvements I can make

jamwaffles:
@bugadani A quick note if/when you pick this up again: crate readmes are now generated from each crate's `lib.rs` doc comment, so you'll want to make changes in `simulator/src/lib.rs` and regenerate the readme by running `./generate_readmes.sh` in the project root.

bugadani:
Sorry for being silent on this one - I'm keeping the PR up to date if I notice, otherwise I think this is ready. Not a lot of change but I think it's still an improvement

simulator/src/lib.rs
```diff
@@ -23,8 +23,13 @@
...
 //!
-//! The Windows install process is a bit more involved, but it _does_ work. See [the SDL2
-//! wiki](https://wiki.libsdl.org/Installation#WinRT.2FWindows_8.2FWinPhone) for instructions.
+//! The Windows install process is a bit more involved, but it _does_ work. See [the Rust-SDL2
+//! crate's README](https://github.com/Rust-SDL2/rust-sdl2) for instructions. There are multiple
+//! ways to get it working, but probably the simplest method is copying the binaries as shown
+//! [here](https://github.com/Rust-SDL2/rust-sdl2#windows-msvc).
+//!
+//! **Note:** The precompiled binaries need to be added to _all_ Rust versions you use. For example,
+//! this means that you will need to copy the files after every update to a new stable Rust version.
```
jamwaffles, line 32: I think this can be removed as it's outside the simulator setup instructions and should be covered by the SDL2 side.

bugadani: Fair enough

jamwaffles:
Thanks!

PR #389, opened 2020-07-11
Update hello-world example

bugadani:
Some of the changes made the hello-world example a bit weird.

## PR description

Before:
image

After:
image

Changes:
 * Draw border inside so it's 3px as comment states
 * Draw circle using inside-aligned stroke
 * Make square 1px bigger so it's aligned properly

Nothing serious, but after I noticed these, I just couldn't unsee that the example got a bit misaligned.


bugadani:
Unfortunately, there's no way to set stroke alignment without the builder, soooooo... :)

rfuest:
I don't mind using `PrimitiveStyleBuilder` in the example. But I don't like the look of a 17 pixel diameter circle, because it doesn't look as round as the previous example. Can we change it to something else, like 16 or 20 pixels and change the other primitives to match?

bugadani:
Sure thing, but now I don't like the triangle that much 😅

bugadani:
I can if the current output looks okay to you

rfuest:
If we don't add a window border we can use the dump feature to generate the example screenshot.

jamwaffles:
> I can if the current output looks okay to you (see PR text for screenshot)

Looks good to me :)

> If we don't add a window border we can use the dump feature to generate the example screenshot.

Good idea. This should do it:

```bash
EG_SIMULATOR_DUMP=assets/hello-world-simulator.png cargo run --example hello-world
```

We should probably also update the example in `embedded-graphics/README.md` as that's a copypasta of `hello-world.rs` as far as I can see, but is now out of date.

rfuest:
> I can if the current output looks okay to you (see PR text for screenshot)

I like the circle and the triangle also looks fine. Can you center the primitives a bit better by moving them to the right a few pixels? And IMO both text and primitives could be also moved down a few pixels (keeping the spacing between them).

bugadani:
> Can you center the primitives a bit better

This is the exact reason embedded-layout exists 😂

PR #390, opened 2020-07-12
Faster triangle point intersection

jamwaffles:

## PR description

Optimises `Triangle`s `ContainsPoint` implementation. I left some comments in the commits which should hopefully be more descriptive. Currently seeing a 92-95% performance improvement.


embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -232,12 +275,29 @@ mod tests {
 
     #[test]
     fn contains() {
-        let triangle = Triangle::new(Point::new(5, 10), Point::new(15, 20), Point::new(10, 15));
+        let triangles = [
+            // Colinear triangle. Mathematically, zero sized, in e-g land however this draws pixels
+            // along a line, therefore it is possible to contain a point.
```
rfuest, line 280: We might want to revisit colinear triangles later. To be consistent with rectangles, filling a zero sized triangles shouldn't draw any pixels. Can you create a new issue or add this to another issue so we remember this?

embedded-graphics/src/primitives/triangle/styled.rs
```diff
@@ -216,4 +216,35 @@ mod tests {
...
     }
+
+    #[test]
+    fn styled_stroke_equals_lines() {
+        let triangle = Triangle::new(Point::new(10, 10), Point::new(20, 20), Point::new(30, -30));
+
+        let styled = triangle.into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 1));
+
+        let mut tri_display: MockDisplay<BinaryColor> = MockDisplay::new();
+        tri_display.set_allow_out_of_bounds_drawing(true);
```
rfuest, line 227: Wouldn't it be better to choose a triangle that is inside the `MockDisplay` bounds and keep the bounds checking enabled? Otherwise parts of the triangle wont be checked.

simulator/examples/triangle-contains.rs
```diff
@@ -0,0 +1,130 @@
+//! # Example: Triangles and `contains()`
```
rfuest, line 1: I would keep the example, but change it to be more generic and demonstrate `contains` for all primitives that support it. But this doesn't need to be in this PR.

simulator/examples/triangle-contains.rs
```diff
@@ -0,0 +1,130 @@
...
+    let inner_radius = Size::new_equal(2);
+
+    Line::new(point - radius.x_axis(), point - inner_radius.x_axis())
+        .points()
+        .chain(Line::new(point + radius.x_axis(), point + inner_radius.x_axis()).points())
+        .chain(Line::new(point - radius.y_axis(), point - inner_radius.y_axis()).points())
+        .chain(Line::new(point + radius.y_axis(), point + inner_radius.y_axis()).points())
+        .map(|p| Pixel(p, color))
+        .draw(display)
+        .unwrap();
```
rfuest, line 31: Should we use iterator chaining in the examples? It's a nice feature, but breaks HW acceleration.

simulator/examples/triangle-contains.rs
```diff
@@ -0,0 +1,130 @@
...
+        //         .into_styled(PrimitiveStyle::with_stroke(Rgb888::YELLOW, 1))
+        //         .draw(display)?;
+        //     Line::new(triangle.p3, triangle.p1)
+        //         .into_styled(PrimitiveStyle::with_stroke(Rgb888::YELLOW, 1))
+        //         .draw(display)?;
+
+        //     Pixel(triangle.p1, Rgb888::CYAN).draw(display)?;
+        //     Pixel(triangle.p2, Rgb888::CYAN).draw(display)?;
+        //     Pixel(triangle.p3, Rgb888::CYAN).draw(display)?;
+        // }
```
rfuest, line 86: If we want to keep the example this needs to be cleaned up a bit. Can the `show_outlines` variable be removed?

simulator/examples/triangle-contains.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // Cursor position
+    empty_crosshair(point, Rgb888::CYAN, display);
```
rfuest, line 90: Cyan on green is really low contrast and should be changed to something else.

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -88,9 +88,48 @@ impl ContainsPoint for Triangle {
...
+        let p = point;
+        let Self { p1, p2, p3 } = *self;
+
+        // Check if point is inside triangle using https://stackoverflow.com/a/20861130/383609.
+        // Works for any point ordering.
+        let is_inside = {
+            let s = p1.y * p3.x - p1.x * p3.y + (p3.y - p1.y) * p.x + (p1.x - p3.x) * p.y;
+            let t = p1.x * p2.y - p1.y * p2.x + (p1.y - p2.y) * p.x + (p2.x - p1.x) * p.y;
+
+            if (s < 0) != (t < 0) {
```
bugadani, line 100: Just a fun thing, this is compiled to exactly the same machine code as `same_signs` in the line joiners PR.

https://godbolt.org/z/EMGGEM

jamwaffles: Cool! I'll likely factor this logic out in that PR as they're doing the same thing.

simulator/examples/triangle-contains.rs
jamwaffles: I've elected to remove this example, at least for now. I think creating an example that demonstrates the `contains()` behaviour specifically, and not a triangle debugger, is a better way to go. That can happen in a new PR.

jamwaffles:
Some benchmark results:

| Bench | Before | After | Change |
| --- | --- | --- | --- |
| triangle contains point (inside) | 111.95 ns | 7.5670 ns | -93.215% |
| triangle contains point (outside) | 2.5961 us | 118.25 ns | -95.490% |

Ryzen 7 3700X OCed to 3.8GHz, 32GB 3000MHz DDR4, Linux Mint 20

Before:

```
triangle contains point (inside)
                        time:   [111.41 ns 111.95 ns 112.53 ns]
                        change: [+0.7986% +1.1838% +1.5368%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe

triangle contains point (outside)
                        time:   [2.5926 us 2.5961 us 2.6005 us]
                        change: [-0.5621% -0.2552% +0.0245%] (p = 0.11 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe
```

After:

```
triangle contains point (inside)
                        time:   [7.5471 ns 7.5670 ns 7.5873 ns]
                        change: [-93.239% -93.215% -93.194%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) high mild
  8 (8.00%) high severe

triangle contains point (outside)
                        time:   [117.58 ns 118.25 ns 118.96 ns]
                        change: [-95.528% -95.490% -95.445%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
```

embedded-graphics/src/primitives/triangle/mod.rs
jamwaffles: I've implemented that in this PR. `contains()` now checks for colinearity by comparing the determinant with zero, as does `ScanlineIterator` which won't return any points for a colinear triangle either.

rfuest: `ScanlineIterator` should still draw a stroke for a colinear triangle, to be consistent with the other primitives.

jamwaffles: Oops, you're right. Fixed!

embedded-graphics/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - **(breaking)** [#357](https://github.com/jamwaffles/embedded-graphics/pull/357) Additional checks for overdraw and out of bounds drawing were added to `MockDisplay` which can cause tests to panic. See the `mock_display` module docs for more information.
 - **(breaking)** [#342](https://github.com/jamwaffles/embedded-graphics/pull/342) Refactored the `DrawTarget` trait to better support common hardware capabilities.
 - **(breaking)** [#360](https://github.com/jamwaffles/embedded-graphics/pull/360) Make the `drawable` module private. `drawable::Drawable` and `drawable::Pixel` are now exported from the crate root.
+- **(breaking)** [#390](https://github.com/jamwaffles/embedded-graphics/pull/390) Triangles that are colinear no longer render any points, and will never contain any points (i.e. `Triangle.contains` always returns false).
```
rfuest, line 43: The first part only applies to styled triangles without a stroke.

embedded-graphics/src/lib.rs
```diff
@@ -537,7 +537,7 @@
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/jamwaffles/embedded-graphics/191fe7f8a0fedc713f9722b9dc59208dacadee7e/assets/logo.svg?sanitize=true"
 )]
-#![no_std]
+// #![no_std]
```
rfuest, line 540:
```suggestion
#![no_std]
```

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -131,6 +175,16 @@ impl Triangle {
             p3: points[2].into(),
         }
     }
+
+    /// Return the [determinant](https://en.wikipedia.org/wiki/Determinant) for this triangle.
```
rfuest, line 179: This method should be renamed IMO. The returned value isn't a generic determinant but a property of the triangle that gets calculated by using a determinant. If my quick google research is right, the returned value is twice the area of the triangle.

jamwaffles: This is my very shallow understanding of the maths showing through. I did see it used as part of the area calculation though. How about I call it `area_doubled`? Open to better suggestions.

embedded-graphics/src/lib.rs
jamwaffles: :facepalm: thanks

embedded-graphics/src/primitives/triangle/mod.rs
rfuest: > How about I call it `area_doubled`?

Sounds good.

jamwaffles:
Thanks!

PR #392, opened 2020-07-13
Split generated examples

jamwaffles:

## PR description

I've split this out as a precursor to #381 as the generated examples have been causing a bunch of issues in that PR. This one changes the generator to output the images to `assets/doc` and adds a script to generate an `EXAMPLES.md` file with all the examples in them. There's some reasonably gnarly `sed` magic in the generation command but it does filter the file properly so...

Some points:

1. Where else should we link to this file from?
2. Should `EXAMPLES.md` be generated into a more sensible location?
3. Should we use full `https://...` URLs for the image assets? This would make testing changes to the file a bit more painful, but would make the `.md` a lot more portable.


rfuest:
> Where else should we link to this file from?

The front page of the docs and the readme would be the places I expect to find something like this. I think that we should make the link stand out visually in the readme to make it easily discoverable, perhaps by adding a button or image for it.

> Should `EXAMPLES.md` be generated into a more sensible location?

IMO it can be generated in the final location and doesn't need to generate in another directory first. Do we want to keep the final file in the root directory? If I'm looking at an unknown repository I tend to check for a `doc` directory, like in the `nom` repo: https://github.com/Geal/nom/tree/master/doc
 
> Should we use full `https://...` URLs for the image assets? This would make testing changes to the file a bit more painful, but would make the `.md` a lot more portable.

Does the md file need to be more portable? If we decide to add a `doc` directory we could also add the images to `doc/assets` to keep the md and image files together.
We could also add a parameter to the generator to set the image base URL, but that would make it harder to make sure the correct version gets committed into the master branch.

jamwaffles:
Thanks for the feedback. I'm now generating the file into `doc/`. I also added an imagemagick `montage` command to produce an eye catching collage of all the examples. This does add a dependency on `imagemagick` for the generator script which I think is ok, but worth pointing out. The README and lib.rs docs use the collage image to link to the doc page.

There's a REVERTME commit I pushed which changes the branch name so the links work, if you want to test them. I'll rebase this out before merge.

generate_drawing_examples.sh
```diff
@@ -0,0 +1,16 @@
...
+mkdir -p $ASSETS
+
+cargo run --bin generate-drawing-examples | \
+    rustfmt +nightly --config-path rustfmt.examples.toml | \
+    sed -E -e 's@//! ?@@g' -e '/^# .*/d' -e '/pub mod dummy \{\}/d' - \
+    > doc/drawing-examples.md
+
+# Generate a collage of all screenshots
+# `imagemagick` must be installed for this to work.
+montage $ASSETS/draw*.png $ASSETS/display*.png -tile 6x2 -background none $ASSETS/all_drawing_ops.png
```
rfuest, line 16:
```suggestion
montage $ASSETS/draw*.png $ASSETS/display*.png -tile 6x2 -background none -geometry 128x128+4+4 miff:- | convert - -trim $ASSETS/all_drawing_ops.png
```
This fixes the slightly different column and row spacing and removes the outer border.

simulator/src/bin/generate-drawing-examples.rs
```diff
@@ -21,64 +29,64 @@ macro_rules! op {
...
-        let path = format!("{}/{}.png", OUTPUT_BASE, $title);
+        let file_path = format!("{}/{}.png", IMAGE_OUTPUT_BASE, cleansed_title);
         $display
             .to_image_buffer(&output_settings)
-            .save(&path)
+            .save(&file_path)
             .unwrap();
 
-        let screenshot = base64::encode(std::fs::read(&path).expect("Couldn't open screenshot"));
+        let md_path = format!("./assets/{}.png", cleansed_title);
```
rfuest, line 45: This isn't the path that contains the md files.
```suggestion
        let doc_assets_path = format!("./assets/{}.png", cleansed_title);
```

doc/primitives.md
```diff
@@ -0,0 +1,268 @@
+Embedded graphics examples
```
rfuest, line 1: Is this an old version/name of the `drawing-examples.md` file?

generate_drawing_examples.sh
jamwaffles: Thank you, that looks much better.

doc/primitives.md
jamwaffles: Oops, yes that shouldn't be in here at all.

jamwaffles:
> But there also seems to be a problem with that. 

That's frustrating :( I'm using `ImageMagick 6.9.10-23 Q16 x86_64 20190101`. We could change the single montage image for an HTML layout in README.md and lib.rs. It's a bit messier, but wouldn't rely on a binary file diffing correctly then. Can you think of a better idea?

simulator/src/bin/generate-drawing-examples.rs
```diff
@@ -21,64 +29,64 @@ macro_rules! op {
...
-            //!
-            //! <div style="display: flex">
-            //! <img style="width: 128px; height: 128px; margin-right: 8px;" alt="{} example screenshot" src="data:image/png;base64,{}" />
-            //! <div style="flex-grow: 1;">
+            r#"//!
+//! ## {}
+//!
+//! {}
+//!
+//! ![{} example screenshot]({})
```
rfuest, line 63: Can we keep the left alignment of the images? IMO that looked better.
According to this https://gist.github.com/DavidWells/7d2e0e1bc78f4ac59a123ddf8b74932d/ it should work by using an `img` tag with an `align` attribute.

rfuest:
> Can you think of a better idea?

Should we just remove this step from being executed when running `build.sh` by moving it into a separate script? The montage is only an visual overview and shouldn't change often, so IMO it's not to bad to have to manually update it when necessary.

simulator/src/bin/generate-drawing-examples.rs
jamwaffles: Sure, done. Much better than my old flexbox solution, which I don't think would've worked anyway as the inline styles are stripped by Github's markdown renderer. The only issue now is I can't get a margin between the image and code block:

image

Doesn't look great, but it's not a major problem. I did try some inline `style=""` attributes but they're all stripped.

jamwaffles:
> Should we just remove this step from being executed when running build.sh by moving it into a separate script?

Yes - turns out imagemagick on my machine isn't idempotent either. I'll move the montage generator into a separate script in the root for now. All these scripts can be cleaned up by #394 when that's implemented.

jamwaffles:
Github finally figured out that I pushed some commits (there's some outage going on atm) but if they all show up in the diff view, please take another look :)

build.sh
```diff
@@ -13,6 +13,13 @@ pushd simulator
 cargo build --release --no-default-features
 popd
 
+# Ensure that examples files are up to date
+./generate_drawing_examples.sh
+git diff --quiet doc/ || (
+    echo "doc/ folder is not up to date"
+    echo "Try running ./generate_drawing_examples.sh"
```
rfuest, line 20: IMO this would also be a good place to mention that you should run the montage script if any image has changed.

simulator/src/bin/generate-drawing-examples.rs
jamwaffles: Oh that's good to know! I've been using a Sublime Text plugin that I believe calls Github's API and shows the page standalone which I've always assumed is accurate to Github rendering. I guess there are some additional styles included on Github.com itself.

build.sh
jamwaffles: Good idea. I added another line.

PR #395, opened 2020-07-15
MockDisplay panic now prints which row width is incorrect

bugadani:

## PR description

I'm not sure if this wording is acceptable, but I find it useful to print which row actually failed the validation.

Also add a dimension to the other panics, just to be overly pedantic.

### DIFF
--- a/embedded-graphics/src/mock_display.rs
+++ b/embedded-graphics/src/mock_display.rs
@@ -428,7 +428,7 @@ where
             assert_eq!(
                 row.len(),
                 pattern_width,
-                "Row #{} is {} characters wide instead of {}",
+                "Row #{} is {} characters wide (must be {} characters to match previous rows)",
                 row_idx + 1,
                 row.len(),
                 pattern_width
--- a/embedded-graphics/src/mock_display.rs
+++ b/embedded-graphics/src/mock_display.rs
@@ -416,19 +416,21 @@ where
         let pattern_height = pattern.len();
         assert!(
             pattern_width <= SIZE,
-            "Test pattern must not be wider than {}",
+            "Test pattern must not be wider than {} columns",
             SIZE
         );
         assert!(
             pattern_height <= SIZE,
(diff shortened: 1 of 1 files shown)


bugadani:
To be fair, I hit this with 0.6.2 where the error message wasn't at all helpful:

```
thread 'layout::test::test_view_location_preserved' panicked at 'assertion failed: `(left == right)`
  left: `6`,
 right: `7`', C:\Users\bugad\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib/rustlib/src/rust\src\libcore\macros\mod.rs:16:9
```

The `master` version at least hints what this panic is about, but I think we can do better by pointing at the exact row that's incorrect.

embedded-graphics/src/mock_display.rs
```diff
@@ -416,19 +416,21 @@ where
...
+            "Test pattern must not be taller than {} rows",
             SIZE
         );
-        for row in pattern {
+        for (row_idx, row) in pattern.iter().enumerate() {
             assert_eq!(
                 row.len(),
                 pattern_width,
-                "Every row in the pattern must be {} characters wide",
+                "Row #{} is {} characters wide instead of {}",
```
rfuest, line 431: The new message doesn't mention that every row needs to be the same width, so that the user can't easily tell where the `instead of X` value comes from.

bugadani: That's a fair point but I wasn't able to come up with a better alternative. Repeating "characters wide" really looks weird and my creativity was/is limited :)

jamwaffles: My submission if you need some inspiration:

```
Row #{} is {} characters wide (must be {} characters to match all other rows).
```

bugadani: Thanks, that was helpful. I modified it a bit, since MockDisplay panics on the first incorrect row and so we can't say the others after it are correct :)

PR #400, opened 2020-07-19
Add OffsetOutline and StyledPrimitiveAreas

### LINKED ISSUE #399: Making expand and shrink methods for primitives public
As part of implementing the `fill_area` and `stroke_area` methods I would also like to make `expand` and `shrink` for the primitives public. 

But I'm not sure how to name the methods and if it's better to merge them into one. Instead of having two methods they could also be implemented as one that takes a signed value as a parameter (>0 expands, <0 shrinks). Would this make more sense or should we keep the current implementation? And if we decide to change it to a single method, how should this method be called? Vector graphics program often refer to the operation as offset, but that wouldn't be a good method name IMO, because it's not clear that the offset is applied to the outline and not the position. Maybe `offset_outline`?

rfuest:

## PR description

This PR adds the `OffsetOutline` trait discussed in #399 and a `StyledPrimitiveAreas` trait that implements the methods that were suggested in https://github.com/jamwaffles/embedded-graphics/pull/307#issuecomment-609475823.

Closes #399

### DIFF
--- a/embedded-graphics/src/primitives/mod.rs
+++ b/embedded-graphics/src/primitives/mod.rs
@@ -50,3 +50,13 @@ pub trait ContainsPoint {
     /// Returns `true` if the given point is inside the shape.
     fn contains(&self, point: Point) -> bool;
 }
+
+/// Offset outline trait.
+pub trait OffsetOutline {
+    /// Offsets the outline of the shape.
+    ///
+    /// The offset is applied perpendicular to each element of the outline.
+    /// Offset values larger than zero will expand the shape and values smaller
+    /// than zero will shrink the shape.
+    fn offset(&self, offset: i32) -> Self;
+}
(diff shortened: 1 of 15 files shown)


embedded-graphics/src/primitives/mod.rs
```diff
@@ -50,3 +50,13 @@ pub trait ContainsPoint {
...
     fn contains(&self, point: Point) -> bool;
 }
+
+/// Offset outline trait.
+pub trait OffsetOutline {
+    /// Offsets the outline of the shape.
+    ///
+    /// The offset is applied perpendicular to each element of the outline.
+    /// Offset values larger than zero will expand the shape and values smaller
+    /// than zero will shrink the shape.
```
jamwaffles, line 60:
```suggestion
    /// Offset values greater than zero will expand the shape and values less
    /// than zero will shrink the shape.
```

jamwaffles, line 60: Should we mention here that the behaviour is to saturate if the numbers are too large to fit in an `i32` or `u32`? It's unlikely to happen so might not be worth adding in.

rfuest: IMO it's not worth mentioning, because it's so unlikely for normal display sizes.

Issue #394, opened 2020-07-14
Replace scripts with the `just` task runner


rfuest:
e-g contains multiple scripts to generate files or to aid with CI. This collection of scripts isn't very well documented and the required parameters aren't obvious.

I would suggest to replace the scripts be using a task runner like [casey/just](https://github.com/casey/just/) or [sagiegurari/cargo-make](https://github.com/sagiegurari/cargo-make). This way all task can be defined and documented in one central file and we can also benefit from other features like dependencies between tasks.

jamwaffles:
> e-g contains multiple scripts to generate files or to aid with CI. This collection of scripts isn't very well documented and the required parameters aren't obvious.

I agree. Thanks for opening this issue.

I had a quick look at both of those linked projects and I'm leaning towards `just` over `cargo-make`. I like the idea of TOML command files, but the syntax is pretty verbose. `just` uses a custom file format, but doesn't look too complex as I think what we kind of want is "bash, but just a little bit more". Do you have a preference/reason not to use `just`?

rfuest:
No, I'm using `just` in another project and I really like it. It's focused on one task and doesn't try to be a solution for everything. The syntax is similar to a Makefile without some of the weirdness.

jamwaffles:
Alright, let's go with `just` then :). If you want to take this on please assign yourself to the issue so I don't step on your toes.

PR #403, opened 2020-07-20
Use associated type for Drawable color type

rfuest:

## PR description

This PR changes `Drawable` to use an associated type for the color type. `DrawTarget` and `IntoIterator` were recently also changed to use associated types and IMO it also makes sense for `Drawable`.

@jamwaffles: Why did you remove `generate_examples_screenshots.sh` in #392? I use it as an additional check before committing  larger changes and have previously found errors, that weren't covered by unit tests, by looking at the generated screenshots.


jamwaffles:
Nice! I'll try and review this later this evening.

> Why did you remove generate_examples_screenshots.sh

My bad! I thought it wasn't used. Definitely add it back in if you need it as part of your flow though.

Changes like this should hopefully be more apparent when #394 is implemented.

rfuest:
The main reason is to align it with the other traits. It doesn't make a huge difference for our own implementations of `Drawable`, but it's more restrictive and might prevent some unexpected implementations in other projects.

PR #407, opened 2020-07-21
Add support for bottom-left image origin

rfuest:

## PR description

This PR adds support for TGA images with bottom-left image origin to `tinytga`. Additionally the pixel iterator now ignores any data present in the `image_data` after `width * height` pixels, which seems to have fixed the `cbw8.rs` and `ubw8.tga` tests. This means that there are no more ignored tests in `tinytga`.

I had to temporarily change the `Drawable` implementation for `Image`, because it assumed the iteration order and ignored the pixel coordinates returned by the image data. I'll revisit this when I work on `SubImage`.

### DIFF
--- a/tinytga/CHANGELOG.md
+++ b/tinytga/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - **(breaking)** [#407](https://github.com/jamwaffles/embedded-graphics/pull/407) The `image_descriptor` in `TgaHeader` was replaced by `image_origin` and `alpha_channel_bits`.
+- **(breaking)** [#407](https://github.com/jamwaffles/embedded-graphics/pull/407) The `Pixel` type returned by `TgaIterator` now uses `u16` coordinates.
 
 ### Added
 
--- a/tinytga/CHANGELOG.md
+++ b/tinytga/CHANGELOG.md
@@ -6,14 +6,23 @@
 
 ## [Unreleased] - ReleaseDate
 
-- **(breaking)** #407 The `image_descriptor` in `TgaHeader` was replaced by `image_origin` and `alpha_channel_bits`.
-- #407 Added support for bottom-left origin images to `TgaIterator`.
+### Changed
+
+- **(breaking)** [#407](https://github.com/jamwaffles/embedded-graphics/pull/407) The `image_descriptor` in `TgaHeader` was replaced by `image_origin` and `alpha_channel_bits`.
+
+### Added
+
+- [#407](https://github.com/jamwaffles/embedded-graphics/pull/407) Added support for bottom-left origin images to `TgaIterator`.
+
+### Fixed
+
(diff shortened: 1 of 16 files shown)


tinytga/src/lib.rs
```diff
@@ -371,7 +386,19 @@ impl<'a> Iterator for TgaIterator<'a> {
...
+
+            if self.tga.header.image_origin.is_bottom() {
+                if self.y > 0 {
+                    self.y -= 1;
+                } else {
+                    self.done = true;
+                }
+            } else {
+                self.y += 1;
+                if self.y >= u32::from(self.tga.header.height) {
```
jamwaffles, line 398: We do quite a lot of `u32::from` in this function and elsewhere. Would it be better to use `u16`s in the iterator and maybe also `Pixel`? This would make the tinytga semantically correct as TGA images can't be larger than a `u16`'s max value in either dimension anyway. The type coercion would be pushed the e-g portion of the crate, but we already do a `as i32` for pixel coordinates there so I don't think that would need any changes.

tinytga/src/header.rs
```diff
@@ -70,16 +101,11 @@ pub struct TgaHeader {
...
-    /// * `00` = bottom left
-    /// * `01` = bottom right
-    /// * `10` = top left
-    /// * `11` = top right
-    pub image_descriptor: u8,
+    /// Image origin
+    pub image_origin: ImageOrigin,
+
+    /// Alpha channel bits
+    pub alpha_channel_bits: u8,
```
jamwaffles, line 108: Would `alpha_channel_depth` be a better name?

rfuest: "Alpha Channel Bits" is used in the TGA specs, but I agree that `alpha_channel_depth` is better and have changed it.

tinytga/src/lib.rs
rfuest: Sounds like a good idea, but isn't directly related to this PR IMO. Or do you want me to include the change in this PR?

tinytga/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ## [Unreleased] - ReleaseDate
 
+- **(breaking)** #407 The `image_descriptor` in `TgaHeader` was replaced by `image_origin` and `alpha_channel_bits`.
+- #407 Added support for bottom-left origin images to `TgaIterator`.
```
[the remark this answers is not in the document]
rfuest: I've added links to the new and all previous entries and added the suggested entries.

tinytga/src/lib.rs
jamwaffles: It's slightly outside the original scope, I agree. If you want to keep it separate can you open a tracking issue please? It might also be worth checking `tinybmp` and doing the same there too.

rfuest: I've changed my mind and included the change in this PR.

tinytga/CHANGELOG.md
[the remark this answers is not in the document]
jamwaffles: Awesome! Thanks for cleaning the whole file up

PR #410, opened 2020-07-22
Add justfile

### LINKED ISSUE #394: Replace scripts with the `just` task runner

rfuest:

## PR description

This PR replaces the scripts in the root directory with a single justfile. I've done some initial testing and it worked as expected, but my PC is getting old and running many complete builds of e-g is time consuming, so I didn't test it very thoroughly.

@jamwaffles I've marked the PR as a draft, because the CI configuration still needs to be updated. Can you add the necessary changes to this PR, because you are more familiar with that aspect of e-g? `./build.sh` needs to be replaced with `just build` and `./build_target.sh --release` with `just build-target $TARGET --release`.

Closes #394


jamwaffles:
On my phone at the moment but yep, I'll add the Circle stuff. I think the best way would be to PR into this branch, right?

rfuest:
I've enabled "Allow edits from maintainers" so you should be able to add commits to my branch.

justfile
```diff
@@ -0,0 +1,138 @@
...
+    cargo readme -r {{crate}} | sed -E -f filter_readme.sed > {{target_dir}}/{{crate}}_README.md
+
+#----------
+# Examples
+#----------
+
+# Generates the drawing examples screenshots and markdown file
+generate-drawing-examples:
+    #!/usr/bin/env bash
+    set -e -o pipefail
```
jamwaffles, line 102: I'm not sure why, but the `-o pipefail` here makes this command fail in CI with the following error

```
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', src/libstd/io/stdio.rs:792:9
```

jamwaffles: We could just remove it, but that might allow some false-positives to pass in CI.

jamwaffles:
I've pushed my CircleCI changes to this branch and left a comment with an issue I found.

justfile
rfuest: I just noticed this in the build log: "error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installed"

rfuest:
The problem existed before it was ported to just. Here is an excerpt from the last successful build log with `build.sh`:
```
+ ./generate_drawing_examples.sh
error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installed
   Compiling libc v0.2.73
   Compiling png v0.16.7
   Compiling embedded-graphics v0.6.2 (/home/circleci/project/embedded-graphics)
   Compiling num_cpus v1.13.0
   Compiling rand_os v0.1.3
   Compiling sdl2-sys v0.32.6
   Compiling rand v0.6.5
   Compiling rayon-core v1.7.1
   Compiling rayon v1.3.1
   Compiling tinytga v0.3.2 (/home/circleci/project/tinytga)
   Compiling sdl2 v0.32.2
   Compiling jpeg-decoder v0.1.20
   Compiling image v0.23.7
   Compiling embedded-graphics-simulator v0.2.0 (/home/circleci/project/simulator)
    Finished dev [unoptimized + debuginfo] target(s) in 17.11s
     Running `target/debug/generate-drawing-examples`
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', src/libstd/io/stdio.rs:792:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

rfuest:
Can we add the entire `/usr/local/cargo` directory to the cache to keep the `just` and `cargo-readme` binaries between runs?

jamwaffles:
Yeah I don't see why not. You'll need to bump the `v` version number in the cache key though as it won't pick the changed config up.

The other option would be to use a custom docker image with all the tooling we want preinstalled.

rfuest:
> The other option would be to use a custom docker image with all the tooling we want preinstalled.

That is a good option, but the cache option would be easier to implement.

Can you to try adding the nightly toolchain and adjust the caching?

justfile
jamwaffles: Can you link the build that failed with that error?

jamwaffles:
Sure thing

justfile
jamwaffles: Oh doy I found it, never mind. Should've just scrolled up...

jamwaffles:
`precheck-stable` passed!

Perhaps the broken pipe error was... because the pipe was broken! That command required nightly rustfmt which, once installed, makes it work again :facepalm: 

I'll rerun the workflow again and see if the cache changes have made a difference.

jamwaffles:
It doesn't like the whole cargo folder cache - on restoring it, it fails on a bunch of permission errors. I've gone back to just caching the `registry` sub folder for now. I think we'll just have to swallow the build times, but I'll look into a custom Docker image to speed it up a bit.

jamwaffles:
Alright, I'm done with the CI changes. Mark the PR as ready for review when you want me to take a look at it. In the short time I've looked at `just`, I'm really enjoying it so far so thanks for adding it to the repo. Definitely much better than a pile of bash scripts.

PR #415, opened 2020-07-29
Replace linkchecker with cargo-deadlinks

bugadani:

## PR description

Let's see how this goes. One downside may be that deadlinks hasn't been updated for a while, though, so this may or may not be acceptable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,8 @@ precheck_steps: &precheck_steps
     - checkout
     - restore_cache:
         key: v5-embedded-graphics-{{ .Environment.CIRCLE_JOB }}-{{ checksum "embedded-graphics/Cargo.toml" }}-{{ checksum "tinybmp/Cargo.toml" }}-{{ checksum "tinytga/Cargo.toml" }}-{{ checksum "simulator/Cargo.toml" }}
-    - run: sudo apt update && sudo apt install -qq libsdl2-dev linkchecker
-    - run: cargo install cargo-readme just
+    - run: sudo apt update && sudo apt install -qq libsdl2-dev
+    - run: cargo install cargo-readme just cargo-deadlinks
     - run: rustup default ${RUST_VERSION:-stable}
     - run: rustup toolchain add nightly -c rustfmt
     - run: rustup component add rustfmt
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -29,8 +29,4 @@ for crate in "${CRATES[@]}"; do
 done
```
rfuest, line 29: It seems that I forgot to remove this file in #410. If you want you can delete it.

bugadani: You mean the whole script?

rfuest: Yes, it was replaced by the `justfile`.

bugadani: Thanks, I'll do that in a different PR.

jamwaffles:
That looks like a _very_ dormant repository which is unfortunate, but it appears the tool does work. This could be a good, platform-agnostic standin until intra-doc links are stabilised. What do you think @rfuest? I'd be happy to move more of the build dependencies into Rust tools.

The CI time has lengthened again, so it might be time to create a custom build image in another PR.

rfuest:
> What do you think @rfuest?

The Python linkchecker seems to be more actively developed but doesn't seem to be in a perfect state either because the Python 3 version needs to be installed from the git repo. So I think it's OK to use `cargo-deadlinks` at the moment.

> The CI time has lengthened again, so it might be time to create a custom build image in another PR.

Yes, I think that's a good idea. On the other side the link check itself seems to be much faster. On my aging system linkcheck takes 41 seconds while deadlinks uses less than 1 second. I hope this improvement doesn't mean that the check isn't as thorough and just means that Rust is far superior to Python :wink:.

jamwaffles:
> The Python linkchecker seems to be more actively developed but doesn't seem to be in a perfect state either

One day, intra-doc links will stabilise and this whole problem will go away :see_no_evil: 

>  I hope this improvement doesn't mean that the check isn't as thorough and just means that Rust is far superior to Python

Hopefully it's the latter :crossed_fingers:

bugadani:
> One day, intra-doc links will stabilise and this whole problem will go away

It's marked for 1.47.0

If something goes bad, this can always be reverted.

jamwaffles:
> It's marked for 1.47.0

TIL. That's exciting!

bugadani:
> > It's marked for 1.47.0
> 
> TIL. That's exciting!

Don't get too excited, someone will have to go through the links :P

jamwaffles:
porting

PR #419, opened 2020-07-30
Move to custom CI Docker image

### LINKED ISSUE #418: Speed up CI build with custom Docker image
After merging #415, the build time has increased significantly due to all the `cargo install`.

It might help to create a custom Docker image with the tooling this project needs preinstalled.

jamwaffles:

## PR description

Closes #418


jamwaffles:
We've gone from ~12m down to ~8m now. Mild progress.

jamwaffles:
I couldn't find any more low hanging fruit so I'll put this up for review. I think we'd see some improvements in removing some of the commands we run in the build itself, but they all seem kind of necessary.

Another option might be to migrate to Github Actions if it supports build instances with more than the 2 cores CircleCI provides. For example, my 8c/16t machine runs `just build` in 1m 27s whereas CircleCI takes over 5m to do the same.

Yet another option would be to split the `just build` command up and take advantage of Circle's build-level parallelism to split up the `precheck-stable` and `precheck-beta` build steps.

I can also start paying for Circle's larger instance sizes at ~$30/mo which I'd be ok with if there's a pressing need.

jamwaffles:
I changed it to build all the targets in one image. We're now under 7m:

image

justfile
```diff
@@ -132,3 +133,8 @@ generate-example-screenshot example:
         just generate-example-screenshot $(basename "$example" .rs); \
     done
 
+build-ci-image:
```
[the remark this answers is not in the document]
jamwaffles: Yeah I really like `just`. My only regret is not using it sooner, so thank you for introducing it.

I've added some comments to those two commands.

rfuest:
The next build should be even faster, now that `all-target` can use the cache.

justfile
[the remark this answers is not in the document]
rfuest: Comments were also missing, but I meant a header above the docker section:
```
#--------
# Docker
#--------
```

jamwaffles:
Spicy

image

justfile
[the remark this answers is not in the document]
jamwaffles: Ahh I see, that's a good idea. Added a header too.

jamwaffles:
On the Github side, should we require all three builds pass before merging? This would include requiring `precheck-beta` to pass. To my knowledge, it's never failed. I don't see why we shouldn't mark all 3 as required, but you might see something I don't.

.circleci/config.yml
```diff
@@ -22,23 +19,25 @@ precheck_steps: &precheck_steps
...
-    - run: cargo install just
+        TARGETS=("arm-unknown-linux-gnueabi" "armv7-unknown-linux-gnueabihf" "x86_64-unknown-linux-gnu" "x86_64-unknown-linux-musl" "thumbv6m-none-eabi" "thumbv7em-none-eabi" "thumbv7em-none-eabihf" "thumbv7m-none-eabi")
+
+        for TARGET in "${TARGETS[@]}"; do
+          if [[ ! "$SYSROOT" =~ "$TARGET" ]]; then
+            rustup target add $TARGET
+          else
+            echo "Target $TARGET is already installed"
+          fi
+        done
```
rfuest, line 38: Should we move this to the `justfile`? This way we wouldn't need to maintain two target lists.

rfuest:
> On the Github side, should we require all three builds pass before merging? This would include requiring `precheck-beta` to pass. To my knowledge, it's never failed. I don't see why we shouldn't mark all 3 as required, but you might see something I don't.

I can't think of any reason not to include it and also can't remember any instance where `beta` failed on its own, unlike `nightly` which we removed for that reason.

.circleci/config.yml
jamwaffles: I did consider that to reduce the duplication. I decided not to originally because I thought it would be strange having a command that installs a bunch of targets, but then again we do have the command to _build_ those targets in the justfile already, so I'll move this script into it too.

jamwaffles:
> I can't think of any reason not to include it and also can't remember any instance where beta failed on its own, unlike nightly which we removed for that reason.

Cool. When this PR is merged I'll change the Github settings to suit.

.circleci/config.yml
rfuest: We could also do the opposite and call `just build-target $TARGET` in a loop from `config.yml`.

jamwaffles: We could, but I think I prefer the symmetry of `just install-targets` and `just build-targets` so I went with that.

justfile
```diff
@@ -43,6 +45,28 @@ build-target target *args:
     cargo build -p tinybmp --target {{target}} {{args}}
     cargo build -p tinybmp --target {{target}} --all-features {{args}}
 
+# Cross compiles embedded-graphics tinytga and tinybmp for all targets
```
rfuest, line 48:
```suggestion
# Cross compiles embedded-graphics, tinytga and tinybmp for all targets
```

justfile
```diff
@@ -43,6 +45,28 @@ build-target target *args:
...
+build-targets *args:
+    #!/usr/bin/env bash
+    set -e
+
+    for target in {{targets}}; do just build-target $target {{args}}; done
```
rfuest, line 53: Bash shouldn't be required for this task:
```suggestion
    for target in {{targets}}; do just build-target $target {{args}}; done
```

jamwaffles:
Thanks for the quick review!

rfuest:
> Thanks for the quick review!

Thanks for the quick CI builds!

jamwaffles:
No problem! This should get PR turnaround times down a bit.

Issue #424, opened 2020-08-10
Support images with unknown color format


rfuest:
The color format for images must currently be known at compile time, which works OK for images that are bundled with an application. But this is inconvenient for user supplied images that are handled by `tinytga` or `tinybmp`, because it requires the user to generate images with a specific color format.

I would suggest to add additional `DynamicTga` and `DynamicBmp` types to the tiny crates, that convert the color internally to `Rgb888` and implement `ImageDrawable<Color = Rgb888>` to provide a generic interface. Drawing these images will require color conversion for non `Rgb888` displays, but this was already intended to be implemented as another `DrawTarget` adapter.

The `Tga` and `Bmp` types will remain to provide a higher performance option for known color formats. But they could be enhanced by checking if the BPP specified in the image header matches the size of `PixelColor::Raw` (also related to #413).

rfuest:
Support for images with unknown color formats is now available in `tinytga` and `tinybmp` by using the new `DynamicTga` and `DynamicTga` types.

PR #430, opened 2020-08-15
Improved error handling for tinytga

rfuest:

## PR description

The reason to create this PR was to overhaul the error handing in tinytga. But it grew in scope and now also contains other changes. Sorry for the large monolithic PR.

Overview of the changes:
- `embedded-graphics` is now required and the `graphics` feature is removed.
- `TgaRaw` is merged into `Tga` and can be used by calling `Tga::from_slice_raw` which creates a `Tga<()>`.
- Unknown BPP values don't cause `panic`s anymore and instead `from_slice` now returns an error.
- `Tga::from_slice` now also checks that the bit depth matches the specified e-g color type and returns an error if not.
- The iterators now handle errors by returning black pixels until the image is complete.
- A `ColorMap` struct was added to allow easier access to the color map.
  This currently only supports raw color values and not e-g types, but this will be added later in the dynamic color type PR.
- `raw_developer_dictionary` and `raw_extension_area` replace the `TgaFooter` type.
  The footer only contained the offset and didn't allow access to the actual data. This could later be extended to parse the areas instead of only returning the raw data.
- The image ID can now be accessed by the `image_id` method.
- To save space in the `Tga` struct some infrequently used methods parse the header or footer each time they are called.
- `width()` and `height()` were replaced by `size` from the `OriginDimensions` trait.
- `IntoIterator`  was replaced by `pixels` and `raw_pixels`.
- `RawPacket` and `RlePacket` are now merged into a single `Packet` struct that also implements `Iterator`.
   I hope this makes the code easier to understand.

The docs could use some more polish, but I would like to work on them in the following PR that will implement #424. And the changelog is also missing, which I will add after the first review pass.


tinytga/src/lib.rs
```diff
@@ -93,324 +110,216 @@
...
-                    self.bytes_to_consume = bytes_to_consume;
-                    self.current_packet_pixel_length = current_packet.len() / self.stride;
-                    self.current_packet = current_packet;
-                }
-                image_type => panic!("Image type {:?} not supported", image_type),
-            };
-            // }
-        }
+impl<'a> Tga<'a, ()> {
+    /// Parse a TGA image from a byte slice
```
jamwaffles, line 285: This could use some clarification around colour depth validation and/or its differences from `from_slice`.

tinytga/src/color_map.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /// Returns the bit depth for the entries in the color map.
+    pub fn entry_bpp(&self) -> Bpp {
+        self.entry_bpp
+    }
+
+    /// Returns the raw color value for a color map entry.
+    pub fn get_raw(&self, index: usize) -> Option<u32> {
+        //TODO: use start_index
```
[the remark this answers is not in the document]
rfuest: I didn't have a TGA file with `start_index != 0` to test this feature. But it is on the TODO list for the next PR.

tinytga/src/header.rs
```diff
@@ -1,13 +1,63 @@
...
     number::complete::{le_u16, le_u8},
     IResult,
 };
 
-/// TGA footer length in bytes
-pub const HEADER_LEN: usize = 18;
+/// Bits per pixel.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[non_exhaustive]
+pub enum Bpp {
```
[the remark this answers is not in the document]
rfuest: It wouldn't be very efficient to use TGA for these bit depths, because each pixel consists of an integer number of bytes. But I guess it should be possible to allow this anyway. I'll add it to the TODO for the next PR.

tinytga/src/lib.rs
rfuest: As I said in the original comment the docs still need some work. Would it be OK to add this in the next PR? The additional `DynamicTga` type will make this even more confusing and there should be one overview to show when to use which type IMO.

jamwaffles: Yeah that's fine. I added some comment suggestions in case you wanted to add some of them in now.

rfuest: Thanks for the suggestions, I've already added them. I didn't mean that these changes weren't appreciated, but I didn't want to cause extra work if I decide to do bigger changes in the next PR which could make some of the added docs invalid.

tinytga/CHANGELOG.md
```diff
@@ -9,16 +9,24 @@
 ### Changed
 
 - **(breaking)** [#407](https://github.com/jamwaffles/embedded-graphics/pull/407) The `image_descriptor` in `TgaHeader` was replaced by `image_origin` and `alpha_channel_bits`.
-- **(breaking)** [#407](https://github.com/jamwaffles/embedded-graphics/pull/407) The `Pixel` type returned by `TgaIterator` now uses `u16` coordinates.
-- **(breaking)** [#420](https://github.com/jamwaffles/embedded-graphics/pull/420) To support the new embedded-graphics 0.7 image API a color type parameter was added to `Tga`. To use this crate without the `graphics` flag enabled replace `Tga` by `TgaRaw`.
+- **(breaking)** [#420](https://github.com/jamwaffles/embedded-graphics/pull/420) To support the new embedded-graphics 0.7 image API a color type parameter was added to `Tga`.
+- **(breaking)** [#430](https://github.com/jamwaffles/embedded-graphics/pull/430) The `graphics` feature was removed and the `embedded-graphics` dependency is now non optional.
+- **(breaking)** [#430](https://github.com/jamwaffles/embedded-graphics/pull/430) `Tga` does no longer implement `IntoIterator`. Pixel iterators can now be created using the `pixels` an `raw_pixels` methods.
```
jamwaffles, line 14:
```suggestion
- **(breaking)** [#430](https://github.com/jamwaffles/embedded-graphics/pull/430) `Tga` no longer implements `IntoIterator`. Pixel iterators can now be created using the `pixels` and `raw_pixels` methods.
```

PR #431, opened 2020-08-16
Stroke dimensions

### LINKED ISSUE #428: Dimensions for styled primitives don't use stroke width and alignment
The `Dimensions` impl for styled primitives returns the same dimensions as the base primitive, but the actual dimensions could be larger if the stroke alignment is set to `Outside` or `Center` and the stroke width is non zero.

jamwaffles:

## PR description

Closes #428 

I've only added a test for `RoundedRect` so far as I'd like @rfuest's feedback on its structure. Once we've got that sorted I'll add tests to the rest of the primitives.

### DIFF
--- a/embedded-graphics/src/mock_display.rs
+++ b/embedded-graphics/src/mock_display.rs
@@ -172,7 +172,7 @@
 use crate::{
     draw_target::DrawTarget,
     drawable::Pixel,
-    geometry::{OriginDimensions, Point, Size},
+    geometry::{Dimensions, OriginDimensions, Point, Size},
     pixelcolor::{
         Bgr555, Bgr565, Bgr888, BinaryColor, Gray2, Gray4, Gray8, GrayColor, PixelColor, Rgb555,
         Rgb565, Rgb888, RgbColor,
@@ -238,21 +238,16 @@ where
         self.pixels[i as usize] = color;
     }
 
-    /// Get the affected area of the display.
+    /// Returns the area that was affected by drawing operations.
     pub fn affected_area(&self) -> Rectangle {
         let (tl, br) = self
-            .pixels
-            .iter()
-            .enumerate()
-            .filter_map(|(idx, p)| p.map(|_| idx))
+            .bounding_box()
+            .points()
+            .zip(self.pixels.iter())
+            .filter_map(|(point, color)| color.map(|_| point))
             .fold(
                 (None, None),
-                |(tl, br): (Option<Point>, Option<Point>), idx| {
-                    let x = idx % SIZE;
-                    let y = idx / SIZE;
-
-                    let point = Point::new(x as i32, y as i32);
-
+                |(tl, br): (Option<Point>, Option<Point>), point| {
                     (
                         tl.map(|tl| tl.component_min(point)).or(Some(point)),
                         br.map(|br| br.component_max(point)).or(Some(point)),
@@ -260,10 +255,11 @@ where
                 },
             );
 
-        Rectangle::with_corners(
-            tl.unwrap_or_else(Point::zero),
-            br.unwrap_or_else(Point::zero),
-        )
+        if let (Some(tl), Some(br)) = (tl, br) {
+            Rectangle::with_corners(tl, br)
+        } else {
+            Rectangle::new(Point::zero(), Size::zero())
+        }
     }
 
     /// Changes the color of a pixel.
@@ -779,5 +775,15 @@ mod tests {
     #[should_panic(expected = "invalid char in pattern: 'G'")]
     fn invalid_gray8_char_g() {
         Gray8::char_to_color('G');
+    }
+
+    #[test]
+    fn zero_sized_affected_area() {
+        let disp: MockDisplay<BinaryColor> = MockDisplay::new();
+
+        assert_eq!(
+            disp.affected_area(),
+            Rectangle::new(Point::zero(), Size::zero())
+        );
     }
 }
--- a/embedded-graphics/src/mock_display.rs
+++ b/embedded-graphics/src/mock_display.rs
@@ -200,7 +200,6 @@ where
     pixels: [Option<C>; SIZE * SIZE],
     allow_overdraw: bool,
     allow_out_of_bounds_drawing: bool,
-    affected_area: Option<Rectangle>,
 }
 
 impl<C> MockDisplay<C>
@@ -235,24 +234,36 @@ where
 
     /// Changes the value of a pixel without bounds checking.
     pub fn set_pixel(&mut self, point: Point, color: Option<C>) {
-        if let Some(ref mut area) = self.affected_area {
-            *area = Rectangle::with_corners(
-                area.top_left.component_min(point),
-                area.bottom_right()
-                    .map(|bottom_right| bottom_right.component_max(point))
-                    .unwrap_or(point),
-            );
-        } else {
-            self.affected_area = Some(Rectangle::new(point, Size::new_equal(1)));
-        }
-
         let i = point.x + point.y * SIZE as i32;
         self.pixels[i as usize] = color;
     }
 
     /// Get the affected area of the display.
-    pub fn affected_area(&self) -> Option<Rectangle> {
-        self.affected_area
+    pub fn affected_area(&self) -> Rectangle {
+        let (tl, br) = self
+            .pixels
+            .iter()
+            .enumerate()
+            .filter_map(|(idx, p)| p.map(|_| idx))
+            .fold(
+                (None, None),
+                |(tl, br): (Option<Point>, Option<Point>), idx| {
+                    let x = idx % SIZE;
+                    let y = idx / SIZE;
+
+                    let point = Point::new(x as i32, y as i32);
+
+                    (
+                        tl.map(|tl| tl.component_min(point)).or(Some(point)),
+                        br.map(|br| br.component_max(point)).or(Some(point)),
+                    )
+                },
+            );
+
+        Rectangle::with_corners(
+            tl.unwrap_or_else(Point::zero),
+            br.unwrap_or_else(Point::zero),
+        )
     }
 
     /// Changes the color of a pixel.
@@ -486,7 +497,6 @@ where
             pixels: [None; SIZE * SIZE],
             allow_overdraw: false,
             allow_out_of_bounds_drawing: false,
-            affected_area: None,
         }
     }
 }
--- a/embedded-graphics/src/mock_display.rs
+++ b/embedded-graphics/src/mock_display.rs
@@ -200,6 +200,7 @@ where
     pixels: [Option<C>; SIZE * SIZE],
     allow_overdraw: bool,
     allow_out_of_bounds_drawing: bool,
+    affected_area: Option<Rectangle>,
 }
 
 impl<C> MockDisplay<C>
@@ -234,8 +235,24 @@ where
 
     /// Changes the value of a pixel without bounds checking.
     pub fn set_pixel(&mut self, point: Point, color: Option<C>) {
+        if let Some(ref mut area) = self.affected_area {
+            *area = Rectangle::with_corners(
+                area.top_left.component_min(point),
+                area.bottom_right()
+                    .map(|bottom_right| bottom_right.component_max(point))
+                    .unwrap_or(point),
+            );
+        } else {
+            self.affected_area = Some(Rectangle::new(point, Size::new_equal(1)));
+        }
+
         let i = point.x + point.y * SIZE as i32;
         self.pixels[i as usize] = color;
+    }
+
+    /// Get the affected area of the display.
+    pub fn affected_area(&self) -> Option<Rectangle> {
+        self.affected_area
     }
 
     /// Changes the color of a pixel.
@@ -469,6 +486,7 @@ where
             pixels: [None; SIZE * SIZE],
             allow_overdraw: false,
             allow_out_of_bounds_drawing: false,
+            affected_area: None,
         }
     }
 }
--- a/embedded-graphics/src/primitives/arc/styled.rs
+++ b/embedded-graphics/src/primitives/arc/styled.rs
@@ -213,7 +213,7 @@ mod tests {
         // TODO: Uncomment when arc bounding box is fixed in #405
         // let mut display = MockDisplay::new();
         // center.draw(&mut display).unwrap();
-        // assert_eq!(display.affected_area().unwrap(), center.bounding_box());
+        // assert_eq!(display.affected_area(), center.bounding_box());
     }
 
     #[test]
--- a/embedded-graphics/src/primitives/arc/styled.rs
+++ b/embedded-graphics/src/primitives/arc/styled.rs
@@ -1,7 +1,7 @@
 use crate::{
     draw_target::DrawTarget,
     drawable::{Drawable, Pixel},
-    geometry::Dimensions,
+    geometry::{Dimensions, Size},
     iterator::IntoPixels,
     pixelcolor::PixelColor,
     primitives::{
@@ -109,9 +109,13 @@ where
 {
     // FIXME: This doesn't take into account start/end angles. This should be fixed to close #405.
     fn bounding_box(&self) -> Rectangle {
-        let offset = self.style.outside_stroke_width().saturating_cast();
+        if self.style.is_transparent() {
+            Rectangle::new(self.primitive.bounding_box().center(), Size::zero())
+        } else {
+            let offset = self.style.outside_stroke_width().saturating_cast();
 
-        self.primitive.bounding_box().offset(offset)
+            self.primitive.bounding_box().offset(offset)
+        }
     }
 }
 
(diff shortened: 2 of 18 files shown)


rfuest:
I don't think it's possible to use a single `Dimension` implementation for all styled primitives. Thick lines, for example, need a `sqrt(2)` times higher offset if they are at 45° and they also ignore the stoke alignment, which the current implementation doesn't.

jamwaffles:
I didn't think about `Line`s... those are going to be tricky. I'll add individual impls for all the primitives and see what that looks like.

jamwaffles:
I've ported (and improved I think) a solution to the line bounding box I developed in the `line-joints` branch: I reuse the Bresenham algorithm to figure out the start points of the left/right edges of the thick line, then compute the end points to get two lines that represent the left and right edges of the thick line. With these 4 points, I can then get the bounding box.

It might be slower than `sqrt()`, but it 1. sticks to integer maths and 2. is pixel-accurate to the rendered thick line. I'm digging into quite a few of the thick line internals for this which is quite messy. You were the last to refactor the Bresenham stuff so some suggestions on how to factor the new code out better would be welcome.

rfuest:
This should work and only requires `ParallelsIterator::new` and `Bresenham::point` to be public in the `line` module:
```rust
    /// Get two lines representing the left and right edges of the thick line.
    fn extents(&self, thickness: i32) -> (Line, Line) {
        let mut it = ParallelsIterator::new(self, thickness);

        let mut l_start = self.start;
        let mut r_start = self.start;

        loop {
            if let Some((bresenham, _)) = it.next() {
                r_start = bresenham.point;
            } else {
                break;
            }

            if let Some((bresenham, _)) = it.next() {
                l_start = bresenham.point;
            } else {
                break;
            }
        }

        let delta = self.end - self.start;

        (
            Line::new(l_start, l_start + delta),
            Line::new(r_start, r_start + delta),
        )
    }
}
```

`l_start` and `r_start` are initialized to `self.start` instead of `Point::zero()` because otherwise 1 px wide lines would extend to the origin on the left side. I haven't tested the behavior for 0px wide lines.

I was thinking about adding an `affected_area` method to `MockDisplay`, which returns a bounding box around all changed pixels. This should make it easier to test if the `bounding_box` impl matches the drawn pixels. Should I open a PR for that feature?

rfuest:
> I've ported (and improved I think) a solution to the line bounding box I developed in the `line-joints` branch: I reuse the Bresenham algorithm to figure out the start points of the left/right edges of the thick line, then compute the end points to get two lines that represent the left and right edges of the thick line. With these 4 points, I can then get the bounding box.
> 
> It might be slower than `sqrt()`, but it 1. sticks to integer maths and 2. is pixel-accurate to the rendered thick line.

Sounds like a good idea and I also prefer to stick to integer math if possible.

jamwaffles:
That's a lot cleaner, thank you!

jamwaffles:
Ready for a first review. The tests could probably use some work but it's something to start from as it is.

jamwaffles:
> Did you consider the addition to MockDisplay I suggested above? This would make sure that the bounding boxes are pixel perfect.

I didn't see that before but it's a good suggestion. I'll add it as part of this PR.

embedded-graphics/src/primitives/polyline/styled.rs
```diff
@@ -72,6 +76,17 @@ where
     }
 }
 
+impl<C> Dimensions for Styled<Polyline<'_>, PrimitiveStyle<C>>
+where
+    C: PixelColor,
+{
+    // NOTE: Polyline currently ignores stroke width, so this delegates to the un-styled bounding
+    // box impl.
```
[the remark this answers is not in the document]
jamwaffles: I don't think so - I'll add this as part of #386

embedded-graphics/src/primitives/arc/styled.rs
```diff
@@ -102,6 +103,18 @@ where
     }
 }
 
+impl<C> Dimensions for Styled<Arc, PrimitiveStyle<C>>
+where
+    C: PixelColor,
+{
+    // FIXME: This doesn't take into account start/end angles. This should be fixed to close #405.
+    fn bounding_box(&self) -> Rectangle {
```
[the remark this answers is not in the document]
jamwaffles: This would also be the same for all the other primitives, no? If they're transparent, the bounding box would be zero sized. For consistency, should that be represented as `Rectangle::new(Point::zero(), Size::zero())` or `Rectangle::new(primitive_top_left, Size::zero())` in all cases?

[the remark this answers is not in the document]
rfuest: > This would also be the same for all the other primitives, no?

Yes, I forgot to consider completely transparent primitives. `Arc` and `Line` are special because they don't use the fill.

> For consistency, should that be represented as `Rectangle::new(Point::zero(), Size::zero())` or `Rectangle::new(primitive_top_left, Size::zero())` in all cases?

I would prefer to have the point be inside the primitive. Normally the `top_left` shouldn't matter for a zero sized rectangle, but I encountered some problems with this some time ago, but I cannot remember the details at the moment. How about `Rectangle::new(primitive.bounding_box().center(), Size::zero())`?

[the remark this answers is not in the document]
jamwaffles: Sounds good. I'll add that in.

jamwaffles:
I've added a bunch of tests to check the affected area in `MockDisplay` as suggested which surfaced a couple of bugs. One, on the thick line, I think might be due to the fact that `extents()` adds a simple delta for the end point of each edge which might not be the same as where the Bresenham point ends up.

The only solution I can think of right now would be to consume the parallel Bresenham iterators for each edge to get their true end points, which seems quite slow. Do you have any other suggestions?

rfuest:
> The only solution I can think of right now would be to consume the parallel Bresenham iterators for each edge to get their true end points, which seems quite slow. Do you have any other suggestions?

I don't have time looked at the code at the moment, but if it is off by 1 px it could be caused by the shorter extra lines. Here is a bit of pseudo code how this could be handled:
```
    fn extents(&self, thickness: u32) -> (Line, Line) {
        let mut it = ParallelsIterator::new(self, thickness.saturating_cast());

        let mut left = (self.start, 0);
        let mut right = (self.start, 0);

        loop {
            if let Some(r) = it.next() {
                right = r;
            } else {
                break;
            }

            if let Some(l) = it.next() {
                left = l;
            } else {
                break;
            }
        }

        let delta = self.end - self.start;

        let left_line = if l.1 == 0 {
            Line::new(l_start, l_start + delta)
        }  else {
           //TODO: reduce line length by l.1 major steps
        }

        // the same for the right line

        (left_line, right_line)
    }
```

jamwaffles:
Thanks for the suggestion - that works with a couple of tweaks.

embedded-graphics/src/primitives/line/styled.rs
```diff
@@ -79,3 +83,117 @@ where
         display.draw_iter(self.into_pixels())
     }
 }
+
+impl<C> Dimensions for Styled<Line, PrimitiveStyle<C>>
+where
+    C: PixelColor,
+{
+    fn bounding_box(&self) -> Rectangle {
+        if self.style.is_transparent() {
```
rfuest, line 92: See comment for `Arc`.

embedded-graphics/src/primitives/polyline/styled.rs
```diff
@@ -72,6 +76,21 @@ where
...
+    fn bounding_box(&self) -> Rectangle {
+        if self.style.is_transparent() {
```
rfuest, line 86: See comment for `Arc`.

simulator/examples/line-thickness.rs
```diff
@@ -47,13 +47,17 @@ fn draw(
...
-            Rgb888::new(0x80, 0xf2, 0x91),
-            stroke_width,
-        ))
+    let l = Line::new(start, position).into_styled(PrimitiveStyle::with_stroke(
+        Rgb888::new(0x80, 0xf2, 0x91),
+        stroke_width,
+    ));
+
+    l.bounding_box()
+        .into_styled(PrimitiveStyle::with_stroke(Rgb888::RED, 1))
```
rfuest, line 56: Do you want to keep the bounding boxes in the examples?

embedded-graphics/src/primitives/line/styled.rs
```diff
@@ -79,3 +83,117 @@ where
...
+    use super::*;
+    use crate::{
+        geometry::{Point, Size},
+        mock_display::MockDisplay,
+        pixelcolor::{Rgb888, RgbColor},
+        primitives::{Primitive, Rectangle},
+    };
+
+    #[test]
+    fn bounding_box() {
```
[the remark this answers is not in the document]
jamwaffles: Good idea. I've added a loop up to a thickness of 15px.

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -78,6 +79,48 @@ impl Line {
...
         Line::new(self.start, self.start + delta)
     }
+
+    /// Get two lines representing the left and right edges of the thick line.
+    ///
+    /// If a thickness of `0` is given, the lines returned will lie on the same points as `self`.
+    fn extents(&self, thickness: u32) -> (Line, Line) {
+        let mut it = ParallelsIterator::new(self, thickness.saturating_cast());
+        let reduce =
+            it.parallel_parameters.position_step.major + it.parallel_parameters.position_step.minor;
```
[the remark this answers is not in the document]
jamwaffles: I discovered this fix by adding a bounding box to the `line-thickness` demo and scientifically dragging the line around with various thicknesses. It doesn't prove 100% that this fix works, but in all the dragging I did I didn't see an error. 

I've also added a loop as suggested below which is still passing. Is there anything else we could do to test this more empirically?

embedded-graphics/src/mock_display.rs
```diff
@@ -238,6 +238,34 @@ where
         self.pixels[i as usize] = color;
     }
 
+    /// Get the affected area of the display.
+    pub fn affected_area(&self) -> Rectangle {
```
rfuest, line 242: This should have a changelog entry, because it's public.
```suggestion
    /// Returns the area that was affected by drawing operations.
    pub fn affected_area(&self) -> Rectangle {
```

embedded-graphics/src/mock_display.rs
```diff
@@ -238,6 +238,34 @@ where
...
+        let (tl, br) = self
+            .pixels
+            .iter()
+            .enumerate()
```
rfuest, line 246: Would it be easier to get the point by using `self.bounding_box().points().zip(self.pixels.iter())`?

embedded-graphics/src/mock_display.rs
```diff
@@ -238,6 +238,34 @@ where
...
+                        tl.map(|tl| tl.component_min(point)).or(Some(point)),
+                        br.map(|br| br.component_max(point)).or(Some(point)),
+                    )
+                },
+            );
+
+        Rectangle::with_corners(
+            tl.unwrap_or_else(Point::zero),
+            br.unwrap_or_else(Point::zero),
+        )
```
rfuest, line 266: This will not produce a zero sized rectangle if `tl` and `br` are `None`. I would also suggest to add a test for this case.

embedded-graphics/src/primitives/arc/styled.rs
```diff
@@ -170,4 +187,55 @@ mod tests {
...
+        const SIZE: u32 = 10;
+
+        let empty = Arc::with_center(CENTER, SIZE - 4, 0.0.deg(), 90.0.deg());
+
+        assert_eq!(
+            empty
+                .into_styled::<BinaryColor>(PrimitiveStyle::new())
+                .bounding_box(),
+            Rectangle::new(CENTER, Size::zero())
+        );
```
rfuest, line 231:
```suggestion
            Rectangle::new(CENTER, Size::zero()),
            "empty"
        );
```

embedded-graphics/src/primitives/circle/styled.rs
```diff
@@ -279,4 +298,49 @@ mod tests {
...
+    #[test]
+    fn bounding_boxes() {
+        const CENTER: Point = Point::new(15, 15);
+        const SIZE: u32 = 10;
+
+        let style = PrimitiveStyle::with_stroke(BinaryColor::On, 3);
+
+        let center = Circle::with_center(CENTER, SIZE).into_styled(style);
+
+        let inside = Circle::with_center(CENTER, SIZE + 2).into_styled(
```
rfuest, line 311: If the test uses the `affected_area` function we don't need to calculate the size anymore.

embedded-graphics/src/primitives/ellipse/styled.rs
```diff
@@ -293,4 +312,50 @@ mod tests {
...
+                .build(),
+        );
+        let outside = Ellipse::with_center(CENTER, SIZE - Size::new_equal(4)).into_styled(
+            PrimitiveStyleBuilder::from(&style)
+                .stroke_alignment(StrokeAlignment::Outside)
+                .build(),
+        );
+
+        assert_eq!(center.bounding_box(), inside.bounding_box());
+        assert_eq!(outside.bounding_box(), inside.bounding_box());
```
rfuest, line 336: I would remove this and the size calculation above.

embedded-graphics/src/primitives/line/mod.rs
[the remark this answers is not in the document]
rfuest: I also couldn't break it, so I guess we just declare it to be correct.

embedded-graphics/src/primitives/line/mod.rs
```diff
@@ -78,6 +79,48 @@ impl Line {
...
+            if let Some((bresenham, reduce)) = it.next() {
+                left = (bresenham.point, reduce);
+            } else {
+                break;
+            }
+        }
+
+        let delta = self.end - self.start;
+
+        let left_line = if left.1 == 0 {
```
rfuest, line 110: Should we change the `Item` type of the `ParallelsIterator`? This code only works for `0` and `1` and it's strange to use an `u32` for that purpose. I would suggest to add an `enum ParallelLineType { Normal, Extra }` and use that instead of the `u32`.

embedded-graphics/src/primitives/rectangle/styled.rs
```diff
@@ -378,4 +394,56 @@ mod tests {
...
+        assert_eq!(
+            inside.bounding_box(),
+            Rectangle::new(Point::new(10, 10), Size::new(15, 20)),
+            "inside"
+        );
+        assert_eq!(
+            outside.bounding_box(),
+            Rectangle::new(Point::new(5, 5), Size::new(25, 30)),
+            "outside"
+        );
```
rfuest, line 425: Do we need these additional assertions? If you want to keep them I would at least use `offset` like in the rounded rectangle test.

embedded-graphics/src/primitives/triangle/styled.rs
```diff
@@ -262,4 +283,29 @@ mod tests {
...
         styled.draw(&mut display).unwrap();
     }
+
+    #[test]
+    fn bounding_box() {
+        let triangle = Triangle::new(Point::new(10, 10), Point::new(30, 20), Point::new(20, 25));
+
+        let styled = triangle.into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 30));
+
+        assert_eq!(triangle.bounding_box(), styled.bounding_box());
```
rfuest, line 293: This should also have a `FIXME` comment.

rfuest (changes requested):
Some more nits.

embedded-graphics/src/primitives/line/mod.rs
jamwaffles: It is a bit strange, but it's not causing any issues at the moment. Maybe change this in a future PR if it becomes more of a papercut?

jamwaffles:
Builds are nice and quick now! 3m 20s isn't something to complain about :)

embedded-graphics/src/primitives/line/mod.rs
rfuest: IMO this change belongs into this PR, because it's the first instance where a length reduction > 1 wouldn't be handled correctly. But if you don't want to dig into the other code at the moment you can leave it and I will create a PR for the change.

jamwaffles: Good point. I've added a new commit to make the change in this PR.

embedded-graphics/src/primitives/line/thick_points.rs
```diff
@@ -27,6 +27,12 @@ impl Side {
     }
 }
 
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub(in crate::primitives::line) enum ParallelLineType {
+    Normal = 0,
+    Extra = 1,
+}
+
```
rfuest, line 35: See comment below.
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
pub(in crate::primitives::line) enum ParallelLineType {
    Normal,
    Extra,
}

```

rfuest (approved):

jamwaffles:
Thanks for the many reviews :). One step closer to 0.7.0-alpha.1!

PR #432, opened 2020-08-17
Fix translating a translated polyline

bugadani:


embedded-graphics/src/primitives/polyline/mod.rs
```diff
@@ -122,13 +122,15 @@ impl<'a> Transform for Polyline<'a> {
     ///
     /// let polyline = Polyline::new(&points);
     /// let moved = polyline.translate(Point::new(10, 12));
+    /// let moved2 = moved.translate(Point::new(10, 12));
```
rfuest, line 125: This shouldn't clutter the usage example IMO. Could you move this to a separate test?

bugadani: Of course, I can do anything :P

embedded-graphics/src/primitives/polyline/mod.rs
```diff
@@ -236,4 +236,22 @@ mod tests {
...
+            Point::new(10, 10),
+        ];
+
+        let polyline = Polyline::new(&points);
+        let moved = polyline.translate(Point::new(10, 12));
+        let moved2 = moved.translate(Point::new(10, 12));
+
+        assert_eq!(polyline.bounding_box().top_left, Point::new(5, 7));
+        assert_eq!(moved.bounding_box().top_left, Point::new(15, 19));
+        assert_eq!(moved2.bounding_box().top_left, Point::new(25, 31));
```
jamwaffles, line 255: IMO these would be better off testing the complete bounding box, not just the `top_left` property. The sizes will be the same but it's good to check.

bugadani: Fair point, I'll do that.

> p.s. please hit the "request review" button when you've made changes - that way we get an email notification and don't lose the PR behind the couch :)

I really wouldn't like to generate noise, but sure :)

jamwaffles:
Thanks!

PR #434, opened 2020-08-18
Line joiners: scanline and lookback

bugadani:
This PR reworks how the triangles in a polyline are rendered. Instead of trying to get the mathematical point rendering to work, I've instead created a fast-ish scanline iterator to draw individual triangles, and introduced a check in ThickPoints, that, for each to-be-rendered pixel checks, if it was part of a previous triangle. This check is done for the last 3 triangles to avoid overdraw on the inside of bevelled corners.

# The scanline iterator

The new scanline iterator (`FillScanlineIterator` for now) returns the points of a triangle from top to bottom, from left to right. This implementation differs from the existing `ScanlineIterator` in that it does not differentiate between edge and inside points, and also only yields unique points, thus it has no overdraw.

Basically, it walk the outside edges, determines which are the limits of a line and for each line, creates a line point iterator to generate the points.

Maybe some of it's implementation can be used to improve the `ScanlineIterator` but I haven't bothered as it was tricky enough to get this to work.

This PR replaces the implementation behind Triangle::points() to use the new `FillScanlineIterator`.

# Lookback

Right now, the lookback in `ThickPoints` contains 3 Option<Triangle> fields and for each of those, checks if the point we want to return is inside any of the 3 previous triangles. This isn't particularly efficient. It should be possible to only store the unique points and reconstruct triangles, so the memory cost can be cut down from 9 points + extra to 5 points + a little less, but the runtime cost can't be reduced this way.

There are some development options:
 - Try to solve the overdraw by tuning the bevel rendering - a lookback of 1 triangle is still necessary to avoid drawing over the
   triangle edges.
 - Try to solve edge overdraw some other way. Maybe it is possible to only track vertical information of the previous triangle and 
   abuse the fact that the scanline iterator returns points in a given order.

---------------

This PR does NOT solve the general overdraw issue.

This PR is opened agains #386 so the code is not master-quality

### DIFF
--- a/embedded-graphics/src/primitives/line_joint.rs
+++ b/embedded-graphics/src/primitives/line_joint.rs
@@ -312,7 +312,7 @@ impl LineJoint {
         }
     }
 
-    /// Returns whether the joint is the end.
+    /// Returns whether the joint is the last one.
     pub fn is_end_joint(&self) -> bool {
         match self.kind {
             JointKind::End => true,
--- a/embedded-graphics/src/primitives/line_joint.rs
+++ b/embedded-graphics/src/primitives/line_joint.rs
@@ -311,4 +311,12 @@ impl LineJoint {
             _ => None,
         }
     }
+
+    /// Returns whether the joint is the end.
+    pub fn is_end_joint(&self) -> bool {
+        match self.kind {
+            JointKind::End => true,
+            _ => false,
+        }
+    }
 }
--- a/embedded-graphics/src/primitives/line_joint.rs
+++ b/embedded-graphics/src/primitives/line_joint.rs
@@ -173,12 +173,12 @@ impl LineJoint {
                 .segment_intersection(&second_edge_right)
                 || second_segment_end_edge.segment_intersection(&first_edge_right);
 
-            // Distance from midpoint to miter end point. This arbitrarily picks the left-side
-            // intersection, but the left/right intersections are symmetrical around `mid` anyway.
-            let miter_length_squared = Line::new(mid, l_intersection).length_squared();
-
             // Normal line: non-overlapping line end caps
             if !self_intersection_r && !self_intersection_l {
+                // Distance from midpoint to miter end point. This arbitrarily picks the left-side
+                // intersection, but the left/right intersections are symmetrical around `mid` anyway.
+                let miter_length_squared = Line::new(mid, l_intersection).length_squared();
+
                 // Intersection is within limit at which it will be chopped off into a bevel, so return
                 // a miter.
                 if miter_length_squared <= miter_limit {
--- a/embedded-graphics/src/primitives/polyline/thick_points.rs
+++ b/embedded-graphics/src/primitives/polyline/thick_points.rs
@@ -2,7 +2,7 @@ use crate::{
     prelude::Point,
     primitives::{
         polyline::triangle_iterator::TriangleIterator,
-        triangle::{self, FillScanlineIterator, Points, Triangle},
+        triangle::{Points, Triangle},
         ContainsPoint, Primitive,
     },
     style::StrokeAlignment,
--- a/embedded-graphics/src/primitives/polyline/thick_points.rs
+++ b/embedded-graphics/src/primitives/polyline/thick_points.rs
@@ -13,6 +13,8 @@ use crate::{
 pub(in crate::primitives) struct ThickPoints<'a> {
     triangle_iter: TriangleIterator<'a>,
     prev_triangle: Option<Triangle>,
+    prev_triangle2: Option<Triangle>,
+    prev_triangle3: Option<Triangle>,
     triangle: Triangle,
     points_iter: Points,
 }
@@ -29,6 +31,8 @@ impl<'a> ThickPoints<'a> {
 
             Self {
                 prev_triangle: None,
+                prev_triangle2: None,
+                prev_triangle3: None,
                 triangle,
                 triangle_iter,
                 points_iter,
@@ -39,6 +43,8 @@ impl<'a> ThickPoints<'a> {
     pub fn empty() -> Self {
         Self {
             prev_triangle: None,
+            prev_triangle2: None,
+            prev_triangle3: None,
             triangle: Triangle::empty(),
             triangle_iter: TriangleIterator::empty(),
             points_iter: Triangle::empty().points(),
@@ -52,10 +58,15 @@ impl<'a> Iterator for ThickPoints<'a> {
     fn next(&mut self) -> Option<Self::Item> {
         loop {
             if let Some(point) = self.points_iter.next() {
-                if !ContainsPoint::contains(&self.prev_triangle, point) {
+                if !ContainsPoint::contains(&self.prev_triangle, point)
+                    && !ContainsPoint::contains(&self.prev_triangle2, point)
+                    && !ContainsPoint::contains(&self.prev_triangle3, point)
+                {
                     return Some(point);
                 }
             } else {
+                self.prev_triangle3 = self.prev_triangle2;
+                self.prev_triangle2 = self.prev_triangle;
                 self.prev_triangle = Some(self.triangle);
                 self.triangle = self.triangle_iter.next()?;
                 self.points_iter = self.triangle.points();
--- a/embedded-graphics/src/primitives/polyline/thick_points.rs
+++ b/embedded-graphics/src/primitives/polyline/thick_points.rs
@@ -1,8 +1,9 @@
 use crate::{
     prelude::Point,
     primitives::{
+        Primitive,
         polyline::triangle_iterator::TriangleIterator,
-        triangle::{MathematicalPoints, Triangle},
+        triangle::{self, FillScanlineIterator, Triangle, Points},
         ContainsPoint,
     },
     style::StrokeAlignment,
@@ -14,7 +15,7 @@ pub(in crate::primitives) struct ThickPoints<'a> {
     triangle_iter: TriangleIterator<'a>,
     prev_triangle: Option<Triangle>,
     triangle: Triangle,
-    points_iter: MathematicalPoints,
+    points_iter: Points,
 }
 
 impl<'a> ThickPoints<'a> {
@@ -25,7 +26,7 @@ impl<'a> ThickPoints<'a> {
             let mut triangle_iter = TriangleIterator::new(points, width, alignment);
 
             let triangle = triangle_iter.next().unwrap_or_else(Triangle::empty);
-            let points_iter = triangle.mathematical_points();
+            let points_iter = triangle.points();
 
             Self {
                 prev_triangle: None,
@@ -41,7 +42,7 @@ impl<'a> ThickPoints<'a> {
             prev_triangle: None,
             triangle: Triangle::empty(),
             triangle_iter: TriangleIterator::empty(),
-            points_iter: MathematicalPoints::empty(),
+            points_iter: Triangle::empty().points(),
         }
     }
 }
@@ -58,7 +59,7 @@ impl<'a> Iterator for ThickPoints<'a> {
             } else {
                 self.prev_triangle = Some(self.triangle);
                 self.triangle = self.triangle_iter.next()?;
-                self.points_iter = self.triangle.mathematical_points();
+                self.points_iter = self.triangle.points();
             }
         }
     }
--- a/embedded-graphics/src/primitives/polyline/thick_points.rs
+++ b/embedded-graphics/src/primitives/polyline/thick_points.rs
@@ -54,16 +54,11 @@ impl<'a> Iterator for ThickPoints<'a> {
     fn next(&mut self) -> Option<Self::Item> {
         loop {
             if let Some(point) = self.points_iter.next() {
-                if !ContainsPoint::contains(&self.prev_triangle, point) {
-                    if self.triangle_iter.is_current_beveled() {
-                        if !ContainsPoint::contains(&self.prev_triangle2, point)
-                            && !ContainsPoint::contains(&self.prev_triangle3, point)
-                        {
-                            return Some(point);
-                        }
-                    } else {
-                        return Some(point);
-                    }
+                if !ContainsPoint::contains(&self.prev_triangle, point)
+                    && !ContainsPoint::contains(&self.prev_triangle2, point)
+                    && !ContainsPoint::contains(&self.prev_triangle3, point)
+                {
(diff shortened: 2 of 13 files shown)


bugadani:
There is still this one:
image

But the overdraw issue in the original PR (https://user-images.githubusercontent.com/800791/90419139-0883ad00-e0ae-11ea-9216-e00955c897ff.png) is gone

embedded-graphics/src/primitives/line_joint.rs
```diff
@@ -200,9 +200,9 @@ impl LineJoint {
                             kind: JointKind::Bevel {
                                 /// Must be counter-clockwise
                                 filler_triangle: Triangle::new(
-                                    first_edge_right.end,
                                     l_intersection,
                                     second_edge_right.start,
+                                    first_edge_right.end,
```
bugadani, line 205: I have no clue at all why this helps overdraw.

bugadani:
@jamwaffles for some reason I can't check CI but I believe the failures are the same as with the parent PR.

embedded-graphics/src/primitives/line_joint.rs
jamwaffles: It changes the winding order to be clockwise which I think has some effect on the point-in-triangle algorithm. There must be some numerical inaccuracies in that which cause issues for some edges with certain winding angles. That's my best guess anyway.

bugadani: Didn't I just rotate the triangle? I did not flip 2 points, but shifted everything.

bugadani:
In this case, the overdraw is caused by the fact that we don't know any more, where the last edge segment started, so the intersection returns a point that is correct segment-wise, but not on the edge of the polilyne.

image

embedded-graphics/src/primitives/line_joint.rs
jamwaffles: Sorry yeah, I've done similar things in the past but swapping the last two points instead of rotating.

jamwaffles:
IMO that overdraw is ok. I think it's pretty difficult to deal with all the self-intersecting cases, so that can be tackled in the future when we add transparency (or not, with a warning in the docs about it).

bugadani:
Hmm I think my first screenshot has the same issue. With that, I think there should be no more (unexplained) overdraws

bugadani:
This is a rollercoaster for sure.

image

But:
image

bugadani:
Scanline iterators are hard.
image

But they are fast(er). Compared to the previous method (checking for each pixel if it's in the triangle), this one runs fast enough. On a 6 core ryzen 5...

I have no idea why the iterator is broken, though and I gave up for today.

bugadani:
Current state of affairs:

I've implemented a scanline iterator that should not overdraw, but it can't differentiate edge points.
Results:
image
Almost as good as the triangle.contains() snail, but with a lot better perf and an added rendering artifact.

jamwaffles:
So close! Is the only remaining issue the overdraw at sharp edges now?

> looking back more triangles

I haven't had time to look at your code in detail, but what do you mean by this? I'd be ok consuming a bit more memory if it means we get no overdraw in ordinary joints. Similarly to the thick lines feature, IMO the first step is to just get it working with a good unit test suite. We can optimise in the future, using the tests to ensure the same behaviour.

FWIW CI is failing on a formatting check.

bugadani:
> I haven't had time to look at your code in detail

I plan on updating the PR description with what I've done so you can get an idea without looking at the code. Right now, I'm rendering the triangles using a new scanline iterator, and I keep track of the last rendered triangle to exclude some points - this works around the overdraw problem almost all the time.

Oh I don't think you'd want to waste 400 bytes of memory just to fix 2 pixels.

jamwaffles:
> Oh I don't think you'd want to waste 400 bytes of memory just to fix 2 pixels.

That's... quite a lot of memory :(. Where does it all come from?

Overdraw will bite us in the ass when we add transparency though so maybe it's good to start with a completely non-overdrawing solution, regardless of memory usage.

bugadani:
> Where does it all come from?

I kinda pulled that number out of thin air and is actually 3xOption<Triangle> which is hmm... 3 * (12 + 4) = 64, I think. But I also think I can do better :)

jamwaffles:
I don't know if this would be useful to you or not, but I did try a trapezium iterator in https://github.com/jamwaffles/embedded-graphics/pull/386/commits/b5dbd99c9bffb8ffefacf9dd21be334b91bf7db0 which might save you some memory. It replaces the two triangles that make up the straight line segment with one shape. Might save some memory if you can shoehorn it into your approach.

bugadani:
Maybe it could, I'll look at it.

Clippy is rather helpful, by the way:

```
warning: large size difference between variants
  --> embedded-graphics\src\primitives\polyline\styled.rs:17:5
   |
17 |     Thick(ThickPoints<'a>),
   |     ^^^^^^^^^^^^^^^^^^^^^^ this variant is 488 bytes
```

bugadani:
```
warning: large size difference between variants
  --> embedded-graphics\src\primitives\polyline\styled.rs:17:5
   |
17 |     Thick(ThickPoints<'a>),
   |     ^^^^^^^^^^^^^^^^^^^^^^ this variant is 400 bytes
```

Getting better.

embedded-graphics/src/primitives/polyline/triangle_iterator.rs
```diff
@@ -24,62 +30,59 @@ impl<'a> TriangleIterator<'a> {
...
-                EdgeCorners {
-                    left: left_start,
-                    right: right_start,
-                },
+            first_edge_end: EdgeCorners { left: left_end, .. },
             ..
+        } = end_joint;
+
+        // NOTE: Winding order is important here to prevent overdraw of the shared edge from
+        // right_start to left_end.
```
jamwaffles, line 79: Can you check if this is still valid?

embedded-graphics/src/primitives/polyline/triangle_iterator.rs
```diff
@@ -92,58 +95,66 @@ impl<'a> TriangleIterator<'a> {
 
         // NOTE: Winding order is important here to prevent overdraw of the shared edge from
         // right_start to left_end.
```
jamwaffles, line 97: Likewise here

embedded-graphics/src/primitives/triangle/fill_scanline_iterator.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+        Some((first, next_c))
+    }
+
+    fn next_scanline(&mut self) -> bool {
+        let a = self.update_ab();
+        let c = self.update_c();
+
+        match (a, c) {
+            (Some((fa, la)), Some((fc, lc))) => {
```
jamwaffles, line 95: Does `f` stand for `first` and `l` for `last` her? It would be better to use longer, more descriptive variable names IMO.

Can you also add a short doc comment description to this method explaining what it does? I think I understand, but a few words would help me and others.

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -102,11 +109,24 @@ impl ContainsPoint for Triangle {
...
+
+        // Skip first points so that they are not checked twice
+        l1.next();
+        l2.next();
+
+        l1.chain(l2).chain(l3).any(|line_point| line_point == point)
+    }
+}
+
+impl ContainsPoint for Option<Triangle> {
```
jamwaffles, line 124: This can be removed now I think?

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -185,10 +205,14 @@ impl Triangle {
     /// Point inside triangle, ignoring pixels that partially lie outside triangle lines.
     pub(self) fn mathematical_contains(&self, point: &Point) -> bool {
         // Skip expensive calculations below if point is outside the bounding box
-        if !self.bounding_box().contains(*point) {
-            return false;
+        if self.bounding_box().contains(*point) {
+            self.in_mathematical_triangle(point)
+        } else {
+            false
         }
```
jamwaffles, line 212: This could be condensed using Rust's short circuiting rules:

```suggestion
        self.bounding_box().contains(*point) && in_mathematical_triangle(point)
```

jamwaffles (changes requested):
Nice work on this! I found a few small things.

embedded-graphics/src/primitives/triangle/mod.rs
bugadani: I didn't want to go fancy :)

jamwaffles:
Just to note that the `thick polyline` benchmark is 270% slower on my machine compared to `jamwaffles/line-joints` :grimacing: 

If the rendering is correct though, who am I to complain...!

bugadani:
> Just to note that the `thick polyline` benchmark is 270% slower on my machine compared to `jamwaffles/line-joints` 😬
> 
> If the rendering is correct though, who am I to complain...!

Now with 3x as much checking per pixel!
Honestly, I'm surprised it's not worse.

embedded-graphics/src/primitives/triangle/fill_scanline_iterator.rs
bugadani: Yeah this is not nice :)

bugadani:
@jamwaffles May or may not be my fault, but stroke alignments other than center aren't beveled correctly.
image

jamwaffles:
That's a bug on my branch as well, whew. There's no overdraw though!

bugadani:
@jamwaffles I've optimized Triangle::contains for the polylines - this cut rendering time in half. However, I think I saw some regression in the triangle benches. Should I move this contains function to a polyline-only place?

bugadani:
Right now, this is much better than it was this morning.

```
Without lookback:
thick polyline          time:   [32.778 us 32.841 us 32.914 us]

With lookback:
thick polyline          time:   [90.987 us 91.333 us 91.737 us]

Original:
thick polyline          time:   [47.859 us 48.020 us 48.240 us]
```

embedded-graphics/src/primitives/polyline/thick_points.rs
```diff
@@ -42,12 +52,21 @@ impl<'a> Iterator for ThickPoints<'a> {
...
-            Some(point)
-        } else {
-            self.points_iter = self.triangle_iter.next().map(|t| t.mathematical_points())?;
-
-            self.next()
+        loop {
+            if let Some(point) = self.points_iter.next() {
+                if !ContainsPoint::contains(&self.prev_triangle, point)
+                    && !ContainsPoint::contains(&self.prev_triangle2, point)
+                    && !ContainsPoint::contains(&self.prev_triangle3, point)
```
bugadani, line 59: Technically speaking, one of this can be elided. When the joint turns right, prev_triangle2 check is redundant, and when it turns left, prev_triangle3 is. Also, probably both can be elided for mitered joints, but I haven't tried that.

prev_triangle is necessary though to avoid overdrawing the joining edges.

bugadani: Although I think it may be better to find a correct solution instead of hammering out the issues with the current naive one.

bugadani:
Darn it, last commits have a regression... :'(

bugadani:
image

bugadani:
I guess (did not confirm but symptoms seem to indicate that) I found a case where Bresenham's algo fails to return a point that is otherwise inside a triangle.

jamwaffles:
I think I've seen that before as well. The _center_ of a point can lie on either side of the true line when rendered using the Bresenham algorithm, up to 0.5px away IIRC. This might be what you're seeing.

bugadani:
@jamwaffles there are 2 regression tests I had to #[ignore] because they require a MockDisplay > 64x64. Can we increase MD's size so these tests fit?

jamwaffles:
We could, but I'd prefer to find a set of points that tests for the same regression, but inside the `MockDisplay` area.

bugadani:
I really should set rustfmt as a precommit hook script.

@jamwaffles I think this is ready for you. I don't want to delay the feature with optimizations that may or may not work.

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -383,6 +425,15 @@ mod tests {
...
+    #[test]
+    fn triangle_contains_edge_point_regression() {
+        // This test is a regression test case found while optimizing Triangle::contains()
+        assert!(
+            Triangle::new(Point::new(30, 30), Point::new(0, 0), Point::new(32, 33))
+                .contains(Point::new(31, 31))
+        );
+    }
+
     // FIXME: Colinear triangles are rendered as a line, so this should also return true. Why not?
```
bugadani, line 437: @jamwaffles what's your issue here? Comment says "should also return true" and then you assert that it returns false.

jamwaffles (changes requested):
Looks pretty good, but I'd like to see more doc/inline comments explaining what things to and how all the pieces fit together. This makes it easier for me to understand and pick up in my PR, and also for other maintainers to hack on.

embedded-graphics/src/primitives/triangle/mod.rs
```diff
@@ -86,27 +88,59 @@ impl Primitive for Triangle {
     }
 }
 
+fn is_inside(a: i32, b: i32) -> bool {
+    (a > 0) == (b > 0)
+}
+
+fn point_on_line(point: Point, p1: Point, p2: Point) -> i32 {
+    (p2.x - p1.x) * (point.y - p1.y) - (p2.y - p1.y) * (point.x - p1.x)
+}
```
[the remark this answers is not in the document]
bugadani: But even I don't know 🤣 j/k will do

jamwaffles (approved):
Looks good now. Thanks for progressing this feature! I was reaching my wits end tbh. In terms of attribution, would you prefer I rebase this onto my branch, or squash and merge it?

bugadani:
> Looks good now. Thanks for progressing this feature! I was reaching my wits end tbh. In terms of attribution, would you prefer I rebase this onto my branch, or squash and merge it?

I'll leave that up to you. As far as I can see, the commit history is linear with your branch so both works.

To quote a colleague of mine: "The problem with github's squash and merge feature is that it's not a squash and it's not a merge" :)

I'm fine with a squash.

jamwaffles:
Well, seeing as we're going for the record for most commits in a single PR I've rebased it ;)

PR #435, opened 2020-08-18
Line joiners optimizations

bugadani:
This PR aims to optimize the memory requirements of the line joiners feature.


bugadani:
Work in progress, currently based on #434, totally incomplete and irrelevant while the feature itself is not ready

PR #436, opened 2020-08-20
Line joints optimizations

bugadani:
I know it can't be merged right now.

A lot of memory savings with some better runtime performance.

WIP because there might be more in there somewhere.


jamwaffles:
On mobile atm but I think this needs a rebase?

bugadani:
Yes, I'm postponing that until the main branch is fixed.

embedded-graphics/src/primitives/polyline/thick_points.rs
```diff
@@ -0,0 +1,152 @@
...
+        self.points[1] = self.points[0];
+        self.points[0] = p;
+        if self.count < 5 {
+            self.count += 1;
+        }
+    }
+
+    /// Returns whether the given point is inside the newest tracked triangle.
+    pub fn prev1_contains(&self, p: Point) -> bool {
+        self.count >= 3 && Triangle::new(self.points[0], self.points[1], self.points[2]).contains(p)
```
bugadani, line 43: This is the most painful check. Any simplification could yield significant perf gains.

bugadani:
This is rather messy... I've modified the scanline iterator to be able to skip generating points for edges, and also created a stateful tessellator to avoid calculating each parallel line segment twice. There's still a lot of cleanup to do, also some code arrangement (i.e. I should probably split up the scanline iterator into a simple variant for returning points when nothing is ignored...), but the performance is... interesting :) I think the benchmark is not good enough, there should be more tricky joints in there other than miters.

jamwaffles:
I'm going to close this now as #386 went in a different direction. I'd definitely appreciate any performance gains on the impl in that PR though! It's pretty complex and slow.

Thanks for contributing to the first attempt at thick polylines!

bugadani:
My pleasure. Glad you found a neater solution :)

PR #438, opened 2020-08-22
Add web colors as list of const exports

jamwaffles:

## PR description

Adds the 147 base and extended named colours as defined in the CSS spec. I for one am getting a little bored of `Rgb888::RED` in all the demos ;)

The list at the moment is separate from the colour types themselves. I'd like to move them into the colour types so we can do e.g. `Rgb888::REBECCAPURPLE`, but I'm not sure how that plays with the current constants. I could add the web colours as constants to the `RgbColor` trait? There's also some overlap between e.g. the existing `RED: Self::MAX_R, 0, 0` as defined in the `rgb_color!` macro and the new `RED: 255, 0, 0`. What do we do here?


jamwaffles:
All the CSS colours are prefixed with `CSS_` now. I've done some macro trait impls with a `WebColors` trait. This still needs to go in the prelude and be exported properly, etc, but to prove the rest of this PR it's ok where it is.

I'm doing some macro magic to convert the colours for non-24 bit colour types which I'm not happy with, so suggestions are gladly accepted.

I've got a single unit test that checks the 24 -> 16 bit reduction on one colour and it seems to work.

rfuest:
Adding a new const constructor, that always uses 8 bit color values, to the color types could make the code cleaner. I think the code duplication could be reduced even further, but here is a quick test of this idea: https://github.com/rfuest/embedded-graphics/commit/8ba196c992d971892dc4a4078779be6e22aa5e9f

jamwaffles:
That's a good approach, thanks. I've cherry picked it into this PR. If only we had const trait fns the two lists of macros could be deduped nicely. Beyond that I'm struggling to come up with anything cleaner than what we have now. Any tricks I'm missing? Otherwise I think this is ready for review.

rfuest:
I've updated my branch with a version that uses only a single list to define the colors and doesn't require an additional string version of each color: https://github.com/rfuest/embedded-graphics/commit/91d2498886eef5b76b012dabe324d829bd394c31

jamwaffles:
Ah very nice, a TT muncher. I hate those :joy:. But that's lots cleaner, so I've cherry picked it into this branch.

Also added a changelog entry and `WebColors` to the prelude.

Do we need any more tests on this feature?

embedded-graphics/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#429](https://github.com/jamwaffles/embedded-graphics/pull/429) Added `ToBytes` trait to convert colors into byte arrays.
 - [#431](https://github.com/jamwaffles/embedded-graphics/pull/431) Added `MockDisplay::affected_area` to get the area affected by previous drawing operations.
 - [#450](https://github.com/jamwaffles/embedded-graphics/pull/450) Added `ColorConverted` and `DrawTargetExt::color_converted` to support color conversion for draw targets.
+- [#438](https://github.com/jamwaffles/embedded-graphics/pull/438) Added 147 CSS web colors as associated consts to the RGB color types.
```
rfuest, line 39: Did we loose some on the way, because there are less than 147 lines in the macro call?

embedded-graphics/src/pixelcolor/conversion.rs
```diff
@@ -1,7 +1,7 @@
 use crate::pixelcolor::{binary_color::*, gray_color::*, rgb_color::*};
 
 /// Convert color channel values from one bit depth to another.
-const fn convert_channel(value: u8, from_max: u8, to_max: u8) -> u8 {
+pub const fn convert_channel(value: u8, from_max: u8, to_max: u8) -> u8 {
```
rfuest, line 4: I don't think this is necessary anymore.

embedded-graphics/src/pixelcolor/mod.rs
```diff
@@ -88,11 +88,13 @@ mod conversion;
 mod gray_color;
 pub mod raw;
 mod rgb_color;
+pub mod web_colors;
```
rfuest, line 91: Does this need to be public? The module docs for `web_colors` could also be added to  `WebColors` instead IMO.

embedded-graphics/CHANGELOG.md
jamwaffles: We removed 6 `grEy` variants. I've changed this to read "most" instead, but I can put the actual number back in if you think that's better.

rfuest (approved):
LGTM

jamwaffles:
Thank you ☺️

PR #439, opened 2020-08-24
Add support for transparent characters

### LINKED ISSUE #408: Transparent characters are not supported
https://github.com/jamwaffles/embedded-graphics/blob/0ebc0d2dc6e210193662373e91faf6cc63fb17ea/embedded-graphics/src/fonts/text.rs#L197

While e-g can render transparent text background just fine, when text_color is not specified, it defaults back to background color. IMHO this is rather pointless, a transparent character has more use cases than drawing boxes where characters would go :)

bugadani:

## PR description

Closes #408 

I think drawing a solid rectangle using the background color is not particularly useful. This PR modifies text rendering so that the text can be transparent in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 - [#420](https://github.com/embedded-graphics/embedded-graphics/pull/420) Added support for `SubImage`s.
 - [#429](https://github.com/embedded-graphics/embedded-graphics/pull/429) Added `ToBytes` trait to convert colors into byte arrays.
 - [#431](https://github.com/embedded-graphics/embedded-graphics/pull/431) Added `MockDisplay::affected_area` to get the area affected by previous drawing operations.
+- [#439](https://github.com/embedded-graphics/embedded-graphics/pull/439) Added support to render transparent characters.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 - [#420](https://github.com/embedded-graphics/embedded-graphics/pull/420) Added support for `SubImage`s.
 - [#429](https://github.com/embedded-graphics/embedded-graphics/pull/429) Added `ToBytes` trait to convert colors into byte arrays.
 - [#431](https://github.com/embedded-graphics/embedded-graphics/pull/431) Added `MockDisplay::affected_area` to get the area affected by previous drawing operations.
-- [#439](https://github.com/embedded-graphics/embedded-graphics/pull/439) Added support to render transparent characters.
+- [#439](https://github.com/embedded-graphics/embedded-graphics/pull/439) Added support to render transparent characters with a colored background.
 
 ### Changed
 
--- a/src/fonts/text.rs
+++ b/src/fonts/text.rs
@@ -110,6 +110,11 @@ where
     F: Font,
 {
     fn bounding_box(&self) -> Rectangle {
+        // If a piece of text is completely transparent, return an empty bounding box
+        if self.style.text_color.is_none() && self.style.background_color.is_none() {
+            return Rectangle::new(self.primitive.position, Size::new(0, 0));
+        }
+
         let width = if !self.primitive.text.is_empty() {
             self.primitive
                 .text
@@ -194,7 +199,7 @@ where
                     self.char_walk_x as u32,
                     self.char_walk_y as u32,
                 ) {
-                    self.style.text_color.or(self.style.background_color)
+                    self.style.text_color
                 } else {
                     self.style.background_color
                 };
@@ -236,6 +241,8 @@ mod tests {
         fonts::{tests::assert_text_from_pattern, Font6x12, Font6x8},
         mock_display::MockDisplay,
         pixelcolor::BinaryColor,
+        prelude::*,
+        style::PrimitiveStyle,
     };
 
     const HELLO_WORLD: &'static str = "Hello World!";
@@ -430,5 +437,59 @@ mod tests {
         text.translate_mut(Point::new(-6, 0));
 
         assert_eq!(text.into_pixels().count(), 6 * 12);
+    }
+
+    #[test]
+    fn transparent_text_color_does_not_overwrite_background() {
+        let style = TextStyle {
+            font: Font6x8,
+            text_color: None,
+            background_color: Some(BinaryColor::On),
+        };
+
+        let mut display = MockDisplay::new();
+        display.set_allow_overdraw(true);
+
+        // Draw a background for the first character
+        Rectangle::new(Point::zero(), Size::new(6, 8))
+            .into_styled(PrimitiveStyle::with_fill(BinaryColor::Off))
+            .draw(&mut display)
+            .unwrap();
+
+        Text::new("AA", Point::zero())
+            .into_styled(style)
+            .draw(&mut display)
+            .unwrap();
+
+        assert_eq!(
+            display,
+            MockDisplay::from_pattern(&[
+                "#...###   ##",
+                ".###.# ### #",
+                ".###.# ### #",
+                ".....#     #",
+                ".###.# ### #",
+                ".###.# ### #",
+                ".###.# ### #",
+                "############",
+            ])
+        );
+    }
+
+    #[test]
+    fn transparent_text_has_zero_size_but_retains_position() {
+        let style: TextStyle<BinaryColor, _> = TextStyle {
+            font: Font6x8,
+            text_color: None,
+            background_color: None,
+        };
+
+        let styled = Text::new(" A", Point::new(7, 11)).into_styled(style);
+
+        assert_eq!(
+            styled.bounding_box(),
+            Rectangle::new(Point::new(7, 11), Size::zero()),
+            "Transparent text is expected to have a zero sized bounding box with the top left corner at the text position",
+        );
     }
 }


embedded-graphics/CHANGELOG.md
```diff
@@ -55,6 +56,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - **(breaking)** [#409](https://github.com/jamwaffles/embedded-graphics/pull/409) The `DrawTarget` trait was moved from the root of the crate to a new `draw_target` module. Projects that use the `prelude` to import `DrawTarget` won't be affected by this change.
 - **(breaking)** [#420](https://github.com/jamwaffles/embedded-graphics/pull/420) To add support for additional image formats now `ImageDrawable` and `OriginDimensions` needs to be implemented instead of `IntoPixelIterator` and `ImageDimensions`.
 - **(breaking)** [#420](https://github.com/jamwaffles/embedded-graphics/pull/420) The color type parameter was removed from the `Image` struct and is now defined by the associated `Color` type of an `ImageDrawable` implementation.
+- **(breaking)** [#439](https://github.com/jamwaffles/embedded-graphics/pull/439) When text color is omitted, the text is rendered as transparent instead of a solid block using the backgorund color.
```
jamwaffles, line 59: IMO this PR fixes a bug so this should go under `### Fixed`, without the **(breaking)** tag. There's also a typo:

```suggestion
- **(breaking)** [#439](https://github.com/jamwaffles/embedded-graphics/pull/439) When text color is omitted, the text is rendered as transparent instead of a solid block using the background color.
```

embedded-graphics/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#420](https://github.com/jamwaffles/embedded-graphics/pull/420) Added support for `SubImage`s.
 - [#429](https://github.com/jamwaffles/embedded-graphics/pull/429) Added `ToBytes` trait to convert colors into byte arrays.
 - [#431](https://github.com/jamwaffles/embedded-graphics/pull/431) Added `MockDisplay::affected_area` to get the area affected by previous drawing operations.
+- [#439](https://github.com/jamwaffles/embedded-graphics/pull/439) Added support to render transparent characters.
```
jamwaffles, line 39: I don't think we need two changelog items for this PR.

jamwaffles (changes requested):
Sorry it took me a while to get to this. The fix looks good aside from some changelog nits.

I think this PR also raises a related issue: if the stroke colour _and_ background colour are `None`, should the bounding box size of `Styled<Text>` be zero?

For e-t I can see it being useful that even transparent text retains its size. On the flip side, styled e-g primitives return a BB of zero when they're transparent as the bounding box is often used to tell a display how much and where to draw pixels. What do you think the behaviour should be @rfuest?

jamwaffles:
p.s. pleas give this a rebase when you have a moment.

bugadani:
> I think this PR also raises a related issue: if the stroke colour _and_ background colour are `None`, should the bounding box size of `Styled<Text>` be zero?

Makes sense, best to be consistent on that one IMO.

jamwaffles:
I just noticed there was some force pushing to this branch. When you're ready for a re-review, can you click the "request review" button so I get an email notification please? That way these PRs don't fall down the back of the couch :grin:

bugadani:
I'll do that as soon as I have the size=0 change.

bugadani:
All yours @jamwaffles

embedded-graphics/src/fonts/text.rs
```diff
@@ -431,4 +437,47 @@ mod tests {
...
+        let style: TextStyle<BinaryColor, _> = TextStyle {
+            font: Font6x8,
+            text_color: None,
+            background_color: None,
+        };
+
+        let styled = Text::new(" A", Point::new(7, 11)).into_styled(style);
+
+        assert_eq!(Size::new(0, 0), styled.bounding_box().size);
+        assert_eq!(Point::new(7, 11), styled.bounding_box().top_left);
```
jamwaffles, line 481:
```suggestion
        assert_eq!(
            styled.bounding_box(),
            Rectangle::new(Point::new(7, 11), Size::zero())
        );
```

jamwaffles (changes requested):
Just to let you know, this PR is going to have to wait until the repo split is complete as it merges into the old workspace structure. Hopefully not long until #459 is merged :)

@rfuest what's your opinion on the zero sized BB change here? As drivers will use the bounding box to figure out what to draw where, I think the change of behaviour is correct, even if it makes layout of transparent text in e-t a bit harder.

embedded-graphics/src/fonts/text.rs
bugadani: I actually prefer testing two properties separately. I think it's much more explicit in what I expect to happen. Merging the tests makes the expected size (*the* thing I want to test) be hidden in a construct that's ... collateral.

bugadani:
> even if it makes layout of transparent text in e-t a bit harder

I don't think it does :)

embedded-graphics/src/fonts/text.rs
```diff
@@ -110,6 +110,11 @@ where
     F: Font,
 {
     fn bounding_box(&self) -> Rectangle {
+        // If a piece of text is completely transparent, return an empty bounding box
+        if self.style.text_color.is_none() && self.style.background_color.is_none() {
+            return Rectangle::new(self.primitive.position, Size::new(0, 0));
```
rfuest, line 115:
```suggestion
            return Rectangle::new(self.primitive.position, Size::zero());
```

embedded-graphics/src/fonts/text.rs
rfuest: I would also prefer the code @jamwaffles has suggested, because it is more consistent with the other tests. The name of the test is a bit misleading, because every text has a size, it just happens to be `Size::zero()` in this case. If I would write the test I would simply call it `transparent_text_bounding_box` or `transparent_text_dimensions` (because of the `Dimenstions` trait) and add a comment that describes the expected behavior, which is often easier than trying to come up with a function name that conveys the same information.

rfuest (changes requested):

bugadani:
For whatever reason I can't reply to the second comment, but:

> transparent_text_bounding_box or transparent_text_dimensions

I can often be lazy with naming my tests, but after listening to some of Kevlin Henney's talks, I'm trying to give somewhat meaningful names to my tests. So that when they fail, the name of the test can often tell what's wrong. For example, "transparent_text_bounding_box" hints at which test is bad, but not *why*. However, "transparent_text_has_no_size_but_has_position" while a bit unwieldy and asserts two things, at least gives a pointer without looking at the test result or test code. If anything, I think I should split the test into 2 :) 

However, I agree with the "no size" part. No extent, maybe, or zero size, but both sound also a bit off to my ears.

rfuest:
Any possible time gain from a good function name, if this test ever fails, has already been eaten up by this discussion :wink:.

If I run the tests by using `cargo test` I don't think the function name on it's own is very important, because it always includes the entire test output. But adding the reason why the test failed is a valid point, so I would suggest this:

```rust
#[test]
fn transparent_bounding_box() {
    let style: TextStyle<BinaryColor, _> = TextStyle {
        font: Font6x8,
        text_color: None,
        background_color: None,
    };

    let styled = Text::new(" A", Point::new(7, 11)).into_styled(style);

    assert_eq!(
        styled.bounding_box(),
        Rectangle::new(Point::new(7, 11), Size::zero(),
        "Transparent text is expected to have a zero sized bounding box with the top left corner at the text position",
    );
}
```

This might not be what you suggested but is consistent with the other tests and shows the reason why the test failed when running `cargo test`. IMO this is not the right place to discuss changing our style of writing test in general, but feel free to open an issue if you think we should improve our tests so we can discuss this further. There certainly is room for improvement.

bugadani:
I don't want the current tests to change, I merely tried to explain why/how I (at least want to) write my tests. I'm sorry that I am so bad at explaining my thoughts.

I still think having a more descriptive test name is as important here as it is with the API itself. But I quite like your suggestion, @rfuest :)

jamwaffles:
Now that #459 is merged, this PR needs updating to change files in the project root - there's no workspace anymore.

I like @rfuest's suggestion to add some descriptive failure text. Beyond that, I think this is getting a bit bikesheddy, so if there are no other issues it would be good to get this PR merged.

bugadani:
@jamwaffles I've rebased, and I've also updated the tests. The first one I've originally added now actually tests different background colors, which, while not strictly necessary, I think shows the intended behaviour better (i.e. no accidental "deletion" of existing pixels [even though that's not even possible]).

rfuest:
> The first one I've originally added now actually tests different background colors, which, while not strictly necessary, I think shows the intended behaviour better (i.e. no accidental "deletion" of existing pixels [even though that's not even possible]).

The correct behavior is already sufficiently shown by a single draw call to a `MockDisplay`. `MockDisplay` records the changes made to the display and not only the final framebuffer content. It's guaranteed that the drawing operation didn't change any ' ' pixels in the pattern. And by having to deactivate the overdraw detection the test is now insensitive to any overdraw that happens while a transparent text is drawn.

bugadani:
Closing in favour of #470

rfuest:
@bugadani I would like to get this merged soon. Did you see my previous comment about the test changes?

bugadani:
@rfuest Yes. I can change it back but I feel like the current one is more informative for a reader. I'm not worried about the overdraw in the test, since the feature kind of (but not completely) implies it anyway. Also, the lack of overdraw is already covered by other tests, and changing a color shouldn't cause any that wasn't there before. I don't really feel strongly either way, so I leave the decision up to you :)

rfuest:
OK, let's not drag this out any longer and just leave it this way.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#420](https://github.com/embedded-graphics/embedded-graphics/pull/420) Added support for `SubImage`s.
 - [#429](https://github.com/embedded-graphics/embedded-graphics/pull/429) Added `ToBytes` trait to convert colors into byte arrays.
 - [#431](https://github.com/embedded-graphics/embedded-graphics/pull/431) Added `MockDisplay::affected_area` to get the area affected by previous drawing operations.
+- [#439](https://github.com/embedded-graphics/embedded-graphics/pull/439) Added support to render transparent characters.
```
rfuest, line 52: IMO it's important to mention the background, because otherwise it might not be clear why someone wants to render transparent text:
```suggestion
- [#439](https://github.com/embedded-graphics/embedded-graphics/pull/439) Added support to render transparent characters on a colored background.
```

rfuest (changes requested):
One small nit left:

rfuest (approved):

rfuest:
@bugadani LGTM. Thanks for you patience ;).
@jamwaffles Your review is still marked as "changes requested" so that I cannot merge the PR. I'll let you merge it.

jamwaffles (approved):
Looks good!

PR #440, opened 2020-09-09
Make DrawTarget trait methods generic across color

BroderickCarlin:

## PR description

This PR attempts to allow the `DrawTarget` trait to be generic across the color type that the trait methods can accept. This allows for crates/code that interface with displays that impl `DrawTarget` to be more agnostic as to the pixel type of said display. 

Noting that this sort of generic approach was possible pre-0.7 release because `DrawTarget<>` could be made to be generic. An example of this can be seen here: https://github.com/rust-rpi-led-matrix/rust-rpi-rgb-led-matrix/blob/5d1334a47b270a6666e310cbf4229bace04fa3d3/src/matrix.rs#L93

### DIFF
--- a/embedded-graphics/src/draw_target/clipped.rs
+++ b/embedded-graphics/src/draw_target/clipped.rs
@@ -2,6 +2,7 @@ use crate::{
     draw_target::DrawTarget,
     geometry::Dimensions,
     iterator::contiguous::Crop,
+    pixelcolor::PixelColor,
     prelude::Transform,
     primitives::{ContainsPoint, Rectangle},
     Pixel,
@@ -41,9 +42,10 @@ where
     type Color = T::Color;
     type Error = T::Error;
 
-    fn draw_iter<I>(&mut self, pixels: I) -> Result<(), Self::Error>
+    fn draw_iter<I, C>(&mut self, pixels: I) -> Result<(), Self::Error>
     where
-        I: IntoIterator<Item = Pixel<Self::Color>>,
+        I: IntoIterator<Item = Pixel<C>>,
+        C: Into<Self::Color> + PixelColor,
     {
         let clip_area = self.clip_area;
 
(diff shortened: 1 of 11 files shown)


jamwaffles:
Thanks for the PR! Apologies for the delay - I'm personally feeling pretty burned out right now, so my OSS projects are suffering a bit.

Anyway, this could work I think but @rfuest might have some deeper thoughts on it. Might using `Into<Self::PixelColor>` this way cause some strange behaviour with colour conversions? AFAIK the conversion rules we have now are pretty solid so perhaps now.

BroderickCarlin:
Yeah, I think the concern with respect to color conversions is valid. My personal opinion on this though is that no color conversion should take place as long as you know the native color type of the display you are writing to. In the case where you are taking advantage of the generic to write more display-agnostic code it becomes (IMHO) the responsibility of a consuming crate to manage color conversions outside of the types provided by this crate.

I've done some basic tests with the color types provided within this crate and the conversions seem to all work well within reason.

rfuest:
> ...but @rfuest might have some deeper thoughts on it.

IMO the methods in `DrawTarget` shouldn't do automatic color conversion. When they are generic over the color type it's much easier to accidentally call them with the wrong color type which impacts drawing performance because of the unnecessary color conversions.

I was planning to implement color conversion as a `DrawTarget` adapter (similar to `TranslatedDrawTarget` etc.). This way you can draw different color types, but need to explicitly opt into this feature.

jamwaffles:
Apologies for the late reply on this!

Good idea - an adapter would allow opt-in for those who need it, but not slow down `DrawTarget` itself.

I've opened #446 to track this feature request. @BroderickCarlin would you be interested in modifying this PR to take an adapter-based approach akin to e.g. [`TranslatedDrawTarget`](https://github.com/jamwaffles/embedded-graphics/blob/master/embedded-graphics/src/draw_target/translated.rs) or shall I close this and we can work on the new approach in another PR?

BroderickCarlin:
@jamwaffles, I am very much light on available time to put towards this for the immediate future. Because of this I'm fine with closing this down and we can work on the new approach in another PR

jamwaffles:
No worries. I'll close this PR. We do have #446 to track this feature so we can pick up against that. Thanks for the original contribution!

PR #441, opened 2020-09-19
Fix arc sector bounding box

### LINKED ISSUE #405: Bounding boxes for Arc and Sector don't take the angles into account
The `Dimensions` trait impl for `Arc` and `Sector` returns the same bounding box as a circle with the same `top_left` and `diameter`. But the bounding box can be much smaller if `angle_sweep` is small.

jamwaffles:

## PR description

Fix Arc/Sector bounding boxes.

TODO

* [ ] Off-by-one on right/bottom edges of Arc bounding box:

  arc-oob

* [ ] Unit tests

Note that there are some errors around stroke widths as can be seen on the right side here:

oob-sector

Due to stroke corners lying outside the bounding box. IMO this should be fixed, but later. Perhaps once thick polylines is merged so the `Line::extents()` method can be reused.

Closes #405


jamwaffles:
I'm getting my ass kicked by something that I'm sure has a simple solution: more off-by-one errors. I've tried various things but as a baseline I have this code in `arc.rs` (not pushed):

```rust
let radius = Real::from(self.diameter) / 2.into();

Point::new(
    i32::from(angle.cos() * radius),
    -i32::from(angle.sin() * radius),
) + Size::new_equal(radius.into())
```

The sin/cos stuff is fine, but as far as I can see the final point rarely lies on the actual arc (or circle, same thing) because the point is transformed not from the true center of the arc but somewhere near it (+/- 1) due to rounding. @rfuest you've done some magic with circles in the same vein as this, so could you offer some guidance perhaps?

rfuest:
I did take a very brief look at this and I think that at least part of the problem is in the existing code. I would expect an 90 degree arc starting at 0 degrees to exactly draw one quadrant of a full circle. But in my test this only worked correctly for arcs with odd diameters. Arcs with even diameters were missing one row of pixels at the bottom.

The circle code internally uses coordinates that are scaled by a factor of 2 (e.g. in `Circle::center_2x`). This way it is possible to use an effective resolution of half a pixel for the calculations, which is required to represent the accurate center of a circle with even diameter.

PR #444, opened 2020-10-19
Apply recent tinytga changes to tinybmp

rfuest:

## PR description

This is the `tinybmp` version of #430. Summary of changes:
* Merge `Bmp` and `BmpRaw`.
* Make `embedded-graphics` dependency mandatory.
* Use enum for `Bpp` and check if bpp is correct in `from_slice` constructor.
* Remove possible panic in iterator.
* Replace `IntoIterator` by `pixels` and `raw_pixels` methods.
* Remove some unused fields from the `Header` struct.


tinybmp/src/lib.rs
```diff
@@ -1,251 +1,243 @@
...
+//! let image = Image::new(&bmp, Point::zero());
+//!
+//! image.draw(&mut display)?;
+//! # Ok::<(), core::convert::Infallible>(()) }
+//! ```
+//!
+//! ## Accessing the raw image data
 //!
-//! ## Load a BMP image and check its [`Header`] and returned pixels.
+//! This example demonstrates how the mage header and raw image data can be accessed to use
```
jamwaffles, line 33:
```suggestion
//! This example demonstrates how the image header and raw image data can be accessed to use
```

tinybmp/src/lib.rs
```diff
@@ -1,251 +1,243 @@
...
+impl<'a, C> Bmp<'a, C>
+where
+    C: PixelColor,
+{
+    /// Create a bitmap object from a byte slice.
+    ///
+    /// The created object keeps a shared reference to the input and does not dynamically allocate
+    /// memory.
+    ///
+    /// The color type needs to be specified explicitly when this method is called, for example by
```
jamwaffles, line 171:
```suggestion
    /// The color type must be explicitly specified when this method is called, for example by
```

tinybmp/src/lib.rs
```diff
@@ -1,251 +1,243 @@
...
 
-    /// Current Y position
-    y: u32,
+        Ok(bmp)
+    }
 
-    /// Start bit index for the current pixel.
+    /// Returns an iterator over the pixels in this image.
+    ///
+    /// The iterator automatically converts the pixel colors into a `embedded-graphics` color type,
```
jamwaffles, line 193:
```suggestion
    /// The iterator automatically converts the pixel colors into an `embedded-graphics` color type,
```

tinybmp/src/raw_pixels.rs
```diff
@@ -0,0 +1,103 @@
+use crate::{Bmp, Bpp};
+use embedded_graphics::prelude::*;
+
+/// Iterator over individual BMP pixels
```
jamwaffles, line 4:
```suggestion
/// Iterator over individual BMP pixels.
```

tinybmp/src/raw_pixels.rs
```diff
@@ -0,0 +1,103 @@
...
+///
+/// Each pixel is returned as a `u32` regardless of the bit depth of the source image.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct RawPixels<'a, 'b, C> {
+    /// Reference to original BMP image
```
jamwaffles, line 9:
```suggestion
    /// Reference to original BMP image.
```

tinybmp/src/raw_pixels.rs
```diff
@@ -0,0 +1,103 @@
...
+    bmp: &'a Bmp<'b, C>,
+
+    /// Image pixel data as a byte slice, little endian ordering
```
jamwaffles, line 12:
```suggestion
    /// Image pixel data as a byte slice, little endian ordering.
```

tinybmp/src/raw_pixels.rs
```diff
@@ -0,0 +1,103 @@
...
+    pixel_data: &'b [u8],
+
+    /// Current position
```
jamwaffles, line 15:
```suggestion
    /// Current position.
```

tinybmp/src/raw_pixels.rs
```diff
@@ -0,0 +1,103 @@
...
+            }),
+        };
+
+        self.bit_idx += usize::from(self.bmp.color_bpp().bits());
+
+        Some(RawPixel::new(p, u32::from_le_bytes(pixel_value)))
+    }
+}
+
+/// Pixel with raw pixel color.
```
jamwaffles, line 88:
```suggestion
/// Pixel with raw pixel color stored as a `u32`.
```

I think it would be worth explaining that this type exists so using `tinybmp` without e-g is still an option.

tinybmp/src/lib.rs
```diff
@@ -1,251 +1,243 @@
...
-        }
+impl<'a> Bmp<'a, ()> {
+    /// Create a bitmap object from a byte slice.
+    ///
+    /// The created object keeps a shared reference to the input and does not dynamically allocate
+    /// memory.
+    ///
+    /// In contrast to the [`from_slice`] constructor no color type needs to be specified when
+    /// calling this method. This will disable all functions that requires a specified color type,
+    /// like the [`pixels`] method.
```
jamwaffles, line 153: Could we mention e-g by name here as well, something about using `from_slice` if using this crate with e-g.

tinybmp/src/lib.rs
```diff
@@ -1,251 +1,243 @@
...
+    C: PixelColor,
+{
+    /// Create a bitmap object from a byte slice.
+    ///
+    /// The created object keeps a shared reference to the input and does not dynamically allocate
+    /// memory.
+    ///
+    /// The color type needs to be specified explicitly when this method is called, for example by
+    /// using the turbofish syntax. An error is returned if the bit depth of the specified color
+    /// type doesn't match the bit depth of the BMP file.
```
jamwaffles, line 173: As above, might it be worth adding something like "to load an image without e-g support/integration, use `from_slice_raw`?

It isn't immediately clear what the difference between these two methods is when looking at the rendered docs, hence the suggestion that naming e-g as the reason for these _two_ methods to exist.

jamwaffles (changes requested):
image

Welcome back!

I left some pedantic nits on the comments, otherwise this looks good! Thanks for keeping the project moving :)

tinybmp/src/lib.rs
rfuest: The docs need some work, but I would prefer to add this to the PR for #424. That PR will add additional ways of using `tinybmp` and it will be easier to write an overview of the different variants when everything is implemented.

rfuest:
> Welcome back!

Thanks, I had to take a break from e-g to finish another project and needed a break from Rust after that. But I'm determined to at least finish the `tinybmp` and `tinytga` issues soon.

tinybmp/src/lib.rs
jamwaffles: Sure, that makes sense.

jamwaffles:
Lockdown in the UK has got me pretty burned out too so I feel your pain :(. Take your time :)

Are you adding more to this PR or would you like me to do a final review?

rfuest:
> Are you adding more to this PR or would you like me to do a final review?

I think this PR is done and ready for review.

jamwaffles (approved):
Looks good. Send it :rocket:

PR #445, opened 2020-10-21
Authenticate CircleCI image pulls for custom image

jamwaffles:
This is due to Docker Hub's new rate limits for unauthenticated pulls -
by IP address. There are still limits for authenticated pulls, but are
limited now per Docker repo, not by IP address, so we don't hit limits
when pulling from CircleCI's IP addresses.

The `DOCKERHUB_PASSWORD` env var has already been added to the CircleCI
env var secrets thing. This uses my personal account, but it's not too
hard to change it to anything else in the future as (afaik) no special
permissions are required beyond login as the image is public.

Edit: according to https://www.docker.com/pricing, we get 200 authenticated pulls per 6 hours. Previously, that would be 100 unauthed pulls per IP address, which is pretty stinky when those IPs have to service all of CircleCI.

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,8 +1,10 @@
-FROM circleci/rust:1.40.0
+FROM cimg/rust:1.40.0
 
-RUN sudo apt update && sudo apt install -y libsdl2-dev && sudo apt clean
+RUN sudo apt update && sudo apt install -y libsdl2-dev libssl-dev && sudo apt clean
 
-RUN cargo install cargo-readme just cargo-deadlinks
+RUN cargo install just --version ^0.6
+RUN cargo install cargo-readme --version ^3.2
+RUN cargo install cargo-deadlinks --version ^0.4
 
 RUN rustup toolchain add nightly -c rustfmt
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,9 @@
 precheck_steps: &precheck_steps
   docker:
     - image: jamwaffles/circleci-embedded-graphics:1.40.0
+      auth:
+        username: jamwaffles
+        password: $DOCKERHUB_PASSWORD
   steps:
     - checkout
     - restore_cache:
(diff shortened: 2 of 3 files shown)


jamwaffles:
Rereading it, that description is a bit of a mess. There's more background and explanation here: https://support.circleci.com/hc/en-us/articles/360050623311-Docker-Hub-rate-limiting-FAQ

jamwaffles:
I've also updated from `circleci/rust` to cimg/rust` which is the newer generation of CircleCI images. Not sure what's different, but yay upgrades. Build times remain about the same which is good.

.circleci/Dockerfile
```diff
@@ -1,8 +1,12 @@
-FROM circleci/rust:1.40.0
+# Build with `docker build -t jamwaffles/circleci-embedded-graphics:<rust_version> -f ./.circleci/Dockerfile .`
+# Push with `docker push jamwaffles/circleci-embedded-graphics:<rust_version>`
```
rfuest, line 2: You did add docker commands to the `justfile` some time ago, are these the same commands?

.circleci/Dockerfile
```diff
@@ -1,8 +1,12 @@
...
 
-RUN sudo apt update && sudo apt install -y libsdl2-dev && sudo apt clean
+FROM cimg/rust:1.40.0
 
-RUN cargo install cargo-readme just cargo-deadlinks
+RUN sudo apt update && sudo apt install -y libsdl2-dev libssl-dev && sudo apt clean
+
+RUN cargo install just --version ^0.6
```
rfuest, line 8: Do newer `just` versions have problems with rust `1.40` or why do you explicitly specify a version for `just`?
Maybe we should specify the version for all cargo tools we use to make the build reproducible.

rfuest (commented):
Looks good, but I have two questions.

.circleci/Dockerfile
jamwaffles: Yeah they do, due to use of the `matches!` macro. I've pushed a commit to pin the other dependencies installed in the image (and replaced the image on Docker Hub under the same tag). I decided to allow patch releases so they're not perfectly pinned. Let me know if you think this is a bad idea and I'll change them to `=x.y.z`.

.circleci/Dockerfile
jamwaffles: :facepalm: totally forgot about that, thank you. I've removed this comment and updated the `justfile`.

rfuest (approved):
The version pinning is fine IMO and should prevent any incompatible updates.

The rest also looks good :+1:

jamwaffles:
Danke schön!

PR #447, opened 2020-10-22
Add 16bpp TGA test images

rfuest:
This PR adds some 16bpp TGA test images in preparation for #424.

The TGA specs mention 15 and 16 bit color map entry sizes, but it seems that TGA uses RGB555 even if the entry size is set to 16 bits. I will assume this is correct and also implement the same behavior for #424. But at some point we will need to revisit this, because the upper unused bit could also be used as a 1 bit alpha channel.

The e-g test functions are now generated using a macro and the `paste` crate. By generating one function per test case it's easier to debug the failing test cases, because they are independent of each other. IMO this will be useful when I add additional tests for #424 and will also reduce code duplication.


tinytga/Cargo.toml
```diff
@@ -18,6 +18,9 @@ exclude = [
 [badges]
 circle-ci = { repository = "jamwaffles/embedded-graphics", branch = "master" }
 
+[dependencies]
+paste = "1.0"
```
jamwaffles, line 22: Looks like this is only used in the tests, so should be a dev-dependency I think.

```suggestion
[dev-dependencies]
paste = "1.0"
```

jamwaffles (changes requested):
Looks good apart from the dev dependency. Why do image formats gotta be so whack.

jamwaffles (approved):

Issue #446, opened 2020-10-21
Add DrawTarget adapter to perform automatic colour conversions


jamwaffles:
#440 introduces changes to make `DrawTarget` generic across colors. This is a useful feature, however in https://github.com/jamwaffles/embedded-graphics/pull/440#issuecomment-693431053 the point is raised that this may accidentally introduce slower automatic conversion if the wrong colour type is used. 

The suggested solution is a `DrawTarget` adapter akin to `TranslatedDrawTarget`, etc, which is an opt-in target which can perform the potentially costly automatic conversion for use cases that need it.

rfuest:
IMO another advantage of this approach is that we can easily add different variants for color conversions that all use a similar API.

For example, e-g doesn't implement `From<RgbXXX>` for `GrayX`, because there are multiple ways of converting colors to grayscale (average of the RGB components, calculating luma, ...). We could later add specialized adapters for these conversion, in addition to the generic adapter that was suggested in the original comment and requires a `From<OtherColor>` impl.

A `map` type adapter, that takes a closure to convert from one color type to another, could also be a useful addition.

PR #450, opened 2020-10-25
Add DynamicTga and ConvertColor

rfuest:

## PR description

This PR adds `DynamicTga` which is the first part of #424. The implementation depends on the color conversion suggested in #446, so I've also included that in this PR.

I've also decided to revert a change I made in #430, where I merged `Tga` and `TgaRaw`. Keeping them separate results in a cleaner API and I'm not sure why I decided to change that in the first place.

### DIFF
--- a/embedded-graphics/src/draw_target/mod.rs
+++ b/embedded-graphics/src/draw_target/mod.rs
@@ -575,7 +575,7 @@ pub trait DrawTargetExt: DrawTarget + Sized {
     /// Creates a color conversion draw target.
     ///
     /// A color conversion draw target is used to draw drawables with a different color type to a
-    /// draw target. The drawable color type needs to implement `Into<C>`, where `C` is the draw
+    /// draw target. The drawable color type must implement `Into<C>`, where `C` is the draw
     /// target color type.
     ///
     /// # Performance
@@ -583,11 +583,11 @@ pub trait DrawTargetExt: DrawTarget + Sized {
     /// Color conversion can be expensive on embedded hardware and should be avoided if possible.
     /// Using the same color type for drawables and the draw target makes sure that no unnecessary
     /// color conversion is used. But in some cases color conversion will be required, for example,
-    /// to draw images.
+    /// to draw images with a color format only known at runtime.
     ///
     /// # Examples
     ///
-    /// This example draws a `BinaryColor` image to a `Rgb888` display.
+    /// This example draws a `BinaryColor` image to an `Rgb888` display.
     ///
     /// ```
     /// use embedded_graphics::{
--- a/embedded-graphics/src/draw_target/mod.rs
+++ b/embedded-graphics/src/draw_target/mod.rs
@@ -1,6 +1,7 @@
 //! A target for embedded-graphics drawing operations.
 
 mod clipped;
+mod convert_color;
 mod cropped;
 mod translated;
 
@@ -12,6 +13,7 @@ use crate::{
 };
 
 pub use clipped::Clipped;
+pub use convert_color::ConvertColor;
 pub use cropped::Cropped;
 pub use translated::Translated;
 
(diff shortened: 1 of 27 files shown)


jamwaffles:
Do we want to revert the same change to `tinybmp` and go back to `Bmp` and `BmpRaw` as well?

rfuest:
Yes, I would like to keep the APIs of `tinybmp` and `tinytga` as similar as possible. But this doesn't belong in this PR IMO and I plan to include that change in the PR where I add `DynamicBmp`.

jamwaffles:
I thought that might be the case :). I agree it doesn't belong in this PR.

embedded-graphics/src/draw_target/mod.rs
```diff
@@ -586,4 +651,11 @@ where
     fn clipped(&mut self, area: &Rectangle) -> Clipped<'_, Self> {
         Clipped::new(self, area)
     }
+
+    fn convert_color<C>(&mut self) -> ConvertColor<'_, Self, C>
```
jamwaffles, line 655: This name sounds a bit strange, especially in the context of `thing.draw(&mut display.convert_color())`. The best alternative I can come up with so far is `color_converted` which would match the `-ed` suffixes of the other display modes. If we change this method name, the trait name should probably be changed to `ColorConverted` too.

Open to better suggestions - `color_converted` isn't great.

embedded-graphics/src/draw_target/mod.rs
```diff
@@ -569,6 +571,69 @@ pub trait DrawTargetExt: DrawTarget + Sized {
     /// # Ok::<(), core::convert::Infallible>(())
     /// ```
     fn clipped(&mut self, area: &Rectangle) -> Clipped<'_, Self>;
+
+    /// Creates a color conversion draw target.
+    ///
+    /// A color conversion draw target is used to draw drawables with a different color type to a
+    /// draw target. The drawable color type needs to implement `Into<C>`, where `C` is the draw
```
jamwaffles, line 578:
```suggestion
    /// draw target. The drawable color type must implement `Into<C>`, where `C` is the draw
```

embedded-graphics/src/draw_target/mod.rs
```diff
@@ -569,6 +571,69 @@ pub trait DrawTargetExt: DrawTarget + Sized {
...
+    /// target color type.
+    ///
+    /// # Performance
+    ///
+    /// Color conversion can be expensive on embedded hardware and should be avoided if possible.
+    /// Using the same color type for drawables and the draw target makes sure that no unnecessary
+    /// color conversion is used. But in some cases color conversion will be required, for example,
+    /// to draw images.
```
jamwaffles, line 586:
```suggestion
    /// to draw images with a color format only known at runtime.
```

embedded-graphics/src/draw_target/mod.rs
```diff
@@ -569,6 +571,69 @@ pub trait DrawTargetExt: DrawTarget + Sized {
...
+    ///
+    /// # Examples
+    ///
+    /// This example draws a `BinaryColor` image to a `Rgb888` display.
```
jamwaffles, line 590:
```suggestion
    /// This example draws a `BinaryColor` image to an `Rgb888` display.
```

tinytga/src/dynamic_tga.rs
```diff
@@ -0,0 +1,92 @@
...
+            target_color_type: PhantomData,
+        })
+    }
+
+    /// Returns a reference to the raw TGA image.
+    ///
+    /// The [`RawTga`] object can be used to access lower level details about the TGA file.
+    ///
+    /// [`RawTga`]: struct.RawTga.html
+    pub fn raw(&self) -> &RawTga<'a> {
```
jamwaffles, line 58:
```suggestion
    pub fn as_raw(&self) -> &RawTga<'a> {
```

tinytga/src/lib.rs
```diff
@@ -7,65 +7,97 @@
...
 //! let image = Image::new(&tga, Point::zero());
 //!
 //! image.draw(&mut display)?;
 //! # Ok::<(), core::convert::Infallible>(()) }
 //! ```
+//! ## Accessing pixels using an embedded-graphics color type
+//!
+//! Even if tinytga is used without using [embedded-graphics] to draw the image the color types
+//! provided by [embedded-graphics] can still be used to access the pixel data using the
+//! [`pixels`](struct.Tga.html#method.pixels) method.
```
jamwaffles, line 62: This took a couple of readthroughs to understand the meaning. I've hopefully clarified it a bit:

```suggestion
//! If [embedded-graphics] is not used to draw the TGA image, the color types provided by 
//! [embedded-graphics] can still be used to access the pixel data using the 
//! [`pixels`](struct.Tga.html#method.pixels) method.
```

tinytga/src/lib.rs
```diff
@@ -7,65 +7,97 @@
...
+//! // Collect pixels into a vector.
+//! let pixels: Vec<_> = img.pixels().collect();
+//! ```
 //!
 //! ## Accessing raw pixel data
 //!
 //! If you do not want to use the color types provided by [embedded-graphics] you can also access
-//! the raw image data.
+//! the raw image data. The iterator returned by the [`pixels`](struct.RawTga.html#method.pixels)
+//! method uses `u32` values to return the raw color value of each pixel.
```
jamwaffles, line 86: This paragraph might be better written as something like:

```
//! If [embedded-graphics] is not used in the target application, the raw image data can be accessed
//! with the [`pixels`](struct.RawTga.html#method.pixels) method on [`RawTga`](struct.RawTga.html). 
//! The returned iterator produces a `u32` for each pixel value.
```

The "you" wording sounds a bit odd compared to the rest of the docs that use an imperative form.

jamwaffles (changes requested):
Looks good. I left some small comment/naming nits.

tinytga/src/lib.rs
```diff
@@ -111,193 +147,93 @@
...
 
-    /// Returns an iterator over the raw pixels in this image.
-    pub fn pixels<'b>(&'b self) -> Pixels<'b, 'a, C> {
-        Pixels::new(self.raw_pixels())
+    /// Returns a reference to the raw TGA image.
+    ///
+    /// The [`RawTga`] object can be used to access lower level details about the TGA file.
+    ///
+    /// [`RawTga`]: struct.RawTga.html
+    pub fn raw(&self) -> &RawTga<'a> {
```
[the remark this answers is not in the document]
rfuest: That makes sense. I had considered `raw` to be a getter, but it could also be considered to be another view of the object.

tinytga/src/lib.rs
rfuest: Thanks for the suggestion, I'll use that.

embedded-graphics/src/draw_target/mod.rs
rfuest: I don't have any better ideas, so I've used the names you've suggested.

jamwaffles (approved):
Looks good!

PR #451, opened 2020-10-26
Detect errors in for loops in the justfile

### LINKED ISSUE #449: Outdated tinybmp readme isn't detected
The master version of the `tinybmp` README is currently outdated, but that problem doesn't seem to be detected by `just build`. After I've added some changes to the `tinytga` README `just build` now reports that both READMEs are outdated, which suggests that there is some problem with the check or the justfile.

rfuest:
This PR should fix #449. The problem was caused by some shell scripting weirdness: Apparently for loops in shell scripts only return the exit code of the last iteration and ignore any other errors.

If everything works as expected CI should break with this fix applied. I didn't want to include the README update in this PR, because it is also included in #450. If we merge that one before this PR no merge conflicts should exist.

### DIFF
--- a/justfile
+++ b/justfile
@@ -97,10 +97,14 @@ generate-readme crate: (_build-readme crate)
 
 # Generate README.md for all crates
 generate-readmes:
+    #!/usr/bin/env bash
+    set -e
     for crate in {{crates}}; do just generate-readme $crate; done
 
 # Checks README.md for all crates
 check-readmes:
+    #!/usr/bin/env bash
+    set -e
     for crate in {{crates}}; do just check-readme $crate; done
 
 # Builds README.md for a single crate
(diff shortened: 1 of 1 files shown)


jamwaffles (approved):
Bash strikes again!

I assume this breakage is fixed in #450?

rfuest:
> Bash strikes again!

Yeah, I _really_ like writing shell scripts :angry:

> I assume this breakage is fixed in #450?

I hope that it is. We should rerun CI after #450 is merged.

jamwaffles:
> Yeah, I really like writing shell scripts 

Makes me wonder how much of the `justfile` could be calls to `cargo run` to custom Rust binaries in the repo. Would help with Windows support.

Do you want me to merge this into master even with failing builds? I can do it as admin but I don't think you have permission to do so.

rfuest:
> Do you want me to merge this into master even with failing builds?

No, I was planning to merge this after #450, so I don't have to rebase that PR. But if you think this should be added first I could also update the README in this PR to check if it fixes the build and then rebase #450.

jamwaffles:
Oh I see, I misunderstood your original comment. No worries, let's leave this for after #450 is merged.

rfuest:
After rebasing and rerunning CI the build error is gone as expected, so this should be ready to be merged.

jamwaffles (approved):
Looks good

PR #456, opened 2020-11-02
Upgrade to Rust 1.47

jamwaffles:

## PR description

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,4 +1,4 @@
-FROM cimg/rust:1.40.0
+FROM cimg/rust:1.47.0
 
 RUN sudo apt update && sudo apt install -y libsdl2-dev libssl-dev && sudo apt clean
 
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -2,7 +2,7 @@ FROM cimg/rust:1.47.0
 
 RUN sudo apt update && sudo apt install -y libsdl2-dev libssl-dev && sudo apt clean
 
-RUN cargo install just --version ^0.6
+RUN cargo install just --version ^0.8
 RUN cargo install cargo-readme --version ^3.2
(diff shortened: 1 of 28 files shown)


jamwaffles:
This isn't really giving us much. Performance is a mixed bag but is largely the same and we still can't `const` lots of things. I'd be tempted to close this and wait until there's a more compelling time to upgrade, as somewhat old as 1.40 is.

rfuest:
`embedded-fonts` needs some newer proc-macro features, which were introduced in 1.44, if I remember correctly. Should we only require a newer rust version for that crate or use the same MSRV for all crates?

jamwaffles:
>  Should we only require a newer rust version for that crate or use the same MSRV for all crates?

Only for that crate, if possible - that way users who don't want `embedded-fonts` don't need to use a recent version of Rust if they don't want to/can't.

rfuest:
OK, but this means that we will also need to maintain multiple versions of the docker image.

jamwaffles:
I think that's ok as each project can evolve separately that way. 

We could move the relevant bits of the `just` Docker build/push commands and CircleCI config changes in e-g into `embedded-fonts` and get it using its own Docker image. I'm happy to do that if you don't want to.

rfuest:
> We could move the relevant bits of the `just` Docker build/push commands and CircleCI config changes in e-g into `embedded-fonts` and get it using its own Docker image. I'm happy to do that if you don't want to.

Or we could create a new repo in the e-g org to contain all docker related files. IMO that would make it easier to maintain than having docker related files spread across multiple repos.

jamwaffles:
Good shout. Is any of this a blocker to the actual e-f work that needs doing?

rfuest:
No, we can work on CI for e-f when it is added to the org.

jamwaffles:
Closing, see #463 for build tools tracking.

PR #457, opened 2020-11-02
Fix filter_readme.sed

rfuest:
The removal of the square bracket links was slightly broken which caused colons at the end of sentences to disappear.


jamwaffles (approved):
Well spotted

Issue #460, opened 2020-11-03
Split core traits


jamwaffles:
This has been discussed before, IIRC with @rfuest raising the suggestion that we split core items like `DrawTarget` out into an `embedded-graphics-core` trait. 

I'm opening this issue because it might fix another issue: `tinytga` is used in the e-g doc examples, but `tinytga` depends on e-g itself, creating a cyclic dependency that breaks as soon as the versions aren't _perfectly_ in sync. This is obviously not good, so along with `DrawTarget`, we may want to consider splitting more traits for images and primitives, as well as the pixel colours, into a new core crate. 

If we go ahead with something like this, should it target 0.7 or go for a 0.8 release?

Discuss.

rfuest:
> I'm opening this issue because it might fix another issue: `tinytga` is used in the e-g doc examples, but `tinytga` depends on e-g itself, creating a cyclic dependency that breaks as soon as the versions aren't _perfectly_ in sync.

I think the easiest solution would be to remove any usage of `tinytga` in e-g itself. My suggestion would be to only use raw image in the e-g docs, which is enough to show the e-g side of the image API. In addition to that we could add links to some e-g doc examples that point to a copy of the example in the `tinytga` or `tinybmp` docs which uses a TGA or BMP file instead of raw data.

The `simulator` currently also has a dependency on `tinytga` and `tinybmp`, which caused some problems when I worked on e-f. But this shouldn't be a problem anymore if we split the examples into another repo.

> This is obviously not good, so along with `DrawTarget`, we may want to consider splitting more traits for images and primitives, as well as the pixel colours, into a new core crate.

I wouldn't want to add the entire `primitives` module to the core crate, because most of the types aren't necessary for the draw target or an image drawable. Instead we could only move the `Rectangle` struct, that is required by `DrawTarget` and `ImageDrawable`, to the core crate. Normal e-g users wouldn't notice this change because the `embedded-graphics` crate can reexport the `Rectangle` type in the `primitives` module.

I think the following types should be included in a core crate:
- `geometry::*`
- `pixelcolor::*`
- `draw_target::*`
  Not only `DrawTarget` is required, but also the other types, because they are used in `ImageDrawable` and the `tiny...` crates
- `image::ImageDrawable`
- `primitives::Rectangle` including all impls for traits that are moved into the core crate
  Other traits like `Primitive` should be implementable in `embedded-graphics`.

[...]

PR #465, opened 2020-11-09
Add benches from simulator crate

rfuest:

## PR description

This PR adds the benches from the simulator create to this crate. I'm not sure why we originally had to move them into the simulator crate, but it seems to work at the moment. IMO they belong in this crate, so I would like to try if this works and only move them somewhere else when we encounter problems again.

I also made some changes to the benches:

I've replaced the `tinybmp` image test by an `ImageRaw` one to remove the `tinybmp` dependency. IMO it would be better to add e-g drawing benches to `tinybmp` and `tinytga` anyway.

The benches used to collect pixels in a `Vec` which isn't the most accurate representation of actual drawing. Because the `Vec` size isn't known in advance the `Vec` might need to increase the capacity several times. The effect of this reallocation isn't huge, but I was able to measure it. This PR instead implements a simple framebuffer that is similar to what an display with an in memory framebuffer would use.

### DIFF
--- a/justfile
+++ b/justfile
@@ -16,7 +16,11 @@ build: check-formatting test test-all build-benches check-readme check-links
 
 # Build the benches
 build-benches:
-    cargo bench --no-run
+    cargo bench --features "criterion" --no-run
+
+# Run the benches
+bench:
+    cargo bench --features "criterion"
 
 # Run cargo test in release mode
 test:
--- a/justfile
+++ b/justfile
@@ -9,7 +9,7 @@ ci_build_image := "jamwaffles/circleci-embedded-graphics:1.40.0-cimg"
 # Building
 #----------
 
-build: check-formatting test test-all check-readme check-links
+build: check-formatting test test-all build-benches check-readme check-links
 
 # Build the benches
 build-benches:
--- a/justfile
+++ b/justfile
@@ -19,8 +19,8 @@ build-benches:
     cargo bench --features "criterion" --no-run
 
 # Run the benches
-bench:
-    cargo bench --features "criterion"
+bench *args:
+    cargo bench --features "criterion" {{args}}
 
 # Run cargo test in release mode
 test:
--- a/justfile
+++ b/justfile
@@ -30,7 +30,8 @@ check-formatting:
 # Cross compiles embedded-graphics for a target
 build-target target *args:
     cargo build --target {{target}} {{args}}
-    cargo build --target {{target}} --all-features {{args}}
+    cargo build --target {{target}} --features nalgebra {{args}}
+    cargo build --target {{target}} --features fixed {{args}}
 
 # Cross compiles embedded-graphics for all targets
 build-targets *args:
--- a/justfile
+++ b/justfile
@@ -5,6 +5,9 @@ target_dir := "target"
 # when upgrading the tag next time.
 ci_build_image := "jamwaffles/circleci-embedded-graphics:1.40.0-cimg"
 
+# list of all features except criterion
+all_features := "nalgebra fixed"
+
 #----------
 # Building
 #----------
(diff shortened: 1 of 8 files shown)


jamwaffles:
Alas this is why they were moved:

image

Is there a clean way to "turn off" the benchmarks for non-std builds? I tried some `cfg()` haxx a while back I think but didn't get very far.

rfuest:
I hope the new feature resolver will be stabilized soon, because I think it did fix the problem last time I tried.

I've updated the PR with a hacky solution, that changes `criterion` to be a normal dependency instead of a dev-dependency. This way it's possible to mark it as optional so that it doesn't affect the build when the `criterion` feature isn't enabled. The biggest problem this hack causes is that `cargo build --all-features` is now broken for any `no_std` target and you need to manually list the required features.

Should we use this hack or search for another solution?

justfile
```diff
@@ -30,7 +30,8 @@ check-formatting:
 # Cross compiles embedded-graphics for a target
 build-target target *args:
     cargo build --target {{target}} {{args}}
-    cargo build --target {{target}} --all-features {{args}}
+    cargo build --target {{target}} --features nalgebra {{args}}
+    cargo build --target {{target}} --features fixed {{args}}
```
rfuest, line 34: If we go with this solution: Should we test both features in one build or use separate ones? I think the total build time has increased, but this would test the features independent of each other.

jamwaffles: I'm on the fence. Rerunning the build leaves it at 2m 10s which is still pretty good.

Cargo features are supposed to be additive, right, so *should* be safe to compile together. We also know that `nalgebra` and `fixed` are unlikely to conflict with each other, so I'd be fine merging these two back together if you wanted to.

jamwaffles:
Something else I noticed is `cargo bench` doesn't do anything by default now. Should we add a `just bench` command that adds `--features criterion` to make life a bit easier?

I also considered a guard file in `benches/` containing something like the following:

```rust
#[cfg(not(feature = "criterion"))]
compile_error!(
    "The `criterion' feature must be enabled to run the benchmarks. Rerun with `just bench'."
);
```

but that might be a step too far. Thoughts?

justfile
rfuest: Sure, they should be sage to compile together, but compiling them separately makes sure they don't depend on each other.

But I think building it once is enough. I've added an `all_features` variable to the justfile and use that in all instances that used `--all-features` before. This should minimize the build time for `build-all` because it doesn't need to build criterion anymore.

rfuest:
> Something else I noticed is `cargo bench` doesn't do anything by default now. Should we add a `just bench` command that adds `--features criterion` to make life a bit easier?

I hadn't event noticed this yet :roll_eyes:. `just bench` and the error messages are good ideas and I've added both.

Do you think we should implement the benches this way for the moment? If we decide to use a single repo for #460 we could try to move the benches to an unpublished child crate to see if this works better. But until then IMO it's better to have them in this crate instead of the simulator one.

jamwaffles (changes requested):
> Do you think we should implement the benches this way for the moment? If we decide to use a single repo for #460 we could try to move the benches to an unpublished child crate to see if this works better. But until then IMO it's better to have them in this crate instead of the simulator one.

I agree. With a monorepo before, it didn't matter as much that the benches were under `simulator`, but now it's a bit more important they live with e-g.

Something else I didn't test before was passing filters to `just bench`, e.g. `just bench triangle`. Could this be added to the `justfile`? After that I think this is good to merge.

rfuest:
> Could this be added to the `justfile`?

Good idea, I've added this to the `justfile`.

jamwaffles (approved):
Looks good!*

* as much as cargo feature hacking can anyway ;)

rfuest:
>  * as much as cargo feature hacking can anyway ;)

Features are easily my least favorite feature of cargo :wink:, at least in the current state. I hope this won't hack won't cause more problems in the future.

PR #466, opened 2020-11-09
Package upgrades

jamwaffles:

## PR description

The result of a [`cargo upgrade`](https://crates.io/crates/cargo-edit#cargo-upgrade) and `cargo update`.

I don't think this needs a changelog entry does it?


jamwaffles:
Argh, I did run the build locally but didn't have the override set after the repo move.

jamwaffles:
`fixed` was failing. It requires an MSRV of 1.44.0 as of [1.0.0](https://gitlab.com/tspiteri/fixed/blob/master/RELEASES.md#version-100-2020-06-04).

rfuest:
> I don't think this needs a changelog entry does it?

The `nalgebra` update from `0.19.0` to `0.23.0` is a breaking change and should IMO be included in the changelog.

jamwaffles:
Good catch. Changelog entry added. Should be good for review.

rfuest (approved):
LGTM

PR #467, opened 2020-11-09
Add OffsetOutline trait to prelude

rfuest:

## PR description

This PR adds the `OffsetOutline` trait to the prelude. The other traits in the `primitive` module are already included in the prelude, so I think it makes sense to also add this trait.


jamwaffles (approved):
Makes sense :)

PR #468, opened 2020-11-10
Use correct feature for nalgebra in justfile

rfuest:

## PR description

The recent dependency update had broken the `nalgebra` integration. This wasn't noticed because the `justfile` was also broken and used a wrong feature for `nalgebra`. This PR should fix both issues.

I've also added `#[allow(dead_code)]` to the `StrokeOffset` to make it easier to spot new errors and warnings.

### DIFF
--- a/src/primitives/line/mod.rs
+++ b/src/primitives/line/mod.rs
@@ -339,7 +339,7 @@ impl Transform for Line {
 }
 
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
-#[allow(dead_code)] // TODO: remove when all variants are used
+#[allow(dead_code)]
 pub(in crate::primitives) enum StrokeOffset {
     /// Stroke is centered around the line skeleton.
     None,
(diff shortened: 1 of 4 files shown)


src/primitives/line/mod.rs
```diff
@@ -339,6 +339,7 @@ impl Transform for Line {
 }
 
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[allow(dead_code)] // TODO: remove when all variants are used
```
jamwaffles, line 342: All going well, this will be removed in the thick triangle work I'm doing at the moment. The warnings are annoying, but could you remove this TODO as I have a feeling we'll forget it in the next review.

jamwaffles (changes requested):
Oops!

src/primitives/line/mod.rs
rfuest: > All going well, this will be removed in the thick triangle work I'm doing at the moment.

Nice, I hope this will be less aggravating than the polylines.

> The warnings are annoying, but could you remove this TODO as I have a feeling we'll forget it in the next review.

OK.

jamwaffles: It's going well so far but as with all of these cursed features I'll probably spend 6 months on the last 0.01%!

jamwaffles (approved):

PR #470, opened 2020-11-11
Text refactor

rfuest:

## PR description

This PR adds support for external text renderers to e-g. `TextStyle` was converted to a trait that can be also implemented in other crates. To make drawing more efficient text renderers get direct access to the draw target and can use optimizations like `fill_contiguous`. Renderers can implement `TextStylePixels` to also provide a pixel iterator, but this isn't required.

Are the names `MonospacedTextStyle` and `MonospacedTextStyleBuilder` too long? Do you have a better suggestion?


jamwaffles:
I'm a bit frazzled at the moment so I'll get round to a review later (sorry), but as for 

> Are the names MonospacedTextStyle and MonospacedTextStyleBuilder too long? Do you have a better suggestion?

How about `MonoTextStyle` and `MonoTextStyleBuilder`? You could change `MonospacedFont` to `MonoFont`, etc as well. `Mono` might be a bit ambiguous at this point, but my hope is that if it's used everywhere for all the monospaced fonts, the name along with all the documentation around it will be clear.

rfuest:
I've also implemented an example of an external text renderer in my fork of the `simulator` repo: https://github.com/rfuest/simulator/blob/text-refactor/examples/text-seven-segment.rs

It can be used to draw seven segment displays with any scale:

grafik

The code is too complex to keep it as an example, but I think it could be useful to release the renderer as a separate crate.

rfuest:
> I'm a bit frazzled at the moment so I'll get round to a review later (sorry), but as for

Never mind, it's only a draft PR and you can review it when it's done.

> How about `MonoTextStyle` and `MonoTextStyleBuilder`? You could change `MonospacedFont` to `MonoFont`, etc as well. `Mono` might be a bit ambiguous at this point, but my hope is that if it's used everywhere for all the monospaced fonts, the name along with all the documentation around it will be clear.

Sounds good to me, thanks for the suggestion.

src/fonts/mod.rs
```diff
@@ -228,17 +202,17 @@ mod tests {
...
         drawable::Drawable, geometry::Point, mock_display::MockDisplay, pixelcolor::BinaryColor,
-        style::TextStyle,
+        style::MonoTextStyle,
     };
 
     /// Draws a text using the given font and checks it against the expected pattern.
     pub(super) fn assert_text_from_pattern<F>(text: &str, font: F, pattern: &[&str])
     where
-        F: Font + Copy,
+        F: MonoFont + Copy,
```
bugadani, line 211: If you're at it, could you require MonoFont to implement Copy?

src/fonts/monospaced_pixels.rs
```diff
@@ -0,0 +1,112 @@
...
+                if let Some(color) = self.style.background_color {
+                    break Some(Pixel(Point::new(x, y), color));
+                }
+            } else if let Some(current_char) = self.current_char {
+                let color = if F::character_pixel(
+                    current_char,
+                    self.char_walk_x as u32,
+                    self.char_walk_y as u32,
+                ) {
+                    self.style.text_color.or(self.style.background_color)
```
bugadani, line 79: Hmm, I already closed #439 assuming you'd done that. Would you incorporate those changes, or should I reopen it after this is merged?

src/fonts/monospaced_pixels.rs
```diff
@@ -0,0 +1,112 @@
...
+                if self.char_walk_y >= F::CHARACTER_SIZE.height as i32 {
+                    self.char_walk_y = 0;
+                    self.char_walk_x += 1;
+                }
+
+                if let Some(color) = self.style.background_color {
+                    break Some(Pixel(Point::new(x, y), color));
+                }
+            } else if let Some(current_char) = self.current_char {
+                let color = if F::character_pixel(
```
bugadani, line 74: While it's not a huge difference, this lookup has a bit of perf impact. In embedded-text, I made a glyph renderer struct that contains the pixel offset instead of the character and avoids some unnecessary re-calculations for each pixel. Also I think it would be nice if this code could be reused in embedded-text, but I think I might implement that after this is merged :)

src/fonts/monospaced_pixels.rs
rfuest: I would suggest to merge #439 first and I'll rebase this PR to incorporate the changes in that PR.

src/fonts/mod.rs
rfuest: I think that should be possible, but I'll need to take a closer look.

src/fonts/monospaced_pixels.rs
rfuest: I want to try to reuse the `RawDataIter` that is used to draw images. This would reduce some code duplication but I'm not sure about the performance impact yet. If that doesn't work out we can look into optimizing this code.

src/fonts/mod.rs
rfuest: Done.

CHANGELOG.md
```diff
@@ -13,10 +13,17 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#386](https://github.com/embedded-graphics/embedded-graphics/pull/386) Added the `delta` method to `Line` to compute the difference between start and end points.
 - [#450](https://github.com/embedded-graphics/embedded-graphics/pull/450) Added `ColorConverted` and `DrawTargetExt::color_converted` to support color conversion for draw targets.
 - [#438](https://github.com/embedded-graphics/embedded-graphics/pull/438) Added majority CSS web colors as associated `const`s to the RGB color types.
+- [#470](https://github.com/embedded-graphics/embedded-graphics/pull/470) Added support for external text renderers. External text renderers can be implemented using the new `TextStyle` trait.
 
 ### Changed
 
 - **(breaking)** [#466](https://github.com/embedded-graphics/embedded-graphics/pull/466) Upgrade Nalgebra from 0.19.0 to 0.23.0.
+- **(breaking)** [#470](https://github.com/embedded-graphics/embedded-graphics/pull/470) Renamed `Font`, `TextStyle` and `TextStyleBuilder` to `MonoFont`, `MonoTextStyle` and `MonoTextStyleBuilder`.
```
jamwaffles, line 21: So we can keep track of what needs to go in the migration guide instead of doing it all at once, could you add a section about these changes? Bullet points are fine if you prefer - I can flesh them out later.

src/fonts/monospaced_pixels.rs
jamwaffles: > While it's not a huge difference, this lookup has a bit of perf impact.

Did you see a perf impact running `just bench font`? I realise they're contrived benchmarks and perhaps not representative of realworld performance, but the text benches improve ~10% for me on this branch compared to `master`.

src/fonts/mod.rs
```diff
@@ -111,47 +111,45 @@
 //!
 //! | Type | Screenshot |
 //! |------|------------|
-//! | [`Font6x6`] | ![6x6 font spritemap screenshot](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/master/data/font6x6.png) |
 //! | [`Font6x8`] | ![6x8 font spritemap screenshot](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/master/data/font6x8.png) |
```
jamwaffles, line 114: FYI: I've opened #479 to track changes around fixing these links.

src/fonts/mod.rs
```diff
@@ -169,33 +167,9 @@ pub trait Font {
     /// on a single line of text.
     const CHARACTER_SPACING: u32 = 0;
 
-    /// Whether characters have a variable width or not.
-    ///
-    /// Variable width characters have a maximum width of CHARACTER_SIZE.x, but the empty columns at
-    /// the right of each characters are ignored, allowing some characters to be smaller than others.
-    const VARIABLE_WIDTH: bool = false;
-
     /// Returns the position a character in the font.
```
jamwaffles, line 170:
```suggestion
    /// Returns the position of a character in the font.
```

src/fonts/monospaced_pixels.rs
```diff
@@ -0,0 +1,112 @@
...
+impl<C, F> Iterator for MonoPixels<'_, C, F>
+where
+    C: PixelColor,
+    F: MonoFont,
+{
+    type Item = Pixel<C>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            if self.current_char == Some('\n') {
```
jamwaffles, line 54: Thanks for keeping this in :)

src/style/mono_text_style.rs
```diff
@@ -0,0 +1,250 @@
...
+///     .background_color(Rgb565::BLUE)
+///     .build();
+///
+/// let text = Text::new("Hello Rust!", Point::new(0, 0)).into_styled(style);
+/// ```
+///
+/// ## Transparent background
+///
+/// If a property is omitted, it will remain at its default value in the resulting
+/// `MonoTextStyle`returned by `.build()`. This example draws white text with no background at
```
jamwaffles, line 138:
```suggestion
/// `MonoTextStyle` returned by `.build()`. This example draws white text with no background at
```

jamwaffles (changes requested):
My eyes have gone a bit square looking at the word `Mono` over and over again, but looks good overall.

One thing that's missing that I think is important is an example of implementing a custom font. I don't know if it's possible to add one that's short enough to show in the docs, but it would really help tie all the traits and structs together.

There's a reference to the `text-custom-font` simulator example somewhere (I can't find it in the diff now for some reason) which might be worth turning into a link too.

src/fonts/monospaced_pixels.rs
rfuest: Sorry, but the final verdict is still open on this one. After this PR each font renderer still needs to handle newlines itself, but I'll look into it after this is merged.

CHANGELOG.md
rfuest: I've added some very basic bullet points to the migration guide. I'll try to remember to update the guide when I add more text features in future PRs.

rfuest:
I couldn't reply to this above:

> Did you see a perf impact running just bench font? I realise they're contrived benchmarks and perhaps not representative of realworld performance, but the text benches improve ~10% for me on this branch compared to master.

I hadn't noticed that before, but I also get some performance improvements. Not quite 10%, but I hadn't expected any performance gains at all, so the 3%-7% are also welcome.

Since the benches were updated to use a framebuffer they should be pretty representative, but only for displays drivers with a framebuffer. Perhaps we should look into also adding benches for other displays, where the number of draw calls or the amount of data that has to be transmitted are better metrics.

src/fonts/monospaced_pixels.rs
jamwaffles: I think that's ok for external fonts. I'd like to keep at as a feature of the builtin mono fonts if it's not too complicated. Otherwise I'll concede and just use multiple `Text` calls :)

rfuest: > I think that's ok for external fonts.

I don't ;). I would like to keep the external text renderer API as simple as possible.

> Otherwise I'll concede and just use multiple `Text` calls :)

I don't think you need to. It should be possible to move the handling of newlines from the text renderers to the styled text drawable. This way each text renderer would only need to render single lines and get the the multiline support for free. But I'll investigate this in another PR.

rfuest:
> One thing that's missing that I think is important is an example of implementing a custom font. I don't know if it's possible to add one that's short enough to show in the docs, but it would really help tie all the traits and structs together.
> 
> There's a reference to the `text-custom-font` simulator example somewhere (I can't find it in the diff now for some reason) which might be worth turning into a link too.

An example in the docs wouldn't be much shorter than `text-custom-font`. If it's OK, I would like to create a new issue for the doc improvements and wait until the text refactor is finished.

src/fonts/monospaced_pixels.rs
jamwaffles: Sounds good!

jamwaffles:
> An example in the docs wouldn't be much shorter than text-custom-font. If it's OK, I would like to create a new issue for the doc improvements and wait until the text refactor is finished.

I thought that might be the case. No worries - we can add more in later.

jamwaffles (approved):
Looks good from my side. I don't know if @bugadani had anything else to add?

bugadani:
Don't wait for me, I can't review in depth :(

rfuest:
> I've also implemented an example of an external text renderer in my fork of the `simulator` repo: https://github.com/rfuest/simulator/blob/text-refactor/examples/text-seven-segment.rs
> 
> It can be used to draw seven segment displays with any scale:
> 
> grafik
> 
> The code is too complex to keep it as an example, but I think it could be useful to release the renderer as a separate crate.

@jamwaffles  If I decide to convert this example into a reusable crate do you think it would be useful enough to be an official e-g project or should I release it myself?

jamwaffles:
I think it'd be right at home in the e-g org if that's what you mean. Would it end up being published on crates.io? 

As complex as it is, it would also be a good example for those looking to implement a custom font.

rfuest:
> I think it'd be right at home in the e-g org if that's what you mean.

Yes, that's what I meant.

> Would it end up being published on crates.io?

Yes. IMO a freely scalable seven segment font could be useful for different projects.

> As complex as it is, it would also be a good example for those looking to implement a custom font.

Just to be clear: It's not a custom font, but a custom text renderer. But I agree that it could still be a good example for how an external renderer can be integrated.

jamwaffles:
> Yes, that's what I meant.

> Yes. IMO a freely scalable seven segment font could be useful for different projects.

Definitely! You should be able to make repositories in the e-g org, so feel free to do so when you have time. Let me know if you run into permissions errors.

> Just to be clear: It's not a custom font, but a custom text renderer. But I agree that it could still be a good example for how an external renderer can be integrated.

Right, renderer not font. I'll do my best to be clearer with the terminology from now on :)

rfuest:
>  I'll do my best to be clearer with the terminology from now on :)

I just wanted to make sure we talk about the same thing ;).

PR #472, opened 2020-11-14
Merge scanlines

jamwaffles:

## PR description

Adds an optimisation/cleanup to merge adjacent scanlines when rasterising thick polygons.

A ~5% performance boost on x64, so not as much as I would've hoped for but it's in the right direction. The complex `fix_overlap` function is removed for a simple accumulator in the iterator that holds the currently growing line.

### DIFF
--- a/src/primitives/polyline/scanline_intersections.rs
+++ b/src/primitives/polyline/scanline_intersections.rs
@@ -19,7 +19,7 @@ pub struct ScanlineIntersections<'a> {
     scanline_y: i32,
     next_start_join: Option<LineJoin>,
     width: u32,
-    prev_line: Option<Line>,
+    accum: Option<Line>,
 }
 
 impl<'a> ScanlineIntersections<'a> {
@@ -45,7 +45,7 @@ impl<'a> ScanlineIntersections<'a> {
             points,
             remaining_points: points,
             scanline_y,
-            prev_line: None,
+            accum: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/primitives/polyline/scanline_intersections.rs
```diff
@@ -89,103 +89,105 @@ impl<'a> ScanlineIntersections<'a> {
...
                     }
-                }
 
-                self.prev_line = Some(line);
-
-                return Some(line);
+                    // If next segment touches current accum, extend accum and continue.
+                    self.accum = Some(extend(accum, next_segment));
+                }
+                // Initialise accumulator with a single line. Next iteration will return it if
```
rfuest, line 133:
```suggestion
                // Initialize accumulator with a single line. Next iteration will return it if
```

src/primitives/polyline/scanline_intersections.rs
```diff
@@ -89,103 +89,105 @@ impl<'a> ScanlineIntersections<'a> {
...
+///
+/// # Overlapping - merge
+/// X---Y+++X+++Y
+///      ⇓
+/// X-----------X
+///
+/// # Separate - leave alone
+/// X---X Y---Y
+///      ⇓
+/// X---X Y---Y
```
rfuest, line 114: I was very briefly confused by X and Y, because I hadn't read the text and thought the lines were using a different orientation. IMO it's better to use other markers:
```suggestion
/// A---AB+++B
///     ⇓
/// A--------A
///
/// # Overlapping - merge
/// A---B+++A+++B
///      ⇓
/// A-----------A
///
/// # Separate - leave alone
/// A---A B---B
///      ⇓
/// A---A B---B
```

src/primitives/polyline/scanline_intersections.rs
```diff
@@ -89,103 +89,105 @@ impl<'a> ScanlineIntersections<'a> {
...
-    // Don't need to draw the second line if it will be completely obscured by first.
-    if l1 == l2 || is_completely_contained {
-        return None;
+        // No more segments - return the final accumulated line.
+        self.accum.take()
     }
+}
 
-    let mut l2 = l2;
+/// Check for lines that are adjacent or overlapping.
```
rfuest, line 146:
```suggestion
/// Check for lines that are adjacent or overlapping.
///
/// This assumes that both lines have the same y coordinate and are horizontal.
```

rfuest (changes requested):
That looks much cleaner than the previous code and I only have a few comment nits.

We could save a few bytes of memory by using `Option<(i32, i32)>` or perhaps `Option<RangeInclusive<i32>>`, because all we don't need to store the y coordinates. But I'm not sure if that's worth the effort.

And I think there has been some rebase issue, because the commit is called: "Add support for transparent characters"

jamwaffles:
> We could save a few bytes of memory by using Option<(i32, i32)> or perhaps Option<RangeInclusive<i32>>, because all we don't need to store the y coordinates. But I'm not sure if that's worth the effort.

We should definitely take a look at this in the future, but maybe it would be a good idea to take a look at thick polylines/thick polygons/thick triangles when all three are supported. For example, I'm working on thick stroked triangles at the moment and there might be some common code/logic we can introduce with a wider picture.

> And I think there has been some rebase issue, because the commit is called: "Add support for transparent characters"

Oops. Let me fix that...

rfuest:
> We should definitely take a look at this in the future, but maybe it would be a good idea to take a look at thick polylines/thick polygons/thick triangles when all three are supported. For example, I'm working on thick stroked triangles at the moment and there might be some common code/logic we can introduce with a wider picture.

Sounds good to me.

rfuest (approved):

PR #475, opened 2020-11-14
Triangles with thick stroke

### LINKED ISSUE #384: Triangles with thick strokes
- Version of embedded-graphics in use (if applicable): 0.6.2
- Target: simulator

Triangle ignores stroke thickness

## Test case (if applicable)

```rust
use embedded_graphics_simulator::{
    BinaryColorTheme, OutputSettingsBuilder, SimulatorDisplay, Window,
};

use embedded_graphics::{
    pixelcolor::BinaryColor, prelude::*, primitives::Triangle, style::PrimitiveStyle,
};

fn main() -> Result<(), core::convert::Infallible> {
    let mut display: SimulatorDisplay<BinaryColor> = SimulatorDisplay::new(Size::new(128, 64));

    let output_settings = OutputSettingsBuilder::new()
        .theme(BinaryColorTheme::OledBlue)
        .build();

    let thick_stroke = PrimitiveStyle::with_stroke(BinaryColor::On, 3);

    Triangle::new(Point::new(0, 0), Point::new(12, 0), Point::new(6, 12))
        .into_styled(thick_stroke)
        .draw(&mut display)
        .unwrap();

    Window::new("Triangle", &output_settings).show_static(&display);
    Ok(())
}
```

screenshot

Haven't checked master very closely, only a cursory glance on the git history, but I haven't seen any explicit commit targeted towards triangles.

jamwaffles:

## PR description

Adds thick stroke support for `Triangle`s. 

Closes #384

### DIFF
--- a/src/primitives/triangle/mod.rs
+++ b/src/primitives/triangle/mod.rs
@@ -1,17 +1,21 @@
 //! The triangle primitive.
 
 mod points;
+mod scanline_intersections;
 mod scanline_iterator;
 mod styled;
 
 use crate::{
     geometry::{Dimensions, Point},
-    primitives::{ContainsPoint, Line, Primitive, Rectangle},
+    primitives::{
+        line::StrokeOffset, line_join::LineJoin, thick_segment::bresenham_scanline_intersection,
+        ContainsPoint, Line, Primitive, Rectangle,
+    },
     transform::Transform,
 };
 use core::{
     borrow::Borrow,
-    cmp::{max, min},
+    cmp::{max, min, Ordering},
 };
 pub use points::Points;
 pub use styled::StyledPixels;
@@ -125,7 +129,7 @@ impl ContainsPoint for Triangle {
 
         // Sort points into same order as `ScanlineIterator` so this check produces the same results
         // as a rendered triangle would.
-        let (p1, p2, p3) = sort_yx(p1, p2, p3);
+        let Triangle { p1, p2, p3 } = self.sorted_yx();
 
         // Special case: due to the Bresenham algorithm being used to render triangles, some pixel
         // centers on a Styled<Triangle> lie outside the mathematical triangle. This check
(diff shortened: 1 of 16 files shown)


jamwaffles:
Not quite as bad as polylines, but still rather a lot of code and a shit load of edge cases :sweat:. I won't be surprised (but I'll be sad) if more come up.

Also, this... isn't great

image

I wouldn't mind leaving performance for another PR, but if 1200% slowdown is too much for other people, I can take a look.

rfuest (changes requested):
I didn't have time to do a very thorough review, but the output looks good an I wasn't able to break it yet.

> I wouldn't mind leaving performance for another PR, but if 1200% slowdown is too much for other people, I can take a look.

That's a pretty big regression, but IMO it's better to have working code than fast code. Let's merge the code after my comments are resolved and handle any possible improvements in other PRs. There are also some `set_allow_overdraw`s in the test that shouldn't be necessary anymore, which is always nice to see.

src/primitives/triangle/mod.rs
```diff
@@ -180,11 +184,122 @@ impl Triangle {
...
+    type Output = Point;
+
+    fn index(&self, idx: usize) -> &Self::Output {
+        match idx {
+            0 => &self.p1,
+            1 => &self.p2,
+            2 => &self.p3,
+            _ => unreachable!(),
+        }
+    }
```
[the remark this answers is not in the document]
jamwaffles: I would rather not change the `Triangle` structure as it's public. This commend did make me realise that the `Index` impl is public, so I've changed that to a more private `.vertex` method on `Triangle`. It still has an `unreachable()` in it, but this time it really should be as the index wrapping is done in the method.

jamwaffles:
I've just pushed a commit to reduce the amount of stored data for line joins. This reduces the `JoinKind` enum storage from 20 bytes to 2 bytes. I didn't test before, but a `LineJoin` is now 36 bytes, so a thick segment is 72. Performance has not changed.

There's more we could do in this area, but I'll leave that for another PR.

jamwaffles:
A little better

image

I'll stop for now. Sorry for the moving target if you were reviewing.

src/primitives/triangle/mod.rs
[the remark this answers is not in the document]
rfuest: OK, that's better. I still think that it might be beneficial to use a points array for `Triangle`, but that's outside the scope of this PR. Or did we already discuss this in the past? I'm not sure, but I vaguely remember something.

src/primitives/triangle/scanline_intersections.rs
```diff
@@ -0,0 +1,230 @@
...
+};
+
+/// Type of scanline.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum PointType {
+    /// Represents part of the stroke.
+    Border,
+
+    /// Represents the interior of the shape.
+    Inside,
```
rfuest, line 21: Couldn't these be named `Stroke` and `Fill`? We also use the same naming convention for `stroke_area` and `fill_area`. Or do these variants have a different meaning?

rfuest (changes requested):

src/primitives/triangle/scanline_intersections.rs
jamwaffles: Ah, no these are just a holdover from the previous triangle implementation. They can changed.

jamwaffles:
Looking at the latest commit, those changes alone regained about 10% of the lost performance against master:

image

We broke the 1000% barrier lol. Yay!

src/primitives/triangle/scanline_iterator.rs
```diff
@@ -1,192 +1,63 @@
...
+        triangle: &Triangle,
+        stroke_width: u32,
+        stroke_offset: StrokeOffset,
+        has_fill: bool,
+        bounding_box: &Rectangle,
+    ) -> Self {
+        let triangle = triangle.sorted_clockwise();
+
+        let scanline_y = bounding_box.top_left.y;
+        let scanline_limit = scanline_y + bounding_box.size.height as i32;
```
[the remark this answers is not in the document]
jamwaffles: I also had to remove a couple of derives from various structs, but they weren't really necessary anyway.

src/primitives/triangle/scanline_intersections.rs
```diff
@@ -0,0 +1,230 @@
...
+                    stroke_width,
+                    stroke_offset,
+                );
+                let end = LineJoin::from_points(
+                    t.vertex(idx + 1),
+                    t.vertex(idx + 2),
+                    t.vertex(idx + 3),
+                    stroke_width,
+                    stroke_offset,
+                );
```
jamwaffles, line 125: Something I'd like to try to improve performance with in the future is to cache the 3 generated edge `ThickSegment`s. I chose not to do it in this PR because it adds `72 * 3 = 216` bytes to the iterator size (minus a few bytes when removing the `LineConfig` struct) which is pretty unacceptable, even for a short-lived iterator struct IMO.

src/primitives/triangle/mod.rs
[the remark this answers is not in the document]
jamwaffles: I vaguely recall some discussions around this. I wouldn't be opposed to throwing another braking change on the pile, but I'm not sure what benefits the change would bring beyond easier indexing.

src/primitives/triangle/scanline_intersections.rs
rfuest: Sounds interesting, do you want to open an issue for this?

src/primitives/triangle/mod.rs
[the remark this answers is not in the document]
rfuest: IMO it would be more consistent with `Polyline` and (in the future) `Polygon`. They use a slice instead of an array, but the syntax is the same in many cases.

src/primitives/triangle/scanline_intersections.rs
jamwaffles: I don't think so. I want to take a closer look at all primitive drawing performance at some point, so I'll fold it into that.

If I do try this trick in the future, do you think 216+ bytes for the iterator alone is too much?

src/primitives/triangle/mod.rs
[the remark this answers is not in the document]
jamwaffles: That's a very good point. I think it would also allow (hopefully) quite a lot of code reuse in the scanline iterators too. I'll open an issue.

src/primitives/triangle/scanline_intersections.rs
rfuest: > If I do try this trick in the future, do you think 216+ bytes for the iterator alone is too much?

It's a lot and we should IMO first see if we can gain performance in other ways. If this is the only way to get a substantial performance gain we might need to consider to add a feature to select between better performance or lower memory consumption. I'm sure there would be more instances where such a feature could be useful.

rfuest (approved):

src/primitives/triangle/scanline_intersections.rs
jamwaffles: Yeah, I'll try other ways first.

> we might need to consider to add a feature to select between better performance or lower memory consumption. 

I'm glad we're thinking of the same things :)

jamwaffles:
:tada: thanks very much!

PR #477, opened 2020-11-14
Fix drawing of 90 degree args

rfuest:

## PR description

This PR makes drawing of 90° arcs consistent to drawing a quarter of a circle. Maybe this will help with #441. `Sector` will probably also need some tweaks, but that will be another PR.

### DIFF
--- a/src/primitives/arc/linear_equation.rs
+++ b/src/primitives/arc/linear_equation.rs
@@ -20,8 +20,8 @@ pub struct LinearEquation {
 impl LinearEquation {
     /// Create a new linear equation based on one point and one angle
     pub fn from_point_angle(point: Point, angle: Angle) -> Self {
-        // FIXE: angle.tan() for 180.0 degrees isn't exactly 0 which causes problems when drawing
-        //       a single quadrant. Is there a better solution to fix this?
+        // FIXME: angle.tan() for 180.0 degrees isn't exactly 0 which causes problems when drawing
+        //        a single quadrant. Is there a better solution to fix this?
         let (a, b) = if angle == Angle::from_degrees(180.0) {
             (Real::from(0.0), Real::from(-1.0))
         } else {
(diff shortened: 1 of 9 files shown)


rfuest:
I've converted this to a draft, because arcs with even diameters aren't handled correctly. There is a test for this case, but it makes the same wrong assumptions than the actual code.

rfuest:
It should now work as expected. The changes to the way the center point is calculated will also have some effect on `Sector` drawing, but I'll add additional tests for that in another dedicated PR.

src/primitives/arc/linear_equation.rs
```diff
@@ -20,10 +20,17 @@ pub struct LinearEquation {
 impl LinearEquation {
     /// Create a new linear equation based on one point and one angle
     pub fn from_point_angle(point: Point, angle: Angle) -> Self {
-        let (a, b) = match angle.tan() {
-            None => (Real::from(1.0), Real::from(0.0)),
-            Some(a) => (-a, Real::from(-1.0)),
+        // FIXE: angle.tan() for 180.0 degrees isn't exactly 0 which causes problems when drawing
+        //       a single quadrant. Is there a better solution to fix this?
```
jamwaffles, line 24:
```suggestion
        // FIXME: angle.tan() for 180.0 degrees isn't exactly 0 which causes problems when drawing
        //        a single quadrant. Is there a better solution to fix this?
```

A stupid nit, but makes for more accurate grepping

jamwaffles (changes requested):
I can't speak to the improvements on the algorithm, but the extra tests are a very welcome addition.

jamwaffles (approved):

jamwaffles:
Please let me know if you decide to pick up #405, otherwise I'll take another look at it.

rfuest:
> Please let me know if you decide to pick up #405, otherwise I'll take another look at it.

I haven't planed to, so give it a go if you like.

PR #485, opened 2020-11-17
Simplify self-intersection checks

jamwaffles:
This is a somewhat noisy PR but the core of it revolves around e04c3fcbeaf2072dbb1ef82f8af8e6ba75fa8dae. Performance doesn't meaningfully change against master which is a shame, but the self-intersection check is a lot simpler now.

The other changes just move some methods out of `Line` as they're only used by `LineJoin`.

The performance difference isn't that surprising. It turns out that a significant amount of time is spent in `Line::extents`, so optimising the code there, and the iterators it calls, would make a good difference I think. @rfuest you mentioned long ago that you hoped never to touch the Bresenham code again. Well... ;)

### DIFF
--- a/src/primitives/line/mod.rs
+++ b/src/primitives/line/mod.rs
@@ -66,13 +66,6 @@ impl Dimensions for Line {
     }
 }
 
-/// Check signs of two signed numbers
-///
-/// Fastest ASM output compared to other methods. See: https://godbolt.org/z/zVx9cD
-fn same_signs(a: i32, b: i32) -> bool {
-    (a >= 0) == (b >= 0)
-}
-
 /// Intersection test result.
 #[derive(Copy, Clone, Debug)]
 pub enum Intersection {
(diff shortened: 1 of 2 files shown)


rfuest:
You assigned this to me instead of requesting a review, does this mean you want me to add to this PR?

> @rfuest you mentioned long ago that you hoped never to touch the Bresenham code again. Well... ;)

I had a feeling that the happy Bresenham free time might not last forever :disappointed:

I would first like to merge some of the code that uses linear equation to reduce some duplication. The `Coefficients` struct in this PR, for example, could reuse the `LinearEquation` struct from the `arc`/`sector` modules with some tweaks.

jamwaffles:
Ah sorry, clicked too fast. Yes a review would be great, but I'm happy to wait for the linear equation stuff.

Something to note with the Bresenham stuff - we're probably going to have to change it to iterate outwards, then step, instead of multiple parallel lines, when we want to support alpha or antialiasing. Not something to tackle in any refactoring, but if a whole rewrite happens, it would be good to preempt alpha support.

rfuest (approved):

rfuest:
> Ah sorry, clicked too fast. Yes a review would be great, but I'm happy to wait for the linear equation stuff.

I did take a quick look at the changes and everything looks OK. I'll open another PR for the linear equation stuff.

> Something to note with the Bresenham stuff - we're probably going to have to change it to iterate outwards, then step, instead of multiple parallel lines, when we want to support alpha or antialiasing. Not something to tackle in any refactoring, but if a whole rewrite happens, it would be good to preempt alpha support.

Interesting point. IMO it would be worth to open an issue for this or add it as a comment to the existing antialiasing issue.

PR #487, opened 2020-11-17
Add private common module to primitives module

rfuest:

## PR description

This PR moves the files that are shared between different primitives into a `common` module. IMO that makes the code a little clearer than having a lot of directories and files inside the `primitives` module, but I think that is subjective. One more objective advantage of the `common` module is that it's harder to accidentally leak an internal type, because the module is private. And it reduces the number of places where the visibility must be set explicitly by using `pub(in ...)`.


src/primitives/common/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+    None,
+
+    /// Stroke is offset to the left of the line.
+    Left,
+
+    /// Stroke is offset to the right of the line.
+    Right,
+}
+
+impl From<StrokeAlignment> for StrokeOffset {
```
rfuest, line 29: I had to move this from `style::primitive_style` to here, because otherwise the `StrokeOffset` type would need to be at least `pub(crate)`.

src/primitives/arc/mod.rs
```diff
@@ -1,17 +1,14 @@
...
     geometry::{Angle, Dimensions, Point, Size},
     primitives::{Circle, Primitive, Rectangle},
     transform::Transform,
 };
-pub(in crate::primitives) use linear_equation::LinearEquation;
-pub(in crate::primitives) use plane_sector::{PlaneSector, PlaneSectorIterator};
+
+mod points;
+mod styled;
+
```
jamwaffles, line 11: This should be above all the `use`s for consistency IMO

jamwaffles (changes requested):
One minor nit.

The common folder is a good idea, especially as we start to reuse things across primitives. It's already a bit of a functionality salad, but it's definitely better to consolidate all the "outlier" stuff.

src/primitives/arc/mod.rs
rfuest: OK, makes sense.

rfuest:
> The common folder is a good idea, especially as we start to reuse things across primitives. It's already a bit of a functionality salad, but it's definitely better to consolidate all the "outlier" stuff.

I'm already working on the next PR that cleans up the code a bit more by merging some stuff and moving some types into the common module.

jamwaffles (approved):

PR #488, opened 2020-11-17
Some cleanup of primitives module

rfuest:

## PR description

This PR contains some cleanup of the primitives module. There will be additional thing that can be merged, but I didn't want to let the PR get even larger.

Summary of the changes:
* I've replaced `Coefficients` with `LinearEquation` which was previously only used by `Arc` and `Sector`.
  I'm sure there will be more code that can be migrated to the `LinearEquation` crate, but I just wanted to get this started in this PR. We might not even need the `Real` and `i32` variants of that type.
* I've removed the old `LineSide` enum and renamed the existing `Side` to `LineSide`.
 They were essentially the same enum and only used different names for the sides. I decided to use `LineSide` as the name, because IMO is a better name if it is shared across multiple primitives.
* I've replaced `Line::side` by `Line::check_side`.
  Instead of returning an integer which needs to be checked by the caller the function now takes a `LineSide` parameter. This way you don't need to remember which side is negative and which one is positive
* I've added a `Scanline` type, which is used to represent a horizontal line.
  This provides a place to put all the scanline related functions and it is a more efficient points iterator than `line::Points`.
* I've done some optimization of the `bresenham_scanline_intersection` method.
  The method now uses the fact that the y coordinate cannot change arbitrarily. After a matching y coordinate was found it only continues checking the iterator until the y coordinate changes.

`thick_polyline` is a bit faster now, but the other benches don't benefit much from the changes:
grafik

### DIFF
--- a/src/primitives/common/scanline.rs
+++ b/src/primitives/common/scanline.rs
@@ -1,8 +1,9 @@
 use core::ops::Range;
 
 use crate::{
-    geometry::{Dimensions, Point},
-    primitives::{ContainsPoint, Line, Primitive, Rectangle},
+    geometry::Point,
+    prelude::Primitive,
+    primitives::{Line, Rectangle},
 };
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
@@ -38,21 +39,27 @@ impl Scanline {
     /// Intersection lines produced by this function are sorted so that the start always lies to the
     /// left of the end.
     pub fn bresenham_intersection(&mut self, line: &Line) {
-        if !line
-            .bounding_box()
-            .contains(Point::new(line.start.x, self.y))
-        {
+        // Check if the scanline is in the y range of the line.
+        let y_range = if line.start.y <= line.end.y {
+            line.start.y..=line.end.y
+        } else {
+            line.end.y..=line.start.y
+        };
+
+        if !y_range.contains(&self.y) {
             return;
         }
 
         let y = self.y;
-        let mut points = line.points().filter(|p| p.y == y);
+        let mut points = line.points();
 
-        if let Some(first) = points.next() {
+        // Find the first point with the same y coordinate.
+        while let Some(first) = points.find(|p| p.y == y) {
             self.extend(first.x);
         }
 
-        if let Some(last) = points.last() {
+        // Check points until the y coordinate changes.
+        if let Some(last) = points.take_while(|p| p.y == y).last() {
             self.extend(last.x);
         }
     }
--- a/src/primitives/common/scanline.rs
+++ b/src/primitives/common/scanline.rs
@@ -6,6 +6,7 @@ use crate::{
     primitives::{Line, Rectangle},
 };
 
+/// Scanline.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Scanline {
     pub y: i32,
@@ -13,15 +14,18 @@ pub struct Scanline {
 }
 
 impl Scanline {
+    /// Creates a new empty scanline.
     pub const fn new(y: i32) -> Self {
         Self { y, x: 0..0 }
     }
 
+    /// Returns `true` if the x range of the scanline is empty.
     pub fn is_empty(&self) -> bool {
         // MSRV: use `Range::is_empty` on version >= 1.47.0
         !(self.x.start < self.x.end)
     }
 
+    /// Extends the scanline to include the given x coordinate.
     fn extend(&mut self, x: i32) {
         if self.is_empty() {
             self.x = x..x + 1;
@@ -114,7 +118,10 @@ impl Scanline {
         Rectangle::new(Point::new(self.x.start, self.y), Size::new(width, 1))
     }
 
-    /// Renamed to try_take to because of conflict with Iterator::take
+    /// Returns a clone of the scanline if it isn't empty.
+    ///
+    /// This method is used similar to `Option::take`, but was renamed to `try_take` because
+    /// of a naming conflict with `Iterator::take`.
     pub fn try_take(&mut self) -> Option<Self> {
         if !self.is_empty() {
             let ret = self.clone();
--- a/src/primitives/common/scanline.rs
+++ b/src/primitives/common/scanline.rs
@@ -1,0 +1,161 @@
+use core::ops::Range;
+
+use crate::{
+    geometry::{Dimensions, Point},
+    primitives::{ContainsPoint, Line, Primitive, Rectangle},
+};
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct Scanline {
+    pub y: i32,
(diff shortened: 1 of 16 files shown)


src/primitives/common/scanline.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        // Find the first point with the same y coordinate.
+        while let Some(first) = points.find(|p| p.y == y) {
+            self.extend(first.x);
+        }
+
+        // Check points until the y coordinate changes.
+        if let Some(last) = points.take_while(|p| p.y == y).last() {
+            self.extend(last.x);
+        }
```
jamwaffles, line 68: This could be shrunk down to:

```suggestion
        line.points()
            .skip_while(|p| p.y != y)
            .take_while(|p| p.y == y)
            .for_each(|p| {
                self.extend(p.x);
            });
```

I was hoping Rust would be able to optimise an iterator a bit better but there's no change on top of the benchmarks you already posted. At least it's more concise :)

src/primitives/triangle/scanline_intersections.rs
```diff
@@ -178,51 +177,55 @@ fn generate_lines(
...
-                        ))
-                    } else {
-                        None
+                    let start_x = first.x.end.min(second.x.end);
+                    let end_x = first.x.start.max(second.x.start);
+
+                    // TODO: check
+                    //  // Line needs to be shrunk by 1px off each end to prevent overdraw.
+                    //  // This can only happen if there's enough room to do so.
+                    //  if (end_x - start_x) > 1 {
```
jamwaffles, line 190: Does this still need checking?

Also, there are two tests in `triangle/styled.rs` that still overdraw: `bounding_box` and `styled_stroke_equals_lines`. Could you remove the other `.set_allow_overdraw(true)` calls in that file as part of this PR? Sorry I didn't pick it up in my thick triangles work. I've opened #489 to track the remaining bugs.

src/primitives/line/mod.rs
```diff
@@ -202,19 +191,59 @@ impl Line {
...
 
-    /// Get which side of a line a point lies on.
+    /// Checks if the point lies on the given side of the line.
     ///
-    /// If the result is zero, the point lies on the line. Otherwise, values greater than zero
-    /// denote that the point is on the left side of the line. Values less than zero denote the
-    /// point lies to the right.
-    pub(in crate::primitives) fn side(&self, point: Point) -> i32 {
+    /// Returns `true` if the point lies on the correct side or on the line.
+    pub(in crate::primitives) fn check_side(&self, point: Point, side: LineSide) -> bool {
```
jamwaffles, line 197: I think the entire body of this fn can be replaced with:

```rust
LinearEquation::from_line(&self).check_side(point, side)
```

Performance for the `thick polyline` bench went up to ~+12% compared to master with this change. `filled triangle` is +~3%.

jamwaffles (changes requested):
I just realised you haven't even asked for a review yet, so sorry if the comments I left are in your TODOs anyway!

That said, it already looks like a good cleanup! I like the `Scanline` struct idea - very neat :). The reuse of `LinearEquation` is great too.

src/primitives/common/scanline.rs
rfuest: You code has a slightly different behavior. My version did only call `extend` for the first and last pixel with the scanline y position and your version does call `extend` for every pixel with the same y coordinate. This would only make a performance difference for lines that are (almost) horizontal, so maybe it's better to just call `extend` for every pixel to be sure the range is correctly calculated in all cases. And your code looks much nicer.

jamwaffles: Ah, yes you're right. I think it's probably better to use your version in that case. There might be a performance cliff for nearly horizontal lines which would be nice to avoid.

src/primitives/triangle/scanline_intersections.rs
rfuest: > Does this still need checking?

I haven't really looked at the triangle code in detail and only wanted to keep it working while I was refactoring the other code. But I guess as long as no unit test fails it should be safe to remove this note.

> Also, there are two tests in `triangle/styled.rs` that still overdraw: `bounding_box`...

I can't remove it without the test failing and IMO this is outside the scope of this PR.

> ...and `styled_stroke_equals_lines`.

`styled_stroke_equals_lines` only uses `set_allow_overdraw` for the `lines_display` which must be enabled, because drawing single lines will always overdraw on the vertices. But I've removed the unnecessary `set_allow_out_of_bounds_drawing` in that test.

src/primitives/common/scanline.rs
rfuest: I just tested your version and it was a bit faster than the old one, nice.

jamwaffles: Did you test with nearly-horizontal long lines? I'd be interested to see if that's any quicker or slower.

src/primitives/line/mod.rs
rfuest: > I think the entire body of this fn can be replaced with:

It can't be replaced at the moment, because `from_line` is only implemented for `LinearEquation<i32>` and `check_side` is only for `LinearEquation<Real>`. But this is something I want to merge in the future.

jamwaffles: Ah, sorry in my local testing I also changed `LinearEquation<i32>` to:

```rust
impl LinearEquation<i32> {
    pub const fn from_line(line: &Line) -> Self {
        Self {
            a: line.end.y - line.start.y,
            b: line.start.x - line.end.x,
            c: line.end.x * line.start.y - line.start.x * line.end.y,
        }
    }

    pub fn check_side(&self, point: Point, side: LineSide) -> bool {
        let t = self.a * point.x + self.b * point.y + self.c;

        match side {
            LineSide::Left => t >= 0,
            LineSide::Right => t <= 0,
        }
    }
}
```

which IIRC is the only change I needed to make. Wouldn't it make sense to change that in this PR?

rfuest: I could add that change, but I would like to get rid of `LinearEquation<Real>` entirely, which is outside the scope of this PR. So I would prefer to do this in a future PR.

jamwaffles: Sure thing. Thanks for entertaining my bikeshedding :)

src/primitives/common/scanline.rs
rfuest: No, I haven't, but that would indeed be interesting. I don't have time at the moment, so if you want to give it a shot go ahead, otherwise I'll test it later.

jamwaffles: I ran a quick scrappy benchmark locally and the iterator approach doesn't seem to change performance. I see a ~+2% speedup in `thick polyline` but that's so small it's not really worth mentioning.

jamwaffles (approved):
Looks good!

PR #490, opened 2020-11-18
Triangle performance

rfuest:

## PR description

I've found an easy optimization for the triangle drawing performance. `is_collapsed` was unnecessarily calculated once per scanline. Calculating it only once in the constructor gives a significant performance improvement:
grafik


jamwaffles (approved):
Good catch! From the original thick triangles PR (first benchmark):



we've now gone down to this result when comparing this PR to 704694b:

image

Which isn't actually too bad... kind of.

jamwaffles:
I'm appreciating these smaller PRs btw, thanks for splitting the changes up like this. I'll attempt to do the same in the future.

rfuest:
Thanks for the quick review!

PR #493, opened 2020-11-20
Add improved MockDisplay assertions

rfuest:

## PR description

This PR adds `assert_eq` and `assert_pattern` methods to `MockDisplay`. They make tests a littler shorter, because you can now directly use `display.assert_pattern(&[ "RGB", ... ])` instead of having to create an expected display first.

But the main reason for implementing these function was to improve the output of failing tests. This wouldn't have been be possible by just updating the `Debug` impl for `MockDisplay`. The new output is optional and requires an terminal that supports 24 bpp colors. It can be enabled by setting the an environment variable at compile time: `EG_FANCY_PANIC=1` (any naming suggestions?). The default output remains unchanged to make sure there are no CI problems or if you need to see the pattern to update a pattern in the code.

Here is a comparison of the old and new output of a failing test:

Without `EG_FANCY_PANIC`:
before

With `EG_FANCY_PANIC`:
after
The diff colors are explained in the docs.

I haven't updated the tests to use this new feature yet, because this would cause a lot of noise and I wanted to keep the PR short. And we might also want to wait to update all test until we update to at least Rust 1.46.0, because otherwise we cannot use the `track_caller` attribute and loose the ability to see the location of a failing test without using a backtrace.

### DIFF
--- a/src/mock_display.rs
+++ b/src/mock_display.rs
@@ -175,14 +175,14 @@ use crate::{
     geometry::{Dimensions, OriginDimensions, Point, Size},
     pixelcolor::{
         Bgr555, Bgr565, Bgr888, BinaryColor, Gray2, Gray4, Gray8, GrayColor, PixelColor, Rgb555,
-        Rgb565, Rgb888, RgbColor,
+        Rgb565, Rgb888, RgbColor, WebColors,
     },
     prelude::Primitive,
     primitives::{ContainsPoint, Rectangle},
 };
 use core::{
     cmp::PartialEq,
-    fmt::{self, Write},
+    fmt::{self, Display, Write},
     iter,
 };
 
@@ -260,6 +260,14 @@ where
         } else {
             Rectangle::new(Point::zero(), Size::zero())
         }
+    }
+
+    /// Returns the `affected_area` with the top left corner extended to `(0, 0)`.
+    fn affected_area_origin(&self) -> Rectangle {
+        self.affected_area()
+            .bottom_right()
+            .map(|bottom_right| Rectangle::with_corners(Point::zero(), bottom_right))
+            .unwrap_or_default()
     }
 
     /// Changes the color of a pixel.
@@ -373,6 +381,36 @@ where
 
         target
     }
+
+    /// Compares the display to another display.
+    ///
+    /// The following color code is used to show the difference between the displays:
+    ///
+    /// | Color               | Description                                                   |
+    /// |---------------------|---------------------------------------------------------------|
+    /// | None                | The color of the pixel is equal in both displays.             |
+    /// | Some(Rgb888::GREEN) | The pixel was only set in `self`                              |
+    /// | Some(Rgb888::RED)   | The pixel was only set in `other`                             |
+    /// | Some(Rgb888::BLUE)  | The pixel was set to a different colors in `self` and `other` |
+    pub fn diff(&self, other: &MockDisplay<C>) -> MockDisplay<Rgb888> {
+        let mut display = MockDisplay::new();
+
+        for point in display.bounding_box().points() {
+            let self_color = self.get_pixel(point);
+            let other_color = other.get_pixel(point);
+
+            let diff_color = match (self_color, other_color) {
+                (Some(_), None) => Some(Rgb888::GREEN),
+                (None, Some(_)) => Some(Rgb888::RED),
+                (Some(s), Some(o)) if s != o => Some(Rgb888::BLUE),
+                _ => None,
+            };
+
+            display.set_pixel(point, diff_color);
+        }
+
+        display
+    }
 }
 
 impl MockDisplay<BinaryColor> {
@@ -482,6 +520,38 @@ where
 
         display
     }
+
+    /// Checks if the displays are equal.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the display aren't equal.
+    // MSRV: add track_caller attribute to get better error messages for rust >= 1.46.0
+    // #[track_caller]
+    pub fn assert_eq(&self, other: &MockDisplay<C>) {
+        if option_env!("EG_FANCY_PANIC") != Some("1") {
+            assert_eq!(self, other);
+            return;
+        }
+
+        if self != other {
+            let fancy_panic = FancyPanic::new(self, other);
+            panic!("{}", fancy_panic);
+        }
+    }
+
+    /// Checks if the displays is equal to to the pattern.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the display content isn't equal to the pattern.
+    // MSRV: add track_caller attribute to get better error messages for rust >= 1.46.0
+    // #[track_caller]
+    pub fn assert_pattern(&self, pattern: &[&str]) {
+        let other = MockDisplay::<C>::from_pattern(pattern);
+
+        self.assert_eq(&other);
+    }
 }
 
 impl<C> Default for MockDisplay<C>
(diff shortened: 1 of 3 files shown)


src/mock_display.rs
```diff
@@ -563,12 +633,195 @@ where
...
+) -> fmt::Result
+where
+    C: PixelColor + ColorMapping,
+{
+    for y in bounding_box.rows() {
+        if y % 2 != 0 {
+            // Skip all odd y coordinates, because `write_row` outputs two pixel rows with
+            // a single row of characters.
+            continue;
+        }
```
jamwaffles, line 666: I think this has the same behaviour, no?

```suggestion
    for y in bounding_box.rows().step_by(2) {
```

src/mock_display.rs
```diff
@@ -563,12 +633,195 @@ where
...
+            Point::zero(),
+            bounding_box_display
+                .size
+                .component_max(bounding_box_expected.size),
+        );
+
+        f.write_char('\n')?;
+
+        // Output the 3 displays in columns if they are less than 30 pixels wide.
+        if bounding_box.size.width <= 30 {
```
jamwaffles, line 772: Nice! I was going to complain that the boxes might not fit in thinner terminals but here's this `if` statement just sitting there!

src/mock_display.rs
```diff
@@ -563,12 +633,195 @@ where
...
+            let column_width = (bounding_box.size.width as usize).max(10);
+
+            write_header(f, column_width)?;
+
+            for y in bounding_box.rows() {
+                if y % 2 != 0 {
+                    // Skip all odd y coordinates, because `write_row` outputs two pixel rows with
+                    // a single row of characters.
+                    continue;
+                }
```
jamwaffles, line 784: Same change here I think?

```suggestion
            for y in bounding_box.rows().step_by(2) {
```

src/mock_display.rs
```diff
@@ -482,6 +520,38 @@ where
...
+            return;
+        }
+
+        if self != other {
+            let fancy_panic = FancyPanic::new(self, other);
+            panic!("{}", fancy_panic);
+        }
+    }
+
+    /// Checks if the displays is equal to to the pattern.
```
jamwaffles, line 543:
```suggestion
    /// Checks if the display is equal to the given pattern.
```

src/mock_display.rs
```diff
@@ -482,6 +520,38 @@ where
 
         display
     }
+
+    /// Checks if the displays are equal.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the display aren't equal.
```
jamwaffles, line 528:
```suggestion
    /// Panics if the displays aren't equal.
```

src/mock_display.rs
```diff
@@ -373,6 +381,36 @@ where
...
+    /// Compares the display to another display.
+    ///
+    /// The following color code is used to show the difference between the displays:
+    ///
+    /// | Color               | Description                                                   |
+    /// |---------------------|---------------------------------------------------------------|
+    /// | None                | The color of the pixel is equal in both displays.             |
+    /// | Some(Rgb888::GREEN) | The pixel was only set in `self`                              |
+    /// | Some(Rgb888::RED)   | The pixel was only set in `other`                             |
+    /// | Some(Rgb888::BLUE)  | The pixel was set to a different colors in `self` and `other` |
```
jamwaffles, line 394: Do you think it would add too much noise to print these rules to the console with the diff table? Not as verbose as the table presented here, but a quick reminder might be useful. A (hopefully) single line sentence would be ideal.

jamwaffles (changes requested):
This is *very* cool! Would a better name for the env var be `EG_PANIC_COLORS` perhaps? Or maybe `EG_GRAPHICAL_PANIC`. Tbh I don't really mind `EG_FANCY_PANIC`.

We may also want to add a `just` command to run the tests graphically to make it easier to remember what to type. Or at least add a note about the env var to the readme.

`#[track_caller]` will be useful in the future for sure, but at least for now we still get told which test failed.

Finally, whilst I'm not too bothered by all the manual escape codes, did you consider using something like [termcolor](https://github.com/BurntSushi/termcolor) to make them a bit more maintainable?

rfuest:
> Tbh I don't really mind `EG_FANCY_PANIC`.

OK, let's keep the name.

> We may also want to add a `just` command to run the tests graphically to make it easier to remember what to type. Or at least add a note about the env var to the readme.

I'll add that in the PR that changes the test to actually use this function.

> `#[track_caller]` will be useful in the future for sure, but at least for now we still get told which test failed.

We do, but some tests contain more than one assertion.

> Finally, whilst I'm not too bothered by all the manual escape codes, did you consider using something like [termcolor](https://github.com/BurntSushi/termcolor) to make them a bit more maintainable?

I did consider this, but I don't think there is a `non_std` crate for color escape codes.

src/mock_display.rs
rfuest: Thanks for the suggestion, that's better. I was too lazy to look up a better solution when I wrote the code.

src/mock_display.rs
rfuest: That's a good idea, but I would like to do this as part of a future PR. The `mock_display.rs` file is getting long and I would like to move the code into a `mock_display` to clean it up a bit.

src/mock_display.rs
rfuest: The current value of 30 pixels will produce an output of up to 110 characters per line. Should we reduce the width further to make sure that it also fits in the default 80 columns?

jamwaffles: I think 110 is fine. Not many of us are developing e-g on a C64 any more (although your old Intel box is getting close ;))

jamwaffles:
> We do, but some tests contain more than one assertion.

This is true. If at some point you get annoyed enough to bump the version, I won't reject the PR ;)

> I did consider this, but I don't think there is a non_std crate for color escape codes.

Ah yes, it has to be no-std doesn't it

src/mock_display.rs
```diff
@@ -10,6 +10,25 @@
...
+//! display and the expected output will be shown:
+//!
+//! ```bash
+//! EG_FANCY_PANIC=1 cargo test
+//! ```
+//!
+//! Enabling the advanced test output requires a terminal that supports 24 BPP colors and a font
+//! that includes the upper half block character `'\u{2580}'`.
+//!
+//! The color code used to show the difference between the display and the expected output is show
```
jamwaffles, line 29:
```suggestion
//! The color code used to show the difference between the display and the expected output is shown
```

src/mock_display.rs
```diff
@@ -482,6 +532,48 @@ where
 
         display
     }
+
+    /// Checks if the displays are equal.
+    ///
+    /// An advanced output for failing tests can be enabled, see the [module-level documentation]
+    /// for more details.
```
jamwaffles, line 539: It can't hurt to mention this here as well IMO. One less link to click for the user, and this is the documentation that pops up when hovering over usages of `assert_eq` in VSCode, etc.

```suggestion
    /// An advanced output for failing tests can be enabled by setting the environment variable 
    /// `EG_FANCY_PANIC=1`. See the [module-level documentation] for more details.
```

jamwaffles (changes requested):
A couple more minor comment nits, otherwise good to go.

I didn't comment because it's not really necessary, but do you think it's a good idea to add a basic unit tests to check the diff output? It might be helpful catching regressions, but it's not particularly important.

src/mock_display.rs
rfuest: > Not many of us are developing e-g on a C64 any more...

That's why I suggested 80 and not 40 :wink:.

> ...(although your old Intel box is getting close ;))

True, but my ultrawide monitor isn't quite as old as the CPU and can easily display more than 400 characters per row in a relatively large font.

jamwaffles (approved):
One thing I think I might miss is being able to add assertion messages, but maybe that could be added later with a `display_assert_eq!()` macro or similar.

This PR looks good!

rfuest:
> I didn't comment because it's not really necessary, but do you think it's a good idea to add a basic unit tests to check the diff output?

`diff` already has a unit test. Do you mean the output of `FancyPanic::fmt`?

jamwaffles:
> Do you mean the output of FancyPanic::fmt?

Yeah

rfuest:
Some basic unit test wouldn't be bad, but with all the escape sequences in the output it would also be messy. And I haven't looked into testing a `fmt` method in a `no_std` environment. Maybe we can add something later.

jamwaffles:
SGTM

rfuest:
> One thing I think I might miss is being able to add assertion messages, but maybe that could be added later with a `display_assert_eq!()` macro or similar.

That's a good point and a macro is probably the best solution for this.

PR #494, opened 2020-11-21
Triangle array

### LINKED ISSUE #482: Change Triangle to store points in an array
See https://github.com/embedded-graphics/embedded-graphics/pull/475#discussion_r524545267

It would make the scanline iterators a bit cleaner, and remove an `unreachable!()`

jamwaffles:

## PR description

Closes #482.

Should the constructor still take 3 separate arguments? We do have the `Triangle::from_points` method to take arrays and slices, so that might be enough, although maybe it should be renamed?

Benchmark performance hasn't changed.

### DIFF
--- a/src/primitives/triangle/mod.rs
+++ b/src/primitives/triangle/mod.rs
@@ -181,30 +181,27 @@ impl Triangle {
 
     /// Create a new triangle with points sorted in a clockwise direction.
     pub(in crate::primitives::triangle) fn sorted_clockwise(&self) -> Self {
-        let [p1, p2, p3] = self.vertices;
-
-        let determinant = -p2.y * p3.x + p1.y * (p3.x - p2.x) + p1.x * (p2.y - p3.y) + p2.x * p3.y;
-
-        match determinant.cmp(&0) {
+        match self.area_doubled().cmp(&0) {
             // Triangle is wound CCW. Swap two points to make it CW.
-            Ordering::Less => Self::new(p2, p1, p3),
+            Ordering::Less => Self::new(self.vertices[1], self.vertices[0], self.vertices[2]),
             // Triangle is already CW, do nothing.
             Ordering::Greater => *self,
             // Triangle is colinear. Sort points so they lie sequentially along the line.
-            Ordering::Equal => {
-                let (p1, p2, p3) = sort_yx(p1, p2, p3);
-
-                Self::new(p1, p2, p3)
-            }
+            Ordering::Equal => self.sorted_yx(),
         }
     }
 
+    /// Sort the 3 vertices of the triangle in order of increasing Y value.
+    ///
+    /// If two points have the same Y value, the one with the lesser X value is put before.
     fn sorted_yx(&self) -> Self {
         let [p1, p2, p3] = self.vertices;
 
-        let (p1, p2, p3) = sort_yx(p1, p2, p3);
+        let (y1, y2) = sort_two_yx(p1, p2);
+        let (y1, y3) = sort_two_yx(p3, y1);
+        let (y2, y3) = sort_two_yx(y3, y2);
 
-        Self::new(p1, p2, p3)
+        Self::new(y1, y2, y3)
     }
 
     pub(in crate::primitives::triangle) fn scanline_intersection(
@@ -295,13 +292,9 @@ impl Transform for Triangle {
     /// );
     /// ```
     fn translate(&self, by: Point) -> Self {
-        Self {
-            vertices: [
-                self.vertices[0] + by,
-                self.vertices[1] + by,
-                self.vertices[2] + by,
-            ],
-        }
+        let mut triangle = *self;
+        triangle.translate_mut(by);
+        triangle
     }
 
     /// Translate the triangle from its current position to a new position by (x, y) pixels.
@@ -332,16 +325,6 @@ fn sort_two_yx(p1: Point, p2: Point) -> (Point, Point) {
     } else {
         (p2, p1)
     }
-}
-
-/// Sort 3 points in order of increasing Y value. If two points have the same Y value, the one with
-/// the lesser X value is put before.
-fn sort_yx(p1: Point, p2: Point, p3: Point) -> (Point, Point, Point) {
-    let (y1, y2) = sort_two_yx(p1, p2);
-    let (y1, y3) = sort_two_yx(p3, y1);
-    let (y2, y3) = sort_two_yx(y3, y2);
-
-    (y1, y2, y3)
 }
 
 #[cfg(test)]
--- a/src/primitives/triangle/mod.rs
+++ b/src/primitives/triangle/mod.rs
@@ -13,10 +13,7 @@ use crate::{
     },
     transform::Transform,
 };
-use core::{
-    borrow::Borrow,
-    cmp::{max, min, Ordering},
-};
+use core::cmp::{max, min, Ordering};
 pub use points::Points;
 pub use styled::StyledPixels;
 
@@ -47,7 +44,10 @@ pub use styled::StyledPixels;
 /// # Ok::<(), core::convert::Infallible>(())
 /// ```
 ///
-/// ## Create a triangle from an array of points
+/// ## Create a triangle from a slice
+///
+/// A triangle can be created from the first 3 items of a `&[Point]` slice. If the slice is less
+/// than 3 elements long, the [`from_slice`] method will panic.
 ///
 /// ```rust
 /// use embedded_graphics::{geometry::Point, primitives::Triangle};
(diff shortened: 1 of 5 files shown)


rfuest:
> Should the constructor still take 3 separate arguments?

IMO there is no big advantage in changing it to use an array. Keeping it as is would be one less breaking API change.

> We do have the `Triangle::from_points` method to take arrays and slices, so that might be enough,...

`Triangle::from_points` is kind of odd, because I think it's more generic than any other constructor. And I think that `AsRef` should be used instead of `Borrow`. I would prefer to remove the `Into<Point>` stuff and only accept `AsRef<[Point; 3]>`.

>  ...although maybe it should be renamed?

I think it should. `from_points` can be confusing because `Triangle::points` also exists, which does something unrelated. Should it be called `from_vertices` or maybe `with_vertices`?

jamwaffles:
> Triangle::from_points is kind of odd, because I think it's more generic than any other constructor. And I think that AsRef should be used instead of Borrow. I would prefer to remove the Into<Point> stuff and only accept AsRef<[Point; 3]>.

I think it needs to remain `Borrow` if we want to keep accepting both `&[]` and `[]` (which would be nice tbh) for [one reason](https://doc.rust-lang.org/std/convert/trait.AsRef.html):

> Unlike `AsRef`, `Borrow` has a blanket impl for any `T`, and can be used to accept either a reference or a value.

The other changes make sense, so I've removed the `Into<Point>` stuff and renamed it to `with_vertices`, which aligns with `Rectangle::with_corners`, etc.

rfuest:
> I think it needs to remain `Borrow` if we want to keep accepting both `&[]` and `[]` (which would be nice tbh) for [one reason](https://doc.rust-lang.org/std/convert/trait.AsRef.html):

OK, I didn't read the docs correctly. I'm just wondering how common this kind of constructor is. Would it be more familiar for users to have separate constructors for arrays and slices? A slice constructor might be more useful than one that only works with references to arrays with 3 elements. Or is there an easy way to convert `&[Point]` into `&[Point; 3]`?

jamwaffles:
We could remove the method for now and add a better version in later if someone approaches us with a more specific use case to guide the impl. IIRC it was originally added as a "nice to have", so there's no specific use case for it.

rfuest:
I would prefer to at least add a constructor that takes an array. IMO it's strange if the user can access the vertices using an array but cannot construct a new triangle using an array.

Being able to construct a triangle from a slice would be useful for constructing triangles from an iterator like `points.chunks_exact(3)`.

jamwaffles:
I could keep the current `from_vertices` to cater for the `[Point; 3]` and `&[Point; 3]` case, but I'm not quite sure how a `from_slice` would behave. It could take the first 3 elements from the slice, but what if the slice is shorter? Return `Err`? Return `None`?

rfuest:
Maybe we could implement `core::convert::TryFrom<&[Points]> for Triangle`?

jamwaffles:
We certainly could, although is it a better convention to have a `from_slice` method? I've seen that in a few places, but I could be wrong. Either way, it should return a `Result<Self, ()>`, right?

rfuest:
I'm not sure how common `from_slice` is. But if it returns a `Result` would it be better to call it `try_from_slice`?

Another option would be to panic in `from_slice` if the slice has the wrong size, which is the behavior some methods in the standard library use: https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice. Or we could provide both `from_slice` and `try_from_slice`. If the user can guarantee that the slice will always have the required size, e.g. in the `chunks_exact` example above, `from_slice` might be easier to use.

jamwaffles:
For now, how about we just go with a panicking `from_slice()`. It satisfies the `chunks_exact(3)`, `windows(3)`, etc, use case, and we can add `try_from_slice` at a later date if users want it.

rfuest:
Sounds good to me.

src/primitives/triangle/mod.rs
```diff
@@ -144,39 +138,37 @@ impl ContainsPoint for Triangle {
...
+        let y_max = max(max(p1.y, p2.y), p3.y);
 
         Rectangle::with_corners(Point::new(x_min, y_min), Point::new(x_max, y_max))
     }
 }
 
 impl Triangle {
-    /// Create a new triangle with a given style
+    /// Create a new triangle with the given vertices.
     pub const fn new(p1: Point, p2: Point, p3: Point) -> Self {
```
rfuest, line 152: This might be better if someone uses code completion to show the arguments of the function:
```suggestion
    pub const fn new(vertex1: Point, vertex2: Point, vertex3: Point) -> Self {
```
Don't worry I don't expect you to change all `p1`...`p3`s in the file :wink:.

src/primitives/triangle/mod.rs
```diff
@@ -144,39 +138,37 @@ impl ContainsPoint for Triangle {
...
     }
 
-    /// Creates a new triangle from an array of points.
+    /// Creates a new triangle from the first three elements of a [`Point`] slice.
+    ///
+    /// # Panics
     ///
-    /// This supports both [`Point`]s, as well as anything that implements `Into<Point>` like
-    /// `(i32, i32)`.
+    /// This method will panic if the given slice is less than 3 items long.
```
rfuest, line 162: Perhaps we should panic for all length other than 3. This would match the behavior of `copy_from_slice`.

rfuest (changes requested):

src/primitives/triangle/mod.rs
jamwaffles: How about `vN`, etc, or is that too ambiguous compared to `vertexN` do you think?

rfuest: Is there a reason why you want to keep the names short? I just prefer more descriptive names in the public API. But I'm also OK with using `vN` if you prefer that.

jamwaffles: No that's a good point actually. I wanted to keep them consistent with `pN` used internally, but didn't consider these args are part of the public API. I've changed them to `vertexN`.

src/primitives/triangle/mod.rs
```diff
@@ -330,14 +313,15 @@ impl Transform for Triangle {
...
+    ///     Triangle::new(Point::new(15, 20), Point::new(25, 30), Point::new(20, 25))
+    /// )
     /// ```
     fn translate_mut(&mut self, by: Point) -> &mut Self {
-        self.p1 += by;
-        self.p2 += by;
-        self.p3 += by;
+        self.vertices[0] += by;
+        self.vertices[1] += by;
+        self.vertices[2] += by;
```
rfuest, line 324:
```suggestion
        self.vertices.iter_mut().for_each(|v| *v += by);
```

src/primitives/triangle/mod.rs
```diff
@@ -185,33 +175,35 @@ impl Triangle {
...
 
         -p2.y * p3.x + p1.y * (p3.x - p2.x) + p1.x * (p2.y - p3.y) + p2.x * p3.y
     }
 
     /// Create a new triangle with points sorted in a clockwise direction.
     pub(in crate::primitives::triangle) fn sorted_clockwise(&self) -> Self {
-        let Self { p1, p2, p3 } = self;
+        let [p1, p2, p3] = self.vertices;
 
         let determinant = -p2.y * p3.x + p1.y * (p3.x - p2.x) + p1.x * (p2.y - p3.y) + p2.x * p3.y;
```
rfuest, line 187:
```suggestion
        let determinant = self.area_doubled();
```

src/primitives/triangle/mod.rs
```diff
@@ -185,33 +175,35 @@ impl Triangle {
...
-            Ordering::Less => Self::new(*p2, *p1, *p3),
+            Ordering::Less => Self::new(p2, p1, p3),
             // Triangle is already CW, do nothing.
             Ordering::Greater => *self,
             // Triangle is colinear. Sort points so they lie sequentially along the line.
             Ordering::Equal => {
-                let (p1, p2, p3) = sort_yx(*p1, *p2, *p3);
+                let (p1, p2, p3) = sort_yx(p1, p2, p3);
 
                 Self::new(p1, p2, p3)
```
rfuest, line 198:
```suggestion
                self.sorted_yx(p1, p2, p3)
```

After this change `sort_yx` isn't required anymore and you can move the code in `sort_yx` into `sorted_yx`.

src/primitives/triangle/mod.rs
```diff
@@ -185,33 +175,35 @@ impl Triangle {
...
     pub(in crate::primitives::triangle) fn sorted_clockwise(&self) -> Self {
-        let Self { p1, p2, p3 } = self;
+        let [p1, p2, p3] = self.vertices;
 
         let determinant = -p2.y * p3.x + p1.y * (p3.x - p2.x) + p1.x * (p2.y - p3.y) + p2.x * p3.y;
 
         match determinant.cmp(&0) {
             // Triangle is wound CCW. Swap two points to make it CW.
-            Ordering::Less => Self::new(*p2, *p1, *p3),
+            Ordering::Less => Self::new(p2, p1, p3),
```
rfuest, line 191: After the other changes are applied this is the only place that still uses `p1`...`p3`, so I would suggest to simply change it to this:
```suggestion
            Ordering::Less => Self::new(self.vertices[1], self.vertices[0], self.vertices[2]),
```
Or move the `let [p1, p2, p3] = self.vertices;` into the `Less` match arm, if you prefer that.

src/primitives/triangle/mod.rs
```diff
@@ -144,39 +135,38 @@ impl ContainsPoint for Triangle {
...
-        I: Into<Point> + Copy,
-        P: Borrow<[I; 3]>,
-    {
-        let points = points.borrow();
-
-        Triangle {
-            p1: points[0].into(),
-            p2: points[1].into(),
-            p3: points[2].into(),
+    // MSRV: Consider subslice patterns for rust >= 1.42.0
```
rfuest, line 165: We don't need subslice patterns anymore, now that we require exactly 3 elements.
```suggestion
```

rfuest (changes requested):

rfuest (approved):
LGTM

jamwaffles:
Thanks!

PR #495, opened 2020-11-21
Fix drawing of translated polylines

### LINKED ISSUE #473: Thick polylines are drawn without translation
Polylines with a stroke width > 1px don't use the `Polyline::translate` field and are always drawn without an offset.

Should we fix this issue or remove the `translate` field entirely, because polylines can also be translated by using a `TranslatedDrawTarget`? This would be a cleaner implementation IMO, but we would loose the ability to get the bounding box of a translated styled polyline.

jamwaffles:

## PR description

Closes #473


src/primitives/polyline/styled.rs
```diff
@@ -10,14 +10,50 @@ use crate::{
...
     style::{PrimitiveStyle, Styled},
+    transform::Transform,
 };
 
+impl<'a, C> Styled<Polyline<'a>, PrimitiveStyle<C>>
+where
+    C: PixelColor,
+{
+    /// Compute the bounding box of the non-translated polyline.
+    pub(in crate::primitives::polyline) fn original_bounding_box(&self) -> Rectangle {
```
rfuest, line 21: This could use a more descriptive function name IMO:
```suggestion
    pub(in crate::primitives::polyline) fn untranslated_bounding_box(&self) -> Rectangle {
```

src/primitives/polyline/styled.rs
```diff
@@ -119,6 +158,8 @@ where
                         .map(|point| Pixel(point, stroke_color)),
                 ),
                 _ => {
+                    let mut display = display.translated(self.primitive.translate);
```
rfuest, line 161: There might be a little performance improvement if this would only use a translated draw target if `translate != Point::zero()`.

rfuest (changes requested):

src/primitives/polyline/styled.rs
jamwaffles: Good shout, however the only way I could get this to work was by copying and pasting most of the block:

```rust
if self.primitive.translate != Point::zero() {
    let mut display = display.translated(self.primitive.translate);

    for line in ScanlineIterator::new(self) {
        let rect = line.to_rectangle();

        if !rect.is_zero_sized() {
            display.fill_solid(&rect, stroke_color)?;
        }
    }
} else {
    for line in ScanlineIterator::new(self) {
        let rect = line.to_rectangle();

        if !rect.is_zero_sized() {
            display.fill_solid(&rect, stroke_color)?;
        }
    }
}
```

Any ideas on deduping this?

rfuest: Move the duplicate part into another method that is generic over any `DrawTarget`. You will probably need to add a new `impl<C> Styled<Polyline<'_>, PrimitiveStyle<C>>`  block, if it doesn't already exist.

jamwaffles: Was hoping there's a better way but ah well :). I've deduped it now.

src/primitives/polyline/styled.rs
```diff
@@ -45,6 +45,21 @@ where
             Rectangle::new(self.primitive.bounding_box().center(), Size::zero())
         }
     }
+
+    fn draw_thick<D>(&self, stroke_color: C, display: &mut D) -> Result<(), D::Error>
```
rfuest, line 49: This is more of general pedantic nit and not specific to this PR:
Can we decide on one name for the `DrawTarget` arguments? `target` and `display` are used in very similar quantities. My vote would be `target`, because it's more generic, but I don't really care as long as it's consistent.

src/primitives/polyline/styled.rs
```diff
@@ -457,6 +558,44 @@ mod tests {
         );
     }
 
+    #[test]
+    fn translated_bounding_box() {
+        let by = Point::new(10, 12);
+        let pl = Polyline::new(&PATTERN).translate(by);
+
+        assert_eq!(
+            pl.into_styled::<Rgb565>(PrimitiveStyle::with_stroke(Rgb565::RED, 1))
```
rfuest, line 567: Is the additional `Rgb565` required?
```suggestion
            pl.into_styled(PrimitiveStyle::with_stroke(Rgb565::RED, 1))
```
This question also applies to the other assertions.

src/primitives/polyline/styled.rs
```diff
@@ -246,6 +299,54 @@ mod tests {
...
+                "  #            ######       ",
+                "                 ##         ",
+            ])
+        );
+    }
+
+    #[test]
+    fn thick_stroke_points() {
+        let mut d1: MockDisplay<BinaryColor> = MockDisplay::new();
+        let mut d2: MockDisplay<BinaryColor> = MockDisplay::new();
```
rfuest, line 337: Another general question about coding style:

I've recently started to use this type code more often:
```suggestion
        let mut d1 = MockDisplay::<BinaryColor>::new();
        let mut d2 = MockDisplay::<BinaryColor>::new();
```
Should we use this in new code or do you prefer the other version?

src/primitives/polyline/styled.rs
```diff
@@ -246,6 +299,54 @@ mod tests {
...
+            .into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 4));
+
+        assert_eq!(
+            styled.bounding_box(),
+            Rectangle::new(Point::zero(), Size::new(28, 10))
+        );
+
+        styled.draw(&mut display).unwrap();
+
+        assert_eq!(
```
rfuest, line 317: Should we start to migrate to `display.assert_pattern(&[...])`?

rfuest (commented):
Here are some more comments, that are about our coding style in general and not specifically this PR. The PR looks good otherwise.

src/primitives/polyline/styled.rs
jamwaffles: `target` makes most sense to me too. I'll try and adhere to that name from now on - consistency is indeed important.

src/primitives/polyline/styled.rs
jamwaffles: I can do! You mentioned you wanted to change it all in another PR so I decided to leave these as is to help with conflicts. I will change only the new tests in this PR to use it though - that should be relatively conflict free.

src/primitives/polyline/styled.rs
jamwaffles: I'll change it to the new one, but I don't really mind either way :)

jamwaffles: Heh, turns out we don't need to specify the types in these tests at all

src/primitives/polyline/styled.rs
rfuest: I haven't started migrating the tests yet, so there is no danger of causing merge conflicts. But I did only meant to use it in the new tests anyway.

src/primitives/polyline/styled.rs
```diff
@@ -246,6 +296,51 @@ mod tests {
...
+
+        let pl = Polyline::new(&PATTERN)
+            .translate(Point::new(2, 3))
+            .into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 4));
+
+        pl.draw(&mut d1).unwrap();
+
+        pl.into_pixels().draw(&mut d2).unwrap();
+
+        assert_eq!(d1, d2);
```
rfuest, line 341:
```suggestion
        d1.assert_eq(d2);
```

rfuest (changes requested):

src/primitives/polyline/styled.rs
jamwaffles: Alright alrightt! You write some cool code I gotta use it!

Joke ;)

rfuest (approved):

jamwaffles:
Jokes aside, I really appreciate the thorough reviews :)

PR #496, opened 2020-11-21
Remove TinyTGA from image examples

jamwaffles:

## PR description

Removes TinyTGA from ignored doc tests, unignores them, and replaces tinytga with some raw images.

### DIFF
--- a/src/image/image_drawable.rs
+++ b/src/image/image_drawable.rs
@@ -80,16 +80,11 @@ pub trait ImageDrawableExt: Sized {
     /// # use embedded_graphics::mock_display::MockDisplay as Display;
     /// let mut display: Display<Rgb565> = Display::default();
     ///
-    /// let data = [
-    ///     0x00, 0x00, 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, //
-    ///     0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F, 0xFF, 0xFF, //
-    ///     // .
-    ///     // .
-    ///     // .
-    ///     0xFF, 0xFF, 0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F,
-    /// ];
+    /// let data = [ 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, /* ... */ ];
+    /// // or: let data = include_bytes!("sprite_atlas.raw");
     ///
-    /// let sprite_atlas: ImageRawBE<Rgb565> = ImageRaw::new(&data, 4, 3);
+    /// # let data = [0u8; 64 * 32 * 2];
+    /// let sprite_atlas: ImageRawBE<Rgb565> = ImageRaw::new(&data, 64, 32);
     ///
     /// let sprite_1 = sprite_atlas.sub_image(&Rectangle::new(Point::new(0, 0), Size::new(32, 32)));
     /// let sprite_2 = sprite_atlas.sub_image(&Rectangle::new(Point::new(32, 0), Size::new(32, 32)));
--- a/src/image/image_drawable.rs
+++ b/src/image/image_drawable.rs
@@ -67,20 +67,29 @@ pub trait ImageDrawableExt: Sized {
     ///
     /// # Examples
     ///
-    /// This example loads an image containing multiple 32x32px sprites and draws two of them to a
-    /// display, with their top-left corners positioned at `(100, 100)` and `(100, 140)`.
+    /// This example loads a raw image containing multiple 32x32px sprites and draws two of them to
+    /// a display, with their top-left corners positioned at `(100, 100)` and `(100, 140)`.
     ///
-    /// ```rust,ignore
-    /// use embedded_graphics::{image::Image, pixelcolor::Rgb888, prelude::*, primitives::Rectangle};
+    /// ```rust
+    /// use embedded_graphics::{
+    ///     image::{Image, ImageRaw, ImageRawBE},
+    ///     pixelcolor::Rgb565,
+    ///     prelude::*,
+    ///     primitives::Rectangle,
+    /// };
     /// # use embedded_graphics::mock_display::MockDisplay as Display;
-    /// use tinytga::Tga;
+    /// let mut display: Display<Rgb565> = Display::default();
     ///
-    /// let mut display: Display<Rgb888> = Display::default();
+    /// let data = [
+    ///     0x00, 0x00, 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, //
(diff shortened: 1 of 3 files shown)


src/image/image_drawable.rs
```diff
@@ -67,20 +67,29 @@ pub trait ImageDrawableExt: Sized {
...
+    /// let data = [
+    ///     0x00, 0x00, 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, //
+    ///     0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F, 0xFF, 0xFF, //
+    ///     // .
+    ///     // .
+    ///     // .
+    ///     0xFF, 0xFF, 0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F,
+    /// ];
+    ///
+    /// let sprite_atlas: ImageRawBE<Rgb565> = ImageRaw::new(&data, 4, 3);
```
rfuest, line 92: The description of the example says that the sprite atlas contains multiple sprites, but the `ImageRaw` is only 4 by 3 pixels.

```suggestion
    /// let data = [ 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, /* ... */ ];
    /// // or: let data = include_bytes!("sprite_atlas.raw");
    ///
    /// # let data = [0u8; 64 * 32 * 2];
    /// let sprite_atlas: ImageRawBE<Rgb565> = ImageRaw::new(&data, 64, 32);
```

src/image/mod.rs
```diff
@@ -11,31 +11,38 @@
...
+//! ## Load an RGB565 raw data imge and display it
 //!
-//! This example loads a TGA-formatted image using the [tinytga] crate and draws it to the display
-//! using an [`Image`] object.
+//! This example loads a small image from a raw data array and displays it. The image is RGB565
+//! encoded, so uses the `Rgb565` color type.
 //!
 //! The `graphics` feature of `tinytga` needs to be enabled in `Cargo.toml` to use the `Tga` object
 //! with embedded-graphics.
 //!
```
rfuest, line 21:
```suggestion
```

src/image/mod.rs
```diff
@@ -11,31 +11,38 @@
...
 //! # Examples
 //!
-//! ## Load a TGA image and draw it to a display
+//! ## Load an RGB565 raw data imge and display it
 //!
-//! This example loads a TGA-formatted image using the [tinytga] crate and draws it to the display
-//! using an [`Image`] object.
+//! This example loads a small image from a raw data array and displays it. The image is RGB565
+//! encoded, so uses the `Rgb565` color type.
```
rfuest, line 17:
```suggestion
//! encoded, so it uses the `Rgb565` color type.
```

src/image/mod.rs
```diff
@@ -11,31 +11,38 @@
...
+//! let mut display: Display<Rgb565> = Display::default();
 //!
-//! // Load the TGA file.
-//! // Note that the color type is set explicitly to match the format used in the TGA file,
+//! // Raw image data for demonstration purposes.
+//! let data = [
+//!     0x00, 0x00, 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, //
+//!     0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F, 0xFF, 0xFF, /*
+//!            * ... */
+//! ];
```
rfuest, line 37: The data array can't be any other size for the given image size. But we could add a comment that  real image data would likely be longer.
```suggestion
//! let data = [
//!     0x00, 0x00, 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, //
//!     0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F, 0xFF, 0xFF, //
//! ];
```

src/image/mod.rs
```diff
@@ -11,31 +11,38 @@
...
+//!
+//! // Load the image file.
```
rfuest, line 39: Is load the correct verb in this context? The data is already in memory and `ImageRaw` is just a wrapper around it.

src/image/mod.rs
```diff
@@ -11,31 +11,38 @@
...
+//! // Note that the color type is set explicitly to match the format used in the image file,
```
rfuest, line 40:
```suggestion
//! // Note that the color type must be set explicitly to match the format used in the image file,
```

src/image/mod.rs
```diff
@@ -50,20 +57,31 @@
...
+//!     // .
+//!     // .
+//!     // .
+//!     0xFF, 0xFF, 0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F,
+//! ];
 //!
-//! // Load the TGA file with the sprite atlas.
-//! // Note that the color type is set explicitly to match the format used in the TGA file,
+//! // Load the image file with the sprite atlas.
+//! // Note that the color type is set explicitly to match the format used in the image file,
```
rfuest, line 82: See comments above.

src/image/mod.rs
```diff
@@ -50,20 +57,31 @@
...
+//!
+//! // Raw image data for demonstration purposes.
+//! let data = [
+//!     0x00, 0x00, 0xF8, 0x00, 0x07, 0xE0, 0xFF, 0xE0, //
+//!     0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F, 0xFF, 0xFF, //
+//!     // .
+//!     // .
+//!     // .
+//!     0xFF, 0xFF, 0x00, 0x1F, 0x07, 0xFF, 0xF8, 0x1F,
+//! ];
```
rfuest, line 79: See comments above.

src/mock_display.rs
```diff
@@ -132,42 +132,35 @@
...
 //! image.draw(&mut display);
 //!
-//! display.assert_pattern(&[,
-//!     "WKRGBYMCW",
-//!     "KKRGBYMCW",
-//!     "WKRGBYMCW",
-//!     "KKKKKKKKK",
-//!     "WKWCMYBGR",
-//! ]);
+//! display.assert_pattern(&["KRGY", "BCMW"]);
```
rfuest, line 163:
```suggestion
//! display.assert_pattern(&[
    "KRGY", //
    "BCMW", //
]);
```

src/image/mod.rs
```diff
@@ -11,31 +11,38 @@
 //!
 //! # Examples
 //!
-//! ## Load a TGA image and draw it to a display
+//! ## Load an RGB565 raw data imge and display it
```
rfuest, line 14: I have only looked at the diff and didn't build the docs, but how often do we mention that other image formats can also be used via other crates? Maybe we should include this as a comment in this first example and point out that only the `raw` object is dependent on the image format and the code after that is common for all image formats.

rfuest (changes requested):

rfuest (approved):
LGTM

Issue #382, opened 2020-06-29
Add release checklist


rfuest:
GitHub still considers 0.2.0 to be the latest release of e-g. The new layout now shows this on the front page of the repository:
![grafik](https://user-images.githubusercontent.com/226123/86019274-aea82480-ba26-11ea-88ba-df88c9497f4e.png)
This could make a visitor think that no new versions were released since 2018.

jamwaffles:
That is _really_ annoying. This new Github interface is... mixed good/bad/horseshit.

I can draft a new release of e-g for 0.6.2 at that given commit which might fix things.

rfuest:
I think you could create a release for the tag at: https://github.com/jamwaffles/embedded-graphics/tags

jamwaffles:
Good tip, thanks. I've created a new release with a low-effort message to ask the user to look at `CHANGELOG.md`.

Looks like this now:

![image](https://user-images.githubusercontent.com/800791/86031159-5cb6cd00-ba2d-11ea-905c-d028a64b5624.png)

rfuest:
That's better.
But this is one more thing we need to remember when doing a release, maybe we should create a release checklist.

jamwaffles:
We should indeed. How would you suggest doing that? An issue template? A markdown document in the project root? Something in the readme?

rfuest:
I wouldn't clutter the readme with this, because it isn't relevant for the normal user. IMO the markdown document in the project root would be the easiest solution.

jamwaffles:
Sounds good to me. I've repurposed this issue to create a checklist in the root.

Items I can think of off the top of my head:

Pre-release:

* `cargo-release` must be used to release a crate in this repo.
* Is the changelog up to date, particularly with breaking changes?
* Is the version in `Cargo.toml` for the crate to be released the same as crates.io? (cargo-release automatically bumps to the next version)
* Double check which level (major, minor, patch) of release this is

Post-release

* Is the tag created by `cargo-release` pushed to Github?
* Is a Github release made for the latest version?

rfuest:
I haven't done any release on crates.io yet, so I cannot comment on the details, but I think it looks good.
Maybe the pre-release checks should also contain that you need to make sure the latest CI build was successful.

jamwaffles:
Yeah that definitely makes sense

rfuest:
Should we announce all new releases on the matrix channel? Then this should also be add to the post-release section.

[...]

PR #498, opened 2020-11-22
Split core traits

### LINKED ISSUE #460: Split core traits

jamwaffles:

## PR description

This is going to be a horrific PR to review, so sorry for that in advance.

Do we want to re-export `embedded-graphics-core` from e-g? I have a `pub use embedded_graphics_core as core;` at the moment which allows things like `use embedded-graphics::core::pixelcolor::Rgb565;` which isn't awful. Another thing I considered is `pub use embedded_graphics_core::{pixelcolor, geometry, drawable::*};` which would result in fewer changes in user code.

I chose to go with the `csv`-ish structure, with the core as a subfolder of the top level. I think this makes sense as there are fewer changes required.

Am I going in the right direction with this @rfuest? 

Closes #460 

TODO

* [x] Flesh out migration guide just a little bit, to reduce required effort later.
* [x] Add documentation to e-g lib.rs to point driver authors toward using e-g-core
* [x] Find other documentation about implementing e-g-core stuff and refactor
* [x] Add e-g-core lib.rs docs.


rfuest:
> Another thing I considered is pub use embedded_graphics_core::{pixelcolor, geometry, drawable::*}; which would result in fewer changes in user code.

Yes, this was my plan. IMO the separation into two crates shouldn't normally be visible for users of e-g and only matter for driver, text renderer or image format authors.

> I chose to go with the `csv`-ish structure, with the core as a subfolder of the top level. I think this makes sense as there are fewer changes required.

Sounds good to me.

jamwaffles:
Good to know I'm on the right path. The crate code will use imports from `embedded-graphics-core`, but should the tests and doc examples use _only_ embedded-graphics exports?

rfuest:
> The crate code will use imports from `embedded-graphics-core`, but should the tests...

I'm not sure about the tests. But if we only use `embedded-graphics` it would create a much less noisy PR.

> ... and doc examples use _only_ embedded-graphics exports?

Most doc examples should only use `embedded-graphics` exports. But examples that are meant to show how display drivers, image formats or text renderes are implemented would use `embedded-graphics-core` exports.

rfuest:
But there might be a problem when we add examples to the types in the core crate. When they are reexported into the main e-g docs, I would expect them to use `embedded-graphics` imports, but these wouldn't be available when the core crate docs are built.

jamwaffles:
I've probably missed a load of things, but not really sure what they are. A cursory look at the structure and stuff would be good. I think the code itself is pretty ordinary as nothing's really changed - just lots of moving around.

rfuest:
Looks good on the first cursory browse through the code!

Should we keep the structure identical between the crates? I think the only type that currently has a different location is `Rectangle`. Moving it into an, otherwise empty, primitives module wouldn't make much sense on it's own, but it would replicate the paths used in `embedded-graphics`. This might make migrating between `embedded-graphics` and `embedded-graphics-core` as easy as replacing all `use crate::embedded_graphics::`s with `use crate::embedded_graphics_core::`.

jamwaffles:
I'd prefer to keep it as a single file. Having a `primitives/` folder suggests more will be added in the future, which is unlikely. True, it is more effort during migration, but not a lot I'd argue.

jamwaffles:
I've done some moving around of the newly added methods from #500 in 5e31eb24ef2c03777c3c2f526e91b3edce2d46d8. This only leaves `center_offset` as a public method on `Size`. We could quite easily move this into a private function in `e-g` as I think it's utility outside of e-g is limited, but I'm interested to hear your thoughts @rfuest

jamwaffles:
Ah, I forgot to mention: e-g-core needs it too! I could make it `pub` and optionally add a `#[doc(hidden])`, or we could just copy and paste the functions between crates (preferably not)

rfuest:
We could move the code from `center_offset` into `Rectangle::with_center` and use `Rectangle::with_center(center, size).top_left` instead of `center_offset`.

jamwaffles:
Good shout, thanks. 

There are a few bits of docs that are still TODO. Is there anything in particular you'd like to see? I'll get some stuff together and then I think this is ready for review. I'll convert to draft for now - I requested review too early.

rfuest:
> Is there anything in particular you'd like to see?

No, nothing comes to my mind.

embedded-graphics-core/src/lib.rs
```diff
@@ -0,0 +1,84 @@
+//! Embedded Graphics Core
+//!
+//! DOC TODO
```
jamwaffles, line 3: @rfuest I could use a hand here. Bullet points are fine, but I'm drawing a blank at the moment.

jamwaffles: I had a cup of tea and the caffeine whispered sweet nothings into my brain. I've written some filler. Not sure if it's really in the right direction, but it's something we can work on.

embedded-graphics-core/src/rectangle/mod.rs
```diff
@@ -0,0 +1,760 @@
...
+}
+
+/// Checks if the two ranges overlap.
+fn overlaps(first: RangeInclusive<i32>, second: RangeInclusive<i32>) -> bool {
+    second.contains(first.start())
+        || second.contains(first.end())
+        || first.start() < second.start() && first.end() > second.end()
+}
+
+/// Anchor point.
```
rfuest, line 450: Should we move this type into the `geometry` module? It might be reused for other primitives in the future.

justfile
```diff
@@ -68,36 +68,52 @@ install-targets:
...
+    cargo deadlinks --dir target/doc/embedded_graphics
+    cargo deadlinks --dir target/doc/embedded_graphics_core
 
 #----------------------
 # README.md generation
 # ---------------------
 
 # Generate README.md for a single crate
-generate-readme: (_build-readme)
+generate-core-readme: (_build-readmes)
```
rfuest, line 83: `_build-readme` had an argument before the other crates were split from the e-g repo: https://github.com/embedded-graphics/embedded-graphics/blob/906ee1180cf31872b03aee037b9022c5fde30630/justfile#L111

tests/core-draw-target-cropped.rs
```diff
@@ -0,0 +1,138 @@
+use embedded_graphics::{
```
rfuest, line 1: Why did you move these tests to here?

embedded-graphics-core/src/geometry/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+pub use point::Point;
+pub use size::Size;
+
+use crate::rectangle::Rectangle;
+
+/// Adds the ability to get the dimensions/position of a graphics object
+///
+/// This **should** be implemented for all builtin embedded-graphics primitives and fonts. Third party
+/// implementations do not have to implement this trait as an object may not have a known size. If
+/// the object _does_ have a known size, this trait **should** be implemented.
```
rfuest, line 15: This needs to be updated. `Dimensions` is also used by `DrawTarget`, where this description doesn't make much sense.

IMO this is one of the places were we need to consider the different target audiences for the docs. One the one hand we have normal users who need to known how to use `Dimensions` to get the dimensions of drawable and draw targets. On the other hand we have authors of third party crates who need to know when and how the need to implement `Dimensions` and `OriginDimensions`. I think this is an important distinction and should also be considered for similar traits. I would suggest to put the user facing docs first and than add a common header like `# Implementing Dimensions` or `# Implementation notes` followed by the third party crate developer docs. I've started to follow this pattern in the docs for `ImageDrawable` (maybe not the greatest example).

rfuest (changes requested):
Here is the result of my first review pass.

.circleci/Dockerfile
```diff
@@ -4,6 +4,8 @@ RUN sudo apt update && sudo apt install -y libsdl2-dev libssl-dev && sudo apt cl
 
 RUN cargo install just --version ^0.6
 RUN cargo install cargo-readme --version ^3.2
+
+# MSRV: Upgrade to 0.6 when >= 1.45.0
```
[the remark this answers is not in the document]
jamwaffles: Yeah that would be good. I only added this comment because I saw the MSRV in the `deadlinks` readme. I did have a quick look at the `just` repo but couldn't find any versioning info there.

MIGRATING-0.6-0.7.md
```diff
@@ -97,6 +97,8 @@ Please search for `DrawTargetExt` on <https://docs.rs/embedded-graphics> for usa
 
 ## For display driver authors
 
+Driver authors should use `DrawTarget` exported by the [`embedded-graphics-core`](https://crates.io/crates/embedded-graphics-core) crate to integrate with embedded-graphics.
```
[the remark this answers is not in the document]
jamwaffles: Are the `For crates that handle images` and `For text rendering crates` sections enough or do you mean I should add more content to this `For display driver authors` section?

tests/core-draw-target-cropped.rs
jamwaffles: IIRC it was to keep the e-g dependency out of core. It should be possible to add it as a dev dependency and move these tests back into their respective files, but I didn't do that in case it causes more dep conflicts in the future. What would you suggest?

embedded-graphics-core/src/geometry/size.rs
```diff
@@ -153,7 +153,7 @@ impl Size {
     /// Saturating addition.
     ///
     /// Returns `u32::max_value()` for `width` and/or `height` instead of overflowing.
-    pub(crate) fn saturating_add(self, other: Self) -> Self {
+    pub fn saturating_add(self, other: Self) -> Self {
```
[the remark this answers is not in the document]
jamwaffles: I've added it to both

embedded-graphics-core/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+//! [`Dimensions`]:  ./geometry/trait.Dimensions.html
+//! [`OriginDimensions`]: ./geometry/trait.OriginDimensions.html
+//! [`prelude`]: ./prelude/index.html
+//! [`pixelcolor`]: ./pixelcolor/index.html
+//! [`BinaryColor`]: ./pixelcolor/enum.BinaryColor.html
+//! [`Rgb888`]: ./pixelcolor/struct.Rgb888.html
+//! [`ImageDrawable`]: ./image_drawable/trait.ImageDrawable.html
+
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/191fe7f8a0fedc713f9722b9dc59208dacadee7e/assets/logo.svg?sanitize=true"
```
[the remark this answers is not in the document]
jamwaffles: I'd like to create a core variant at some point, but we can do that later.

embedded-graphics-core/src/geometry/mod.rs
jamwaffles: I like the `implementation notes` idea, although I could use some suggestions for what goes above that heading, and below it.

embedded-graphics-core/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+        if self < 0x8000_0000 {
+            -(self as i32)
+        } else {
+            i32::min_value()
+        }
+    }
+}
+
+pub use drawable::{Drawable, Pixel};
+pub use rectangle::Rectangle;
```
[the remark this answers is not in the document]
jamwaffles: We need `Drawable` and `Pixel` to match e-g, but I've removed `Rectangle`.

embedded-graphics-core/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+pub mod drawable;
+pub mod geometry;
+pub mod image_drawable;
+pub mod iterator;
+pub mod pixelcolor;
+pub mod prelude;
+pub mod rectangle;
+
+/// Trait to convert unsigned into signed integer.
+#[doc(hidden)]
```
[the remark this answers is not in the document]
jamwaffles: I've copied and pasted it

jamwaffles:
My apologies if I've missed some comments - there were a lot, this PR is huge and it's Friday night :grimacing:

MIGRATING-0.6-0.7.md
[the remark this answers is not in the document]
rfuest: That's enough. I just wanted to make sure we remember to mention the core crate when we add these sections.

tests/core-draw-target-cropped.rs
rfuest: But doesn't the core crate already have a dev-dependency on e-g?

embedded-graphics-core/src/geometry/mod.rs
rfuest: Above:

* What objects the can this trait be used on.
* A description of which dimension is defined by this trait: For styled primitives it's the bounding box around all drawn pixels, but for draw target it's the drawable area.
* The difference between the dimensions of styled and unstyled primitives.
* Maybe some suggestions how the result can be used, like being able to use `.bounding_box().points()`.
* At some point we might also want to add examples to this, but that's even more out of scope of this PR than the rest of this suggestion :wink:.

Below:

* That some traits require this trait to be implemented, perhaps with a list of other traits.
* Your description of `OriginDimensions` isn't accurate: The `top_left` is also known and is guaranteed to be `Point::zero()`.
* Maybe a suggestion that the impl of `Dimensions` for drawables can be tested by using `MockDisplay::affeted_area`.

tests/core-draw-target-cropped.rs
jamwaffles: Yes it does :facepalm:. Not sure where I was going with the tests moved here, but I can move them back. With one strange quirk: all their dependencies must come from e-g to compile. I can't use a mix of `crate::` and `embedded_graphics::`. Unsure why, but they work now so :shrug:

rfuest: I think this is be caused by the cyclic dev-dependency. Cargo will have to break the cycle somewhere and this makes the types from `crate` and `embedded_graphics` incompatible. 

The cyclic dev-dependency will also cause a problem when we want to publish the crate, because we can't publish one create without the other already existing. But this can be fixed by removing the `version` from the dev-dependency in the core crate: See the note at https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

jamwaffles: Another TIL on Cargo's behaviour. Is it just me or is it really complex when you get into the weeds? Anyway, I've removed the version so now the dev dependency should hopefully not cause any problems.

jamwaffles:
When this PR is merged, I want to release core onto crates.io. Should we release as `1.0.0` or `0.1.0`? If we want to go for a 1.0.0 release, we could release `0.1.0` now to make sure everything works ok, then release 1.0.0 when e-g 0.7 is cut. That gives us time to make breaking changes before committing to a major version. 

We could also just release core as < 1.0.0 for e-g 0.7. Thoughts?

core/src/primitives/rectangle/points.rs
```diff
@@ -1,4 +1,6 @@
-use crate::{geometry::Point, primitives::rectangle::Rectangle};
+//! Iterator over all points inside the rectangle.
+
```
rfuest, line 2: IMO we don't need to add doc comments to private modules, because they will never be seen in to docs.
```suggestion
```

src/geometry/mod.rs
```diff
@@ -1,42 +1,73 @@
...
+    fn length_squared(self) -> i32;
 }
 
-/// Dimensions with `top_left` of the bounding box at `(0, 0)`.
-///
-/// A blanket implementation of `Dimensions` is provided for all types that implement this trait.
-pub trait OriginDimensions {
-    /// Returns the size of the bounding box.
-    fn size(&self) -> Size;
+impl PointExt for Point {
```
rfuest, line 24: The doc comments should be moves to `PointExt`, otherwise they don't show up when using code completion.

src/lib.rs
```diff
@@ -303,26 +297,3 @@ impl SaturatingCast<i32> for u32 {
         }
     }
 }
-
```
rfuest, line 306: If we duplicate the trait I would also duplicate the test.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#475](https://github.com/embedded-graphics/embedded-graphics/pull/475) `Triangle`s can now be drawn with stroke widths greater than 1.
 - [#478](https://github.com/embedded-graphics/embedded-graphics/pull/478) Added `resized`, `anchor_point`, `rows`, `columns` and `is_zero_sized` methods to `Rectangle`.
 - [#493](https://github.com/embedded-graphics/embedded-graphics/pull/493) Added `assert_eq`, `assert_pattern` and `diff` methods to `MockDisplay`. Improved error messages for failing assertions can be enabled by setting the `EG_FANCY_PANIC` environment variable to `1` at compile time.
+- [#498](https://github.com/embedded-graphics/embedded-graphics/pull/498) Added `Size::saturating_add` and `Size::saturating_sub`.
```
rfuest, line 20: There should also be an entry in the changelog(s) that `Primitive::points` no longer exists and was replaced by `PointsIter::points`.

rfuest (changes requested):
Here are some more comments

rfuest:
> We could also just release core as < 1.0.0 for e-g 0.7. Thoughts?

I'm not sure yet. I would suggest we decide this when it's time to release 0.7. Until then we can release core as 0.1.0, as you suggested.

core/src/primitives/rectangle/mod.rs
```diff
@@ -0,0 +1,748 @@
...
+                .top_left
+                .x
+                .saturating_add(self.size.width.saturating_cast())
+    }
+
+    /// Create a row-first, top-down iterator over all points in the rectangle.
+    pub fn points(&self) -> Points {
+        Points::new(self)
+    }
+
```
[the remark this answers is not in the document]
jamwaffles: It was indeed. Should we also move `PointIter` out of the `primitives` module and into `core/src/iterator/points.rs`?

jamwaffles:
> I'm not sure yet. I would suggest we decide this when it's time to release 0.7. Until then we can release core as 0.1.0, as you suggested.

SGTM.

rfuest (approved):
There are still some thing I'd like to tweak before the release, but I would suggest to merge this PR first and handle all further changes in smaller PRs that are easier to review.

Changes I would like to make or at least try out:
* Moving the `PointsIter` trait into the `iterator` module seems like a good idea, but I think the `iterator` module could use a cleanup anyway and perhaps some types could be moved out of the core crate.
* I would like to remove `RawDataIter` from `core`, because it's more likely to change than the other stuff in the `pixelcolor` module. Perhaps it can also be moved into `iterator`.
* We currently need to include the draw target adapters, like `Clipped` or `ColorConverted` in `core`, but IMO they don't really belong there. But I think there are several ways to fix this, which we can discuss later.

jamwaffles:
Thanks for the reviews! Those points make sense. Please make an issue if you think it's worth tracking.

With those changes in mind, should I hold off on a `0.1.0` release of core?

rfuest:
It might be a good idea to release `0.1.0` now, so we can start testing it with other crates. The suggested changes will be breaking, but as long as the core crate isn't widely adopted there is no harm in breaking the API a few times until the `0.7` release.

Do you also want to release a new e-g alpha that is compatible with core `0.1.0`?

jamwaffles:
Yep, I'll get both crates released. Hopefully it goes smoothly!

rfuest:
> Hopefully it goes smoothly!

:crossed_fingers:

jamwaffles:
core 0.1.0 is released: https://crates.io/crates/embedded-graphics-core
e-g 0.7.0-alpha.2 is released: https://crates.io/crates/embedded-graphics/0.7.0-alpha.2

`cargo-release` didn't update the changelogs which is annoying. I'll fix those now.

jamwaffles:
Everything is fixed. I force pushed to master as the release commits/tags generated by `cargo-release` were incorrect. Everything should be good now. I haven't gone through the checklist in #382 but I've created some GH releases, and I'll throw some announcements in the e-g Matrix chat.

Issue #484, opened 2020-11-16
Sector drawing improvements


rfuest:
Drawing of sectors has some issues that should be fixed at some point:
- There is no line join in the center:
  ![grafik](https://user-images.githubusercontent.com/226123/99317238-045e5b80-2866-11eb-809b-ac71905da5a7.png)
- The lines can overlap the outer edge of the sector, if the stroke width is large relative to the radius of the sector:
  ![grafik](https://user-images.githubusercontent.com/226123/99317390-48516080-2866-11eb-80fc-5f77c288c0ab.png)
- 180° degree sectors should produce a straight line, but they sometimes have a 1px offset:
   ![grafik](https://user-images.githubusercontent.com/226123/99317584-97979100-2866-11eb-9b10-a419cda1292b.png)
- The fill sometimes overlaps the stroke and the background also shows through:
   ![grafik](https://user-images.githubusercontent.com/226123/99318117-67042700-2867-11eb-8d18-f5f2deb622a8.png)
  diameter = 50
- Sectors with a sweep angle > 360° should be drawn as a simple circle IMO:
   ![grafik](https://user-images.githubusercontent.com/226123/99318295-b77b8480-2867-11eb-9bcb-dff270c107cd.png)
- The sector gets flipped around when `angle_start` is exactly 360°:
![grafik](https://user-images.githubusercontent.com/226123/99319667-3a054380-286a-11eb-9f85-799081cc4570.png)

jamwaffles:
Who'd've thought doing a graphics could be so difficult lol

> There is no line join in the center:

We're lucky I did all the work for thick polylines!

rfuest:
> Who'd've thought doing a graphics could be so difficult lol

This was supposed to be a short issue with a single screenshot for the line joins, but somehow that didn't work out as expected ;).

> We're lucky I did all the work for thick polylines!

That should make this much easier. :+1:

jamwaffles:
I was thinking of renaming the crate to `embedded-edgecases-and-rabbit-holes-with-maybe-some-pixels` so it's more accurate.

> That should make this much easier.

Some of it, hopefully :)

jamwaffles:
I've edited the original post to hide the completed issues. This makes it easier to see what's left at a glance.

Issue #491, opened 2020-11-18
Add additional benchmarks for primitives


rfuest:
The current benches don't cover a big enough sample of different styles that can be used to draw primitives IMO. `Circle`, for example, has only a single bench. Many other primitives are only tested with `stroke_width == 1`.

I would suggest to at least add benches for each filled shape that use thin stroke, thick stroke, thin stroke with fill, thick stroke + fill and only fill styles.

PR #500, opened 2020-11-22
 Merge LinearEquation variants 

rfuest:

## PR description

This PR merges `LinearEquation<Real>` and `LinearEquation<i32>` into one variant that uses `i32` values. I've also changed the way the coefficients are stored inside `LinearEquation` to make the formulas that use these values more readable. But this means that the PR turned out to be longer than expected.

I had to rewrite some code for sectors, so there might be some changes for #484, but I haven't checked yet.

### DIFF
--- a/src/geometry/angle.rs
+++ b/src/geometry/angle.rs
@@ -7,10 +7,10 @@ use float_cmp::{ApproxEq, F32Margin};
 #[allow(unused_imports)]
 use micromath::F32Ext;
 
-#[allow(dead_code)]
 pub(crate) mod angle_consts {
     use super::{real, Angle};
 
+    #[allow(dead_code)]
     pub(crate) const ANGLE_90DEG: Angle = Angle(real::FRAC_PI_2);
     pub(crate) const ANGLE_180DEG: Angle = Angle(real::PI);
     pub(crate) const ANGLE_360DEG: Angle = Angle(real::TAU);
--- a/src/geometry/angle.rs
+++ b/src/geometry/angle.rs
@@ -7,6 +7,7 @@ use float_cmp::{ApproxEq, F32Margin};
 #[allow(unused_imports)]
 use micromath::F32Ext;
 
+#[allow(dead_code)]
 pub(crate) mod angle_consts {
     use super::{real, Angle};
 
@@ -62,12 +63,6 @@ impl Angle {
     /// Normalize the angle to less than one full rotation (ie. in the range 0..360).
     pub fn normalize(self) -> Self {
         Angle(self.0.rem_euclid((2.0 * PI).into()))
-    }
-
-    /// Normalize the angle to less than one full rotation, starting from angle.
-    /// (ie. in the range angle..(angle+360)).
-    pub(crate) fn normalize_from(self, angle: Self) -> Self {
-        Angle((self.0 - angle.0).rem_euclid((2.0 * PI).into()) + angle.0)
     }
 
     /// Return numerical value of the angle in degree
--- a/src/geometry/point.rs
+++ b/src/geometry/point.rs
@@ -232,6 +232,27 @@ impl Point {
     pub fn component_div(self, other: Self) -> Self {
         Self::new(self.x / other.x, self.y / other.y)
     }
+
+    /// Returns a point that is rotated by 90° relative to the origin.
+    pub(crate) const fn rotate_90(self) -> Self {
+        Self::new(self.y, -self.x)
+    }
+
+    /// Calculates the dot product of two points.
+    pub(crate) const fn dot_product(self, other: Point) -> i32 {
+        self.x * other.x + self.y * other.y
+    }
+
+    /// Calculates the determinant of a 2x2 matrix formed by this and another point.
+    ///
+    /// ```text
+    ///          | self.x  self.y  |
+    /// result = |                 |
+    ///          | other.x other.y |
+    ///
+    pub(crate) const fn determinant(self, other: Point) -> i32 {
+        self.x * other.y - self.y * other.x
+    }
 }
 
 impl Add for Point {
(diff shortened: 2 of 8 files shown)


jamwaffles:
Awesome! Also, congrats on #500! :tada: 

> I had to rewrite some code for sectors, so there might be some changes for #484, but I haven't checked yet.

When you do check, could you either post the code to render the screenshots in that issue, or maybe add them as unit tests? It would be helpful for me if I pick any of the bugs up to try and fix.

rfuest:
> ...or maybe add them as unit tests?

I had started to write some unit tests for that issue when I got distracted by the fancy panic output. I'll open a PR when they are finished.

src/geometry/angle.rs
```diff
@@ -7,6 +7,7 @@ use float_cmp::{ApproxEq, F32Margin};
 #[allow(unused_imports)]
 use micromath::F32Ext;
 
+#[allow(dead_code)]
```
jamwaffles, line 10: Perhaps a bit too nitpicky, but maybe this would be better on the `const ANGLE_90DEG` as that's the only item that triggers a warning.

src/geometry/point.rs
```diff
@@ -232,6 +232,27 @@ impl Point {
...
+        self.x * other.x + self.y * other.y
+    }
+
+    /// Calculates the determinant of a 2x2 matrix formed by this and another point.
+    ///
+    /// ```text
+    ///          | self.x  self.y  |
+    /// result = |                 |
+    ///          | other.x other.y |
+    ///
```
jamwaffles, line 252: Not strictly necessary but good for consistency.

```suggestion
    /// ```
```

src/primitives/common/plane_sector.rs
```diff
@@ -14,65 +24,59 @@ use crate::{
...
-            side_a,
-            side_b,
-            sweep: angle_sweep.abs(),
+            half_plane_left: LinearEquation::from_point_angle(center_2x, angle_start),
+            half_plane_right: LinearEquation::from_point_angle(center_2x, angle_end),
+            operation,
         }
     }
 
+    /// TODO: This method doesn't really return an empty plane sector. Does this matter?
```
jamwaffles, line 60: > Does this matter?

Not really. From looking at its usage in `PlaneSectorIterator::empty`, the intent is clear from that context so I'd be happy to keep this as `empty`. If you do want to change the name, would `zero` make sense? `infinity`?

src/primitives/line/mod.rs
```diff
@@ -217,28 +199,42 @@ impl Line {
...
         let line2 = LinearEquation::from_line(other);
 
-        let denom = line1.a * line2.b - line2.a * line1.b;
+        // Calculate the determinant to solve the system of linear equations using Cramer's rule.
+        let denominator = line1.normal_vector.determinant(line2.normal_vector);
 
-        // Lines are colinear or parallel
-        if denom == 0 {
+        // The system of linear equations has no solutions if the determinant is zero and the lines
+        // have to be colinear.
```
jamwaffles, line 206:
```suggestion
        // The system of linear equations has no solutions if the determinant is zero. In this case, 
        // the lines must be colinear.
```

src/primitives/common/linear_equation.rs
```diff
@@ -3,62 +3,194 @@ use crate::{
...
     }
 
-    /// Create a horizontal line equation
+    /// Creates a horizontal line equation.
     pub fn new_horizontal() -> Self {
         LinearEquation {
-            a: Real::from(0.0),
-            b: Real::from(1.0),
-            c: Real::from(0.0),
+            normal_vector: Point::new(0, -16384),
```
jamwaffles, line 62: Would this make sense to be

```suggestion
            normal_vector: Point::new(0, i32::MIN),
```

I'm not sure what the significance of -16384 is specifically.

src/primitives/common/linear_equation.rs
```diff
@@ -3,62 +3,194 @@ use crate::{
...
+            origin_distance,
         }
     }
-}
 
-impl LinearEquation<Real> {
-    /// Create a new linear equation based on one point and one angle
+    /// Creates a new linear equation based on one point and one angle.
     pub fn from_point_angle(point: Point, angle: Angle) -> Self {
         // FIXME: angle.tan() for 180.0 degrees isn't exactly 0 which causes problems when drawing
```
jamwaffles, line 40: I assume this comment is still valid and will get fixed later, right?

jamwaffles (changes requested):
I'm not familiar with the maths here, which probably shows in my comments, but these names are _much_ clearer for one thing.

Unless my benchmarks are broken, it also looks like we're ~12% up on the `filled_triangle` benchmark which is nice.

I left some small nits, otherwise this looks good!

src/primitives/common/linear_equation.rs
rfuest: It is still valid and it might get fixed later.

src/primitives/common/linear_equation.rs
rfuest: I did forget to update this. It was supposed to be `-NORMAL_VECTOR_SCALE`, which makes the normal vector have the same length as normal vectors generated by `from_point_angle`.

src/primitives/common/plane_sector.rs
rfuest: I've removed the method and replaced its only use with a call to `PlaneSector::new`. `PlaneSectorIterator::empty` now uses a `PlaneSector` with `operation == Operation::EntirePlane`, which is the fastest to construct and doesn't matter because the points iter is also empty.

src/geometry/angle.rs
rfuest: I would never be so nitpicky... Oh wait, I would, never mind :wink: I've applied your suggestion.

jamwaffles: I do learn from the best to be fair. I noticed you had this attribute on the `const`s themselves in other places too, so I feel vindicated in making this comment :grin:

jamwaffles (approved):
Très bien, merci!

rfuest:
> Unless my benchmarks are broken, it also looks like we're ~12% up on the `filled_triangle` benchmark which is nice.

I couldn't verify that. My benchmark results for this branch are actually slightly worse then `master`.

jamwaffles:
> I couldn't verify that. My benchmark results for this branch are actually slightly worse then master.

Mine must be broken then. I'm using Criterion's `--baseline` feature which means my master baseline might be stale. No worries either way. We can do a more thorough perf pass later, starting with #491

PR #503, opened 2020-11-26
Improve sector drawing

rfuest:

## PR description

This PR fixes most of the issues in #484. Instead of using the existing thick line drawing code the sectors are entirely drawn by using linear equations.

Some notes:
* The drawing performance for 1px stroke sectors is slower, but faster for larger stroke widths.
* There is no more overdraw or gaps between fill and stroke.
  This would have been hard to achieve if the lines were drawn using the thick line drawing code.
* Miter and bevel line joins are drawn (with some limitations).

Problems that need to be fixed in later PRs:
* Line joins for sweep angles close to 360° are always drawn as a miter, but should be beveled.
* In `issue_484_stroke_should_not_overlap_outer_edge` both sides of the radial line stoke are clipped to the circle.
  The left side of the vertical line and the bottom side of the horizontal line should be rectangular. This will also make the bounding box calculation harder, because these corners can be outside the circle bounding box for some angles.
* The behavior of line joins in `Sector` and `Polyline` should be equal. The miter limit is probably different and sectors use this method of switching between miter and bevel: http://tavmjong.free.fr/SVG/MITER_LIMIT/index.html.

### DIFF
--- a/src/primitives/common/plane_sector.rs
+++ b/src/primitives/common/plane_sector.rs
@@ -1,9 +1,6 @@
 use crate::{
-    geometry::{angle_consts::*, Angle, Dimensions, Point},
-    primitives::{
-        common::{LineSide, OriginLinearEquation},
-        rectangle, Primitive, Rectangle,
-    },
+    geometry::{angle_consts::*, Angle, Point},
+    primitives::common::{LineSide, OriginLinearEquation, PointType},
 };
 
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
@@ -14,6 +11,17 @@ enum Operation {
     Union,
     /// Return the entire plane.
     EntirePlane,
+}
+
+impl Operation {
+    /// Executes the operation.
+    fn execute(self, first: bool, second: bool) -> bool {
+        match self {
+            Operation::Intersection => first && second,
+            Operation::Union => first || second,
+            Operation::EntirePlane => true,
+        }
+    }
 }
 
 /// Sector shaped part of a plane.
@@ -28,11 +36,15 @@ pub struct PlaneSector {
     ///
     /// The origin of the half planes is equal to the center point of the arc or circle scaled
     /// up by a factor of 2.
-    origin: Point,
+    // TODO: make private
+    pub(crate) origin: Point,
+
     /// Half plane on the left side of a line.
     half_plane_left: OriginLinearEquation,
+
     /// Half plane on the right side of a line.
     half_plane_right: OriginLinearEquation,
+
     /// The operation used to combine the two half planes.
     operation: Operation,
 }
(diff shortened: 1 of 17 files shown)


src/primitives/common/plane_sector.rs
```diff
@@ -24,10 +32,19 @@ enum Operation {
 /// half-planes.
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
 pub struct PlaneSector {
+    /// Origin of the two half planes.
+    ///
+    /// The origin of the half planes is equal to the center point of the arc or circle scaled
+    /// up by a factor of 2.
+    // TODO: make private
+    pub(crate) origin: Point,
```
jamwaffles, line 40: I think this still needs to be some sort of `pub`:

```suggestion
    pub(in crate::primitives) origin: Point,
```

src/primitives/sector/styled.rs
```diff
@@ -86,38 +110,42 @@ where
...
-                            None
-                        };
-
-                        if let Some(color) = color {
-                            return Some(Pixel(point, color));
+            let (point, distance) = self
+                .iter
+                .find(|(_, distance)| *distance < outer_threshold)?;
+
+            // TODO: only scale point once
```
jamwaffles, line 120: Should this be handled in this PR or in a future one? If it's purely for performance, should we start tracking these things with `// PERF: ...` comments, like the MSRV ones, perhaps?

src/primitives/common/plane_sector.rs
```diff
@@ -57,87 +75,56 @@ impl PlaneSector {
...
-            half_plane_right: LinearEquation::from_point_angle(center_2x, angle_end),
+            origin: center_2x,
+            half_plane_left: OriginLinearEquation::with_angle(angle_start),
+            half_plane_right: OriginLinearEquation::with_angle(angle_end),
             operation,
         }
     }
 
     pub fn contains(&self, point: Point) -> bool {
         // `PlaneSector` uses scaled coordinates for an increased resolution.
```
jamwaffles, line 86: Is it just a 2x scaling or is it more/less? It would help to define what the scale is (at least for me).

src/primitives/common/plane_sector.rs
```diff
@@ -57,87 +75,56 @@ impl PlaneSector {
...
-            plane_sector: PlaneSector::new(Point::zero(), Angle::zero(), ANGLE_360DEG),
-            points: Rectangle::zero().points(),
+    /// Checks if a point is inside the stroke or fill area.
+    pub fn point_type(
+        &self,
+        point: Point,
+        inside_threshold: i32,
+        outside_threshold: i32,
+    ) -> Option<PointType> {
+        // `PlaneSector` uses scaled coordinates for an increased resolution.
```
jamwaffles, line 102: Likewise here

src/primitives/sector/styled.rs
```diff
@@ -257,44 +275,93 @@ mod tests {
...
+    ) {
+        let style = PrimitiveStyleBuilder::new()
+            .stroke_color(BinaryColor::On)
+            .stroke_width(3)
+            .stroke_alignment(stroke_alignment)
+            .build();
 
-        let style = PrimitiveStyle::with_stroke(BinaryColor::On, 3);
+        let mut display = MockDisplay::new();
+        display.set_allow_out_of_bounds_drawing(true);
```
jamwaffles, line 290: This can be removed - the test passes when I remove it on my machine.

src/primitives/sector/styled.rs
```diff
@@ -179,73 +207,63 @@ mod tests {
...
-        Ok(())
+        with_stroke_width.assert_eq(&expected);
     }
 
     // Check the rendering of a simple sector
     #[test]
-    fn tiny_sector() -> Result<(), core::convert::Infallible> {
+    fn tiny_sector() {
         let mut display = MockDisplay::new();
         display.set_allow_overdraw(true);
```
jamwaffles, line 232:
```suggestion
```

jamwaffles (changes requested):
Looks good aside from a couple of small things

jamwaffles:
> Line joins for sweep angles close to 360° are always drawn as a miter, but should be beveled.

Does this result in the lines drawing out to infinity, or are they clipped by a bounding box?

> The behavior of line joins in Sector and Polyline should be equal. The miter limit is probably different and sectors use this method of switching between miter and bevel: http://tavmjong.free.fr/SVG/MITER_LIMIT/index.html.

That's a very neat solution. It would definitely be good to be able to reuse the line join code between polylines and sectors.

src/primitives/common/plane_sector.rs
rfuest: This will be changed in the next PR and the field might be removed completely.

src/primitives/sector/styled.rs
rfuest: This will also be included in the next PR, so it might not be worth changing it now. But I like the idea of using PERF for longer term comments.

src/primitives/common/plane_sector.rs
rfuest: The scaling will be moved out of this struct in the next PR. I'll try to remember to include some more comments about the scale in that PR.

rfuest:
> Does this result in the lines drawing out to infinity, or are they clipped by a bounding box?

The miter is clipped at the circular outline.

jamwaffles (approved):

PR #504, opened 2020-11-27
Improve sector drawing performance

rfuest:

## PR description

This PR adds an additional `delta` value to the `DistanceIterator` item so that the value doesn't need to be calculated multiple times. The performance difference is much larger than I expected, with ~30% or ~17% depending on the sweep angle, but perhaps my measurement is wrong.


jamwaffles (approved):
LGTM.

I see similar performance improvements, so either the benches are broken or we're just leaving that much perf on the table.

PR #508, opened 2020-11-30
Migrate tests to new MockDisplay assertions

### LINKED ISSUE #501: Migrate tests to new MockDisplay assertion functions
All tests, that compare a `MockDisplay` with an expected pattern, should use `MockDisplay::assert_pattern` or `MockDisplay::assert_eq`.

Some suggestion that came up in PR #493 should also be implemented:
* A basic unit test for the output https://github.com/embedded-graphics/embedded-graphics/pull/493#pullrequestreview-535978137
* Macros to allow assertions with messages https://github.com/embedded-graphics/embedded-graphics/pull/493#pullrequestreview-535980029
* A legend for the diff colors in the output https://github.com/embedded-graphics/embedded-graphics/pull/493#discussion_r528191078

rfuest:

## PR description

This PR changes all tests to use the new `MockDisplay` assertions.

Additional changes:
* Some cleanup of the code.
  The `mock_display.rs` file did get too long IMO, so I've moved the code into a module.
* I've added the suggested legend for the diff colors:
  grafik
* I decided not to include macros like `display_assert_eq`. They aren't used by most users and we would need to make them public to use them in our crates, which would add clutter to the root of the e-g crate docs.
 I've instead added `assert_eq_with_message` and `assert_pattern_with_message` methods. They need more boilerplate code to use them, but don't anything public outside the `MockDisplay` type.
  Or do you prefer the macros?

Closes #501


jamwaffles (approved):
The macros aren't necessary. I wanted a way to output a message with the assertion. Macros would be a marginally cleaner way to do that, but it's certainly good not to pollute the crate. The solution in this PR looks good!

PR #509, opened 2020-12-01
Multiline text

rfuest:

## PR description

This PR changes the way multiline text is handled. By moving the multiline code from `MonoTextStyle` to `Text` the multiline support can be shared by all text renderers.

I decided to remove `TextStylePixels` to make the refactor easier. We can add pixel iterators later, if we decide that they are necessary.

The removal of the pixel iterators made it possible to optimize `Drawable` and the benches are now 10%-30% faster. The drawing code now also uses `fill_contiguous` if both `text_color` and `background_color` are set, which could lead to even greater improvements in drawing performance for drivers without a framebuffer.


src/style/text_style.rs
```diff
@@ -12,20 +12,18 @@ pub trait TextStyle {
     /// Color type.
     type Color: PixelColor;
 
-    /// Render a text object using this style.
-    fn render_text<D>(&self, text: &Text<'_>, target: &mut D) -> Result<(), D::Error>
+    /// Renders a single line of text using this style.
+    ///
+    /// Returns the offset from the current position to the start of the next line.
```
jamwaffles, line 17: I think this doc could use some `# Implementation notes`;

* For the first line, what's the starting `position`?
* For subsequent lines, isn't the position just `position + Point::new(0, <line_y_height>)`? It would be good to clarify this behaviour.
* Does the `position` within a line matter? E.g. should it be the top left corner of the line, the baseline, etc?

Edit: After writing the other comment, should this be renamed to `render_first_line` (or something consistent with the other bounding box method name)?

src/style/text_style.rs
```diff
@@ -12,20 +12,18 @@ pub trait TextStyle {
...
-
-/// Pixels iterator.
-pub trait TextStylePixels<'a>: TextStyle {
-    /// Iterator type.
-    type Iter: Iterator<Item = Pixel<Self::Color>> + 'a;
-
-    /// Returns an iterator over the drawn pixels.
-    fn pixels(&self, text: &Text<'a>) -> Self::Iter;
+    /// Returns the bounding box of a single line of text rendered using this style.
+    fn bounding_box(&self, text: &str, position: Point) -> (Rectangle, Point);
```
jamwaffles, line 28: Should we perhaps name this `first_line_bounding_box` or something like that to be more specific? If I didn't read the docs, I'd assume this `bounding_box` method would behave like all the others in e-g and give me the bounding box of the complete, potentially multiline, string.

I'm also not sure what the `Point` being returned does, so explaining that would help me and others.

Presumably an impl should only test the first line of any given text? Either way, defining this behaviour in another `# Implementation notes` section would help.

src/style/mono_text_style.rs
```diff
@@ -59,50 +60,87 @@ where
...
         D: DrawTarget<Color = Self::Color>,
     {
-        target.draw_iter(MonoPixels::new(text, *self))
+        let mut first = true;
+
+        for c in text.chars() {
+            if first {
+                first = false;
+            } else if F::CHARACTER_SPACING > 0 {
+                // File space between characters if background color is set.
```
jamwaffles, line 78:
```suggestion
                // Fill space between characters if background color is set.
```

jamwaffles (changes requested):
Looks good, but the behaviour of `TextStyle` was confusing and a few more doc comments around the methods would help implementers.

src/style/text_style.rs
rfuest: There will be some changes to the trait in future PR to add additional formatting options, like horizontal and vertical alignment.

> * For the first line, what's the starting `position`?

It only renders a single line. The meaning of `position` will change in the future, so I decided not to docment it right now.

>* For subsequent lines, isn't the position just `position + Point::new(0, <line_y_height>)`?

Yes, it should. The method returns the offset, because `Text` doesn't known the line height.

> * Does the `position` within a line matter? E.g. should it be the top left corner of the line, the baseline, etc?

This will be added later.

> After writing the other comment, should this be renamed to `render_first_line`

It's used for all lines, so `render_first_line` wouldn't be a good name. But I'll rename it to `render_line`.

src/style/mono_text_style.rs
rfuest: :grimacing: Another stupid typo, I shouldn't write comments at the moment.

src/style/text_style.rs
rfuest: > Should we perhaps name this `first_line_bounding_box` or something like that to be more specific?

I've renamed it to `line_bounding_box`.

> I'm also not sure what the `Point` being returned does, so explaining that would help me and others.

This might change in future PR and I've only added a TODO to remember to document this later.

> Presumably an impl should only test the first line of any given text?

It will only be called with single line strings.

src/fonts/text.rs
```diff
@@ -99,21 +90,46 @@ where
...
+        let mut position = self.primitive.position;
+
+        let mut min_max: Option<(Point, Point)> = None;
+
+        for line in self.primitive.text.lines() {
+            let (bounding_box, position_delta) = self.style.bounding_box(line, position);
+
+            if let Some(bottom_right) = bounding_box.bottom_right() {
+                if let Some((min, max)) = &mut min_max {
+                    min.x = min.x.min(bounding_box.top_left.x);
```
bugadani, line 102: Preparing for RTL and bottom-to-top text? :) I would assume `min == position` otherwise, is there something I don't see?

bugadani (commented):
Neat, now people have even fewer reasons to use e-t 😅

src/fonts/text.rs
rfuest: I wanted to keep it as generic as possible. And I think this will be required if we add horizontal alignment support, even without RTL.

jamwaffles (approved):
LGTM

Issue #480, opened 2020-11-15
Improve fonts module documentation


rfuest:
From https://github.com/embedded-graphics/embedded-graphics/pull/470#pullrequestreview-530789384:
> One thing that's missing that I think is important is an example of implementing a custom font. I don't know if it's possible to add one that's short enough to show in the docs, but it would really help tie all the traits and structs together.
>
> There's a reference to the text-custom-font simulator example somewhere (I can't find it in the diff now for some reason) which might be worth turning into a link too.

The `text-custom-font` example is barely documented. At least the constants in the `MonoFont` trait should be documented.

Are there any other aspects of the documentation that could be improved?

rfuest:
Regarding the `Baseline` enum variants:
> I think these could/should be explained better than just echoing the variant names.
>
> For example I'm not entirely sure what `Alphabetic` refers to, although I have a guess. Also, does Top refer to the topmost pixel of the character, or something else? Perhaps we could adapt the text from the linked canvas docs.

_Originally posted by @bugadani in https://github.com/embedded-graphics/embedded-graphics/pull/561#discussion_r588148338_

bugadani:
To clarify, my comment tried to point out an ambiguity with regards to the Top alignment (top of the letter, vs top of the "em box").

cbc02009:
This would be very helpful for me. I'm learning embedded rust by trying to program a simple read sensor and print result to display project, and while it works great I'd love to use a font that takes up more of the screen to make it easier to see.

rfuest:
The need for a bigger font has come up multiple times in the past. embedded-graphics 0.7 will have support for including BDF fonts, which allows users to include bigger fonts. But perhaps it would be a good idea to include at least one larger font in e-g itself, which would be easier to use.

Which font are you currently using and on what type/resolution of display? And approximately how many times bigger should the font be for your application?

cbc02009:
> 
> 
> The need for a bigger font has come up multiple times in the past. embedded-graphics 0.7 will have support for including BDF fonts, which allows users to include bigger fonts. But perhaps it would be a good idea to include at least one larger font in e-g itself, which would be easier to use.
> 
> Which font are you currently using and on what type/resolution of display? And approximately how many times bigger should the font be for your application?

[...]

PR #510, opened 2020-12-02
Text alignment

rfuest:

## PR description

This PR adds vertical and horizontal text alignment settings to `MonoTextStyle`. The default vertical alignment is now `Baseline` instead of `Top` and all builtin fonts now set the baseline.

I've also changed the implementation of `MonoTextStyleBuilder` to be more similar to the `PrimitiveStyleBuilder`. All settings can now be changed using setters and it's no longer required to pass the font to the `new` constructor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,14 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 ### Added
 
 - [#508](https://github.com/embedded-graphics/embedded-graphics/pull/508) Added `MockDisplay::assert_eq_with_message` and `MockDisplay::assert_pattern_with_message`.
+- [#XXX](https://github.com/embedded-graphics/embedded-graphics/pull/XXX) Added `vertical_alignment` and `horizontal_alignment` to `MonoTextStyle`.
+- [#XXX](https://github.com/embedded-graphics/embedded-graphics/pull/XXX) Added `From` impl to convert an existing `MonoTextStyle` into a `MonoTextStyleBuilder`.
 
 ### Removed
 
 - **(breaking)** [#508](https://github.com/embedded-graphics/embedded-graphics/pull/508) `MockDisplay` no longer implements `PartialEq`, use `MockDisplay::assert_eq` instead.
 - **(breaking)** [#509](https://github.com/embedded-graphics/embedded-graphics/pull/509) Styled text can no longer be converted into a pixel iterator, because `IntoPixels` is no longer implemented for `Styled<Text<'_>, S>`.
+- **(breaking)** [#XXX](https://github.com/embedded-graphics/embedded-graphics/pull/XXX) `MonoTextStyleBuilder::new` no longer takes a font as an argument, use `MonoTextStyleBuilder::new().font(SomeFont)` instead.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,14 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 ### Added
 
 - [#508](https://github.com/embedded-graphics/embedded-graphics/pull/508) Added `MockDisplay::assert_eq_with_message` and `MockDisplay::assert_pattern_with_message`.
-- [#XXX](https://github.com/embedded-graphics/embedded-graphics/pull/XXX) Added `vertical_alignment` and `horizontal_alignment` to `MonoTextStyle`.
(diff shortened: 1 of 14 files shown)


rfuest:
The docs aren't great and some of this implementation might change when I implement BDF support. But if you find anything that needs additional documentation please add a note to #480 or #511.

bugadani:
I'd prefer to have the explicitly required parameters (font, in this case) to remain in the builder object's constructor. One less thing I can forget about while using the API. Was this change made so that all the style builders have a parameterless `new()`?

rfuest:
> Was this change made so that all the style builders have a parameterless `new()`?

Yes.

bugadani:
But in this case, since I must specify a font, the actual "constructor" a user has to call in more like `MonoTextStyle::new().font(Font)`. While it's cleaner in terms of "hey, what's that parameter?", I think the API should prevent me creating an invalid style by omitting the font call.

rfuest:
> ... I think the API should prevent me creating an invalid style by omitting the font call.

What is your definition of an "invalid style"? One that cannot be used at all, because of the type, or one that isn't useful, because it doesn't draw anything?

The API doesn't allow `build` to be called, if no font was sent previously, so that it isn't possible to build an invalid `MonotTextStye`. But we could also change the behavior of the builder so that you would always be able to call `build` by setting a default font. The default font could be a dummy empty font or one of the built in ones. Do you think this would be better?

bugadani:
> The API doesn't allow build to be called, if no font was sent previously

Sorry, I missed that. I guess that makes my comment a lot less valid (and my problem a lot more subjective), then. Still, I'd like to ramble some more, just in case I make some sense.

On the objective side, I definitely prefer not allowing to create a style without a font. If something is obviously incorrect, let the compiler complain, instead of surprising the user by drawing nothing. So in this regard, what you made is something I would look for.

(This will probably be a bit weak, as the API is simple enough.) On the subjective side, and thinking as a user, if I have to set a font to draw a piece of text, I would be surprised to see an API that _seemingly_ lets me not do that. I am allowed to write the following piece of code, that is syntactically correct, and *not* have it compile:

```rust
let my_style = MonoTextStyle::new().build();
```

I would be greeted with a message that (grain of salt applies, I didn't check explicitely and I can't remember these things) `build()` is not implemented for the MonoTextStyle struct. The error message would, I guess mention `UndefinedFont` somewhere in the type, but not that I _have to call `font()`_ to fix the issue. In this regard, I think I would prefer a missing argument error - I think it makes it easier to spot the mistake.

rfuest:
> On the objective side, I definitely prefer not allowing to create a style without a font. If something is obviously incorrect, let the compiler complain, instead of surprising the user by drawing nothing. So in this regard, what you made is something I would look for.

We also allow users to build empty `PrimitiveStyles` (which would be hard to prevent on the type level). So it would also be reasonable to apply the same logic to `MonoTextStyle` and allow missing a missing font setting. Perhaps setting a default font would be a good compromise. It would draw something, as long as the `text_color` is set and it should be obvious if the font isn't correct. And this would make sure that `build` can always be called.

> The error message would, I guess mention `UndefinedFont` somewhere in the type, but not that I _have to call `font()`_ to fix the issue.

It would be nice to be able to provide a better error message for this, but I don't think that it's possible.

> I think I would prefer a missing argument error - I think it makes it easier to spot the mistake.

It would, but I prefer consistency between the builders in this case.

Setting the font in the `new` constructor would mean that there are two methods of setting the font, which could also be confusing. It would make the API different for new styles vs styles that are based on an existing style:
```rust
// with new argument
let style = MonoTextStyle::new(Font6x8).build();
let other_style = MonoTextStyle::from(&style).font(Font12x16).build();

// without new argument
let style = MonoTextStyle::new().font(Font6x8).build();
let other_style = MonoTextStyle::from(&style).font(Font12x16).build();
```

bugadani:
>  It [...] should be obvious if the font isn't correct.

I guess there are arguments both for and against a default font, but I don't want to be the one who guesses which font that should be. I would personally go with not providing a default - the iteration time is shorter with a compile error, compared to a manual test ("hmm, that doesn't look right") especially if the text in question is buried under a lot of necessary interactions.

Let's leave this as is, then and thank you for explaining your reasons :)

jamwaffles:
> I would personally go with not providing a default

Agreed

rfuest:
@jamwaffles Is there anything I can do to help to get this merged? If you don't have time to review it at the moment I can find other things to work on, but the text improvements are blocked by this PR.

jamwaffles (approved):
Apologies, this fell down the back of the couch. All looks good to me.

PR #512, opened 2020-12-03
Multi-stage Dockerfile

rfuest:
This PR changes the Dockerfile to be a multi-stage build. This makes it possible to use a different version of rust to compile the tools.

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,12 +1,26 @@
+FROM cimg/rust:1.40.0 AS tools-builder
+
+# Update rust using rustup.
+#
+# We cannot use a newer cimg/rust version to build the tools,
+# because newer images use an incompatible version of libc.
+RUN rustup default 1.48.0
+
+RUN cargo install --root /home/circleci/project just --version ^0.8
+RUN cargo install --root /home/circleci/project cargo-readme --version ^3.2
+RUN cargo install --root /home/circleci/project cargo-deadlinks --version ^0.6
+
 FROM cimg/rust:1.40.0
 
-RUN sudo apt update && sudo apt install -y libsdl2-dev libssl-dev && sudo apt clean
+USER root
 
-RUN cargo install just --version ^0.6
-RUN cargo install cargo-readme --version ^3.2
+RUN apt-get update && apt-get install -y \
+    libsdl2-dev \
+    libssl-dev \
+    && rm -rf /var/lib/apt/lists/*
 
-# MSRV: Upgrade to 0.6 when >= 1.45.0
-RUN cargo install cargo-deadlinks --version ^0.4
+USER circleci
 
-RUN rustup toolchain add nightly -c rustfmt
+RUN rustup toolchain add nightly --profile minimal --component rustfmt
 
+COPY --from=tools-builder /home/circleci/project/* /home/circleci/.cargo/bin/
--- a/justfile
+++ b/justfile
@@ -1,9 +1,7 @@
 targets := "arm-unknown-linux-gnueabi armv7-unknown-linux-gnueabihf x86_64-unknown-linux-gnu x86_64-unknown-linux-musl thumbv6m-none-eabi thumbv7em-none-eabi thumbv7em-none-eabihf thumbv7m-none-eabi"
 
 target_dir := "target"
-# FIXME: `-cimg` suffix is temporary while moving to new CircleCI base images. Remove this suffix
-# when upgrading the tag next time.
-ci_build_image := "jamwaffles/circleci-embedded-graphics:1.40.0-cimg"
+ci_build_image := "jamwaffles/circleci-embedded-graphics:1.40.0-2"
 
 # list of all features except criterion
 all_features := "nalgebra_support fixed"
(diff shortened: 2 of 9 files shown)


.circleci/Dockerfile
```diff
@@ -1,12 +1,26 @@
...
+    libsdl2-dev \
+    libssl-dev \
+    && rm -rf /var/lib/apt/lists/*
 
-# MSRV: Upgrade to 0.6 when >= 1.45.0
-RUN cargo install cargo-deadlinks --version ^0.4
+USER circleci
 
-RUN rustup toolchain add nightly -c rustfmt
+RUN rustup toolchain add nightly --profile minimal --component rustfmt
```
rfuest, line 24: I don't think we use the nighlty toolchain for anything else than `rustfmt` and the `minimal` profile only seems to remove the docs, which saves a few hundred MB in the image.

justfile
```diff
@@ -68,7 +68,7 @@ install-targets:
 # Generates the docs
 generate-docs:
     cargo clean --doc
-    cargo doc --workspace --features "{{all_features}}"
+    cargo doc --workspace --features "{{all_features}}" --no-deps
```
rfuest, line 71: I had to disable the deps, because they also contain broken links, which are detected by the newer `cargo-deadlinks` version.

src/geometry/angle.rs
```diff
@@ -312,18 +311,18 @@ impl Neg for Angle {
     }
 }
 
-impl ApproxEq for Angle {
```
rfuest, line 315: The docs for this trait also caused problems with the linkchecker. We don't use `approx_eq` anywhere else than in the test, so I decided to just move the trait into the test module.

rfuest:
The Dockerfile should also be moved into the new `build-tools` repo, but I couldn't open a PR for an empty repo and don't have the permission to push to that repo.

.circleci/Dockerfile
jamwaffles: I don't think we need nightly in the build at all. The only nightly requirement is `rustfmt.nightly.toml` which we don't use in an automated way anywhere.

jamwaffles:
I pushed an initial commit to `build-tools` and added the e-g core team to the maintainers so you should be able to work on it now.

.circleci/Dockerfile
rfuest: `check-drawing-examples` uses nightly rustfmt and is run automated.

rfuest:
Should we merge this PR into the repo first or should I remove the docker changes and only leave the link fixes in this PR?

rfuest:
> I pushed an initial commit to `build-tools` and added the e-g core team to the maintainers so you should be able to work on it now.

Thanks! Can you do the same for the `examples` repo?

jamwaffles:
Done! I don't know if it's necessary but I pushed an initial commit too. 

For the record, I don't want to gatekeep any of these repos - if there are permissions issues in the future please let me know asap and I'll change them.

jamwaffles:
The image name is currently defined as `jamwaffles/circleci-embedded-graphics:1.40.0-cimg`. Should we call it `jamwaffles/circleci-embedded-graphics:1.40.0-cimg-multi` perhaps? When we change MSRV, this would shrink down to simply `jamwaffles/circleci-embedded-graphics:1.x.y`.

If you can change the image name in the justfile and `.circleci/config.yml`, I can push a new image which should allow this branch to build in CI with it for testing.

rfuest:
Should we just always append a version number to the tag, to make it easier to name updates in the future? Like `jamwaffles/circleci-embedded-graphics:1.40.0-0`, `jamwaffles/circleci-embedded-graphics:1.40.0-1`...

jamwaffles:
Good idea. I used `cimg` originally as it was supposed to be a one off until an upgrade to a newer version. Numbers look better though.

rfuest:
I've changed the tag and the CI build now fails, but it should hopefully start working again once the new image is uploaded.

jamwaffles:
I pushed the new image and everything is passing except beta, with more doc comment issues. Does this branch need a rebase or something?

rfuest:
There were some links that were actually broken and this was never detected before, because rustdoc doesn't output an error if you use `[Something]` in the docs but forget to add `[Something]: URL` at the bottom.

Not sure about the new error yet, the broken link doesn't exist in a stable build of the docs. I guess I'll need to install the beta toolchain to figure this out.

jamwaffles:
That makes sense. There's been a lot of work with intra doc links recently so I guess it's good that all this stuff is surfacing, as annoying as it is.

Looks like there's one more link you missed.

rfuest:
> Looks like there's one more link you missed.

I had to remove some links from the `IntoPixels` trait to make this work.  IMO the missing links are no big problem in this case, because the docs for `IntoPixels` are very short and you can easily fit all functions and types on one screen.

jamwaffles (approved):
No worries. `into_pixels` is also a <kbd>Ctrl</kbd> + <kbd>F</kbd> or <kbd>S</kbd> away anyway.

Rest looks good.

PR #518, opened 2020-12-15
Publish UndefinedFont

bugadani:

## PR description

This change makes it possible to create wrapper objects that contain `MonoTextStyleBuilder`s without the need to specify the font in the constructor. This allows users to create APIs consistent with the one provided by e-g.

### DIFF
--- a/src/style/mod.rs
+++ b/src/style/mod.rs
@@ -6,7 +6,7 @@ mod styled;
 mod text_style;
 
 pub use mono_text_style::{
-    HorizontalAlignment, MonoTextStyle, MonoTextStyleBuilder, VerticalAlignment,
+    HorizontalAlignment, MonoTextStyle, MonoTextStyleBuilder, UndefinedFont, VerticalAlignment,
 };
 pub use primitive_style::{PrimitiveStyle, PrimitiveStyleBuilder, StrokeAlignment};
 pub use styled::{Styled, StyledPrimitiveAreas};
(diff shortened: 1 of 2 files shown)


bugadani:
cc @rfuest I think you might have had a reason to not do this?

rfuest (changes requested):
I considered `UndefinedFont` to be an internal type that doesn't need to be public. But if you need it to be public to use it in e-t I don't mind to make it public.

`UndefinedFont` wasn't documented very well, because it's wasn't meant to be visible in the docs. Can you at least add a link to `MonoTextStyleBuilder`?

bugadani:
Well, if I want to provide a style builder API consistent with e-g, I must be able to create a style builder object with an `UndefinedFont`. Therefore I think `UndefinedFont` should be public.

I can try and add some docs.

bugadani:
Sorry for taking this long.

So as it stands, this probably won't be necessary for e-t, instead the library will have two different styles. But maybe third parties may find themselves wanting to wrap a MonoTextStyleBuilder and so I think there's still some value in this PR :)

In general, I find myself annoyed when a library relies on a private type anywhere in its public API. While there might be good reasons to do it, but in the end, personally, I'm just going to hit a wall somewhere because of those. A different approach to flat-out publishing UndefinedFont would be to instead export it in a `dev` module, so that it doesn't pollute the "commonly used" API, but is still accessible if needed. But I wouldn't do this for a single type.

bugadani:
I'm a dummy, I can provide my own struct instead of polluting e-g. D'oh.

Issue #505, opened 2020-11-29
Finish crate split


rfuest:
From https://github.com/embedded-graphics/embedded-graphics/pull/498#pullrequestreview-540484691:

- [ ] Move `PointsIter` from into `embedded_graphics_core::iterator`
- [ ] Move `RawDataIter` into `embedded_graphics::iterator`
- [x] Check which types in the `iterator` module need to be in `core`
- [x] Try to remove draw target adapters, like `Clipped` or `ColorCoverted`, from `core`
- [x] Move `TextStyle` ~and `TextStylePixels`~ to `core`
- [x] Fix `tinybmp` and `tinytga` after new core release

rfuest:
How will we document which e-g-core version is compatible with which e-g versions? Should a README file of an external crate mention the e-g-core version it's compatible with instead of the e-g version? The core version number might not be familiar for application developers who only depend on e-g itself.

rfuest:
I just noticed that we missed to move the external font renderer support to the core crate.

jamwaffles:
Dangit. What needs moving over though? I can see the `Text` struct and `TextStyle`/`TextStylePixels` traits, but I'm not sure what else.

rfuest:
`TextStyle` and `TextStylePoints` should be in core and `Text` should stay in e-g. Additional traits might be added in the future to add frame text support.

jamwaffles:
Both `TextStyle` and `TextStylePixels` (which is what I assume you mean by `TextStylePoints`) require `Text` though, so we'd have to move `Text` into core as well.

rfuest:
Good point. I'm not sure yet if it's better to move `Text` into core or replace `Text` by `&str` and `Point` in `TextStyle` and `TextStylePixels`.

jamwaffles:
I'm not sure either. At any rate, moving the text stuff into core won't be a breaking change as far as I can see, so we can do it at our leisure.

Maybe @bugadani can add some comments on what's required from e-g-core in terms of text.

bugadani:
I think TextStyle and co. is rather renderer-specific, so I wouldn't necessarily move them. To be fair, I haven't had the time to dive into 0.7 and how it affects e-t. In the current versions, although I use TextStyle and TextStyleBuilder, they aren't strictly necessary. I treat e-t as it's own text renderer, and for that the most important dependency is the Font trait, besides the mechanisms to create my own pixel renderers.

[...]

PR #522, opened 2021-01-08
Move DrawTargetExt and adapters from core to e-g

rfuest:

## PR description

This PR moves `DrawTargetExt` and the draw target adapters, like `Cropped` and `Translated`, from `core` to `e-g`. IMO they don't need to be included in `core`, because the aren't part of the image, font or display APIs.

The reason these were included in the first place was the default impl of `draw_sub_image` in `ImageDrawable`. This PR removes the default impl, so that each image format crate will need to implement this method separately. I think this is a reasonable compromise to keep the core crate small. `tinybmp` and `tinytga` will need to depend on e-g instead of core to implement this feature, but they should still be compatible across different e-g versions, because all types in the `ImageDrawables` are included in `core`.


jamwaffles (changes requested):
It's a shame to break core so early, but nobody is using it so that's fine.

Looks good to me apart from any necessary changelog items for core and e-g.

To confirm, once merged I'll:

* Release core 0.2.0
* Release e-g 0.7.0-alpha.3

The simulator, tinytga and tinybmp will all need new alphas as well once the above is done.

rfuest:
> It's a shame to break core so early, but nobody is using it so that's fine.

This change is part of #505 and we had discussed the breaking changes in the core crate in the linked discussion https://github.com/embedded-graphics/embedded-graphics/pull/498#pullrequestreview-540484691.

Regarding the changelog entries: Does e-g need an entry for this change, because the API didn't change? The e-g changelog doesn't mention the core split at all. We should probably start to document the compatibility of e-g and e-g-core versions at some point, but that's outside the scope of this PR.

rfuest:
> To confirm, once merged I'll:
> 
> * Release core 0.2.0
>  * Release e-g 0.7.0-alpha.3

Would it be better to wait until #505 is finished?

jamwaffles (approved):
> Does e-g need an entry for this change, because the API didn't change?

No, that changelog can be left alone.

> Would it be better to wait until #505 is finished?

Ah yes, it would. It's been a while since I went through the issues again.

Issue #511, opened 2020-12-02
Tracking issue for text rendering changes


rfuest:
- ~[ ] Add `FrameText` #483~
- [x] Import BDF code into `embedded-bdf` repo
- [x] Fix text examples
  The changes in PR #510 will break the examples
- ~[ ] Add text alignment example~ -> Moved to https://github.com/embedded-graphics/examples/issues/8
- ~[ ] Use `FrameText` in analog clock example~
- [x] Update builtin monospaced fonts or add additional fonts as separate repo
- ~[ ] Try to distribute BDF fonts as crates~ -> Issue moved to eg-bdf repo embedded-graphics/embedded-bdf#15
- [ ] Improve docs #480
- ~[ ] Add simple external font renderer example to e-g docs~
  IMO this should be added as an external example instead of adding it directly to the e-g docs. A font renderer example would probably be pretty long and a runnable external example might be a better starting point for custom implementations. And depending on the location of the docs it might clutter the docs for a feature that has a relatively small target audience.
- [x] Create repo for seven segment font renderer #481
- [x] Add underline and strikethrough
- [x] Add method to render whitespace
- [x] Add `CharacterStyle` trait: https://github.com/embedded-graphics/embedded-text/pull/90#discussion_r555892950
- [x] Rename `VerticalAlignment` to `Baseline`? #561
- [x] Generate font overview table in the docs
- [x] Rename bold and oblique font variants https://github.com/embedded-graphics/embedded-graphics/pull/535#pullrequestreview-579381754
- [ ] Check if underline and strikethrough for new fonts is correct
- [ ] Add scaling (maybe also: https://github.com/embedded-graphics/embedded-graphics/issues/533#issuecomment-769887633)
- [x] Add return type to `Text` `Drawable` impl https://github.com/embedded-graphics/embedded-graphics/issues/548#issue-807321731 #566 
- [x] Add line height setting to `TextStyle` #569 
- [ ] Add larger built-in font?

bugadani:
I've been thinking about strikethrough and underlined text support in e-g. Because fonts can look whatever, just dashing a 1px line through them is suboptimal, and so this should be supported on a per-renderer basis. This would remove the responsibility out of e-t to try and guess these decorations, which I consider a good thing. This also means I don't have to knock on your doors to publish MonoFont's pixel iterator :) 

For MonoFont, at least for 6x8 it means the following:
 - For underlined text to look good, underlined characters are 1 px taller than otherwise
[...]

PR #523, opened 2021-01-10
Improved text renderer API

rfuest:

## PR description

This draft PR includes changes for issues #505 and #511. The code isn't finished and hasn't been tested with other text renderers, but it should give an first idea how the implementation could look like.

Summary of changes:
* Split of the `fonts` module in `text` and `mono_font`
  The `fonts` module contained code which was specific to the monospaced font support and generic types which can be used with all text renderers. I've also moved `MoveTextStyle` and `MonoTextStyleBuilder` to have all monospaced font related code in one place.
* The text renderer API was moved into the core crate
  The `TextStyle` trait is now called `TextRenderer` and was extended to provide additional methods for e-t
* Underline and strokethrough support was added to `MonoTextStyle`
* The `Text` specific alignment attributes were removed from `MonoTextStyle`
  A new `TextStyle` type was added to set the alignment for `Text` drawables. To make the API easier to use, if the default alignment is used, `MonoTextStyle` can still be used for text styleds:
   ```rust
   // default alignment

   let style = MonoTextStyle::new(Font6x8, BinaryColor::On);
   Text::new("Text", Point::zero()).into_styled(style).draw(&mut display)?;

   // other alignment

   let character_style = MonoTextStyle::new(Font6x8, BinaryColor::On);

   let text_style = TextStyleBuilder::new()
       .character_style(character_style)
       .horizontal_alignment(HorizontalAlignment::Center)
       .vertical_alignment(VerticalAlignment::Top)
       .build();

   Text::new("Text", Point::zero()).into_styled(text_style).draw(&mut display)?;
   ```

### DIFF
--- a/src/mono_font/mono_text_style.rs
+++ b/src/mono_font/mono_text_style.rs
@@ -53,21 +53,6 @@ where
             .build()
     }
 
-    /// Calculates the offset between the line position and the top edge of the bounding box.
-    fn vertical_offset(&self, vertical_alignment: VerticalAlignment) -> i32 {
-        match vertical_alignment {
-            VerticalAlignment::Top => 0,
-            VerticalAlignment::Bottom => {
-                F::CHARACTER_SIZE.height.saturating_sub(1).saturating_cast()
-            }
-            VerticalAlignment::Center => {
-                (F::CHARACTER_SIZE.height.saturating_sub(1) / 2).saturating_cast()
-            }
-            VerticalAlignment::Baseline => F::BASELINE
-                .unwrap_or_else(|| F::CHARACTER_SIZE.height.saturating_sub(1).saturating_cast()),
-        }
-    }
-
     /// Resolves a decoration color.
     fn resolve_decoration_color(&self, color: DecorationColor<C>) -> Option<C> {
         match color {
@@ -77,22 +62,40 @@ where
         }
     }
 
+    fn draw_background<D>(
+        &self,
+        width: u32,
+        position: Point,
+        target: &mut D,
+    ) -> Result<(), D::Error>
+    where
+        D: DrawTarget<Color = C>,
+    {
+        if width == 0 {
+            return Ok(());
+        }
+
+        if let Some(background_color) = self.background_color {
+            target.fill_solid(
+                &Rectangle::new(position, Size::new(width, F::CHARACTER_SIZE.height)),
+                background_color,
+            )?;
+        }
+
+        Ok(())
+    }
+
     fn draw_strikethrough<D>(
         &self,
         width: u32,
         position: Point,
-        vertical_alignment: VerticalAlignment,
         target: &mut D,
     ) -> Result<(), D::Error>
     where
         D: DrawTarget<Color = C>,
     {
         if let Some(strikethrough_color) = self.resolve_decoration_color(self.strikethrough_color) {
-            let top_left = position
-                + Point::new(
-                    0,
-                    F::STRIKETHROUGH_OFFSET - self.vertical_offset(vertical_alignment),
-                );
+            let top_left = position + Point::new(0, F::STRIKETHROUGH_OFFSET);
             let size = Size::new(width, F::STRIKETHROUGH_HEIGHT);
 
             target.fill_solid(&Rectangle::new(top_left, size), strikethrough_color)?;
@@ -101,22 +104,12 @@ where
         Ok(())
     }
 
-    fn draw_underline<D>(
-        &self,
-        width: u32,
-        position: Point,
-        vertical_alignment: VerticalAlignment,
-        target: &mut D,
-    ) -> Result<(), D::Error>
+    fn draw_underline<D>(&self, width: u32, position: Point, target: &mut D) -> Result<(), D::Error>
     where
         D: DrawTarget<Color = C>,
     {
         if let Some(underline_color) = self.resolve_decoration_color(self.underline_color) {
-            let top_left = position
-                + Point::new(
-                    0,
-                    F::UNDERLINE_OFFSET - self.vertical_offset(vertical_alignment),
-                );
+            let top_left = position + Point::new(0, F::UNDERLINE_OFFSET);
             let size = Size::new(width, F::UNDERLINE_HEIGHT);
 
             target.fill_solid(&Rectangle::new(top_left, size), underline_color)?;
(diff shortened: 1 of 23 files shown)


bugadani:
Huh, that's bigger than expected, you just made embedded-text a lot simpler 😅
At this point, by the way, embedded-text is no longer a renderer, but more of a layout engine.

I think I'd prefer the name `.font_style()` over `.character_style()` but that's just a nit.
Let me get back to you after a first round of integration.

rfuest:
> I think I'd prefer the name `.font_style()` over `.character_style()` but that's just a nit.

Can you give some reasons why you prefer `font_style`? I did choose `character_style`, because it might be familiar from word processors or DTP applications. `TextStyle` would be similar to a paragraph style, but not quite the same.

bugadani:
Completely subjective, so it's fine as is.

rfuest:
> Completely subjective, so it's fine as is.

OK. But let me know if you find a reason why it shouldn't be named `character_style`. The name isn't set in stone yet and there might be valid reasons to rename it.

rfuest:
@bugadani How should we handle transparent text? You did add some logic to the internal font renderer that returns an zero sized BB when the text is completely transparent. This makes sense for `Text`, but I don't think it makes sense for e-t. If only a part of the text is transparent I would expect that there would be a gap in the text instead of moving the adjacent texts closer together.

Should we extend the renderer API to make it possible to detect transparent text or return the same BB regardless of the color setting?

rfuest:
I did mark this PR as ready for review. I'm sure that the API isn't final yet, but I think it should be a good base for some initial testing and we can refine it in future PRs.

bugadani:
@rfuest Good question. I think the size of a piece of text (i.e. MonoTextStyle) shouldn't depend on whether the contents are transparent, but the size of the object drawn (i.e. (styled) Text) should, so my "if transparent, return empty" should live in Text, and will probably not be present in TextBox.

rfuest:
I think I've found a way to implement this with minimal changes to the API and will update the PR soon.

src/mono_font/mono_text_style.rs
```diff
@@ -203,15 +203,20 @@ where
...
 
-    fn measure_text(&self, text: &str, position: Point) -> TextMetrics {
-        // TODO: ignore control characters in `text`
-
+    fn measure_string(&self, text: &str, position: Point) -> TextMetrics {
         let width = (text.len() as u32 * (F::CHARACTER_SIZE.width + F::CHARACTER_SPACING))
             .saturating_sub(F::CHARACTER_SPACING);
         let size = Size::new(width, F::CHARACTER_SIZE.height);
 
+        // Return a zero sized bounding box is the text is completely transparent.
```
bugadani, line 211: The typo ("box is the text is ..."), sounds like slang from Mistborn, lol, I approve :D

rfuest: Fixed.

bugadani:
@rfuest since you wrote it's ready for review, I think I'm missing how to measure transparent text. Can you point my tired brain at it, please? :)

rfuest:
You can measure the width of transparent text by using the `next_position` field in `TextMetrics`. Like in `Text`: https://github.com/rfuest/embedded-graphics/blob/1b6a9316d357c016fedcf8a6781bdccb9b6eebb2/src/text/text.rs#L83-L118

bugadani:
That is private, though, isn't it? :)

rfuest:
The `lines` method is private and would probably need to be extended for e-t anyway, but the `TextMetrics` shouldn't be.

bugadani:
Ah, all right, feels a bit weird but I guess if it works, it works.

rfuest:
It is weird, but I hope it works. I'll take another look at this once the BDF support is implemented fully. For italic fonts the bounding box of a string can be wider than the distance from the start point to `next_position` which can easily cause some confusion.

src/mono_font/mono_text_style.rs
```diff
@@ -0,0 +1,834 @@
...
+#[non_exhaustive]
+pub struct MonoTextStyle<C, F> {
+    /// Text color.
+    pub text_color: Option<C>,
+
+    /// Background color.
+    pub background_color: Option<C>,
+
+    /// Underline color.
+    pub underline_color: DecorationColor<C>,
```
bugadani, line 34: Silly question - once set, how do I turn these off?

bugadani: Actually, I don't see how I turn them on if I have a style object already. I guess [this comment](https://github.com/embedded-graphics/embedded-text/pull/90#discussion_r555892950) isn't implemented yet.

bugadani: Normally it would be possible to convert the style to a builder and set the decorations to None color, but unfortunately I don't think I can do that through the TextRenderer trait.

rfuest: You are correct that there currently is no way to set any color back to `None`. This limitation also exists in `PrimitiveStyleBuilder` and should be fixed in both (in a future PR). There was some discussion about this when we introduced `PrimitiveStyleBuilder`, but I don't remember the details and why we decided not to add any method to reset colors.

> Actually, I don't see how I turn them on if I have a style object already. I guess this comment isn't implemented yet.

I want to get this PR merged first and add additional traits in the next PRs. The suggested trait can be used by e-t or any other library, but we will need separate implementations for the user facing API. I don't want to use the same trait, because it contains methods that aren't supported by every text renderer (like `set_bold`).

> Normally it would be possible to convert the style to a builder and set the decorations to None color, but unfortunately I don't think I can do that through the TextRenderer trait.

That's the reason I want to add the `CharacterStyle` trait.

bugadani: Resetting is a must-have for SGR parameters. It's not just the colors, but the decorations themselves as well. I guess it's not an urgent blocker though, there is a lot of work still beside escape sequences.

I'm fine with the idea of an additional, dev-facing trait, I just hope I don't have to drop the fancy features :P

rfuest: > It's not just the colors, but the decorations themselves as well.

Setting `underline_color` or `strikethrough_color` is the way to disable these decorations. The `CharacterStyle` example did just use the wrong types in the setter functions.

> I guess it's not an urgent blocker though, there is a lot of work still beside escape sequences.

I'll try to work on this as soon as this PR is merged. I'm just waiting for @jamwaffles review.

> I'm fine with the idea of an additional, dev-facing trait, I just hope I don't have to drop the fancy features :P

I think this should be possible and we can still break the API until 0.7 is released.

jamwaffles (approved):
I didn't look at this too closely, but I want to unblock @rfuest so I'm approving this PR to let work continue.

A couple of comments:

* I'm a bit concerened that e-g's builtin text will eventually try to do too much or overlap too heavily with e-t, but I'm jumping the gun here - let's see where things lead over time.
* The `TextRenderer` docs could use some work, but I don't think they're a focus for this PR.
* Is there a reason `character_style` has to be a nested property of the builder? It would be nicer to be able to call `.font(...).text_color(...)` instead of constructing another object, however I'm probably missing something about how other renderers integrate (I did look at the BDF PR though).

rfuest:
> I'm a bit concerened that e-g's builtin text will eventually try to do too much or overlap too heavily with e-t, but I'm jumping the gun here - let's see where things lead over time.

In which regard? I thought that this PR would reduce overlap.

> The `TextRenderer` docs could use some work, but I don't think they're a focus for this PR.

The text support in general needs more docs, but I think it would be easier to wait until the API is stabilized.

> Is there a reason `character_style` has to be a nested property of the builder?

The `character_style` is the text renderer dependent part of the styling. By keeping it separate `TextStyle` and e-t's `TextBoxStyle` (and the associated builders) can be independent of the text renderer. It also resembles the split between character and paragraph styling in DTP applications.

bugadani:
> In which regard? I thought that this PR would reduce overlap.

Also, it allows for a LOT more flexibility as well :)

jamwaffles:
Overlap was probably the wrong word. Ignore me anyway - that comment was more of a brain dump than anything useful :)

PR #527, opened 2021-01-22
Detect nearly-parallel numerical inaccuracy

### LINKED ISSUE #471: Polyline end segment out of bounds draw
- Version of embedded-graphics in use (if applicable): [version here]
- Target: simulator

For _some_ positions and thicknesses of polyline points, the end segment has screwed up rendering:

image

Only happens at `stroke: 5` as far as I can tell. Ignore the triangle and horizontal blue line - that's from other test stuff.

Do bear in mind this is on a WIP branch where I've been messing around with line joints, so may not be reproducible in master - I'll check again when I'm in a clean enough state to `git checkout master`.

## Test case (if applicable)

```rust
let line_style = PrimitiveStyle::with_stroke(Rgb888::GREEN, 5);

let h = 100;

let mouse_pos = Point::new(137, 15);

let points = [
    Point::new(10, h / 2),
    Point::new(50, h / 2),
    Point::new(60, h / 2 - 20),
    Point::new(70, h / 2),
    Point::new(80, h / 2),
    Point::new(90, h / 2 + 10),
    Point::new(100, 10),
    Point::new(110, h / 2 + 20),
    Point::new(120, h / 2),
    mouse_pos,
];

Polyline::new(&points)
    .into_styled(line_style)
    .draw(display)?;
```

jamwaffles:

## PR description

Not too happy with this solution, but it fixes the original problem so
at least there's that.

I don't understand the maths behind it at _all_, this just seemed to
work for the cases in #471. Hopefully no other bugs are introduced by
this change.

I'm very open to a better solution, or at least an explanation of why my
fix works.

Closes #471.

### DIFF
--- a/src/primitives/line/mod.rs
+++ b/src/primitives/line/mod.rs
@@ -1,15 +1,19 @@
 //! The line primitive
 
 mod bresenham;
+pub(in crate::primitives) mod intersection_params;
 mod points;
 mod styled;
 mod thick_points;
 
 use crate::{
-    geometry::{Dimensions, Point, PointExt},
+    geometry::{Dimensions, Point},
     primitives::{
-        common::{LineSide, LinearEquation, StrokeOffset},
-        line::thick_points::{ParallelLineType, ParallelsIterator},
+        common::{LineSide, StrokeOffset},
+        line::{
+            intersection_params::IntersectionParams,
+            thick_points::{ParallelLineType, ParallelsIterator},
+        },
         PointsIter, Primitive, Rectangle,
     },
     transform::Transform,
@@ -196,62 +200,9 @@ impl Line {
     ///
     /// Inspired from https://stackoverflow.com/a/61485959/383609, which links to
     /// https://webdocs.cs.ualberta.ca/~graphics/books/GraphicsGems/gemsii/xlines.c
+    #[allow(unused)]
     pub(in crate::primitives) fn intersection(&self, other: &Line) -> Intersection {
-        let line1 = LinearEquation::from_line(self);
-        let line2 = LinearEquation::from_line(other);
-
-        // Calculate the determinant to solve the system of linear equations using Cramer's rule.
-        let denominator = line1.normal_vector.determinant(line2.normal_vector);
-
-        // The system of linear equations has no solutions if the determinant is zero. In this case,
-        // the lines must be colinear.
-        if denominator == 0 {
-            return Intersection::Colinear;
-        }
-
-        let outer_side = if denominator > 0 {
-            LineSide::Right
-        } else {
-            LineSide::Left
-        };
-
-        // Special case: If the two lines are almost parallel, return the average point between
-        // them.
-        if denominator.pow(2) < self.delta().dot_product(other.delta()) {
-            return Intersection::Point {
-                point: (self.end + other.start) / 2,
-                outer_side,
-            };
-        }
-
-        // If we got here, line segments intersect. Compute intersection point using method similar
-        // to that described here: http://paulbourke.net/geometry/pointlineplane/#i2l
-
-        // The denominator/2 is to get rounding instead of truncating.
-        let offset = denominator.abs() / 2;
-
-        let origin_distances = Point::new(line1.origin_distance, line2.origin_distance);
-
-        let numerator =
-            origin_distances.determinant(Point::new(line1.normal_vector.y, line2.normal_vector.y));
-        let x_numerator = if numerator < 0 {
-            numerator - offset
-        } else {
-            numerator + offset
-        };
-
-        let numerator =
-            Point::new(line1.normal_vector.x, line2.normal_vector.x).determinant(origin_distances);
-        let y_numerator = if numerator < 0 {
-            numerator - offset
-        } else {
-            numerator + offset
-        };
-
-        Intersection::Point {
-            point: Point::new(x_numerator, y_numerator) / denominator,
-            outer_side,
-        }
+        IntersectionParams::from_lines(self, other).intersection()
     }
 }
 
--- a/src/primitives/line/mod.rs
+++ b/src/primitives/line/mod.rs
@@ -9,11 +9,8 @@ mod thick_points;
 use crate::{
     geometry::{Dimensions, Point},
     primitives::{
-        common::{LineSide, StrokeOffset},
-        line::{
-            intersection_params::IntersectionParams,
-            thick_points::{ParallelLineType, ParallelsIterator},
-        },
+        common::StrokeOffset,
+        line::thick_points::{ParallelLineType, ParallelsIterator},
         PointsIter, Primitive, Rectangle,
     },
     transform::Transform,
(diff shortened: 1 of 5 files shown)


src/primitives/line/mod.rs
```diff
@@ -209,6 +209,21 @@ impl Line {
...
 
+        let outer_side = if denominator > 0 {
+            LineSide::Right
+        } else {
+            LineSide::Left
+        };
+
+        // Special case: If the two lines are almost parallel, return the average point between
+        // them.
+        if denominator.pow(2) < self.delta().dot_product(other.delta()) {
```
rfuest, line 220: You did mention that you don't know the math behind this, but can you explain how you came up with this formula?

src/primitives/line/mod.rs
```diff
@@ -209,6 +209,21 @@ impl Line {
...
+            return Intersection::Point {
+                point: (self.end + other.start) / 2,
```
rfuest, line 222: Isn't this dependent on the order of `star` and `end`? If, for example, `self.end` is close to the intersection and `other.start` is further away than `other.end` the average won't be in the expected place.

rfuest (changes requested):

src/primitives/line/mod.rs
jamwaffles: I had a test line from #471 set up in the simulator and printed a bunch of values whilst dragging the points around. I tried this formula and the spurs went away when I introduced it. It just seemed to work...

src/primitives/line/mod.rs
jamwaffles: That's true, however for the nearly-parallel case*, each point will only be 1 or 2 px away from the other so it doesn't matter. I originally just went for using `point: self.end`.

* Which I'm hoping is the only time the `if` is entered into.

rfuest: I've added a PR for a debug tool to the `debug-tools` repo. If you move `l2_end` vertically, with all other points at their default value, you'll notice that the intersection point doesn't move, because it averages the wrong points.

rfuest: And this case isn't correct either:
grafik

jamwaffles:
I've pushed a commit that separates the intersection logic into a new struct. This allows `Line::intersection` to use the normal intersection logic without the magic check in it, and the `LineJoin` intersections to use it by calling an extra method. The API is designed so there is minimum recomputation of values used, and supports "ordinary" line intersections without the magic check if we need that functionality in the future, or want to expose `Line::intersection` publicly later. 

Should we remove `Line::intersection` now, seeing as it's not `pub` and is no longer used inside e-g?

I still don't have an explanation of how the magic check works, so apologies for that, but I think the separation of the magic check from normal line intersections is much better now. Open to suggestions as always.

rfuest:
I'll comment on the rest of the PR soon, but I just noticed something that could be fixed in this PR: The `Intersection` enum is public, but it shouldn't be.

rfuest:
Could we just return `Intersection::Colinear` when the angle between the lines is small and not only if its exactly 0°?

If I change the check for colinearity in `Line::intersection` (in master) to:
```rust
if denominator.pow(2) < (self.delta().length_squared() * other.delta().length_squared()) / 2048 {
    return Intersection::Colinear;
}
```
The spurs tests passes, but some other tests fail, which I haven't looked into yet. The rule above returns `Colinear` if the angle between the lines is `arcsin(sqrt(1/2048))` which is ~1.3°.

rfuest:
I've updated the code in the previous comment (replaced `<` with `<=`) and all test should pass now.

jamwaffles:
In the code above, have you removed the `nearly_colinear_has_error` check completely? If I add that `if` to `IntersectionParams::intersection` I still see a small spur for some lines:

image

Whereas with the original code in this PR I don't see it:

image

I'll add a test case for this and add it to the PR.

rfuest:
I did change the code in `master` and not in this branch.

jamwaffles:
This case fails for me in `master` with your `if` applied:

```rust
#[test]
fn issue_471_spurs_2() {
    let points = [Point::new(13, 65), Point::new(20, 50), Point::new(31, 30)];

    let line = Polyline::new(&points)
        .translate(Point::new(0, -15))
        .into_styled(PrimitiveStyle::with_stroke(Rgb565::RED, 18));

    let bb = line.bounding_box();

    // Check bounding box is correct
    assert_eq!(bb, Rectangle::new(Point::new(4, 26), Size::new(36, 44)));

    line.draw(&mut display).unwrap();

    // Check no pixels are drawn outside bounding box
    assert_eq!(display.affected_area(), bb);
}
```

Can you give it a check on your machine in case I did something wrong on my side?

rfuest:
The test also fails on my machine. The angle between the lines is larger than the threshold and doesn't trigger the `Colinear` code. But I don't think that simply increasing the threshold will fix the problems and I don't think your version of `nearly_colinear_has_error` does anything else than comparing the angle of the lines.

Here is another fun test case, which breaks with both versions:
Peek 2021-01-24 21-08

rfuest:
Perhaps we could fix this by using a different algorithm to determine the line join points. Instead of intersecting the two edges we could intersect one edge and the angle bisector of both lines. These lines would be nearly perpendicular for these small angles and shouldn't be a problem for the intersection algorithm.

This should show demonstrate the idea behind this. Instead of intersecting two blue edges we would intersect one of the blue edges and the red line:
line

jamwaffles:
The best I could get with that method is this:

```rust
/// Bisecting line between two lines, tip to tail.
fn bisector(l1: &Line, l2: &Line) -> Line {
    let d1 = l1.delta();
    let d2 = l2.delta();

    let len1 = micromath::F32Ext::sqrt(d1.length_squared() as f32);
    let len2 = micromath::F32Ext::sqrt(d2.length_squared() as f32);

    let x1 = (d1.x as f32) / len1;
    let y1 = (d1.y as f32) / len1;
    let x2 = (d2.x as f32) / len2;
    let y2 = (d2.y as f32) / len2;

    let x = x1 - x2;
    let y = y1 - y2;

    let p = Point::new((x * 1000.0) as i32, (y * 1000.0) as i32);

    Line::new(l2.start, l2.start + p)
}
```

I couldn't figure a decent way to not use floating point, and using `micromath` leads to some inaccuracies in `.sqrt()` which look like they might produce spurs in very short polyline segments - during debugging, nearly-parallel line segments would produce a bisector that wasn't perpendicular so we're back at square one...

Also, when I check for intersections with the angle bisector line, the thickness of some segments also changes slightly along their lengths. This might happen in master as well, but I haven't checked yet. It's a subtle but visible change.

rfuest (changes requested):
Sorry for leading you down a wrong path, I hadn't considered that the line length would need to be normalized.

I would suggest to merge this partial solution and revisit this after the 0.7 release. If you want to continue working on this now go ahead, but I think I should try to not get distracted by this and continue the work on the text renderers.

rfuest:
Could you open an issue or multiple issues for any other polyline drawing problems you find? IMO a combined issue similar to #484 would give a good overview of the remaining problems. 

Here are two issues I've found:

https://user-images.githubusercontent.com/226123/105753878-a62ab480-5f49-11eb-8d58-5a461f60e5a8.mp4


https://user-images.githubusercontent.com/226123/105753885-a7f47800-5f49-11eb-945a-fe7cccbeddff.mp4

And zero sized segments are rendered as a vertical line.

jamwaffles:
> Sorry for leading you down a wrong path, I hadn't considered that the line length would need to be normalized.

No problem. Some issues only show up during implementation :)

rfuest (approved):
Looks good!

jamwaffles:
Thanks!

PR #531, opened 2021-01-26
Add CharacterStyle trait

rfuest:
This PR adds the `CharacterStyle` trait that was discussed in https://github.com/embedded-graphics/embedded-text/pull/90#discussion_r555892950. @bugadani should be able to use it to make SGRs in e-t work independent of the renderer.

### DIFF
--- a/core/src/text/character_style.rs
+++ b/core/src/text/character_style.rs
@@ -1,0 +1,44 @@
+use crate::pixelcolor::PixelColor;
+
+/// Character style.
+///
+/// This trait is used to modify character styles programmatically, for example, to implement
+/// rendering of text with multiple colors. Applications shouldn't use this trait and instead use
+/// the character style types that are provided by the text renderer, like [`MonoTextStyle`] and
+/// [`MonoTextStyleBuilder`] for the integrated font support.
+///
+///  # Implementation notes
+///
+/// Text renderers don't need to support all settings in this trait. All calls to unsupported
+/// setters should be ignored by the implementation. The trait provided empty default
+/// implementations for all setters.
+///
+/// [`MonoTextStyle`]: ../mono_font/struct.MonoTextStyle.html
+/// [`MonoTextStyleBuilder`]: ../mono_font/struct.MonoTextStyleBuilder.html
+pub trait CharacterStyle {
+    /// The color type.
+    type Color: PixelColor;
+
+    /// Sets the text color.
+    fn set_text_color(&mut self, _text_color: Option<Self::Color>) {}
+
+    /// Sets the background color.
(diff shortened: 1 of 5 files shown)


core/src/text/character_style.rs
```diff
@@ -0,0 +1,41 @@
...
+/// rendering of text with multiple colors. Applications shouldn't use this trait and instead use
+/// the character style types that are provided by the text renderer, like `MonoTextStyle` and
+/// `MonoTextStyleBuilder` for the integrated font support.
+///
+///  # Implementation notes
+///
+/// Text renderers don't need to support all settings in this trait. All calls to unsupported
+/// setters should be ignored by the implementation. The trait provided empty default
+/// implementations for all setters.
+pub trait CharacterStyle {
```
jamwaffles, line 15: I might be missing some context from the previous discussions that answers this but: would it be useful if each of these methods returned `&mut Self` for easier chaining?

bugadani: I guess it wouldn't hurt. But the intended use-case (e-t calling these one at a time) doesn't call for it.

rfuest: I don't think this is a common pattern for setters. And if @bugadani doesn't need it anyway I would prefer to keep it this way.

jamwaffles: No worries, just checking. I had the builder pattern in mind but that doesn't really apply here.

jamwaffles (approved):

rfuest:
Thanks for the review!

core/src/text/character_style.rs
bugadani, line 15: @rfuest is there any chance we could require `CharacterStyle: Clone` (or even Copy)?

because e-t can change the char style, and `draw` takes a shared reference, I think e-t should be able to draw the same thing when `draw` is called multiple times. Because of this, I need to clone the character style (well, the whole TextRenderer in fact). This requires me to add a `Clone` bound so there's a chance certain font renderers won't be compatible with e-t.

bugadani: @rfuest friendly ping. Nothing urgent, but maybe this fell off the radar.

rfuest: Sorry, I hadn't noticed this.

The main disadvantage of adding `Clone` as a supertrait to a trait is that it would be no longer object safe. But because `CharacterStyle` has an associated type it isn't object safe anyway. So I think there should be no harm in adding `Clone`. I would choose `Clone` over `Copy`, because it's more universal and `style.clone()` should do the same as `*style` if `Copy` is also implemented.

PR #532, opened 2021-01-26
Add font conversion tools

rfuest:
This PR adds some font conversion tools that allow us to use monospaced BDF fonts with the internal monospaced font renderer.  At the moment the converter only includes the 7 bit ASCII characters in the generated image file, which we might want to extend later on. Perhaps we could at generate a ASCII and a ISO 8859-1 variant of each font. Any other encodings can be handled by the BDF crate.

I've included some new public domain BDF fonts from the xorg project. The necessary e-g types in the `mono_font` module aren't included yet, but here is an overview of the fonts:
new_fonts


jamwaffles (approved):
LGTM, assuming e-g integration will come in a future PR.

rfuest:
> LGTM, assuming e-g integration will come in a future PR.

I'll open an issue for this, because there are some details we will need to discuss before this can be added.

PR #534, opened 2021-01-27
measure_string: factor in underline

bugadani:
I think text measurement should include decoration as well.
Also, embedded-text's automatic fit-to-text function would cut off the underline without this.

Should `line_height()` also depend on the decoration?

### DIFF
--- a/src/mono_font/mono_text_style.rs
+++ b/src/mono_font/mono_text_style.rs
@@ -206,7 +206,14 @@ where
     fn measure_string(&self, text: &str, position: Point) -> TextMetrics {
         let width = (text.len() as u32 * (F::CHARACTER_SIZE.width + F::CHARACTER_SPACING))
             .saturating_sub(F::CHARACTER_SPACING);
-        let size = Size::new(width, F::CHARACTER_SIZE.height);
+
+        let height = if self.underline_color != DecorationColor::None {
+            F::UNDERLINE_HEIGHT + F::UNDERLINE_OFFSET as u32
+        } else {
+            F::CHARACTER_SIZE.height
+        };
+
+        let size = Size::new(width, height);
 
         // Return a zero sized bounding box if the text is completely transparent.
         let bb_size = if self.text_color.is_some() || self.background_color.is_some() {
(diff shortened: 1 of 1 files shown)


bugadani:
cc @rfuest

rfuest:
The decorations should be included in the measurement and this was on my TODO list. Thanks for implementing this and adding the tests I was too lazy to add in the previous PR ;).

> Should `line_height()` also depend on the decoration?

I'm undecided about this one, but lines shouldn't overlap if underlining is enabled. There are at least two ways to implement this: One would be return a fixed line height that is large enough for text with or without underlining. And the other one is to only increase the line height if underlining is enabled.

bugadani:
> lines shouldn't overlap if underlining is enabled
> And the other one is to only increase the line height if underlining is enabled.

Consider the case where underline is enabled for a single word in a single line. If that single line would be 2px taller, I think I'd get a bunch of issues about it :) On the other hand, detecting a dynamically set underline before measurement requires a second parse of the text, which I'd definitely want to avoid.

Previously, e-t used a line_height same as today, and a line_spacing of 0 by default. This resulted in overdraw, but I think the 6x8 font looked the best this way.

`line_height` is (I think) the only value e-t can use to step in the y direction. I'd probably leave the default line spacing as 0, so I think I'll (with your approval) go for increasing what's returned by `line_height`.

rfuest:
OK, that makes sense, let's increase the line height to take the decorations into account. We might want to add a `LINE_HEIGHT` constant to the `MonoFont` trait and return that value in `line_height`, because the line height should be larger than `F::UNDERLINE_HEIGHT + F::UNDERLINE_OFFSET as u32` to make sure the underlining doesn't touch the characters on the next line.

> `line_height` is (I think) the only value e-t can use to step in the y direction. I'd probably leave the default line spacing as 0, so I think I'll (with your approval) go for increasing what's returned by `line_height`.

Why does e-t use line spacing instead of line height? I might want to add a `line_height` setting to `TextStyle` and it would be good if the settings were similar between e-t and `Text`. I prefer a line height over a line spacing setting because it's independent of the font and IMO more common.

bugadani:
Um, no reason at all. I can be persuaded to change that. But we are working with pixels and for each font the default line height is rather "random".

rfuest:
We could also add a new enum for the line height setting:
```rust
enum LineHeight {
    Default,
    Relative(i32),
    Absolute(u32),
}
```

bugadani:
Huh I wouldn't want to complicate this.

rfuest:
Can you update the branch so that we can merge this? We can implement the `line_height` changes in a later PR.

bugadani:
Github really should know if the branch is conflict-free.

rfuest (approved):
Thanks!

PR #535, opened 2021-01-29
Generate ASCII and Latin-1 font subsets

rfuest:
This PR adds the new monospaced fonts to e-g.

The conversion tool now also creates the necessary rust types for the fonts. I had first tried a macro based solution, but adding this to the conversion tool is more flexible and, for example, allowed me to inline the generated PNG files into the docs.

There are some loose ends, like an updated fonts overview table in the docs and correct strikethrough/underline parameters. But most of these depend on some other changes (e.g. in the BDF parser) and might be better left for another PR (and this one is already pretty long).

Should we keep the current naming scheme, which just prefixes the BDF file name with `Font` or change it for the bold and oblique variants? `Font6x13B` -> `Font6x13Bold`, `Font6x13O` -> `Font6x13Oblique`


src/mono_font/mod.rs
```diff
@@ -113,39 +113,26 @@
 //!
 //! | Type | Screenshot |
 //! |------|------------|
-//! | [`Font6x8`] | ![6x8 font spritemap screenshot](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/master/data/font6x8.png) |
-//! | [`Font6x12`] | ![6x12 font spritemap screenshot](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/master/data/font6x12.png) |
-//! | [`Font8x16`] | ![8x16 font spritemap screenshot](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/master/data/font8x16.png) |
-//! | [`Font12x16`] | ![12x16 font spritemap screenshot](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/master/data/font12x16.png) |
-//! | [`Font24x32`] | The 24x32 font is a pixel doubled version of the 12x16 font. |
+//! | `Font6x8` | ![6x8 font spritemap screenshot](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/master/data/font6x8.png) |
```
jamwaffles, line 116: Could you add a `//! TODO: Replace with generated fonts table` or similar comment to this doc? That way it's visible when grepping for `TODO`, as well as sticking out when people view the docs if we forget to update them before release.

If you're planning on changing this very soon anyway, you can probably ignore this comment.

jamwaffles (changes requested):
Not sure if this is finished, but I had a quick look through anyway. Don't forget a changelog item :)

> Should we keep the current naming scheme, which just prefixes the BDF file name with Font or change it for the bold and oblique variants? Font6x13B -> Font6x13Bold, Font6x13O -> Font6x13Oblique

I think it would be clearer to use the full names, so `Font6x13Bold`, `Font6x13Oblique`, etc. Also, if you don't have a strong preference, can you change `Oblique` to `Italic`?

src/mono_font/mod.rs
rfuest: I hope to get this done soon, but I've added the comment anyway and will also add it to the tracking issue.

rfuest:
> Not sure if this is finished, but I had a quick look through anyway. Don't forget a changelog item :)

I've added changelog entry and some notes in the migration guide.

> I think it would be clearer to use the full names, so `Font6x13Bold`, `Font6x13Oblique`, etc. Also, if you don't have a strong preference, can you change `Oblique` to `Italic`?

I agree that we should use the full name, but I want to try if `Font6x13<Bold>` would also be possible. But I would prefer to leave this for another PR. While oblique is more precise I wouldn't mind changing it to italic.

jamwaffles (approved):
Looks good! The inline images for each font are great.

`Font<Variant>` would make for a nice API. Hopefully you can get it working, otherwise `*Bold`, `*Italic` suffixes aren't so bad.

bugadani:
What happened to the old 6x8 font? My tests are sad to see that go, the 6x9 looks a bit different.

rfuest:
I had to adjust a lot of tests in e-g, so I now your pain ;). We have decided to only include fonts that are generated from a single source of BDF files. This gives us a more consistent set of fonts than the old fonts.

bugadani:
Sounds reasonable, although I must say I find the 6x9 a bit harder to read than the old 6x8.

rfuest:
The 6x10 font is pretty close to the old 6x8 font. The 6x9 font has a 1px smaller cap height.

PR #537, opened 2021-02-02
Rename fonts

rfuest:
This PR renames the bold and italic font variants.

I tried to use generics for  the font variants, but I don't think this made the API better. So I decided to keep the simple suffixes.


jamwaffles (approved):
LGTM

> I tried to use generics for the font variants, but I don't think this made the API better. So I decided to keep the simple suffixes.

I think the current list of fonts with variants is manageable from a docs perspective, so this is fine unless we decide to add a pile more fonts, which I think is unlikely?

rfuest:
> I think the current list of fonts with variants is manageable from a docs perspective, so this is fine unless we decide to add a pile more fonts, which I think is unlikely?

IMO the current set of builtin fonts provides a good variety to get started implementing something with e-g. All other fonts should be packaged as external crates, by us or third parties.

If we decide to implement the tooling to create BDF font crates, we could also extend it to create `MonoFont` crates for monospaced BDF files.

rfuest:
Thanks for the review!

jamwaffles:
> IMO the current set of builtin fonts provides a good variety to get started implementing something with e-g. All other fonts should be packaged as external crates, by us or third parties.

That was exactly my thinking too. Glad we're aligned :)

> If we decide to implement the tooling to create BDF font crates, we could also extend it to create MonoFont crates for monospaced BDF files.

Good point.

> Thanks for the review!

Always a pleasure :)

Issue #541, opened 2021-02-03
New core and e-g alpha release


rfuest:
IMO the core crate should be near the final state for 0.7. @jamwaffles do you think it's time for a new core version release and a new e-g alpha? We might need to release another core version before 0.7, but I would like to get rid of all the git dependencies.

jamwaffles:
Sounds good to me! As I see it, the new releases would be:

* Core 0.2.0 (because breaking changes)
* e-g 0.7.0-alpha.3

However:

1. We could release core 0.1.2 as I don't think anyone is using it yet
2. I assume we'd want to update e-g to point to core 0.2.0 before releasing it's new alpha, right?

rfuest:
> We could release core 0.1.2 as I don't think anyone is using it yet

I think we should increase the version number according to the semver standard (to 0.2.0) regardless if it's used or not.

> I assume we'd want to update e-g to point to core 0.2.0 before releasing it's new alpha, right?

:+1:

jamwaffles:
> I think we should increase the version number according to the semver standard (to 0.2.0) regardless if it's used or not.

I agree, I just wanted to see what your thoughts were.

I'll go and do a new 0.2.0 core release now.

jamwaffles:
0.2.0 and 0.7.0-alpha.3 are released

PR #542, opened 2021-02-03
Fix dependent version on core release

jamwaffles:

## PR description

Fixes `just release-core [ver]` command - it was previously not reading `[ver]` and would try to release the current version. I believe this is caused by it not being able to update the version in `e-g`, which `--dependent-version fix` allows it to do.

`cargo-release` is great... when it works :(

### DIFF
--- a/justfile
+++ b/justfile
@@ -128,8 +128,8 @@ push-ci-image:
 # -------
 # Release embedded-graphics-core
 release-core *args:
-    cargo release --workspace --exclude embedded-graphics {{args}}
+    cargo release --workspace --exclude embedded-graphics --dependent-version fix {{args}}
 
 # Release embedded-graphics
 release-e-g *args:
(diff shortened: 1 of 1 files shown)


jamwaffles:
We shouldn't need this for `e-g` releases as there's no dependency the other way but who knows, that might break too :grinning:

jamwaffles:
Scratch that. For some reason `--dependent-version fix` kicks `cargo-release` into doing something properly. This PR now contains the two fixed release commands I just used in #541 so they worked at least once.

rfuest (approved):
I still don't know much about the release process, so I'll just trust you that this is correct. Thanks for releasing the new crate versions!

jamwaffles:
It's _supposed_ to be as easy as `just release-core` but I've had so many problems with cargo-release over time :facepalm:

PR #543, opened 2021-02-04
Add generate-drawing-examples tool

rfuest:
This PR adds a new `generate-drawing-examples` crate to the tools directory. It is an replacement for the previous tool in the `simulator` crate, which currently causes some dependency problems. Because of this problems the tool currently uses e-g `0.7.0-alpha.2` instead of a later release.

The new tool should be easier to use, because examples can now be written as normal functions and the `{}` hack for newlines isn't necessary anymore. Comments inside the example code are now also preserved, which wasn't the case in the old tool, but I hadn't noticed this problem until now.


jamwaffles:
Will the dependency issues be resolved once embedded-graphics/simulator#21 is merged?

justfile
```diff
@@ -73,6 +76,26 @@ check-links: generate-docs
...
+# Generate drawing examples in the doc directory
+generate-drawing-examples:
+    cd tools/generate-drawing-examples && cargo run
+
+# Checks if drawing examples are up to date
+check-drawing-examples: generate-drawing-examples
+    git diff --quiet doc/ || ( \
+        echo "doc/ folder is not up to date" \
+        echo "Try running 'just generate-drawing-examles'." \
+        echo "If any images have changed, run just generate-drawing-examples-montage' to update the collage image too" \
```
jamwaffles, line 88:
```suggestion
        echo "If any images have changed, run 'just generate-drawing-examples-montage' to update the collage image too" \
```

jamwaffles (changes requested):
A tiny nit on an error message, otherwise LGTM. Very clever use of macros!

justfile
rfuest: Good catch! This suggestion also made me notice a typo in the line above.

rfuest:
> Will the dependency issues be resolved once [embedded-graphics/simulator#21](https://github.com/embedded-graphics/simulator/pull/21) is merged?

Yes, I think they should be resolved after that PR is merged. We should release a new simulator alpha after that PR.

jamwaffles (approved):
:+1:

PR #545, opened 2021-02-05
Use the same docker image for all build steps

rfuest:
`target_steps` was still using the old docker image. I've changed the `config.yml` to use YAML anchors, which reduces code duplication and prevents the steps to get out of sync in the future.


jamwaffles (approved):
Awesome, much tidier!

PR #546, opened 2021-02-05
Remove unused test function

jamwaffles:

## PR description

Removes a function that was giving a warning during compilation. I don't know if it's public for another reason, but removing it hasn't made anything fail so...


rfuest (approved):
I did leave this in because I might want to reuse this later. But the warning is kind of annoying, we can remove it now and I'll add it back in later, if I need it.

jamwaffles:
Ah that's why. I thought it might have been used in another crate or something.

PR #547, opened 2021-02-05
Migration guide tweaks

jamwaffles:

## PR description

Complete the migration guide.

I could definitely use a hand with this from @rfuest as I haven't been too closely involved with all the restructurings.

### DIFF
--- a/MIGRATING-0.6-0.7.md
+++ b/MIGRATING-0.6-0.7.md
@@ -142,12 +142,12 @@ All fonts are provided in `embedded_graphics::mono_font::[ascii|latin1]::[font n
   - Also available in **bold** (`Font9x18Bold`)
 - `Font10x20`
 
-Two character sets are now provided for each font:
+Two character sets are now provided for each font. The `ascii` set can be use to reduce memory consumption by the font data, at the cost of losing extra characters.
 
-- `embedded_graphics::mono_font::ascii` for the ASCII character codes `0x20` to `0x7f`.
-- `embedded_graphics::mono_font::latin1` for the character codes `0x20` to `0x7f`, `0xa0` to `0xff`.
+- `embedded_graphics::mono_font::ascii` provides the characters U+0020 to U+007F in the [Basic Latin code block](https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)), excluding control characters.
+- `embedded_graphics::mono_font::latin1` provides all all characters in the `ascii` variant with the addition of U+00A0 to U+00FF in the [Latin-1 Supplement code block](https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)) (excluding control characters).
 
-The `ascii` set can be use to reduce memory consumption by the font data, at the cost of losing extra characters.
+`ascii` has full coverage of the English language. For other languages, `latin1` has complete coverage for [the languages listed here](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Modern_languages_with_complete_coverage), and partial coverage for [these languages](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Languages_with_incomplete_coverage).
 
 ## For display driver authors
 
@@ -390,3 +390,5 @@ The collection of builtin fonts are now sourced from public domain BDF fonts in 
 - `fonts::Font24x32` -> `mono_font::[ascii|latin1]::Font10x20` (this is the largest font available, however the old `Font24x32` was just a scaling of `Font12x16`)
 
 To style fonts, use the `MonoTextStyle` struct. `TextStyle` still exists, but has been repurposed to provide a more general interface to text styling. This more closely mirrors the way primitives are built and styled.
+
+TODO: Discuss baseline font alignment, etc, as per GH comment.
--- a/MIGRATING-0.6-0.7.md
+++ b/MIGRATING-0.6-0.7.md
@@ -331,10 +331,13 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
 
 ## Style module
 
-The `style` module has been removed.
+The `style` module has been removed. The items in it have been moved:
 
-TODO: Improve this section before release.
+- `PrimitiveStyle` and `PrimitiveStyleBuilder` are now available under `embedded_graphics::primitives::{PrimitiveStyle, PrimitiveStyleBuilder}`.
+- `TextStyle` and `TextStyleBuilder` are now available under `embedded_graphics::text::{TextStyle, TextStyleBuilder}`.
+
+  Note that usage with `Text` has changed. Please see [the text changes section](#TODO) for more.
 
 - `PrimitiveStyle` & `PrimitiveStyleBuilder` -> `primitives` module
 - `TextStyle` & `TextStyleBuilder` -> `text` module + see text changes
-- `Styled` -> crate root
+- `Styled` is now exported from the crate root, e.g. `use embedded_graphics::Styled`.
--- a/MIGRATING-0.6-0.7.md
+++ b/MIGRATING-0.6-0.7.md
@@ -115,8 +115,6 @@ Support for external font renderers has been added. TODO: Expand
 - New default baseline is alphabetic
 - New fonts with `ascii` and `latin1` glyph subsets
 - `MonoFont` is now a struct instead of a trait
-- Added `TextStyle` to set the horizontal and vertical alignment for `Text` drawables
-- New default vertical alignment is baseline
 
 The list of fonts has changed to the following:
 
@@ -370,12 +368,10 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
 The `style` module has been removed. The items in it have been moved:
 
 - `PrimitiveStyle` and `PrimitiveStyleBuilder` are now available under `embedded_graphics::primitives::{PrimitiveStyle, PrimitiveStyleBuilder}`.
-- `TextStyle` and `TextStyleBuilder` are now available under `embedded_graphics::text::{TextStyle, TextStyleBuilder}`.
+- `TextStyle` and `TextStyleBuilder` were renamed are now available under `embedded_graphics::mono_font::{MonoTextStyle, MonoTextStyleBuilder}`.
 
   Note that usage with `Text` has changed. See [the text changes section](#Text-rendering) for more.
 
-- `PrimitiveStyle` & `PrimitiveStyleBuilder` -> `primitives` module
-- `TextStyle` & `TextStyleBuilder` -> `text` module + see text changes
 - `Styled` is now exported from the crate root, e.g. `use embedded_graphics::Styled`.
 
 ## Text and fonts
--- a/MIGRATING-0.6-0.7.md
+++ b/MIGRATING-0.6-0.7.md
@@ -372,14 +372,12 @@ The `style` module has been removed. The items in it have been moved:
 
   Note that usage with `Text` has changed. See [the text changes section](#Text-rendering) for more.
 
-- `Styled` is now exported from the crate root, e.g. `use embedded_graphics::Styled`.
-
 ## Text and fonts
 
 The collection of builtin fonts are now sourced from public domain BDF fonts in the XOrg project. Due to this, they have slightly different dimensions and glyphs and so have changed names. Some sizes are not the same in the new set, but a rough mapping is as follows:
 
 - `fonts::Font6x6` -> Removed
-- `fonts::Font6x8` -> `mono_font::[ascii|latin1]::Font5x8`
+- `fonts::Font6x8` -> `mono_font::[ascii|latin1]::Font6x10`. If a display performs better when using 8px high fonts, `mono_font::[ascii|latin1]::Font5x8` is suggested instead.
 - `fonts::Font6x12` -> `mono_font::[ascii|latin1]::Font6x12`
 - `fonts::Font8x16` -> `mono_font::[ascii|latin1]::Font8x13`
 - `fonts::Font12x16` -> `mono_font::[ascii|latin1]::Font9x15`
--- a/MIGRATING-0.6-0.7.md
+++ b/MIGRATING-0.6-0.7.md
@@ -10,7 +10,7 @@
     - [Geometry](#geometry)
     - [Color](#color)
     - [Sub draw targets](#sub-draw-targets)
-    - [Text rendering](#text-rendering)
+    - [Fonts and text](#fonts-and-text)
   - [For display driver authors](#for-display-driver-authors)
   - [For crates that handle images](#for-crates-that-handle-images)
   - [For text rendering crates](#for-text-rendering-crates)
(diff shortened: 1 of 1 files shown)


rfuest (commented):
Here is some initial feedback on this. I hope it makes sense, as I'm not in the correct state of mind to write docs at the moment.

MIGRATING-0.6-0.7.md
```diff
@@ -317,10 +350,24 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
+## Text and fonts
+
+The collection of builtin fonts are now sourced from public domain BDF fonts in the XOrg project. Due to this, they have slightly different dimensions and glyphs and so have changed names. Some sizes are not the same in the new set, but a rough mapping is as follows:
+
+- `fonts::Font6x6` -> Removed
+- `fonts::Font6x8` -> `mono_font::[ascii|latin1]::Font5x8`
+- `fonts::Font6x12` -> `mono_font::[ascii|latin1]::Font6x12`
+- `fonts::Font8x16` -> `mono_font::[ascii|latin1]::Font8x13`
+- `fonts::Font12x16` -> `mono_font::[ascii|latin1]::Font9x15`
+- `fonts::Font24x32` -> `mono_font::[ascii|latin1]::Font10x20` (this is the largest font available, however the old `Font24x32` was just a scaling of `Font12x16`)
```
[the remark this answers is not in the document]
jamwaffles: For the 6x8 example specifically, I decided to use one with the same 8 pixel height. Displays like the SSD1306 use one byte to represent a column of 8 pixels, so suggesting a font that continues to fit within that bound may be useful for drivers for that display, and others like it.

For the others, I went with fonts at the closest size as opposed to closest look so there's definitely room for improvement. Can you think of any changes off the top of your head?

jamwaffles:
> as I'm not in the correct state of mind to write docs at the moment.

I wasn't when I wrote them either! At any rate, I hope your foggy state of mind was only temporary :)

I've addressed some of the changes, so feel free to take a look at those, but I'll mark this for another proper review when I've expanded a couple more of the TODOs.

MIGRATING-0.6-0.7.md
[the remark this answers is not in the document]
rfuest: Interesting point. I would prefer to suggest `Font6x10` as the closest match and mention the possible performance impact on some displays and suggest the 8 px high font as an alternative.

> Can you think of any changes off the top of your head?

Not off the top of my head. I would need to compare the old and new PNG files to get the closest matches.

MIGRATING-0.6-0.7.md
```diff
@@ -98,17 +100,48 @@ The `DrawTargetExt` trait is introduced to allow a translated, cropped or clippe
...
+- `Font9x15`
+  - Also available in **bold** (`Font9x15Bold`)
+- `Font9x18`
+  - Also available in **bold** (`Font9x18Bold`)
+- `Font10x20`
+
+Two character sets are now provided for each font:
+
+- `embedded_graphics::mono_font::ascii` for the ASCII character codes `0x20` to `0x7f`.
+- `embedded_graphics::mono_font::latin1` for the character codes `0x20` to `0x7f`, `0xa0` to `0xff`.
```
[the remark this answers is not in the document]
bugadani: A bit of an off topic, but that lonely non-breaking space (`U+00A0`) won't get much love if the current status quo remains (e-g doesn't have word wrapping, so it's equivalent to a normal space, and e-t basically replaces it with a normal space for compatibility reasons).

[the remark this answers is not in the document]
rfuest: That's correct, the NBSP is only included to fill the gap in the character map so that all rows are 16 glyphs long.

jamwaffles:
Should we merge this as is and keep iterating in smaller/other PRs, close this PR completely, or work on the migration guide here? I'm happy to allow pushes to this branch from other people if we want to collaborate here.

I don't have the energy to write it all myself but will gladly help/review/copyedit wherever we end up doing the work.

rfuest:
Merging this as is and keep iterating in other PRs sounds good to me.

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +367,28 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
 
 ## Style module
 
-The `style` module has been removed.
+The `style` module has been removed. The items in it have been moved:
 
-TODO: Improve this section before release.
+- `PrimitiveStyle` and `PrimitiveStyleBuilder` are now available under `embedded_graphics::primitives::{PrimitiveStyle, PrimitiveStyleBuilder}`.
+- `TextStyle` and `TextStyleBuilder` are now available under `embedded_graphics::text::{TextStyle, TextStyleBuilder}`.
```
rfuest, line 373: The replacement for the old `TextStyle` is `MonoTextStyle`:

```suggestion
- `TextStyle` and `TextStyleBuilder` were renamed are now available under `embedded_graphics::mono_font::{MonoTextStyle, MonoTextStyleBuilder}`.
```

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +367,28 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
+
+  Note that usage with `Text` has changed. See [the text changes section](#Text-rendering) for more.
 
 - `PrimitiveStyle` & `PrimitiveStyleBuilder` -> `primitives` module
 - `TextStyle` & `TextStyleBuilder` -> `text` module + see text changes
```
rfuest, line 378:
```suggestion
```

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +367,28 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
-- `Styled` -> crate root
+- `Styled` is now exported from the crate root, e.g. `use embedded_graphics::Styled`.
```
rfuest, line 379: `Styled` is now only used for primitives and is included in the `primitives` module.

rfuest (changes requested):
I haven't reviewed the text related sections thoroughly, because they need to be tweaked later anyway.

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +365,24 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
+- `TextStyle` and `TextStyleBuilder` were renamed are now available under `embedded_graphics::mono_font::{MonoTextStyle, MonoTextStyleBuilder}`.
+
+  Note that usage with `Text` has changed. See [the text changes section](#Text-rendering) for more.
+
+## Text and fonts
+
+The collection of builtin fonts are now sourced from public domain BDF fonts in the XOrg project. Due to this, they have slightly different dimensions and glyphs and so have changed names. Some sizes are not the same in the new set, but a rough mapping is as follows:
+
+- `fonts::Font6x6` -> Removed
+- `fonts::Font6x8` -> `mono_font::[ascii|latin1]::Font6x10`. If a display performs better when using 8px high fonts, `mono_font::[ascii|latin1]::Font5x8` is suggested instead.
```
rfuest, line 380: I'm unsure if we need to mention this here. This is very specific to some displays and also assumes that the text is vertically aligned to the 8px page borders. And many glyphs in the new 6x10 font are less than 8 px high, so that there might not be a performance difference at all.

As an alternative I would suggest to mention that the new fonts use higher bitmaps for the same cap height below the substitution table. And that drawing performance might be lower if the text isn't properly aligned with page borders, but for details about this we should refer to the driver docs.

rfuest (changes requested):

MIGRATING-0.6-0.7.md
```diff
@@ -113,6 +116,37 @@ TODO: Improve this section before release.
...
+  - Also available in **bold** (`Font7x14`)
+- `Font8x13`
+  - Also available in **bold** (`Font8x13Bold`) and _italic_ (`Font8x13Italic`)
+- `Font9x15`
+  - Also available in **bold** (`Font9x15Bold`)
+- `Font9x18`
+  - Also available in **bold** (`Font9x18Bold`)
+- `Font10x20`
+
+Two character sets are now provided for each font. The `ascii` set can be use to reduce memory consumption by the font data, at the cost of losing extra characters.
```
bugadani, line 143: I think I would drop the "losing" part, instead of stating the pros of each, something along these lines:

```suggestion
Two character sets are now provided for each font. The `ascii` set has a reduced memory footprint, while `latin1` contains more characters.
```

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +365,24 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
+## Text and fonts
+
+The collection of builtin fonts are now sourced from public domain BDF fonts in the XOrg project. Due to this, they have slightly different dimensions and glyphs and so have changed names. Some sizes are not the same in the new set, but a rough mapping is as follows:
+
+- `fonts::Font6x6` -> Removed
+- `fonts::Font6x8` -> `mono_font::[ascii|latin1]::Font6x10`. If a display performs better when using 8px high fonts, `mono_font::[ascii|latin1]::Font5x8` is suggested instead.
+- `fonts::Font6x12` -> `mono_font::[ascii|latin1]::Font6x12`
+- `fonts::Font8x16` -> `mono_font::[ascii|latin1]::Font8x13`
+- `fonts::Font12x16` -> `mono_font::[ascii|latin1]::Font9x15`
+- `fonts::Font24x32` -> `mono_font::[ascii|latin1]::Font10x20` (this is the largest font available, however the old `Font24x32` was just a scaling of `Font12x16`)
```
bugadani, line 384: I'm not sold on the note. While true, "just a scaling" doesn't help if someone needs a bigger font. This also means the replacement isn't a good one (as rfuest's picture shows). I think we should just state that `fonts::Font24x32` has been removed.

bugadani (commented):
A few nits

jamwaffles:
Changes addressed. Not perfect, but I'd suggest merging this rather old PR and improving in new, smaller PRs.

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +365,30 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
rfuest, line 383: Data URLs are stripped by the GitHub markdown renderer. But these images also exist as PNG files in the `fonts` directory.

rfuest (changes requested):

rfuest:
> Changes addressed. Not perfect, but I'd suggest merging this rather old PR and improving in new, smaller PRs.

Should we open an issue to track what still needs to be added to the migration guide or should we rely on the TODOs in the guide itself?

MIGRATING-0.6-0.7.md
jamwaffles: Ah, I didn't realise we had the source bitmaps in the repo - I could only find the BDFs. I've changed the URLs to point to the images in the repo, pinned to the 0.7.0-beta.1 tag. Also makes the table a lot more managable.

jamwaffles:
I think TODOs in the guide are enough. We may want to add more of them as we fill out sections as well.

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +365,30 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
+
+| Old font                                                                                                                                                                             | Visually closest new font                                                                                                                                                                                      |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `fonts::Font6x6`<br>![Font6x6 glpyh bitmap](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/embedded-graphics-v0.6.2/embedded-graphics/data/font6x6.png)       | `mono_font::[ascii/latin1]::FONT_4X6`<br>![FONT_4X6 glyph bitmap](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/embedded-graphics-v0.7.0-beta.1/fonts/ascii/png/4x6.png)               |
```
rfuest, line 383: We use `[ascii|latin1]` above and here we use `[ascii/latin1]`. I'm not too concerned which one we use, but I think we should use the same throughout the document. Or we could switch to a syntax similar to a use statement: `mono_font::{ascii, latin1}::FONT_4X6` or `mono_font::{ascii | latin1}::FONT_4X6`.

MIGRATING-0.6-0.7.md
```diff
@@ -331,10 +365,30 @@ It now also disallows out of bounds drawing by default. This behaviour can be ch
...
+| `fonts::Font8x16`<br>![Font8x16 glpyh bitmap](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/embedded-graphics-v0.6.2/embedded-graphics/data/font8x16.png)    | `mono_font::[ascii/latin1]::FONT_9X15_BOLD`<br>![FONT_9X15_BOLD glyph bitmap](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/embedded-graphics-v0.7.0-beta.1/fonts/ascii/png/9x15B.png) |
+| `fonts::Font12x16`<br>![Font12x16 glpyh bitmap](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/embedded-graphics-v0.6.2/embedded-graphics/data/font12x16.png) | `mono_font::[ascii/latin1]::FONT_10X20`<br>![FONT_10X20 glyph bitmap](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/embedded-graphics-v0.7.0-beta.1/fonts/ascii/png/10x20.png)         |
+| `fonts::Font24x32`                                                                                                                                                                   | (removed)                                                                                                                                                                                                      |
```
rfuest, line 388: IMO we should provide some guidance what alternatives are available in this case:

```suggestion
| `fonts::Font24x32`                                                                                                                                                                   | The largest available new font is `FONT_10X20`, which is significantly smaller than the old `Font24x32`. Larger fonts are available in external crates: (Link to list with other fonts/renderers in the README)                                                                                                                                                                                                   |
```

rfuest (changes requested):

MIGRATING-0.6-0.7.md
jamwaffles: I changed this to work with Markdown tables - my autoformatter gets confused if I use a pipe. I've changed it to `{ascii, latin1}` which looks a lot better than either previous solution - thanks.

MIGRATING-0.6-0.7.md
jamwaffles: Good suggestion. I linked to the the external crates section for now, but that can be changed if/when we add the fonts list.

MIGRATING-0.6-0.7.md
```diff
@@ -113,6 +116,37 @@ TODO: Improve this section before release.
 - New fonts with `ascii` and `latin1` glyph subsets
 - `MonoFont` is now a struct instead of a trait
 
+The list of fonts has changed to the following:
+
+All fonts are provided in `embedded_graphics::mono_font::{ascii latin1}::[font name]`
```
rfuest, line 121:
```suggestion
All fonts are provided in `embedded_graphics::mono_font::{ascii, latin1}::[font name]`
```

rfuest (changes requested):
One last nit. And I didn't mentioned it yet, but I really like the font comparison table!

MIGRATING-0.6-0.7.md
jamwaffles: Argh, thanks for catching this crap. Fixed!

rfuest (approved):

jamwaffles:
> I really like the font comparison table!

Yeah, came out better than I was expecting! Thanks for the reviews!

PR #549, opened 2021-02-14
Improve performance of triangle contains test

rfuest:
The `primitives::triangle::test::contains` test was very inefficient and caused a significant increase in overall unit test runtime. The refactored test is about 70x faster on my system and the runtime of all unit tests is about 30x faster.


jamwaffles (approved):
Ah thank you! This has been bugging me for ages. I'm glad someone got round to it :)

PR #550, opened 2021-02-14
Don't check formatting in beta CI builds

rfuest:
CI did break in #549 because the output of `rustfmt` has changed in the latest Rust beta. I don't know if this is a temporary change in the latest beta or if all future `rustfmt` versions will have the same behavior. But I think the most robust solution is to only test the formatting with one Rust version.

This PR removes the formatting check from `precheck-beta`. I've also renamed `precheck-stable` to `precheck-msrv`, because that's more accurate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
 
   # Check that everything (tests, benches, etc) builds using the latest stable Rust version
   precheck-stable:
-    docker: &docker
+    docker: *docker
     steps:
       - run: rustup default stable
       - eg_init
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,10 +34,6 @@ jobs:
   # Check that everything (tests, benches, etc) builds using the latest stable Rust version
   precheck-stable:
     docker: &docker
-      - image: jamwaffles/circleci-embedded-graphics:1.40.0-3
-        auth:
-          username: jamwaffles
-          password: $DOCKERHUB_PASSWORD
     steps:
       - run: rustup default stable
       - eg_init
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,19 @@ jobs:
       - run: just build
       - eg_finish
 
+  # Check that everything (tests, benches, etc) builds using the latest stable Rust version
+  precheck-stable:
+    docker: &docker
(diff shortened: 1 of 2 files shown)


rfuest:
The renaming of `precheck-stable` to `precheck-msrv` will need require a change in the GH configuration for this repo. Let me know if you want to keep the new name or if I should revert it back to `stable`.

And the latest rust beta seems to have even more problems, because we now get an ICE. But regardless of the issues with this particular beta version, I still think that testing the formatting with only one version is a good idea.

rfuest:
The ICE we get with the current beta seems to be known problem, there are multiple open issues in the rust repo.

jamwaffles (approved):
Looks good, thanks. I think the job name change does make sense. I'll change the repo settings when this PR is merged.

jamwaffles:
Actually, I changed it just now so this PR can be merged. I also removed the required flag precheck-beta as it's mostly supposed to be a canary for upcoming breaking changes, but seeing as we pin to an older version anyway, this is less of an issue.

Should we add a job to build against current stable to make sure we haven't broken anything for the current version?

jamwaffles:
On second thoughts, maybe it would be a good idea to run the same build in beta as everywhere else, but allow it to fail? This ICE seems temporary, but any errors on the beta job shouldn't stop the rest of the build failing.

rfuest:
> On second thoughts, maybe it would be a good idea to run the same build in beta as everywhere else, but allow it to fail? This ICE seems temporary, but any errors on the beta job shouldn't stop the rest of the build failing.

The ICE will be temporary, but I'm not sure about the rustfmt output. It has been pretty pretty stable until now, but I don't know if there is any guarantee that the output is stable across versions. If there is a permanent change in the rustfmt output the beta build would always fail and this could mask other code problems. That's why I prefer to only test the formatting against a single rustfmt version. And we will notice any permanent changes to rustfmt as soon as PRs are opened which fail the formatting test.

I like the idea of making the beta job non required. And I think that adding a `precheck-stable` which actually uses the latest stable would be a good I idea.

rfuest:
I've added the `precheck-stable` job. Could you configure it as required?

jamwaffles:
You make a very good point about rustfmt, and checking it against our MSRV is the best way of doing it. I think this is ready to merge then, although I see a red merge button so do let me know if you're not allowed to merge it for some reason.

PR #552, opened 2021-02-15
Add output associated type to Drawable

### LINKED ISSUE #548: Add return type to Drawable::draw
`Drawable::draw` currently returns `Result<(), D::Error>` and doesn't allow to return any value. This is OK for most draw calls, but a return value would have an advantage for some.

Here is an example how this feature could be used to chain text drawing with different styles:
```rust
let next_point = Text::new("Label ", Point::new(10, 20)).into_styled(label_style).draw(&mut display)?;
Text::new("Value", next_point).into_styled(value_style).draw(&mut display)?;
```
By using the `next_point` value the size of the first text doesn't need to be calculated twice. This will only work for left aligned text, but for everything else we have e-t.

jamwaffles:

## PR description

Perhaps a very little interpretation of #548, but here's my go at a return type for `Drawable`.

Closes #548

### DIFF
--- a/MIGRATING-0.6-0.7.md
+++ b/MIGRATING-0.6-0.7.md
@@ -10,6 +10,7 @@
     - [Geometry](#geometry)
     - [Color](#color)
     - [Sub draw targets](#sub-draw-targets)
+    - [Text rendering](#text-rendering)
   - [For display driver authors](#for-display-driver-authors)
   - [For crates that handle images](#for-crates-that-handle-images)
   - [For text rendering crates](#for-text-rendering-crates)
@@ -24,7 +25,7 @@
     - [Triangle](#triangle)
   - [Geometry](#geometry-1)
   - [Mock display](#mock-display-1)
-  - [style module](#style-module)
+  - [Style module](#style-module)
 
 ## New features
 
@@ -162,6 +163,9 @@ To reduce duplication, please search the `DrawTarget` documentation on <https://
 
 The `Drawable` trait now uses an associated type for its pixel color instead of a type parameters.
 
+An associated type, `Output`, has also been added which can be used to return intermediate values
+from drawing operations. To maintain compatibility with 0.6, this can be set to `type Output = ();`
+
 ```diff
 - impl<'a, C: 'a> Drawable<C> for &Button<'a, C>
 - where
@@ -177,7 +181,9 @@ The `Drawable` trait now uses an associated type for its pixel color instead of 
 + {
 +     type Color = C;
 +
-+     fn draw<D>(&self, display: &mut D) -> Result<(), D::Error>
++     type Output = ();
++
++     fn draw<D>(&self, display: &mut D) -> Result<Self::Output, D::Error>
 +     where
 +         D: DrawTarget<Color = C>,
 +     {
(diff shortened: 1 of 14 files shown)


MIGRATING-0.6-0.7.md
```diff
@@ -162,6 +163,9 @@ To reduce duplication, please search the `DrawTarget` documentation on <https://
 
 The `Drawable` trait now uses an associated type for its pixel color instead of a type parameters.
 
+An associated type, `Output`, has also been added which can be used to return intermediate values
+from drawing operations. To maintain compatibility with 0.6, this can be set to `type Output = ();`
```
rfuest, line 167: Is "compatibility with 0.6" the right way to put this? The source code won't be compatible with e-g 0.6 anymore if the additional type is added. IMO it would be enough to mention that `()` can be used as the `Output` type if the `draw` method doesn't need to return any value.

core/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- **(breaking)** [#552](https://github.com/embedded-graphics/embedded-graphics/pull/552) Added the `Output` associated type to `Drawable` to return intermediate values from drawing operations.
```
rfuest, line 11: Should we specify that the returned value is an "intermediate value"? IMO a `Drawable` should be allowed to return any value that might be useful to the caller.

core/src/drawable.rs
```diff
@@ -65,8 +66,25 @@ pub trait Drawable {
     /// The pixel color type.
     type Color: PixelColor;
 
+    /// The return type of the [`draw`] method.
+    ///
+    /// The `Output` type can be used to return results and values produced from the drawing of the
+    /// current item. For example, rendering two differently styled text items next to each other
+    /// can make use of a returned value, allowing the next text to be positioned after the first:
+    ///
+    /// ```rust,ignore
```
rfuest, line 75: Should we add a TODO comment to remember to enable this when the `Text` `Drawable` impl is updated?

rfuest (changes requested):
Looks good, except for a few small doc nits.

jamwaffles:
All good points - changes made in the most recent commit.

MIGRATING-0.6-0.7.md
```diff
@@ -162,6 +163,10 @@ To reduce duplication, please search the `DrawTarget` documentation on <https://
 
 The `Drawable` trait now uses an associated type for its pixel color instead of a type parameters.
 
+An associated type, `Output`, has also been added which can be used to return intermediate values
+from drawing operations. The nil type `()` can be used if the `draw` method doesn't need to return
```
rfuest, line 167: I think it's called unit type in Rust.
```suggestion
from drawing operations. The unit type `()` can be used if the `draw` method doesn't need to return
```

MIGRATING-0.6-0.7.md
```diff
@@ -162,6 +163,10 @@ To reduce duplication, please search the `DrawTarget` documentation on <https://
 
 The `Drawable` trait now uses an associated type for its pixel color instead of a type parameters.
 
+An associated type, `Output`, has also been added which can be used to return intermediate values
```
rfuest, line 166:
```suggestion
An associated type, `Output`, has also been added which can be used to return values
```

rfuest (changes requested):

rfuest (approved):
LGTM

jamwaffles:
Thanks!

bugadani:
OMG I think this just made it possible to support flowing text into different textboxes. Need to think about it some more but the possibilities...

PR #555, opened 2021-02-26
Enable generation of fonts with variable bounding boxes

twistedfall:
Hi,

I needed to have some fancier fonts for my project so I tried to use the `bdf-to-mono` helper utility to convert them. It lacked support for the variable bounding box fonts though so I added it. With these changes I was able to successfully convert [Liberation Mono](https://github.com/liberationfonts/liberation-fonts), [Linux Libertine Mono](https://sourceforge.net/projects/linuxlibertine/) and [Digital-7](https://www.1001fonts.com/digital-7-font.html). The code modifications don't affect the already existing built-in fonts. Conversion of the TTF fonts to BDF were done following [this guide](https://learn.adafruit.com/custom-fonts-for-pyportal-circuitpython-display/conversion).


rfuest:
Thanks for your contribution.

The `bdf-to-mono` tool was only intended for monospaced fonts, because the internal font renderer doesn't support proportional fonts. https://github.com/embedded-graphics/embedded-bdf will render these fancier fonts much better in the future. It's not quite finished yet, but you could give it a try: https://github.com/embedded-graphics/embedded-bdf/blob/master/embedded-bdf/examples/hello.rs There is also the WIP https://github.com/embedded-graphics/eg-seven-segment, which might be a replacement for the Digital-7 font.

But a more flexible BDF support in the `bdf-to-mono` tool could be useful anyway and I'll review the PR soon.

twistedfall:
Well, those fonts I mentioned *are* monospaced, it's just that their bounding boxes are not set in stone. While they all have the same size, their offsets differ.

rfuest:
Sorry, I was in a hurry when I wrote the last comment and didn't check the linked fonts. I did only look at the "Digital-7" font and had missed that there is also a "Digital-7 Mono" variant.

Could you upload the BDF files you have conveted, so I can try your code without having to convert the fonts myself?

twistedfall:
Here they are: https://gist.github.com/twistedfall/452c4d8157580bacc6d140913bf84f6a (there are 2 font files there, search the page for "Libertine" to jump to the second font)

rfuest:
Thanks for the test files. Looks good so far. There are some tweaks I would like to make to the code, but considering the early development stage of `bdf-to-mono` I think these can be implemented later.

If you rebase the code to the latest master version I can merge this.

twistedfall:
What do you have in mind? If it isn’t something super-heavy I might as well implement it too.

rfuest:
We should take the `DWIDTH` value into account. If all glyph bounding boxes are smaller than `DWIDTH` the `CHARACTER_SPACING` in the `MonoFont` impl should be set to `DWIDTH - (width of the BB)` to maintain the correct kerning. But this would make more sense if `bdf-to-mono` would also generate the necessary Rust code and not only the raw image data. 

I'm planning to move some features, like PNG and Rust code generation, from the `convert-fonts` tool to `bdf-to-mono`. `bdf-to-mono` is a quick hack in it's current form and should get a better command line interface to make it easier to use.

twistedfall:
I've rebased the branch on master and added the support for calculating the character offset. I'm not 100% sure if that's what you had in mind so please do review.

tools/bdf-to-mono/src/lib.rs
```diff
@@ -34,32 +28,106 @@ pub fn bdf_to_bitmap(font: &BdfFont, encoding: Encoding) -> Result<Bitmap> {
...
-        if let Some(glyph) = font.glyphs.iter().find(|glyph| glyph.encoding == Some(c)) {
-            ensure!(
-                glyph.bounding_box == glyph_bb,
-                "glyph '{}' has a different bounding",
-                c
-            );
+    let fallback_glyph = font
+        .glyphs
+        .get('?')
+        .ok_or_else(|| anyhow!("font doesn't contain the fallback '?' glyph"))?;
```
rfuest, line 99: I would move this into the `map` function for `selected_glyphs`. This way the question mark will be included in the bounding box calculation, if it is used. This won't make a difference at the moment, because all subsets include the `?` anyway, but if we decide other subsets in the future it could make a difference.

rfuest (changes requested):
> I'm not 100% sure if that's what you had in mind so please do review.

It's exactly what I had in mind, thanks again!  I have one suggestion regarding the bounding box calculation, but it looks good otherwise.

tools/bdf-to-mono/src/lib.rs
twistedfall: Done!

tools/bdf-to-mono/src/lib.rs
```diff
@@ -34,32 +27,108 @@ pub fn bdf_to_bitmap(font: &BdfFont, encoding: Encoding) -> Result<Bitmap> {
...
+    // to maintain the correct kerning.
+    let mut char_spacing = None;
+
+    // find the maximum bounding box that can fit all glyphs taking offsets into account
+    // glyph_bb.offset.y will contain the baseline
+    // additionally build the `selected_glyphs` vector to save on the glyph lookup later in the code
+    let selected_glyphs = chars
+        .into_iter()
+        .enumerate()
+        .map(|(index, c)| {
```
rfuest, line 49: Here is an untested snippet how I would implement this:
```suggestion
    let glyph = chars
        .iter()
        .map(|c| {
            font.glyphs.get(c)
            .or_else(|| font.glyphs.get('?'))
            .ok_or_else(|| anyhow!("font doesn't contain the '{}' glyph and no fallback '?' glyph", c));
        })
        .collect::<Result<Vec<_>, _>>()?;
        
   /// TODO: calculate BB
```

IMO the code is a little easier to understand if you split the generation of the `selected_glyphs` vector and the bounding box calculation.

This would also replace missing `DEL` glyphs with the fallback glyph, which is a different behavior than before, but IMO this is also a valid replacement.

rfuest (changes requested):
I think I didn't explain this very well. If the fallback glyphs would already be included in the `selected_glyphs` vector the rest of the method wouldn't need to deal with fallback glyphs anymore.

tools/bdf-to-mono/src/lib.rs
twistedfall: Got it, fixed. Now, with the different handling of missing `DEL` char, the built-it fonts have changed (there is now the question mark in place of `DEL`). I'm not sure whether I should include the new .bin and .png files or not.

rfuest: > Got it, fixed.

Looks good, thanks for your patience with the comments.

> I'm not sure whether I should include the new .bin and .png files or not.

They should always reflect the current state of the font conversion tooling and therefore should be included in this PR.

twistedfall: I've updated the built-in fonts

rfuest (approved):

rfuest:
Thanks!

PR #556, opened 2021-02-27
CharacterStyle: Add Clone supertrait

bugadani:
This change would ensure that all text renderers can be used with embedded-text.

e-t requires `CharacterStyle: Clone` because ANSI escape sequence support can change various style options during rendering, where otherwise the style is only accessable through a shared (immutable) reference. While current renderers are all `Clone`, it's possible to create new ones that aren't, which would in turn be incompatible with embedded-text.


rfuest (approved):
LGTM

PR #561, opened 2021-03-04
Rename text alignment settings

rfuest:
This PR changes the names of the `HorizontalAlignment` and `VerticalAlignment` enums to `Alignment` and `Baseline`. The reason for this change is that `VerticalAlignment` might be misleading, because the vertical alignment does only apply to the first line of multiline text:
Bildschirmfoto von 2021-03-04 18-19-46
The text on the left is the actual result for `Baseline::Middle` and the right is an example of what users might expect for `VerticalAlignment::Center`. This also matches the names used in the HTML5 canvas API: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline, https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textAlign

The PR also included some tweaks to the `TextRenderer` trait. I've removed the `vertical_offset` method, which was intended as a performance optimization. The potential grain of saving a few additions and subtractions  per text object isn't worth having an more error prone API.

### DIFF
--- a/core/src/text/mod.rs
+++ b/core/src/text/mod.rs
@@ -18,10 +18,6 @@ pub trait TextRenderer {
 
     /// Draws a string.
     ///
-    /// The interpretation of the y coordinate of `position` is dependent on the implementation and
-    /// can, for example, be the top edge of the bounding box or a point on the baseline. The
-    /// caller must ensure that the coordinate is first converted with the `vertical_offset` method.
-    ///
     /// The method returns the start position of the next character to allow chaining of multiple
     /// draw calls.
     ///
@@ -34,6 +30,7 @@ pub trait TextRenderer {
         &self,
         text: &str,
         position: Point,
+        baseline: Baseline,
         target: &mut D,
     ) -> Result<Point, D::Error>
     where
@@ -41,16 +38,13 @@ pub trait TextRenderer {
 
     /// Draws whitespace of the given width.
     ///
-    /// The interpretation of the y coordinate of `position` is dependent on the implementation and
-    /// can, for example, be the top edge of the bounding box or a point on the baseline. The
-    /// caller must ensure that the coordinate is first converted with the `vertical_offset` method.
-    ///
     /// The method returns the start position of the next character to allow chaining of multiple
     /// draw calls.
     fn draw_whitespace<D>(
         &self,
         width: u32,
         position: Point,
+        baseline: Baseline,
         target: &mut D,
     ) -> Result<Point, D::Error>
     where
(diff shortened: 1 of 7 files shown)


src/text/text.rs
```diff
@@ -423,6 +424,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
```
rfuest, line 427: Ignored because of #562

jamwaffles (approved):
LGTM, and the reasoning makes sense.

I haven't been following the text changes too closely, but does this need a breaking changelog entry for any behavioural changes since the last alpha?

rfuest:
> I haven't been following the text changes too closely, but does this need a breaking changelog entry for any behavioural changes since the last alpha?

The behavior shouldn't have changes, but I've added changelog entries for the naming changes and the changed text renderer API.

core/src/text/mod.rs
```diff
@@ -95,22 +85,22 @@ pub struct TextMetrics {
...
     Top,
     /// Bottom.
     Bottom,
-    /// Center.
-    Center,
-    /// Baseline.
-    Baseline,
+    /// Middle.
+    Middle,
+    /// Alphabetic baseline.
```
bugadani, line 97: I think these could/should be explained better than just echoing the variant names.

For example I'm not entirely sure what `Alphabetic` refers to, although I have a guess. Also, does Top refer to the topmost pixel of the character, or something else? Perhaps we could adapt the text from the linked canvas docs.

src/text/text.rs
```diff
@@ -565,27 +570,27 @@ mod tests {
...
+            .baseline(Baseline::Alphabetic)
             .build();
 
         Text::new("t", Point::new(0, 8))
             .into_styled(style_top)
             .draw(&mut display)
             .unwrap();
-        Text::new("c", Point::new(6, 8))
-            .into_styled(style_center)
+        Text::new("m", Point::new(6, 8))
```
bugadani, line 592: Nit: I'm not sure if changing "c" to "m" has a purpose. Since otherwise it would just inflate the diff, I'm assuming it has, I just can't figure out what. Could you please explain? :)

bugadani (commented):
Just a few nits. Otherwise, looks good.

src/text/text.rs
rfuest: It was on purpose: `VerticalAlignment::Center` was renamed to `Baseline::Middle`.

bugadani: I did not at all manage to associate the letters with the alignments, lol.

core/src/text/mod.rs
rfuest: The docs need some work in general, I'll add a note to the tracking issue to revisit this later.

> For example I'm not entirely sure what Alphabetic refers to, although I have a guess.

The main reason I changed it from `Baseline` to `Alphabetic` is that `Baseline::Baseline` would look weird. But it also matches the canvas API. I don't think we will ever support ideographic or hanging baselines in the internal text renderer.

> Also, does Top refer to the topmost pixel of the character, or something else?

It refers to the top of the glyph bounding box. This can include additional blank pixels above the caps if the font includes them.

bugadani: I mean, I know what these are, but it's not obvious from the docs :)

rfuest: > I know what these are, ... ,but it's not obvious from the docs :)

The docs for the font/text stuff really need some work. Contributions to the docs are always welcome :wink:

PR #563, opened 2021-03-05
Always take stroke_width into account for bounding boxes

### LINKED ISSUE #562: Changing the behavior of Dimensions for transparent drawables
Implementations of the `Dimensions` trait for drawables are currently expected to return the smallest bounding box that contains all drawn pixels. This means that transparent drawables are expected to return a zero sized rectangle.

But this might not be the desirable behavior if you want to layout drawables by using the result of `Dimensions::bounding_box`. An application could toggle the visibility of an element setting the colors to transparent without affecting the layout, if we always return the same bounding box.

The changed `Dimensions` trait would also be easier to implement for `Text` drawables, which have never returned the correct bounding box. `Text::new("    ", ...)...` would need to return an empty bounding box (if the background is transparent), which would be hard to add to the text renderer API.

And I think this would make `Dimensions` trait more consistent in general, because it's also used for `DrawTarget`s, where it always returns the same bounding box, regardless of the display content.

rfuest:
This  PR changes the way `stroke_width` is handled if `stroke_color` is `None`. The bounding boxes that are returned by the `Dimensions` trait will ignore `stroke_color`. The `fill_area` will also be reduced regardless of `stoke_color`. See #562 for more details.

Closes #562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,18 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- **(breaking)** [#552](https://github.com/embedded-graphics/embedded-graphics/pull/552) Added the `Output` associated type to `Drawable` to allow returning non-`()` values from drawing operations.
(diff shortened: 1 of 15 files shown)


jamwaffles (approved):
There's a conflict in `CHANGELOG.md`, otherwise lgtm.

PR #565, opened 2021-03-06
Fix drawing examples and add png-target crate

rfuest:
This PR contains miscellaneous improvements for the tool crates:
* The drawing examples were outdated and not compatible with the latest e-g version.
* `generate-drawing-examples` and `convert-fonts` generated PNG files using different methods. I've added a new `png-target` crate to the tools directory to share code in both crates. This has the additional advantage that the tools no longer need to depend on the simulator, which reduces the change of problems with cyclic dependencies.
* `bdf-to-mono` was broken by recent changes in `bdf-parser`.
* The order of fonts in the `convert-fonts` output wasn't reproducible, because it was dependent on the order returned by `read_dir`, which is OS and file system dependent according to the Rust docs. The file list is now sorted before it is used, which should fix this problem.


jamwaffles (approved):
LGTM

PR #566, opened 2021-03-07
Add return value to Drawable impl for texts

rfuest:
This PR adds a return value to the `Drawable` impl for styled texts, which makes it easier to combine texts with different styles.


jamwaffles (approved):
Looks good! A small but quite impactful change I think :)

PR #568, opened 2021-03-07
Replace anonymous magic constants with named equivalents in tests

regexident:

## PR description

Since we have `MAX_R`/`MAX_G`/`MAX_B` defined on `RgbColor` we might as well use them, rather than equivalent anonymous magic constants. We already do this in all other tests in the same file.


jamwaffles (approved):
Thanks for the fix!

PR #569, opened 2021-03-08
Add line height to TextStyle

### LINKED ISSUE #564: Remove text rendering from core
The support for external text renderers is still pretty new and IMO more likely to break in the future than the other parts of the core crate.

Should we remove the text rendering API from core for the 0.7 release? I think that keeping display drivers compatible is more important, because this affects a lot more crates by multiple authors.

Adding the text renderer API back into core would be a non breaking change. So that we could move the API back into core when we release e-g 0.8 or even earlier if we notice that the API was already perfect (very unlikely).

rfuest:
This PR adds a `line_height` field to `TextStyle`. I've not included an extension trait to construct `LineHeight`s to this PR, because I want to wait until https://github.com/embedded-graphics/embedded-text/issues/100 is resolved.

I've also removed the text rendering API from the core crate and moved into a separate submodule. This makes the docs clearer, because the regular text API and the text renderer API aren't mixed in a single module anymore.

Closes #564


src/text/mod.rs
```diff
@@ -80,6 +80,45 @@ impl<C> DecorationColor<C> {
...
+    /// ```
+    /// use embedded_graphics::text::LineHeight;
+    ///
+    /// let relative_height = LineHeight::Percent(150);
+    /// assert_eq!(relative_height.to_absolute(20), 30);
+    /// ```
+    pub fn to_absolute(self, base_line_height: u32) -> u32 {
+        match self {
+            Self::Pixels(px) => px,
+            Self::Percent(100) => base_line_height,
```
bugadani, line 110: I'm quite sure this special case is not necessary.

bugadani (approved):
One small nit. I agree with not including the extension traits yet - it's always simpler to add them later than to remove them if they prove inconvenient.

src/text/mod.rs
rfuest: That was a weird performance optimization, but `to_absolute` is called so infrequently that this probably doesn't make any sense. I'll remove it.

jamwaffles (approved):
Apologies for the delay again. LGTM though.

PR #571, opened 2021-03-13
Use scanlines to draw rounded shapes

rfuest:
This PR changes the `Drawable` impls for `Circle`, `Ellipse` and `RoundedRectangle` to use scanlines for improved drawing performance.


jamwaffles (approved):
Sorry for the delay! This all looks good to me. Thanks for crossing another item off the imaginary todo list.

Does this PR close #491 or are there still more to add?

PR #572, opened 2021-03-25
Optimitzed raw data iterators

rfuest:
This PR improves the performance of the raw data iterators, especially for drawing sub images.

### DIFF
--- a/src/image/image_raw.rs
+++ b/src/image/image_raw.rs
@@ -4,8 +4,9 @@ use crate::{
     draw_target::{DrawTarget, DrawTargetExt},
     geometry::{Dimensions, OriginDimensions, Size},
     image::ImageDrawable,
+    iterator::raw::RawDataIter,
     pixelcolor::{
-        raw::{BigEndian, ByteOrder, LittleEndian, RawData, RawDataIter},
+        raw::{BigEndian, ByteOrder, LittleEndian, RawData},
         PixelColor,
     },
 };
--- a/src/image/image_raw.rs
+++ b/src/image/image_raw.rs
@@ -136,7 +136,9 @@ where
     /// Creates a new image.
     ///
     /// Only the width of the image needs to be specified. The height of the image will be
-    /// calculated based on the length of the given image data.
+    /// calculated based on the length of the given image data. If the length of the image data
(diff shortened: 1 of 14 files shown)


src/image/image_raw.rs
```diff
@@ -131,48 +135,119 @@ where
...
+        if width == 0 {
+            return Self {
+                data: &[],
+                size: Size::zero(),
+                pixel_type: PhantomData,
+                byte_order: PhantomData,
+            };
+        }
+
+        let height = data.len() / bytes_per_row(width, C::Raw::BITS_PER_PIXEL);
```
jamwaffles, line 151: If a user gives data that would result in a partial final row, how should we handle this? For example, this test fails:

```rust
#[test]
fn partial_row() {
    let data = [
        0xAA, 0x00, //
        0x55, 0xFF, //
        0xAA, 0x80, //
        0xBB,
    ];
    let image_data: ImageRaw<BinaryColor> = ImageRaw::new(&data, 9);

    assert_eq!(image_data.size(), Size::new(9, 4));
}
```

With a height of `3` but if I got my logic right it should indeed have a height of `4` (if we want to handle partial rows). If the desired behaviour is to ignore that final row, we should mention that in the docs I think.

jamwaffles (commented):
Looks good to me, aside from a behaviour clarification I left a comment about.

src/image/image_raw.rs
rfuest: Good point. I'm not entirely sure what the best behavior is in this case. I chose to ignore any partial rows, because it would make sure that the iterator always outputs a full rectangle. But perhaps drawing an incomplete image would be easier to debug for the user.

I've added a note to describe the current behavior. But if you like I could also try to change it to draw incomplete rows instead.

jamwaffles: I'm also unsure, so let's leave it as is with the note. I think it's easier to "fix" this behaviour later in a point release than it would be to remove partial row rendering if for some reason that ends up being the wrong approach.

jamwaffles (approved):

src/image/image_raw.rs
rfuest: OK. Thanks for the review.

PR #574, opened 2021-03-31
Inline drawing examples images in docs

### LINKED ISSUE #421: Versioning of linked drawing examples
The docs of `lib.rs` includes a link to `drawing-examples.md` in the master branch of the repository. This will cause problems in the future if we introduce breaking API changes, because all versions on docs.rs would link to the latest version of the markdown document. IMO it's important to keep the older docs intact and link to the correct version of the markdown file.

I'm not sure if there is a good way to automatically link to a specific commit or if we can handle this by using branches for each major version.

rfuest:
This PR changes the `generate-drawing-examples` tool to generate a Rust file instead of Markdown + PNG files. By inlining the images we no longer have external links that could get out of sync.

One remaining problem is the links to the examples from `lib.rs` and `README.md`. Both are generated form the same source, but the docs should contain a local link to the `drawing_examples` module docs and the readme needs to contains a link to GH or docs.rs. Any suggestions?

Closes #421


jamwaffles (commented):
> Both are generated form the same source, but the docs should contain a local link to the drawing_examples module docs and the readme needs to contains a link to GH or docs.rs.

I can't think of a clean way to make that link work in both environments, so IMO we should go for a link to docs.rs/GH. This benefits users of the crate whilst causing a mild annoyance for devs, but I think that's the right balance.

I also have a small nit: should we call the module `examples` instead, and link to it from further up the docs in `lib.rs`. These are both changes that might make the drawing examples more discoverable.

rfuest:
> I can't think of a clean way to make that link work in both environments, so IMO we should go for a link to docs.rs/GH. This benefits users of the crate whilst causing a mild annoyance for devs, but I think that's the right balance.

IMO this is a bigger problem for users than it is for devs. If a user is looking at the docs for an older version of e-g on docs.rs the examples links will always point to the latest release. This could get confusing because the user might not notice that they are now looking at another version.

I've just updated the PR to try another approach. Instead of removing links using `sed` the `generate-readmes` recipe now converts local links to links to docs.rs. Do you think this is better? One disadvantage is that some links might not work, if we add features to the master version that are not included in the latest version on docs.rs (like the `Arcs` links in the primitives list). But we can retain more of the links in general.

core/README.md
```diff
@@ -25,29 +25,42 @@ version of embedded-graphics-core may be used for multiple major versions of emb
...
 ## Images
 
-The `ImageDrawable` trait should be implemented for any image or image-like item, for example
+The [`ImageDrawable`] trait should be implemented for any image or image-like item, for example
 a spritemap.
 
 [embedded-graphics]: https://docs.rs/embedded-graphics
+[`Pixel`]: https://docs.rs/embedded-graphics-core/latest/embedded_graphics_core/drawable/struct.Pixel.html
+[`Point`]:  ./geometry/struct.Point.html
+[`Size`]:  ./geometry/struct.Size.html
```
jamwaffles, line 53: These should be modified too right?

core/README.md
```diff
@@ -25,29 +25,42 @@ version of embedded-graphics-core may be used for multiple major versions of emb
...
+[`Drawable`]: https://docs.rs/embedded-graphics-core/latest/embedded_graphics_core/drawable/trait.Drawable.html
+[`DrawTarget`]: https://docs.rs/embedded-graphics-core/latest/embedded_graphics_core/draw_target/trait.DrawTarget.html
+[`Rectangle`]: https://docs.rs/embedded-graphics-core/latest/embedded_graphics_core/primitives/rectangle/struct.Rectangle.html
+[`Dimensions`]:  ./geometry/trait.Dimensions.html
```
jamwaffles, line 57: Same here

jamwaffles (changes requested):
That seems like a better approach for sure. I left some comments on some links that look like they weren't picked up by `sed`.

core/README.md
rfuest: Good catch. The regex wasn't accepting more than one space in front of the link. I've made the regex less strict and updated `core/lib.rs`.

jamwaffles (approved):

PR #575, opened 2021-04-02
Make MonoFont fields public

rfuest:
The changes in PR #573 made the parameters of `MonoFont`s inaccessible outside the e-g crate. But some crates need access to this information (e-t uses the values in tests). This PR makes the fields public and adds a `non_exhaustive` attribute to `MonoFont`.

### DIFF
--- a/src/mono_font/mod.rs
+++ b/src/mono_font/mod.rs
@@ -145,33 +145,43 @@ use crate::{
 };
 
 /// Monospaced bitmap font.
+///
+/// See the [module documentation] for more information about using fonts.
+///
+/// # Implementation notes
+///
+/// Use [`MonoFontBuilder`] to create new font objects.
+///
+/// [module documentation]: index.html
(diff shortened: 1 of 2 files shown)


jamwaffles (approved):
On my phone atm so I may have missed something obvious, but does DecorationDimensions need a changelog entry? I can't check if it's lib-pub or mod-pub right now but thought I'd mention it just in case it's the former.

rfuest:
`Decoration` was added as a private struct a few commits ago and I had to make it lib-pub now because it is used in a public field in `MonoFont`. 

I'm unsure if this needs a changelog entry. While it is an addition to the public API this type will never be used on its own and the `MonoFont` changes are already mentioned in the changelog. Do you think we should include these small changes or would adding these increase the noise in the changelog unnecessarily?

jamwaffles:
Thanks for clarifying. Nah I think it's ok to leave it as is, so feel free to merge when ready.

PR #576, opened 2021-04-02
Add reset methods to builders

### LINKED ISSUE #554: Add methods to reset settings in builders
`MonoTextStyleBuilder` doesn't allow to reset `text_color`, `background_color`, `underline_color` or `strikethrough_color` to a transparent color. This isn't a problem when the builder is used to create a new style, but doesn't allow the builder to be used to modify an existing style: `MonoTextStyleBuilder::from(&old_style).text_color(...)`.

I didn't have time to check the other builders, but there will probably be other instances which are also affected by a similar problem. We should make sure that the all builders provide methods to generate any object regardless of the initial values. And the API should be similar for all builders.

rfuest:
This PR adds the reset methods to builders, that were discussed in #554.

Closes #554


jamwaffles (changes requested):
Looks good, although it's missing a changelog entry.

rfuest:
OK, I've added the missing changelog entry.

jamwaffles (approved):
Thanks! Good to go :)

PR #577, opened 2021-04-02
Use bounding_box instead of Rectangle::new

### LINKED ISSUE #448: Implement From<Size> for Rectangle
Should we implement `From<Size> for Rectangle`? This would be equivalent to calling `Rectangle::new(Point::zero(), size)`, which seems to be a common operation (used 40 times in e-g according to a quick grep).

Or do you think the original version is easier to understand and the additional code verbosity doesn't matter?

rfuest:
This PR changes some `Rectangle::new(Point::zero(), ...)` expressions to use `bounding_box` instead, which makes the intent of these statements clearer.

These changes are a replacement for #448. While not all instances of `Rectangle::new(Point::zero(), ...)` can be replaced by `bounding_box` some of the other code is clearer without using `Size::to_rectangle`.

Closes #448

### DIFF
--- a/core/src/draw_target/mod.rs
+++ b/core/src/draw_target/mod.rs
@@ -203,7 +203,7 @@ use crate::{
 ///         // Clamp the rectangle coordinates to the valid range by determining
 ///         // the intersection of the fill area and the visible display area
 ///         // by using Rectangle::intersection.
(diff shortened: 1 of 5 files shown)


jamwaffles (approved):
LGTM :)

Issue #499, opened 2020-11-22
Changing the implementation of styleds


rfuest:
To implement a new styled drawable we currently implement, for example, `impl <C: PixelColor> Drawable for Styled<Rectangle, PrimitiveStyle<C>>`. This does only work inside the main e-g crate and makes it impossible to resuse the `Styled` struct in other crates. I think e-t implements its own `Styled`.

I did implement a prototype a while ago, where I added a new `StyledDrawable` trait, which roughly looked like this:
```rust
pub trait StyledDrawable<S> {
    type Color: PixelColor;

    fn draw_styled<D>(&self, target: &mut D, style: &S) -> Result<(), D::Error>
    where
        D: DrawTarget<Color = Self::Color>;
}
```
This makes it possible to implement new styleds in other crates. And I think this change is even non breaking, because `Styled` can be change to use the new `StyledDrawable` impls.

Now to the semi-breaking part of this suggestion: The new `draw_styled` method could also be used directly:
```rust
let style = PrimitiveStyle::with_stroke(BinaryColor::On, 1);

// before:
Line::new(start, end)
    .into_styled(style)
    .draw(target)?;

// after:
Line::new(start, end)
    .draw_styled(target, &style)?;
```
Would we want to advice users to use `draw_styled` instead of the `into_styled` & `draw` combination, if we implement this change? The old code would still work and could be useful to store a styled primitive for a longer time. But for the short lived object, that are often used in e-g code, the new variant might be easier to use. And there might be a slight performance improvement, because the style is passed using a reference and doesn't need to be cloned into a `Styled` object first.

jamwaffles:
> I did implement a prototype a while ago, where I added a new StyledDrawable trait, which roughly looked like this:

This would mean there are two different ways to do mostly the same thing, with about the same lines of code. This would be more complex to document and maintain and I'm not sure I see the benefit.

I've typed out multiple half replies to the rest of the post now and nothing has become much clearer, so could use some concrete examples of how this breaking change would fix some problems.

bugadani:
> I think e-t implements its own Styled.

[...]

PR #578, opened 2021-04-06
Add support for target specific drawables and text renderers

rfuest:
This PR adds support for target specific drawables and text renderers. These can be used to implement accelerated drawing and text rendering functions, which aren't implementable with the default `DrawTarget` trait.

One example use case for a target specific drawable would be to add support for `Draw Line` commands for displays with a `SSD1331` controller:
```rust
// Stripped down version of `PrimitiveStyle`, which only includes settings that are supported by the `Draw Line` command.
struct SsdLineStyle {
    stroke_color: Rgb565,
}

impl <SPI, DC> TargetSpecificDrawable<Ssd1331<SPI, DC>> for Styled<Line, SsdLineStyle> { ... }

// Using this accelerated line drawing is very similar to regular e-g code:

let style = SsdLineStyle { stroke_color: Rgb565::RED };

Line::new(Point::new(1, 2), Point::new(3, 4))
    .into_styled(style)
    .draw(&mut d)?;
```
(Reusing `Styled` won't quite work yet because of #499, but this can be fixed after this PR)

@ReeceStevens You mentioned that you wanted to try to implement accelerated text rendering for `RA8875`, which should be made possible by this PR. If you have some time to try it out I would appreciate some feedback.

### DIFF
--- a/tests/text_renderer_target_specific.rs
+++ b/tests/text_renderer_target_specific.rs
@@ -1,0 +1,119 @@
+use embedded_graphics::{mock_display::MockDisplay, pixelcolor::Rgb888, prelude::*, primitives::Rectangle, text::{Alignment, Baseline, Text, TextStyleBuilder, renderer::{CharacterStyle, TargetSpecificTextRenderer, TextMetrics}}};
+
+type Error = <MockDisplay<Rgb888> as DrawTarget>::Error;
+
+struct TestDisplay(MockDisplay<Rgb888>);
+
+impl DrawTarget for TestDisplay {
+    type Color = Rgb888;
+    type Error = Error;
+
+    fn draw_iter<I>(&mut self, pixels: I) -> Result<(), Self::Error>
+    where
+        I: IntoIterator<Item = Pixel<Self::Color>>,
+    {
+        self.0.draw_iter(pixels)
+    }
+}
+
+impl OriginDimensions for TestDisplay {
+    fn size(&self) -> Size {
+        Size::new_equal(128)
+    }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+struct DummyTargetTextStyle<T>(T);
+
+impl<T: PixelColor> CharacterStyle for DummyTargetTextStyle<T> {
+    fn measure_string(&self, text: &str, position: Point, _baseline: Baseline) -> TextMetrics {
+        // The `baseline` is ignored in this test, but must be used in a real impl.
+
+        let width = text.len() as u32 * 4;
+
+        TextMetrics {
+            bounding_box: Rectangle::new(position, Size::new(width, 4)),
+            next_position: position + Size::new(width, 0),
+        }
+    }
+
+    fn line_height(&self) -> u32 {
+        5
+    }
+}
+
+impl TargetSpecificTextRenderer<TestDisplay> for DummyTargetTextStyle<Rgb888> {
+    fn draw_string(
+        &self,
+        text: &str,
+        position: Point,
+        baseline: Baseline,
+        target: &mut TestDisplay,
+    ) -> Result<Point, Error> {
+        let metrics = self.measure_string(text, position, baseline);
+
+        target.fill_solid(&metrics.bounding_box, self.0)?;
+
+        Ok(metrics.next_position)
+    }
+
+    fn draw_whitespace(
+        &self,
+        _width: u32,
+        _position: Point,
+        _baseline: Baseline,
+        _target: &mut TestDisplay,
+    ) -> Result<Point, Error> {
+        todo!()
+    }
+}
+
+#[test]
+fn target_specific_text_renderer() {
+    let mut target = TestDisplay(MockDisplay::new());
+
+    Text::new("ab\nc", Point::zero())
+        .into_styled(DummyTargetTextStyle(Rgb888::GREEN))
+        .draw(&mut target)
+        .unwrap();
+
+    target.0.assert_pattern(&[
+        "GGGGGGGG", //
+        "GGGGGGGG", //
+        "GGGGGGGG", //
+        "GGGGGGGG", //
+        "        ", //
+        "GGGG    ", //
+        "GGGG    ", //
+        "GGGG    ", //
+        "GGGG    ", //
+    ]);
+}
+
+#[test]
+fn target_specific_text_renderer_with_text_style() {
+    let mut target = TestDisplay(MockDisplay::new());
+
+    let character_style = DummyTargetTextStyle(Rgb888::BLUE);
+    let text_style = TextStyleBuilder::new()
+        .character_style(character_style)
+        .alignment(Alignment::Right)
+        .build();
+
+    Text::new("ab\nc", Point::new(7, 0))
+        .into_styled(text_style)
+        .draw(&mut target)
+        .unwrap();
+
+    target.0.assert_pattern(&[
+        "BBBBBBBB", //
+        "BBBBBBBB", //
+        "BBBBBBBB", //
+        "BBBBBBBB", //
+        "        ", //
+        "    BBBB", //
+        "    BBBB", //
+        "    BBBB", //
+        "    BBBB", //
+    ]);
+}
--- a/core/src/drawable.rs
+++ b/core/src/drawable.rs
@@ -68,31 +68,69 @@ pub trait Drawable {
     /// The pixel color type.
     type Color: PixelColor;
 
-    // TODO: Remove `ignore` from example code when the `Text` drawable is updated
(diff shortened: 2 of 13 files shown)


tests/text_renderer_target_specific.rs
```diff
@@ -0,0 +1,119 @@
...
+
+#[test]
+fn target_specific_text_renderer_with_text_style() {
+    let mut target = TestDisplay(MockDisplay::new());
+
+    let character_style = DummyTargetTextStyle(Rgb888::BLUE);
+    let text_style = TextStyleBuilder::new()
+        .character_style(character_style)
+        .alignment(Alignment::Right)
+        .build();
```
ReeceStevens, line 101: For my use case, this is the biggest limitation of the current approach. In order to use target-specific font rendering techniques, you need to use the target-specific `DummyTargetTextStyle`. This makes the definition of a UI non-portable across e-g- compatible backends. For instance, if I wanted to define a UI library that was display-independent, I would have no way of allowing a user of that library to provide an accelerated font rendering method if one is available.

Put another way, it seems like there is no way to use display-specific font rendering methods if the display is not known ahead of time. I might be wrong though-- I am still getting up to speed on the font system in e-g.

Is there any way the text rendering functionality could be defined as a part of the `DrawTarget` implementation? If it could be defined as part of the `DrawTarget` implementatoin, then a display implementer could define their font rendering method (or not, and therefore fall back to the default) and a library would not have to know any details.

I am aware my suggestion of putting functionality back into `DrawTarget` is in direct tension with the push to separate `DrawTarget` out from dependency on e-g primitives, so I understand if that's not a direction y'all would like to take.

ReeceStevens (commented):
Hey @rfuest!

Firstly, thank you for continuing to follow up on this topic and putting up a PR to address it! I really appreciate you taking the time to look through this issue and formulate a solution.

I unfortunately haven't yet had a chance to test this live on hardware. My application was running on e-g 0.6.2 and has a pretty non-trivial UI at this point, so porting it over to 0.7.0 is a quite sizeable task that will take more than an evening. I also think I managed to widely use almost every single now-removed API ;)

While I work on porting things over, though, I wanted to post some initial thoughts on the approach here. They mostly boil down to: 

- Pro: this approach allows you to hook specific driver functionality into the e-g system, which is great
- Con: using this approach, you cannot access driver-specific functionality in a driver-generic context

My particular use case for e-g is somewhat specific, in that I am compiling an application to two different targets, one of which uses the RA8875 driver as a `DrawTarget` and the other which uses the `SimulatorDisplay`. However, it seems like this problem would present itself for any generic library that uses the text functionality in e-g.

bugadani:
I don't think I'd separate styles into target-specific and generic parts. IMO e-g should define what's possible to draw and if driver's can't provide that, they should fall back to the current software rasterization/rendering.

At the risk of again posing as the smartass, I'd personally go this route: I think the drivers should be able to decide if they can draw a particular drawable hardware accelerated, or using the software rendering e-g provides. The best I can think of is a `DisplayDriver` trait with a single method (I'd prefer associated types that each implement a renderer, but assoc. type defaults aren't stable yet) to render each styled drawable, where the default implementation is the currently implemented pixel-pushing, and drivers could override them if they offer hardware acceleration.
There's quite a bit of boilerplate this would need and I'm sure there's a better solution. What I have in mind also has implications wrt. the current composable draw target implementations, but I think even target-specific drawables have some interactivity with those - i.e. how does a cropping draw target work with a hardware-accelerated sprite renderer?

These are just my .02€ - maybe some of my thoughts that are worth considering, though, I don't know.

ReeceStevens:
> We want to keep consistency everywhere which means some compromises in the current API unfortunately.

That makes sense from a design goal perspective. I think the RA8875 might be a slightly more niche use case as compared to other displays because it is so large (I'm working with the 800x480 size) and the SPI interface is rather slow, so it's pushing a lot of pixels around and these performance areas become more visible.

> It may be worth a quick experiment to see if the current 0.7 API could work for you

I will work on setting up a minimal experiment case soon, although it might take me about a week to get the time freed up. 

> A quick comment from the peanut gallery: presumably the main issue is performance, right? Are there specific items that are particularly slow? I saw the RA8875 has a builtin font. Do you want a good way to use that for performance/styling reasons, or is something else slow?

Yes, the main issue is performance, and specifically related to text rendering. Right now, rendering text happens so slowly that I can watch all the individual pixels for each letter get drawn, and 2-3 characters will take about a second. I've used the built-in font before for the RA8875 and knew that it was a lot faster, so I've been trying to find ways to use that to accelerate the UI updates.

I did some more brainstorming for my application last night and was able to work around the "display generic" issue I mentioned in my PR comment by creating a custom wrapper trait around `DrawTarget` that has an exposed text handling method, then implementing this on each target (`RA8875` and `SimulatorDisplay`).  It's basically bolting on new methods to `DrawTarget` but it does the trick, and given that e-g is prioritizing consistency, I think it makes sense that rendering workarounds like this fall on the shoulders of the implementer to fit in rather that this crate.

tests/text_renderer_target_specific.rs
rfuest: If I understand the RA8875 the controller supports rendering of fonts from an internal or external ROM. How could this accelerate the drawing of `MonoFont`s, which aren't included in the ROM?

> Put another way, it seems like there is no way to use display-specific font rendering methods if the display is not known ahead of time. I might be wrong though-- I am still getting up to speed on the font system in e-g.

An UI library could be made generic, so that the library user can either use a generic font or a target specific one. This should provide a way to implement both variants with some conditional compilation in the application. But there might be additional problems I haven't thought about yet.

> Is there any way the text rendering functionality could be defined as a part of the `DrawTarget` implementation?

I don't think there is, if we want to continue to support external text renderers. An external text renderer can use very different draw calls than the internal `MonoFont` text renderer. The `eg-seven-segment` renderer, for example, draws all glyphs using rectangles.

rfuest:
> I don't think I'd separate styles into target-specific and generic parts. IMO e-g should define what's possible to draw and if driver's can't provide that, they should fall back to the current software rasterization/rendering.

But in this case the controller can draw something that e-g can't, because of the builtin fonts in the controller.

> Put another way, it seems like there is no way to use display-specific font rendering methods if the display is not known ahead of time. I might be wrong though-- I am still getting up to speed on the font system in e-g.

As @jamwaffles already mentioned, we had tried this previously and it failed. The first external implementation of styled rectangle drawing contained a bug which broke rendering of rectangles with a stroke and no fill. This bug still exists until tody. And if this happens for relatively simple cases like rectangles it would be much harder to get consistent rendering for something like a polyline.

rfuest:
> Yes, the main issue is performance, and specifically related to text rendering. Right now, rendering text happens so slowly that I can watch all the individual pixels for each letter get drawn, and 2-3 characters will take about a second.

That seems very slow, even for unoptimized drawing. What is your SPI clock rate? Even at 1 MHz you should be able to draw ~10000 Pixels/s (rough estimate based on the code in your `ra8875` repo).

> I've used the built-in font before for the RA8875 and knew that it was a lot faster, so I've been trying to find ways to use that to accelerate the UI updates.

Internal fonts will always be faster than drawing text via SPI, but especially drawing fonts with opaque backgrounds can be accelerated a lot by implementing `DrawTarget::fill_contiguous` more efficiently.

ReeceStevens:
@rfuest @jamwaffles I was able to get a minimal example up and running, SPI clock rate of 2MHz. I'm still seeing quite slow speed, even after implementing a more efficient `fill_contiguous` that pushes rows of pixels at a time (although it is slightly better than before, especially if the fonts have a background).

Here's the minimal example: https://github.com/ReeceStevens/display-test

This seems to still be significantly slower than I would expect (2200 pixels, or 11 10x20 characters, in ~1.7 seconds), so I am beginning to wonder if there are some inefficiencies in the way I've implemented the RA8875 driver that I'm not picking up on yet. Either way, this PR isn't the forum to hash out those issues with my implementation. I'll open an issue against my own repository for now, and if I can get to the bottom of it, I will be sure to let y'all know!

Many thanks!

ReeceStevens:
As I mentioned in Matrix, it turns out my surprisingly slow speed was because of the overhead of running in debug mode. Sorry for making so much channel noise because I forgot to check a release build!

After implementing `fill_contiguous`, as y'all predicted, I'm seeing _much_ better performance now.

jamwaffles:
Great to hear you got it working! Not to diminish the work @rfuest has already done but do people think the feature proposed in this PR is still necessary?

rfuest:
IMO there might still be some value in supporting target specific drawing in the future, but I'll close this PR for now.

jamwaffles:
SGTM. Let's keep things a little simpler for 0.7 :)

PR #579, opened 2021-04-09
Generate font overview table

rfuest:
This PR updates the `convert-fonts` tool to generate an overview of the built-in fonts in the `mono_font` module docs. It also updates the `bdf-to-mono` tool to make it easier to use it outside of e-g.


jamwaffles (changes requested):
Looks good, although should we bring attention to the `latin1` encodings too? I don't think we should show both `ascii` and `latin` bitmaps next to each other as the table is already pretty long, but some way of telling the user both variants are available would be useful. Perhaps links next to each font name like

```
| <FONTAXB> - [`ascii`](...), [`latin1`](...) | [ image code here ] |
```

We could also go the simpler route and add a sentence at the top of the table asking the user to see the `Modules` section for all variants. A final option would be to essentially duplicate the `modules` section at the top of the page with a bulletted list linking to all the charsets at the top of the page.

My discovery path was clicking `Monospace fonts` from the docs root, then clicking `built-in fonts` at the top of the `embedded_graphics::mono_font` module. This took me to the samples table (good!) where only the ascii encodings are shown (not so good)

rfuest:
The main focus of this PR was the tooling and I haven't spend much time on the actual docs. But you make some really good points and this should be revisited as part of #480. Most of the current docs in the `mono_font` module need be moved to the `text` module, because they aren't specific to `MonoFont`s. This moves the overview table up and should make it more visible. We could copy some of the description of the `ascii` and `latin1` variants in the migration guide to the docs.

I've updated the PR some more:
- The generator now also outputs overview tables in the `ascii` and `latin1` modules
  We could later link to the encoding specific versions from the main `mono_font` docs. Should we also keep the table in `mono_font` or only add links to the modules?
- The table now uses two columns for a more compact layout
- The table is now correctly sorted by glyph size and not by font name (which put `FONT_10X20` first)
- I've renamed the bold and italic font variants to use an underscore between the size and the suffix
  (`FONT_9X18BOLD`-> `FONT_9X18_BOLD`)
  I think this makes more sense and I'm not sure why I didn't change this as part of #573. Do we need a separate changelog entry for this or is this included in the entry for #573, which had already changed the names?

jamwaffles (approved):
LGTM! I forgot about the doc improvements issue, hence my comments here. The changes after my last review are a good step in the right direction though.

> Do we need a separate changelog entry for this or is this included in the entry for #573, which had already changed the names?

No, I think it's fine to reuse that other entry. Generally I don't think we need changelog items for things that revert/modify changes that haven't been released yet.

PR #580, opened 2021-04-11
Don't use Styled for Text

rfuest:
This PR is an experiment to limit the use of `Styled` to primitives and not use it for `Text`s.

Unstyled primitives clearly define a geometric shape and they can be used on their own, for example, for points iterators or the `contains` method. `Text` objects, on the other hand, aren't really useful without an applied style, because there is no definition how an unstyled text would look.

By moving the styling directly into `Text` we gain some flexibility. Instead of only applying a single style to a `Text` object this makes it possible to have two separate style properties. This way the character and text styles can be handled independently and `TextStyle` doesn't need to be a wrapper around a character style anymore.

@bugadani If we decide to change this in e-g it might be a good idea to also change the e-t API to match.

# Examples

The following examples give an overview of the API changes and show some constructors that were added to simplify the creation of common `TextStyle`s.

## Special constructors for baseline or alignment

Before:
```rust
let character_style = MonoTextStyle::new(&FONT_6X9, Rgb888::WHITE);

let text_style = TextStyleBuilder::new()
    .character_style(character_style)
    .baseline(Baseline::Middle)
    .build();

Text::new("ABC", Point::new(0, 6))
    .into_styled(text_style)
    .draw(&mut display)?;
```

After:
```rust
let character_style = MonoTextStyle::new(&FONT_6X9, Rgb888::WHITE);

Text::with_baseline("ABC", Point::new(0, 6), character_style, Baseline::Middle)
    .draw(&mut display)?;
```

## `TextStyleBuilder` for more complex styling

```rust
let character_style = MonoTextStyle::new(&FONT_6X9, Rgb888::WHITE);

let text_style = TextStyleBuilder::new()
    .alignment(Alignment::Right)
    .baseline(Baseline::Middle)
    .line_height(LineHeight::Percent(200))
    .build();

Text::with_text_style("ABC", Point::new(0, 6), character_style, text_style)
    .draw(&mut display)?;
```


bugadani:
No objections here. In fact, this will reduce some potential confusion (i.e. "why can't I use TextStyle with TextBox?").

e-t will be updated once e-g's API gets somewhat stabilized.

src/text/text.rs
```diff
@@ -7,49 +7,102 @@ use crate::{
...
+            character_style,
+            text_style: TextStyleBuilder::new().build(),
+        }
     }
 
-    /// Attaches a text style to the text object.
-    pub fn into_styled<S>(self, style: S) -> Styled<Self, S> {
-        Styled::new(self, style)
+    /// Creates a text.
+    pub const fn with_text_style(
```
jamwaffles, line 56: This should be called `with_style` IMO; the `_text` part is redundant because we're already inside a `Text`.

src/text/text.rs
```diff
@@ -7,49 +7,102 @@ use crate::{
...
+            position,
+            character_style,
+            text_style: TextStyleBuilder::new().build(),
+        }
     }
 
-    /// Attaches a text style to the text object.
-    pub fn into_styled<S>(self, style: S) -> Styled<Self, S> {
-        Styled::new(self, style)
+    /// Creates a text.
```
jamwaffles, line 55: Almost pointless nit 

```suggestion
    /// Creates a text with a given style.
```

src/text/text.rs
```diff
@@ -7,49 +7,102 @@ use crate::{
...
+    ) -> Self {
+        Self {
+            text,
+            position,
+            character_style,
+            text_style,
+        }
+    }
+
+    /// Creates a text.
```
jamwaffles, line 70:
```suggestion
    /// Creates a text with a given baseline.
```

src/text/text.rs
```diff
@@ -7,49 +7,102 @@ use crate::{
...
+    ) -> Self {
+        Self {
+            text,
+            position,
+            character_style,
+            text_style: TextStyle::with_baseline(baseline),
+        }
+    }
+
+    /// Creates a text.
```
jamwaffles, line 85: I think it's horizontal?

```suggestion
    /// Creates a text with given horizontal alignment.
```

jamwaffles (changes requested):
Oof these PRs are huge. A positive symptom of having many tests/examples I guess!

I left some small nits.

Why is `baseline` a special property? My immediate though when I saw the examples in the PR description was "why isn't baseline in the style struct?" is it because it can be used for e.g. layout, whereas styling not so much?

src/text/text.rs
rfuest: I did name it `with_text_style` on purpose, because all `Text` constructors take a character style argument and only this version takes an additional text style argument.

src/text/text.rs
rfuest: Not pointless at all. A similar change is already included in #581.

src/text/text.rs
rfuest: Also included in #581.

rfuest:
> Oof these PRs are huge. A positive symptom of having many tests/examples I guess!

Sorry for that, but there wasn't a good way to split them any further. But the other PR will get much more manageable when it is rebased after this PR is merged.
 
> Why is `baseline` a special property? My immediate though when I saw the examples in the PR description was "why isn't baseline in the style struct?" is it because it can be used for e.g. layout, whereas styling not so much?

Yes, the layout options are included in the text style and the styling of the individual characters is defined by the character style. This way the layout options can be handled by the `Text` drawable independently of the used text renderer.

jamwaffles (approved):
I looked through your comments and they all make sense, so LGTM! Don't worry about the long PRs. They're a bit of a mission to review but as you said there's not really a good way around it.

bugadani:
@rfuest why does TextStyle only have constructors for baseline and alignment, and not for line height? embedded-text has more options and I'd like to understand the reasoning behind this decision to know what I need to add to the API.

rfuest:
The baseline and alignment settings are the most commonly used ones and often only one of them is set. But TBH the constructors in `TextStyle` aren't really necessary anymore, because they were replaced by `Text::with_alignment` and `Text::with_baseline`. And all more complex text styles need to be build using `TextStyleBuilder` anyway.

bugadani:
> But TBH the constructors in TextStyle aren't really necessary anymore, because they were replaced by Text...

Same problem applies to Text as well :) In my case I don't think I can tell which option is used most commonly, though, so I might just add a constructor for each. Thanks!

PR #581, opened 2021-04-12
Text module docs

rfuest:
This PR adds some docs for the `text` and `mono_font` modules. They are still not great, but I hope they are not too confusing.

I've marked it as a draft PR, because it depends on #580.

### DIFF
--- a/src/mono_font/mod.rs
+++ b/src/mono_font/mod.rs
@@ -12,7 +12,7 @@
 //!
 //! # Built-in fonts
 //!
-//! Each build-in font is provided in an ASCII and a Latin 1 (ISO 8859-1) variant. The ASCII variant
+//! Each built-in font is provided in an ASCII and a Latin 1 (ISO 8859-1) variant. The ASCII variant
 //! contains a smaller subset of characters which saves memory in embedded application, but only
 //! covers all characters of the English language. The Latin 1 variant has complete coverage for
 //! [the languages listed here](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Modern_languages_with_complete_coverage),
--- a/src/mono_font/mod.rs
+++ b/src/mono_font/mod.rs
@@ -4,111 +4,22 @@
 //! several [built-in fonts].
 //!
 //! Additional custom fonts can be added by the application or other crates. This
-//! is demonstrated in the `text-custom-font` example in the simulator crate.
+//! is demonstrated in the `text-custom-font` example in the [examples repository].
 //!
 //! # Examples
 //!
-//! The examples below use the `Font6x8` font, however any of the [built-in fonts]
-//! in this module or custom fonts can be substituted.
-//!
-//! ## Print styled "Hello Rust!"
-//!
-//! TODO: `Text` no longer use `Styled`.
-//!
-//! Text can be drawn to a display by creating a [`Text`] object and attaching a
-//! text style to it by using a `Styled` object. This example prints
-//! "Hello Rust" with a yellow text on a blue background.
-//!
-//! ```rust
-//! use embedded_graphics::{
-//!     mono_font::{ascii::FONT_6X9, MonoTextStyle, MonoTextStyleBuilder},
-//!     pixelcolor::Rgb565,
-//!     prelude::*,
-//!     text::Text,
-//! };
-//! # use embedded_graphics::mock_display::MockDisplay;
-//! # let mut display: MockDisplay<Rgb565> = MockDisplay::default();
-//! # display.set_allow_out_of_bounds_drawing(true);
-//!
-//! // Create a new character style
-//! let style = MonoTextStyleBuilder::new()
-//!     .font(&FONT_6X9)
-//!     .text_color(Rgb565::YELLOW)
-//!     .background_color(Rgb565::BLUE)
-//!     .build();
-//!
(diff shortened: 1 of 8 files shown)


src/mono_font/mod.rs
```diff
@@ -4,111 +4,22 @@
...
-//!
-//! Text::new(&buf, Point::zero(), style).draw(&mut display)?;
-//! # Ok::<(), core::convert::Infallible>(())
-//! ```
+//! The [`text` module] contains examples how these fonts can be used in an application.
 //!
 //! # Built-in fonts
 //!
-//! TODO: Describe the difference between `ascii` and `latin1` and mention the `latin1` table in the `latin1` module.
+//! Each build-in font is provided in an ASCII and a Latin 1 (ISO 8859-1) variant. The ASCII variant
```
jamwaffles, line 15:
```suggestion
//! Each built-in font is provided in an ASCII and a Latin 1 (ISO 8859-1) variant. The ASCII variant
```

src/text/mod.rs
```diff
@@ -1,14 +1,173 @@
...
+//!
+//! In addition to styling the individual characters the [`Text`] drawable also contains a
+//! [`TextStyle`] setting. The text style is used to set the alignment and line spacing of text
+//! objects.
+//!
+//! The [`alignment`] setting sets the horizontal alignment of the text. With the default value
+//! `Left` the text will be rendered to the right of the given text position. Analogously `Right`
+//! aligned text will be rendered to the left of the given position. `Center`ed text will extend
+//! equally to the left and right of the text position. If the text contains multiple lines each
+//! line will be aligned independently of the other lines.
```
jamwaffles, line 30: I'm not sure what 

> If the text contains multiple lines each line will be aligned independently of the other lines.

means. It seems to imply I'd have to set the alignment for each line of text. I don't think that's correct, but that's how I understood this sentence.

src/text/text_style.rs
```diff
@@ -1,6 +1,15 @@
 use crate::text::{Alignment, Baseline, LineHeight};
 
 /// Text style.
+///
+/// A text style is used to set how text lines are layed out in a text drawable.
+///
+/// Use [`TextStyleBuilder`] to build text style object.
```
jamwaffles, line 7:
```suggestion
/// Use [`TextStyleBuilder`] to build a text style object.
```

src/text/mod.rs
```diff
@@ -1,14 +1,173 @@
...
+//! # display.set_allow_out_of_bounds_drawing(true);
+//!
+//! // Create a small and a large character style.
+//! let small_style = MonoTextStyle::new(&FONT_6X10, Rgb565::WHITE);
+//! let large_style = MonoTextStyle::new(&FONT_10X20, Rgb565::WHITE);
+//!
+//! // Draw the first text at (20, 30) using the small character style.
+//! let next = Text::new("small ", Point::new(20, 30), small_style).draw(&mut display)?;
+//!
+//! // Draw the second text after the first text using the large  character style.
```
jamwaffles, line 152:
```suggestion
//! // Draw the second text after the first text using the large character style.
```

src/text/text_style.rs
```diff
@@ -1,6 +1,15 @@
 use crate::text::{Alignment, Baseline, LineHeight};
 
 /// Text style.
+///
+/// A text style is used to set how text lines are layed out in a text drawable.
```
jamwaffles, line 5:
```suggestion
/// A text style is used to set how text lines are laid out in a text drawable.
```

src/text/mod.rs
```diff
@@ -1,14 +1,173 @@
...
+//! line will be aligned independently of the other lines.
+//!
+//! The [`baseline`] setting defines the vertical alignment of the first line of text. With the default
+//! setting of `Alphabetic` the glyphs will be drawn with their descenders below the given position.
+//! This means that the bottom of glyphs without descender (like 'A') will be on the same Y
+//! coordinate as the given position. The other baseline settings will position the glyphs relative
+//! to the EM box, without considering the baseline.
+//!
+//! If the text contains multiple lines only the first line will be vertically aligned based on the
+//! baseline setting. All following lines will be spaced according to the [`line_height`] setting.
```
jamwaffles, line 39: Maybe a small clarification that they're positioned relative to wherever the first line was rendered?

```suggestion
//! baseline setting. All following lines will be spaced relative to the first line, according to the [`line_height`] setting.
```

jamwaffles (changes requested):
Left a few small comments but otherwise LGTM.

I'll admit I was a bit confused by the examples/text style/character styles at first, but I think these docs are a good starting point for release. I think I'd have to use the new text system in a project before being able to suggest improvements to the docs so let's get this merged so there's one less thing to do for 0.7. We can always improve the docs in a patch release later.

src/text/mod.rs
rfuest: It was supposed to mean that each line would be centered independently and not handled as a centered block of left aligned lines:
grafik
The green line is the X position passed to `Text::new`.

But I guess that everyone would expect centering to work this way, so I've just removed the confusing sentence.

jamwaffles: Ah I see now, yeah it's probably best to let assumptions do the work here.

jamwaffles (approved):

PR #587, opened 2021-04-22
Add From impls for MonoFontBuilder and TextStyleBuilder

rfuest:
This PR adds the missing `From` impl for `TextStyleBuilder` that was reported by @bugadani  in https://github.com/embedded-graphics/embedded-graphics/issues/511#issuecomment-823582640.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- [#587](https://github.com/embedded-graphics/embedded-graphics/pull/587) Added `From<&TextStyle>` impl for `TextStyleBuilder` and `From<&MonoFont>` for `MonoFontBuilder`.
+
 ## [0.7.0-beta.1] - 2021-04-19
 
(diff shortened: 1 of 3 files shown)


bugadani:
Could you also add reset methods, please? I think those are also missing, but I only made a cursory glance.

rfuest:
Do you mean for the `TextStyle` settings? Reset methods currently only exist for colors settings. These cannot be reset by other methods, because the builder methods take `C` instead of `Option<C>` arguments.

bugadani:
Ooooh, my bad, sorry!

rfuest:
No problem. In some cases reset methods for other settings might be useful.

jamwaffles (approved):
LGTM, nice and simple :)

Would it make sense to also add `From<MonoFont>` and `From<TextStyle>` (i.e. non-ref) versions as well?

rfuest:
> Would it make sense to also add `From<MonoFont>` and `From<TextStyle>` (i.e. non-ref) versions as well?

Maybe. I did actually think about this briefly while implementing this PR. But because I'm unsure if this would have any benefits I decided to simply mimic the existing `From` impls  for the other builders.

PR #589, opened 2021-05-01
Add missing color conversions

rfuest:
This PR fills the gaps in the color conversions.

We had previously decided not to implement any conversions which can be handled in different ways. But this makes it harder to use color types generically, because some colors don't implement `From` for other color types. One example where this causes problems is `DynamicTga`, which can currently only be used with RGB targets.

While the conversions implemented in this PR might not be ideal for every use case, I think it's a good idea to provide an implementation that should work in most cases. And as long as we document the conversions (in a later PR) this shouldn't be confusing to the user.

@bugadani: I think this should let you replace `Rgb` with `Rgb888` in e-t and should also make it work with `BgrXXX` targets.

### DIFF
--- a/core/src/pixelcolor/gray_color.rs
+++ b/core/src/pixelcolor/gray_color.rs
@@ -11,9 +11,6 @@ pub trait GrayColor: PixelColor {
     /// Black color (0% luma).
     const BLACK: Self;
 
-    /// Middle gray color (50% luma).
-    const GRAY_50: Self;
-
     /// White color (100% luma).
     const WHITE: Self;
 }
@@ -26,6 +23,8 @@ macro_rules! gray_color {
         pub struct $type($raw_type);
 
         impl $type {
+            pub(crate) const GRAY_50: Self = Self::new(0x80 >> (8 - $raw_type::BITS_PER_PIXEL));
+
             /// Creates a new grayscale color.
             ///
             /// Too large luma values are masked to the valid range by setting
@@ -45,7 +44,6 @@ macro_rules! gray_color {
             }
 
             const BLACK: Self = Self::new(0);
-            const GRAY_50: Self = Self::new(0x80 >> (8 - $raw_type::BITS_PER_PIXEL));
             const WHITE: Self = Self::new(255);
         }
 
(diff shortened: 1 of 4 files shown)


bugadani:
At this rate, we won't have any code in e-t by august. :) Thanks for this, `Rgb` was certainly a pain point, especially now with the prelude removed.

core/src/pixelcolor/gray_color.rs
```diff
@@ -8,10 +8,13 @@ pub trait GrayColor: PixelColor {
     /// Returns the luma channel value.
     fn luma(&self) -> u8;
 
-    /// Black color (0% luma)
+    /// Black color (0% luma).
     const BLACK: Self;
 
-    /// White color (100% luma)
+    /// Middle gray color (50% luma).
+    const GRAY_50: Self;
```
jamwaffles, line 15: This might be worth an `Added` changelog entry, unless you think it's too small to bother with?

jamwaffles (approved):
Lots of macro soup so I didn't look at this super closely, but LGTM anyway.

Concerning the `**(breaking)**` in the changelog, in terms of actual user code, what kind of pattern/usage is broken by this PR?

core/src/pixelcolor/gray_color.rs
rfuest: I've removed the constant again. `GRAY_50` wasn't really at exactly 50%, because each color has an even number of possible luma values. And this change would be a breaking API change, because the `GrayColor` trait isn't sealed and could theoretically be implemented externally (very unlikely but possible).

rfuest:
> Concerning the `**(breaking)**` in the changelog, in terms of actual user code, what kind of pattern/usage is broken by this PR?

I did forget to remove the breaking annotation from the entry when I copied another changelog entry. I've removed the breaking annotation and copied the entry to the e-g changelog.

rfuest:
Can you release core `0.3.1` after this is merged?

jamwaffles (approved):
Makes sense! Yep, I'll get a new release out as soon as this is merged.

I assume the duplicated changelog entries in core and e-g are deliberate?

rfuest:
> I assume the duplicated changelog entries in core and e-g are deliberate?

Yes. I thought we had decided to duplicate the entries which are relevant to e-g users, because they probably don't check the core changelog.

jamwaffles:
Ah yes, that rings a bell. Thanks for reminding me.

jamwaffles:
e-g-core 0.3.1 just went out :shipit:

PR #591, opened 2021-05-10
Clarify `DrawTarget` API

GrantM11235:

## PR description

The docs for `DrawTarget` are a bit ambiguous about how to handle off-screen pixels, this makes the API contract more explicit.

### DIFF
--- a/core/src/draw_target/mod.rs
+++ b/core/src/draw_target/mod.rs
@@ -299,6 +299,9 @@ pub trait DrawTarget: Dimensions {
     /// drawing method for a display that writes data to the hardware immediately. If possible, the
     /// other methods in this trait should be implemented to improve performance when rendering
     /// more contiguous pixel patterns.
+    ///
+    /// This method should ignore any pixels that fall outside the drawable area of the
+    /// target display without returning an error. This can be checked with the [`Rectangle::contains`] method
     fn draw_iter<I>(&mut self, pixels: I) -> Result<(), Self::Error>
     where
         I: IntoIterator<Item = Pixel<Self::Color>>;
@@ -316,8 +319,11 @@ pub trait DrawTarget: Dimensions {
     /// The provided iterator is not required to provide `width * height` pixels to completely fill
     /// the area. In this case, `fill_contiguous` should return without error.
     ///
-    /// This method should not attempt to draw any pixels that fall outside the drawable area of the
-    /// target display. The `area` argument can be clipped to the drawable area using the
+    /// The provided iterator is not required to be limited to `width * height` pixels.
+    /// In this case, `fill_contiguous` should ignore the additional pixels and return without error.
+    ///
+    /// This method should ignore any pixels that fall outside the drawable area of the
+    /// target display without returning an error. The `area` argument can be clipped to the drawable area using the
     /// [`Rectangle::intersection`] method.
     ///
     /// The default implementation of this method delegates to [`draw_iter`](#tymethod.draw_iter).
@@ -397,6 +403,10 @@ pub trait DrawTarget: Dimensions {
     }
 
     /// Fill a given area with a solid color.
+    ///
+    /// This method should ignore any pixels that fall outside the drawable area of the
+    /// target display without returning an error. The `area` argument can be clipped to the drawable area using the
+    /// [`Rectangle::intersection`] method.
     ///
     /// If the target display provides optimized hardware commands for filling a rectangular area of
     /// the display with a solid color, this method should be overridden to use those commands to


core/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 - [#589](https://github.com/embedded-graphics/embedded-graphics/pull/589) Implemented `From` trait to convert from RGB colors to grayscale colors, between different grayscale colors and from grayscale and RGB colors to `BinaryColor`.
 
+### Fixed
+
+- [#591](https://github.com/embedded-graphics/embedded-graphics/pull/591) Clarifed DrawTarget API.
+
```
jamwaffles, line 18: This section should be under the `## [Unreleased] - ReleaseDate` heading on line 7.

jamwaffles (changes requested):
I left a CHANGELOG comment, however for the rest of the changes I'm not sure if we should recommend/restrict the default behaviour as described in the added comments. The additions certainly make sense from a user expectations point of view, but it might be the job of individual display drivers to decide how to limit things. 

@rfuest what would you suggest? Perhaps a rewording to recommend some sensible default behaviour, but make clear that individual device drivers may deviate?

rfuest:
I agree with @jamwaffles that the new clarifications aren't valid for all `DrawTarget` impls. Draw target adapters, like `Translated`, implement `DrawTarget`, but don't do any bounds checking.

I would suggest to move the display driver specific information to a new `# Implementation notes for display drivers`  (or  similar) section in the trait docs, instead of the individual methods.

core/CHANGELOG.md
rfuest: As long as only the docs are changed I don't think this needs any changelog entry at all.

core/src/draw_target/mod.rs
```diff
@@ -316,8 +319,11 @@ pub trait DrawTarget: Dimensions {
     /// The provided iterator is not required to provide `width * height` pixels to completely fill
     /// the area. In this case, `fill_contiguous` should return without error.
     ///
-    /// This method should not attempt to draw any pixels that fall outside the drawable area of the
-    /// target display. The `area` argument can be clipped to the drawable area using the
+    /// The provided iterator is not required to be limited to `width * height` pixels.
+    /// In this case, `fill_contiguous` should ignore the additional pixels and return without error.
```
rfuest, line 323: Thanks for clarifying this! I think this could be merged with the previous paragraph to reduce some repetition.

core/src/draw_target/mod.rs
```diff
@@ -316,8 +319,11 @@ pub trait DrawTarget: Dimensions {
...
+    ///
+    /// This method should ignore any pixels that fall outside the drawable area of the
+    /// target display without returning an error. The `area` argument can be clipped to the drawable area using the
```
rfuest, line 326: This is valid, but in this case extra care must be take. If `area` is clipped and the color iterator is not this can corrupt drawing.

rfuest (changes requested):

core/CHANGELOG.md
GrantM11235: In that case I will remove it

GrantM11235:
> I agree with @jamwaffles that the new clarifications aren't valid for all `DrawTarget` impls. Draw target adapters, like `Translated`, implement `DrawTarget`, but don't do any bounds checking.

Perhaps it would be more accurate to simply say that `DrawTarget` impls must handle "out of bounds" pixels without panicking, returning an error, or causing display corruption. Draw target adapters fulfill this requirement, assuming that the parent `DrawTarget` fulfills this requirement as well.

> I would suggest to move the display driver specific information to a new `# Implementation notes for display drivers` (or similar) section in the trait docs, instead of the individual methods.

The `DrawTarget` doc page seems quite cluttered in general, with lots of somewhat-redundant information and examples, but I'm not quite sure how to clean it up without removing potentially useful information. I think separating the "For implementers" docs from the "For users" docs would help. I'll give it a shot when I when I have some free time, let me know if you have any other ideas.

rfuest:
> Perhaps it would be more accurate to simply say that `DrawTarget` impls must handle "out of bounds" pixels without panicking, returning an error, or causing display corruption. Draw target adapters fulfill this requirement, assuming that the parent `DrawTarget` fulfills this requirement as well.

Sounds good! The docs for the `Error` type could also need an update. The current "This error is returned if an error occurred during a drawing operation." could be interpreted that out of bounds drawing should cause an error, but it should be a noop.

> The `DrawTarget` doc page seems quite cluttered in general, with lots of somewhat-redundant information and examples, but I'm not quite sure how to clean it up without removing potentially useful information. I think separating the "For implementers" docs from the "For users" docs would help. I'll give it a shot when I when I have some free time, let me know if you have any other ideas.

Thanks for taking the time to clean up the docs. We only recently started to distinguish between user and implementer facing docs, which should make the docs easier to understand. `DrawTarget` is a slightly special case, because it is only targeted at implementers of `DrawTarget` and `Drawable`s.

I agree that having two display driver examples is a bit redundant. IMO the minimal example isn't required, if we make sure that the description states that overriding the `fill_solid` isn't required if the target hardware doesn't support this.

PR #592, opened 2021-05-15
Implement `defmt::Format` for lots of types.

derekdreery:

## PR description

The functionality is behind a flag so you don't pay the compile time if you're not using `defmt`.

~~It relies on some patches to `defmt` so I will wait until they are released to make this a non-draft PR. I'm opening it early for feedback and interest.~~ Works with latest `defmt`

Although it touches almost all the files it should be easy to review, since the same line of code is added repeatedly.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -21,6 +21,7 @@ circle-ci = { repository = "embedded-graphics/embedded-graphics", branch = "mast
 az = "1.1"
 nalgebra = { version = "0.30.1", optional = true, default-features = false }
 byteorder = { version = "1.3.4", default-features = false }
+defmt = { version = "0.3.2", optional = true}
 
 [dev-dependencies]
 embedded-graphics = { path = ".." }
--- a/core/src/geometry/mod.rs
+++ b/core/src/geometry/mod.rs
@@ -93,6 +93,7 @@ where
 
 /// Anchor point.
 #[derive(Debug, Ord, PartialOrd, Eq, PartialEq, Hash, Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(::defmt::Format))]
 pub enum AnchorPoint {
     /// Top left.
     TopLeft,
--- a/src/geometry/real.rs
+++ b/src/geometry/real.rs
@@ -71,7 +71,6 @@ mod real_impl {
     pub(crate) const TAU: Real = Real(I16F16::from_bits(411775));
 
     #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
-    #[cfg_attr(feature = "defmt", derive(::defmt::Format))]
     pub(crate) struct Real(pub(super) I16F16);
 
     impl Real {
@@ -119,6 +118,13 @@ mod real_impl {
     impl From<Real> for u32 {
         fn from(src: Real) -> Self {
             src.0.to_num::<u32>()
+        }
+    }
+
+    #[cfg(feature = "defmt")]
+    impl ::defmt::Format for Real {
+        fn format(&self, fmt: ::defmt::Formatter<'_>) {
+            ::defmt::write!(fmt, "Read({=i32:X})", self.0.to_bits())
         }
     }
 }
--- a/src/geometry/real.rs
+++ b/src/geometry/real.rs
@@ -124,7 +124,12 @@ mod real_impl {
     #[cfg(feature = "defmt")]
     impl ::defmt::Format for Real {
         fn format(&self, fmt: ::defmt::Formatter<'_>) {
-            ::defmt::write!(fmt, "Read({=i32:X})", self.0.to_bits())
+            ::defmt::write!(
+                fmt,
+                "Real({=i32:#X} ≈{=f32})",
+                self.0.to_bits(),
+                self.0.to_num()
+            )
         }
     }
 }
--- a/src/geometry/real.rs
+++ b/src/geometry/real.rs
@@ -21,6 +21,7 @@ mod real_impl {
     pub(crate) const TAU: Real = Real(2.0 * f32::consts::PI);
 
     #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+    #[cfg_attr(feature = "defmt", derive(::defmt::Format))]
     pub(crate) struct Real(pub(super) f32);
 
     impl From<f32> for Real {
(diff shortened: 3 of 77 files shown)


core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ circle-ci = { repository = "embedded-graphics/embedded-graphics", branch = "mast
 [dependencies]
 nalgebra = { version = "0.23.0", optional = true, default-features = false }
 byteorder = { version = "1.3.4", default-features = false }
+defmt = { version = "0.2.1", optional = true, git = "https://github.com/derekdreery/defmt", branch = "impl_format_for_range"}
```
jamwaffles, line 23: For consistency with e-g itself, IMO it would be a good idea to add a `defmt_support = [ "defmt" ]` feature to `core` as well.

jamwaffles (commented):
One small suggestion on feature naming/consistency, otherwise this looks like a good addition to me! Considering it's opt-in, can you see any issues that could arise from using defmt @rfuest?

rfuest:
> Considering it's opt-in, can you see any issues that could arise from using defmt @rfuest?

The only issue I can think of is that `defmt` doesn't build on our MSRV. But as long as it's opt-in this shouldn't be a problem.

core/Cargo.toml
rfuest: Could we just rename the future to `defmt`? https://rust-lang.github.io/api-guidelines/naming.html#c-feature

jamwaffles:
That's a good point. The CI should check that the feature compiles on the stable/beta/all-targets builds though.

core/Cargo.toml
jamwaffles: I tried this, but a feature can't have the same name as a crate. I guess we could alias the crate to free up the feature name?

core/src/geometry/mod.rs
```diff
@@ -94,6 +94,7 @@ where
 
 /// Anchor point.
 #[derive(Debug, Ord, PartialOrd, Eq, PartialEq, Hash, Copy, Clone)]
+#[cfg_attr(feature = "defmt_support", derive(::defmt::Format))]
```
rfuest, line 97: There are three different variants of this attribute in this PR:
```rust
#[cfg_attr(feature = "defmt", derive(::defmt::Format))]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[cfg_attr(feature = "defmt_support", derive(::defmt::Format))]
```
Could you use the same variant everywhere?

rfuest (changes requested):

core/Cargo.toml
rfuest: An optional crate can be used as a feature without explicitly defining a feature in `Cargo.toml`.

jamwaffles: Sure, but the feature in the not-`core` `Cargo.toml` is there because it turns on `core`'s `defmt` feature too, so we can't just use the `defmt` optional dependency there. 

I don't think there's a solution to requiring the `defmt_support` feature in the root `Cargo.toml`, hence me suggesting we add the `defmt_support` feature here too.

rfuest: OK, I hadn't that the `defmt` feature in e-g would also need to enable the same feature in `core`. So forget what I said and just use a `defmt_support´ flag in both crates.

core/src/geometry/mod.rs
derekdreery: Note to self: use `defmt_support` as the name, and `derive(::defmt::Format)` as the path, since it is a little bit more specific than `derive(defmt::Format)`.

derekdreery:
It will be a while before this can be landed, because it needs a new release of `defmt`.

jamwaffles:
Hey @derekdreery, defmt has had a couple of releases since your last comment. Are you still interested in adding defmt support?

derekdreery:
This can be landed now. I can't see the CI errors because when I click on Details it wants me to log in.

derekdreery:
No idea why there are doc link errors or how to fix them.

core/Cargo.toml
ithinuel: Now we can have `defmt = [ "dep:defmt", "embedded-graphics-core/defmt" ]`

derekdreery:
I've rebased on the latest master, and have changed the feature name to `defmt`, which requires the ability to use feature dependencies like "dep:defmt". This means that the minimum rust compiler version goes up.

src/geometry/real.rs
```diff
@@ -119,6 +120,13 @@ mod real_impl {
             src.0.to_num::<u32>()
         }
     }
+
+    #[cfg(feature = "defmt")]
+    impl ::defmt::Format for Real {
+        fn format(&self, fmt: ::defmt::Formatter<'_>) {
+            ::defmt::write!(fmt, "Read({=i32:X})", self.0.to_bits())
```
rfuest, line 127: The hex value is hard to decipher. Also printing the approximate value by converting into floating point would make the output more useful IMO. Something like this might work:
```suggestion
            ::defmt::write!(fmt, "Real({=i32:#X} ≈{=f32})", self.0.to_bits(), self.0.to_num())
```

src/lib.rs
```diff
@@ -113,6 +113,8 @@
 //! * `fixed_point` - use fixed point arithmetic instead of floating point for all trigonometric
 //! calculation.
 //!
+//! * `defmt` - provide implementations of `defmt::Format` for all types where possible.
```
rfuest, line 116: A link to the the defmt would be useful for users who aren't familiar with defmt.

rfuest (changes requested):
Two small nits, otherwise it looks good.

src/geometry/real.rs
derekdreery: Good idea! I didn't use display/debug methods, as the whole point of defmt is to not pull those in, but hadn't considered converting it to a float.

src/lib.rs
derekdreery: :+1:

derekdreery:
Ready for review again.

Now I know about the fixed point `to_num` method I could just make that the logged value. There are a few times you might need the exact byte repr, for example if you want to cut'n'paste it for a bug report, so I've left it in for now. It shouldn't be considered a breaking change to change the format of logs like defmt - those needing a stable robust serialization should use serialization tools instead.

rfuest (approved):
LGTM! @jamwaffles do you have any more feedback?

jamwaffles (approved):
Looks good to me :). Thanks for keeping this alive since May last year!

Issue #590, opened 2021-05-03
Add documentation for new color conversions


rfuest:
The color conversions that were added in #589 need to be explained in the docs. There are multiple ways of e.g. converting RGB to grayscale and the docs should clearly state how colors are converted by the `From` impls.

PR #595, opened 2021-05-16
Some more progress on the migration guide

jamwaffles:

## PR description

At this point I just want 0.7 out the door, so the changes in this PR are on an MVP basis. I'd prefer to only do changes which fix obviously incorrect content. The migration guide can be improved later as people with confusions start to mention things in issues/Element.


jamwaffles:
I've got rid of the final TODOs in the migration guide. Some sections are a bit light on details, but I think it would be useful to collect migration pains from the community once 0.7 is released and add to the guide at that point.

Once this is merged in, should we go for a 0.7 not-beta release? The other release blocker is #590 but maybe it's not super important to get in?

rfuest:
I can't review this at the moment, but at least the "Fonts and texts" section still contains a TODO with my name.

> Once this is merged in, should we go for a 0.7 not-beta release? The other release blocker is #590 but maybe it's not super important to get in?

#590 isn't very important, which is why I hadn't tagged it with the `0.7-release-blocker` label. But I would like to get one last text change in 0.7, that will should be ready this week.

I also think we should update the examples before the 0.7 release (https://github.com/embedded-graphics/examples/issues/5 and https://github.com/embedded-graphics/examples/issues/8), so that the new features can easily be checked out. IMO this shouldn't happen sometime after the release, because potential new users will look at the examples after they read the 0.7 release announcement.

And someone needs to check if `tinytga` and `tinybmp` still work with the latest e-g beta.

jamwaffles:
I shouldn't be writing comments just before work :D. I'll take a look at the TODOs and see if I can teach myself enough about how the new text system works to flesh those sections out.

Thanks for the list of other things to do as well - I can take a look through those in the next day or so.

jamwaffles:
> And someone needs to check if tinytga and tinybmp still work with the latest e-g beta.

I just confirmed that they do indeed work with no changes other than the version bump in `Cargo.toml`. I'll cut releases of both these crates when 0.7.0 is released.

rfuest:
OK, great, at least something we didn't break.

MIGRATING-0.6-0.7.md
```diff
@@ -161,7 +160,58 @@ Driver authors should use `DrawTarget` exported by the [`embedded-graphics-core`
 
 `DrawTarget` now uses an associated type for the target color instead of a type parameter.
 
-`DrawTarget`s must also implement the `Dimensions` trait.
+`DrawTarget`s must also implement the `OriginDimensions` trait.
```
rfuest, line 163: `DrawTarget`s aren't required to implement `OriginDimensions`. But it is recommended to implement `OriginDimensions` for most display drivers, because the drawable area starts in the origin.

MIGRATING-0.6-0.7.md
```diff
@@ -181,25 +231,129 @@ All `draw_*` methods to draw specific primitives (`draw_circle`, `draw_triangle`
...
 
 These methods aim to be more compatible with hardware-accelerated drawing commands. Where possible, embedded-graphics primitives will use `fill_contiguous` and `fill_solid` to improve performance, however may fall back to `draw_iter` by default.
 
-To reduce duplication, please search the `DrawTarget` documentation on <https://docs.rs/embedded-graphics> for more details on the usage and arguments of the above methods.
+To reduce duplication, please search the `DrawTarget` documentation on <https://docs.rs/embedded-graphics-core> for more details on the usage and arguments of the above methods.
 
 ## For crates that handle images
 
-Crates that handle images should now implement items exported by the [`embedded-graphics-core`](https://crates.io/crates/embedded-graphics-core) crate to integrate with embedded-graphics.
+Crates that handle images should now implement the `ImageDrawable` and `OriginDimensions` traits to integrate with embedded-graphics.
```
rfuest, line 242: Should we replace "should" with "must", because the two traits are required to be implemented to integrate with e-g?

MIGRATING-0.6-0.7.md
```diff
@@ -181,25 +231,129 @@ All `draw_*` methods to draw specific primitives (`draw_circle`, `draw_triangle`
...
+    image::ImageDrawable,
+    pixelcolor::{PixelColor, Rgb888},
+    primitives::Rectangle,
+};
+
+struct MyRgb888Image {
+  // ...
+}
+
+impl ImageDrawable<Rgb888> for MyRgb888Image {
```
rfuest, line 259:
```suggestion
impl ImageDrawable for MyRgb888Image {
```

MIGRATING-0.6-0.7.md
```diff
@@ -181,25 +231,129 @@ All `draw_*` methods to draw specific primitives (`draw_circle`, `draw_triangle`
...
+    fn size(&self) -> Size {
+        // Return image width and height in pixels
+    }
+}
+```
 
 ## For text rendering crates
 
-Crates that handle text rendering should now implement items exported by the [`embedded-graphics-core`](https://crates.io/crates/embedded-graphics-core) crate to integrate with embedded-graphics.
+### Monospace fonts
```
rfuest, line 287: This section needs to be updated when #596 gets merged.

MIGRATING-0.6-0.7.md
```diff
@@ -181,25 +231,129 @@ All `draw_*` methods to draw specific primitives (`draw_circle`, `draw_triangle`
...
+
+### More complex fonts
+
+Crates that handle text rendering more complex than simple monospace fonts should now implement the
+`CharacterStyle` and `TextRenderer` traits. These are used for both text styling and layout.
+
+Please refer to their respective docs for implementation details.
 
-TODO: Improve this section before release. Text rendering isn't included in `core` anymore.
+An implementation of more complex font rendering using BDF font files is available in the [eg-bdf](https://github.com/embedded-graphics/bdf/tree/4eb67f99e5f3a1eba6aea4d57cb4642727a95f8f/eg-bdf) crate, which may be useful as a reference for other implementations.
```
rfuest, line 356: Should we link to a specific revision in this case? eg-bdf will get more commits until e-g is released. I would suggest to simply link to the bdf repo. Once the README in the bdf repo is updated eg-bdf should be easy to find.

MIGRATING-0.6-0.7.md
```diff
@@ -281,10 +435,9 @@ For example, a styled rectangle is now built like this:
 
 ## Primitives
 
-### Styling
+Previously, drawing a filled shape with a transparent stroke of non-zero width would bleed the fill under the stroke. This is changed in 0.7 to honor the stroke color, even if it is `None`, allowing for filled shapes with transparent borders.
 
-TODO: Drawing primitives with transparent strokes will now affect the fill area.
-TODO: The behavior of the `Dimensions` impls has changed
+A stroke or fill color of `None` now does not affect the primitive's dimensions.
```
rfuest, line 440: IMO we should emphasize that the "dimensions" are the return value of `Dimensions::bounding_box`:
```suggestion
The color of stroke and fill no longer affects the primitive's bounding, that is returned by `Dimensions::bounding_box`. The stroke width is now always considered even if the stroke is transparent.
```

MIGRATING-0.6-0.7.md
```diff
@@ -281,10 +435,9 @@ For example, a styled rectangle is now built like this:
 
 ## Primitives
 
-### Styling
+Previously, drawing a filled shape with a transparent stroke of non-zero width would bleed the fill under the stroke. This is changed in 0.7 to honor the stroke color, even if it is `None`, allowing for filled shapes with transparent borders.
```
rfuest, line 438:
```suggestion
Previously, drawing a filled shape with a transparent stroke of non-zero width would bleed the fill under the stroke. This is changed in 0.7 to honor the stroke width and alignment, even if it is the stroke color is `None`, allowing for filled shapes with transparent borders.
```

MIGRATING-0.6-0.7.md
```diff
@@ -396,4 +549,14 @@ Note that the table above shows the new fonts' `ascii` variants only. Some new f
 
 To style fonts, use the `MonoTextStyle` struct. `TextStyle` still exists, but has been repurposed to provide a more general interface to text styling. This more closely mirrors the way primitives are built and styled.
 
-TODO: Discuss baseline font alignment, etc, as per GH comment.
+The default vertical alignment for fonts is now the font's baseline. To retain the 0.6 behaviour and position text using its top-left corner, set the `baseline` property to `Baseline::Top`:
```
rfuest, line 552: We should use "baseline" instead of "vertical alignment" to be consistent with the other docs.

src/text/renderer/mod.rs
```diff
@@ -1,7 +1,13 @@
 //! Text renderer.
 //!
-//! TODO: Describe how this API can be implemented and add an example to the docs or link to an
-//!       external example.
+//! For more complex font rendering cases that are not covered by the [`mono_font`] module, the
+//! `TextRenderer` trait can be implemented for a font.
```
rfuest, line 4:
```suggestion
//! For more complex text rendering cases that are not covered by the [`mono_font`] module, the
//! `TextRenderer` trait can be implemented by external text renderers.
```

src/text/renderer/mod.rs
```diff
@@ -1,7 +1,13 @@
...
+//!
+//! An implementation can be found in the [embedded-bdf] repository and may be useful as a reference
+//! for other implementations.
```
rfuest, line 7:
```suggestion
//! Implementations of this trait can be found in the [bdf] and [eg-seven-segment] repositories, which might be useful as a reference of other implementations.
```

src/text/renderer/mod.rs
```diff
@@ -1,7 +1,13 @@
...
+//!
+//! [`mono_font`]: ../../mono_font/index.html [embedded-bdf]:
+//! https://github.com/embedded-graphics/bdf/tree/a73a34cf45a5ef90cc7441afc12ec611cfe15563/eg-bdf
```
rfuest, line 10:
```suggestion
//! [`mono_font`]: ../../mono_font/index.html
//! [bdf]: https://github.com/embedded-graphics/bdf
//! [eg-seven-segment]: https://github.com/embedded-graphics/eg-seven-segment
```

rfuest (changes requested):
Some notes

MIGRATING-0.6-0.7.md
jamwaffles: Thank you! I was struggling to find good wording for this behaviour.

MIGRATING-0.6-0.7.md
jamwaffles: "The default baseline for fonts is now the font's baseline" is strange to read here though. I'd be ok keeping it as "vertical alignment" as that describes the behaviour as opposed to the property. Can you think of a better way of bringing this across?

MIGRATING-0.6-0.7.md
jamwaffles: Good clarification, thanks. I added some text about the `Dimensions` trait bound and suggesting `OriginDimensions` instead. Let me know what you think - maybe it's superfluous and we should just mention `OriginDimensions`?

jamwaffles:
> If we release 0.7 with a MVP version, should we add some note about this to the migration guide?

That's a good idea. I added a line to the top of the guide. The only TODO left in the guide is the text stuff which I can have a crack at when I've got a spare minute, although I'm very out of touch with the text/font changes so I'll probably end up saying stupid things.

The new features section also needs improving, but I'll do that in a new PR as it needs moving out of the migration guide. I'd be ok with it just being in the GH release notes when 0.7.0 goes out, but we may also want to consider a `RELEASES.md` file that consolidates all changes for major versions. It's more to maintain but could be useful. Thoughts?

MIGRATING-0.6-0.7.md
```diff
@@ -1,26 +1,29 @@
 # Migrating from embedded-graphics 0.6.x to 0.7.0
 
+> Please note that this migration guide may be incomplete in some sections. If any missing or incorrect information is found, please [open an issue](https://github.com/embedded-graphics/embedded-graphics/issues/new) or [join the Element chatroom](https://matrix.to/#/#rust-embedded-graphics:matrix.org) to bring it to our attention.
```
rfuest, line 3: Nit:
```suggestion
> Please note that this migration guide may be incomplete in some sections. If any missing or incorrect information is found, please [open an issue](https://github.com/embedded-graphics/embedded-graphics/issues/new) or [join the Matrix chatroom](https://matrix.to/#/#rust-embedded-graphics:matrix.org) to bring it to our attention.
```
But otherwise I like it.

rfuest (changes requested):

rfuest:
I noticed that some new MockDisplay features are missing:
- All color types in e-g are now supported in patterns
- EG_FANCY_PANIC
- It doesn't impl `PartialEq` anymore. `eq` or `assert_...` can be instead.
- Should bounds and overdraw checking also be mentioned in the new features section?

If you don't want to work on these in the PR you can simply add them as TODOs. I didn't want to open another PR for the same file.

rfuest:
> That's a good idea. I added a line to the top of the guide. The only TODO left in the guide is the text stuff which I can have a crack at when I've got a spare minute, although I'm very out of touch with the text/font changes so I'll probably end up saying stupid things.

If you need any assistance let me know.

> The new features section also needs improving, but I'll do that in a new PR as it needs moving out of the migration guide. I'd be ok with it just being in the GH release notes when 0.7.0 goes out, but we may also want to consider a `RELEASES.md` file that consolidates all changes for major versions. It's more to maintain but could be useful. Thoughts?

The rust-lang `RELEASES.md` is really useful. So copying this would be a good idea IMO.

jamwaffles:
I think this is ready for a proper review now. I've moved the new stuff section into a `RELEASES.md` file. Can we make improvements to that in a new PR and ignore it for this one?

@rfuest would you be open to some sort of call to pair on finishing up these final bits of documentation? No worries if you don't feel comfortable/don't want to, but it could help get these last bits tied up :)

rfuest:
Sorry for taking so long to review this PR, I'm really busy this week. I'll try to find some time to review this.

jamwaffles:
Take your time! No sense burning yourself out for unpaid work ;)

<sub>or paid work for that matter</sub>

MIGRATING-0.6-0.7.md
```diff
@@ -368,6 +392,76 @@ The three methods in the `Dimensions` trait were replaced by a single `bounding_
...
+        .into_styled(PrimitiveStyle::with_fill(BinaryColor::On))
+        .draw(&mut display)?;
+
+    display.assert_eq(&expected);
+}
+```
+
+The `assert_pattern` and `assert_pattern_with_message` methods are also useful for checking display state in tests.
+
+```rust
```
rfuest, line 428:
```suggestion
```diff
```

MIGRATING-0.6-0.7.md
```diff
@@ -368,6 +392,76 @@ The three methods in the `Dimensions` trait were replaced by a single `bounding_
...
++
++     display.assert_pattern(&[
++         " # ",
++         "###",
++         " # ",
++     ]);
++ }
+```
+
+To see a visual representation of failing equality tests, set `EG_FANCY_PANIC=1` when running the tests.
```
rfuest, line 463: See comment above.

```suggestion
```

RELEASES.md
```diff
@@ -0,0 +1,137 @@
...
+
+The `DrawTargetExt` trait is introduced to allow a translated, cropped or clipped sub-area of a `DrawTarget` to be drawn to.
+
+`DrawTargetExt` is implemented for `DrawTarget`.
+
+Please search for `DrawTargetExt` at <https://docs.rs/embedded-graphics> for usage examples.
+
+### Fonts and text
+
+TODO(rfuest): Improve this section before release:
```
rfuest, line 73:
```suggestion
TODO(rfuest): Improve this section before release:
TODO(rfuest): Replace `ascii` and `latin1` with the new subsets
```

rfuest (changes requested):
Next review pass...

I have skipped the text/font related sections in my review, which I can update in a future PR.

MIGRATING-0.6-0.7.md
```diff
@@ -1,168 +1,89 @@
 # Migrating from embedded-graphics 0.6.x to 0.7.0
 
+> Please note that this migration guide may be incomplete in some sections. If any missing or incorrect information is found, please [open an issue](https://github.com/embedded-graphics/embedded-graphics/issues/new) or [join the Matrix chatroom](https://matrix.to/#/#rust-embedded-graphics:matrix.org) to bring it to our attention.
+
 ## Table of contents
 
 - [Migrating from embedded-graphics 0.6.x to 0.7.0](#migrating-from-embedded-graphics-06x-to-070)
   - [Table of contents](#table-of-contents)
```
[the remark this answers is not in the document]
jamwaffles: This is autogenerated by whatever VSCode plugin I'm using unfortunately. I could remove that and manually update the ToC but it would be nice (and reduce mistakes) to have it generated. Would you strongly prefer to remove these items?

MIGRATING-0.6-0.7.md
```diff
@@ -1,168 +1,89 @@
 # Migrating from embedded-graphics 0.6.x to 0.7.0
 
+> Please note that this migration guide may be incomplete in some sections. If any missing or incorrect information is found, please [open an issue](https://github.com/embedded-graphics/embedded-graphics/issues/new) or [join the Matrix chatroom](https://matrix.to/#/#rust-embedded-graphics:matrix.org) to bring it to our attention.
+
 ## Table of contents
```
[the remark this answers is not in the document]
jamwaffles: Similar to above comment, although we could reorder the sections below the ToC?

MIGRATING-0.6-0.7.md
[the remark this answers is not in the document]
rfuest: I would prefer it, but not enough to warrant any extra manual work. Just leave the items in the TOC, if there is no option in the plugin to remove them.

MIGRATING-0.6-0.7.md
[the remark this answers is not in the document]
rfuest: Sorry my comment wasn't very cIear. I meant that the sections should be reorder and not only the ToC.

MIGRATING-0.6-0.7.md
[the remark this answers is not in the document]
jamwaffles: I found some options to remove the top level heading and the recursive item for the ToC so thankfully this is sorted now.

MIGRATING-0.6-0.7.md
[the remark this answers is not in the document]
jamwaffles: That makes more sense, thanks :). I've moved those sections to the bottom of the guide. This also puts `# General` as the first item which is a much better position for it.

MIGRATING-0.6-0.7.md
```diff
@@ -1,168 +1,89 @@
...
-Primitives now support stroke alignment using the `StrokeAlignment` enum.
-
-### Geometry
-
-Adds the `Point::x_axis`, `Point::y_axis`, `Size::x_axis` and `Size::y_axis` methods to get one axis of a point or size, with the other set to zero.
+## For display driver authors
 
-```rust
-use embedded_graphics::geometry::Point;
+Driver authors should use `DrawTarget` exported by the [`embedded-graphics-core`](https://crates.io/crates/embedded-graphics-core) crate to integrate with embedded-graphics.
```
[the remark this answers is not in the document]
jamwaffles: I've added a short section about `core`. I think it could use some improvement but it's a start. Eager to see your suggestions :)

MIGRATING-0.6-0.7.md
jamwaffles: I've updated the section. Let me know what you think.

MIGRATING-0.6-0.7.md
jamwaffles: Just bumping this in case you missed it :)

MIGRATING-0.6-0.7.md
```diff
@@ -396,4 +288,259 @@ Note that the table above shows the new fonts' `ascii` variants only. Some new f
...
+
+These methods aim to be more compatible with hardware-accelerated drawing commands. Where possible, embedded-graphics drawables will use `fill_contiguous` and `fill_solid` to improve performance, however may fall back to `draw_iter` by default.
+
+To reduce duplication, please search the `DrawTarget` documentation on <https://docs.rs/embedded-graphics-core> for more details on the usage and arguments of the above methods.
+
+## For crates that handle images
+
+Crates that handle images must now implement the `ImageDrawable` and `OriginDimensions` traits from [`embedded-graphics-core`](#the-embedded-graphics-core-crate) to integrate with embedded-graphics.
+
+The below examples shows an implementation for an imaginary `MyRgb888Image` which uses 24 bit color and draws to targets that support the same.
```
rfuest, line 400: The second part of the sentence could be interpreted as a limitation of the example, but it's an inherent property of the image API that the color types must be the same.

```suggestion
The below examples shows an implementation for an imaginary `MyRgb888Image` which uses 24 bit RGB color.
```

MIGRATING-0.6-0.7.md
```diff
@@ -396,4 +288,259 @@ Note that the table above shows the new fonts' `ascii` variants only. Some new f
...
+        // Return image width and height in pixels
+    }
+}
+```
+
+## For text rendering crates
+
+### Monospace fonts
+
+Monospaced fonts should now be built by creatign a `MonoFont` struct and assigning it to a `const`. The following example shows a migration of a font with 5x9 pixel characters using the Latin-1 encoding.
```
rfuest, line 445:
```suggestion
Monospaced fonts are no longer using a separate type per font and are now defined by using a `MonoFont` object. In most applications fonts will be declared as a compile time constant, but fonts can now also be loaded or generated at runtime.

The following example shows a migration of a font with 5x9 pixel characters using the ISO-8859-1 (Latin-1) encoding.
```

MIGRATING-0.6-0.7.md
```diff
@@ -396,4 +288,259 @@ Note that the table above shows the new fonts' `ascii` variants only. Some new f
...
++         160,
++     ),
++     // A glyph mapping with base ASCII range and Latin1 range.
++     glyph_mapping: &StrGlyphMapping::new("\0 ~\0\u{00A0}ÿ", '?' as usize - ' ' as usize),
++     character_size: Size::new(5, 9),
++     character_spacing: 0,
++     baseline: 7,
++     underline: DecorationDimensions::new(8, 1),
++     strikethrough: DecorationDimensions::new(4, 1),
++ };
```
rfuest, line 498: I would suggest to tweak the example a bit, so that we can use a predefined mapping:

```suggestion
- // Map a given character to an index in the glyph bitmap
- fn char_offset_impl(c: char) -> u32 {
-     let fallback = '?' as u32 - ' ' as u32;
-     if c < ' ' {
-         return fallback;
-     }
-     if c <= '\u{007f}' {
-         return c as u32 - ' ' as u32;
-     }
-     if c < '\u{00A0}' || c > 'ÿ' {
-         return fallback;
-     }
-     c as u32 - ' ' as u32 - 32
- }
-
- #[derive(Debug, Copy, Clone)]
- pub struct ExampleFont {}
- impl Font for ExampleFont {
-     const FONT_IMAGE: &'static [u8] = include_bytes!("../data/ExampleFont.raw");
-     const CHARACTER_SIZE: Size = Size::new(5, 9);
-     const FONT_IMAGE_WIDTH: u32 = Self::CHARACTER_SIZE.width * CHARS_PER_ROW;
-
-     fn char_offset(c: char) -> u32 {
-         char_offset_impl(c)
-     }
- }
+ use embedded_graphics::{
+     geometry::Size,
+     image::ImageRaw,
+     mono_font::{mapping::ISO_8859_1, DecorationDimensions, MonoFont},
+ };
+
+ pub const EXAMPLE_FONT: MonoFont = MonoFont {
+     image: ImageRaw::new_binary(
+         // This example uses 32 characters per row, each character 5px across.
+         32 * 5,
+     ),
+     glyph_mapping: &ISO_8859_1,
+     character_size: Size::new(5, 9),
+     character_spacing: 0,
+     baseline: 7,
+     underline: DecorationDimensions::new(8, 1),
+     strikethrough: DecorationDimensions::new(4, 1),
+ };
```

MIGRATING-0.6-0.7.md
```diff
@@ -396,4 +288,259 @@ Note that the table above shows the new fonts' `ascii` variants only. Some new f
...
+    }
+    if c <= '~' {
+        return c as usize - ' ' as usize;
+    }
+    if c < '\u{00A0}' || c > 'ÿ' {
+        return fallback;
+    }
+    c as usize - ' ' as usize - 33
+}
+```
```
rfuest, line 537: Now that the example above doesn't use `StrGlyphMapping` anymore I would suggest to use a single example to demonstrate a string mapping and using a function. I've also changed the mapping to one that isn't included as a predefined mapping.

```suggestion
Custom mappings between characters and glyph positions can be used by using `StrGlyphMapping`, using a function or implementing the `GlyphMapping` trait:

```rust
use embedded_graphics::{
    geometry::Size,
    image::ImageRaw,
    mono_font::{mapping::StrGlyphMapping, DecorationDimensions, MonoFont},
};

pub const EXAMPLE_FONT: MonoFont = MonoFont {
    image: ImageRaw::new_binary(
        include_bytes!("../data/digits.raw"),
        // In this example, this equals 32 characters per row, each character 15px across.
        10 * 15,
    ),
    glyph_mapping: &StrGlyphMapping::new("\009", 0),
    // or use a function:
    // glyph_mapping: &digit_mapping,
    character_size: Size::new(15, 30),
    character_spacing: 5,
    baseline: 29,
    underline: DecorationDimensions::default_underline(30),
    strikethrough: DecorationDimensions::default_strikethrough(30),
};

fn digit_mapping(c: char) -> usize {
    if c >= '0' || c <= '9' {
        c as usize - '0' as usize
    } else {
        0
    }
}
```

rfuest (changes requested):
Some more progress on the review.

MIGRATING-0.6-0.7.md
```diff
@@ -396,4 +288,259 @@ Note that the table above shows the new fonts' `ascii` variants only. Some new f
...
+
+- `clear`
+
+  Fills the entire display with a solid color.
+
+These methods aim to be more compatible with hardware-accelerated drawing commands. Where possible, embedded-graphics drawables will use `fill_contiguous` and `fill_solid` to improve performance, however may fall back to `draw_iter` by default.
+
+To reduce duplication, please search the `DrawTarget` documentation on <https://docs.rs/embedded-graphics-core> for more details on the usage and arguments of the above methods.
+
+## For crates that handle images
```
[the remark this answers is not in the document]
jamwaffles: I've clearly struggled with this too. Any of these?

- Image format support crates
- Image decoding crates
- Image crates

I'm not keen on any of them but it might inspire something better from your side.

jamwaffles:
Thanks for the next pass! Suggestions applied, aside from the image crate naming comment.

MIGRATING-0.6-0.7.md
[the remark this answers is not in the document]
rfuest: Out of these three I think I would use "Image format support creates". It's long, but is the most accurate description. Should we just go with that for now and change it later if we find a better description?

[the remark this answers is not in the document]
jamwaffles: There's a golden phrase out there for sure, but yeah for now I've gone with the verbose option.

rfuest (approved):

jamwaffles:
Many thanks! What's left for the 0.7.0 release? We need to fill out RELEASES.md for sure. Anything else you can think of?

rfuest:
I would like to get at least a MVP version of `eg-bdf` and `eg-seven-segment` ready to be released together with e-g 0.7. But I haven't made a lot of progress recently and don't want to delay 0.7 any further.

I'll take another look at e-g on the weekend to check if there are any other issues I would like to fix before 0.7. There was one text rendering related thing that recently came up with e-t, but that might be solvable by only updating the documentation.

jamwaffles:
Sounds good. I should be around if you need any assistance. Getting close!

jamwaffles:
Looking through the issue list, I couldn't find anything that looks like what you mentioned. Can you assign the 0.7 release blocker tag if/when you find it?

PR #596, opened 2021-05-18
Add additional glyph subsets

rfuest:
This PR adds additional glyph subsets for the bundled `MonoFont`s.

While the existing `ascii` and `latin1` covered many languages many others weren't supported at all. This especially applied to non roman alphabets, like Cyrillic or Greek. With this PR support for ISO 8859 variants that don't require support for RTL or composite characters is added. (@bugadani won't need to substitute "ŐőŰű" with umlauts anymore) The 8859-15 subset will also be useful for anyone who needs the euro sign.

I've also added JIS_X0201 to support Japanese katakana signs. I don't know much about the Japanese writing system, but I hope this is useful and this is at least very similar to the character set from HD44780 based displays.

Some additional notes:
* Most code is generated by a new tool that will be added to the `bdf` repo.
 The conversion tool in this repo is deprecated and will be removed in a future PR. With the new tooling it will be possible to generate `MonoFont`s and/or fonts for `eg-bdf`. The tool works similar to `bindgen` and can be used on the command line or in build scripts.
* I had to remove `MonoFontBuilder` because of limitations for const functions.
  The builder did calculate some default values, which made it easier to use for basic fonts. But IMO most fonts will be generated, so this shouldn't be a big problem.
* Some fonts don't include the glyphs for the JIS_X0201 subset.
 They are currently included anyway and only contain question marks in the non ASCII region. These will be removed in a future PR.


src/mono_font/mapping.rs
```diff
@@ -0,0 +1,377 @@
...
+        pub enum Mapping {
+            $(
+                $(#[$meta])*
+                $enum_variant,
+            )*
+        }
+
+        impl Mapping {
+            /// Returns an iterator over all mappings.
+            pub fn all() -> impl Iterator<Item = Self> {
```
bugadani, line 171: I'd prefer `into_iter()` here.

bugadani:
> @bugadani won't need to substitute "ŐőŰű" with umlauts anymore

We still have í and stuff ;) But I'm guessing that was just the non-exhaustive example.
Have to say, I love the fact this PR is so large github can't even handle file filtering.

rfuest:
> We still have í and stuff ;)

https://en.wikipedia.org/wiki/ISO/IEC_8859-2 should be enough to fix to editor example in e-t ;).

src/mono_font/mapping.rs
```diff
@@ -0,0 +1,377 @@
...
+//! let mapping = StrGlyphMapping::new("abcdef1234", 0);
+//! assert_eq!(mapping.index('a'), 0);
+//! assert_eq!(mapping.index('b'), 1);
+//! assert_eq!(mapping.index('1'), 6);
+//! assert_eq!(mapping.index('2'), 7);
+//! ```
+//!
+//! This direct mapping is inefficient for mappings that map consecutive ranges of characters to
+//! consecutive index ranges. To define a range of characters a `\0` character followed by the
+//! start and end characters of the inclusive range can be used. This way the mapping in previous
```
jamwaffles, line 31:
```suggestion
//! start and end characters of the inclusive range can be used. This way the mapping in the previous
```

src/mono_font/mod.rs
```diff
@@ -12,14 +12,14 @@
 //!
 //! # Built-in fonts
 //!
-//! Each built-in font is provided in an ASCII and a Latin 1 (ISO 8859-1) variant. The ASCII variant
-//! contains a smaller subset of characters which saves memory in embedded application, but only
-//! covers all characters of the English language. The Latin 1 variant has complete coverage for
-//! [the languages listed here](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Modern_languages_with_complete_coverage),
-//! and partial coverage for [these languages](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Languages_with_incomplete_coverage).
+//! Each built-in font is provided in different glyph subsets. The ASCII variant is the smallest
+//! subset which saves memory in embedded application, but only covers all characters of the English
```
jamwaffles, line 16:
```suggestion
//! subset which saves memory in embedded applications, but only covers all characters of the English
```

src/mono_font/mod.rs
```diff
@@ -12,14 +12,14 @@
...
+//! language. The ISO 8859 subsets support a wide range of languages, see
+//! [Wikipedia](https://en.wikipedia.org/wiki/ISO/IEC_8859#The_parts_of_ISO/IEC_8859) for a list of
+//! languages.
 //!
-//! The table below shows the ASCII variant of the built-in fonts. See the [`latin1` module] for
-//! an overview of the complete character set included in the Latin 1 variants.
+//! The table below shows the ASCII variant of the built-in fonts. See the subset modules for
```
jamwaffles, line 21:
```suggestion
//! The table below shows the ASCII variant of the built-in fonts. See the [subset modules](#modules) for
```

src/mono_font/mono_text_style.rs
```diff
@@ -1170,4 +1154,30 @@ mod tests {
...
+    #[test]
+    fn builder_change_font() {
+        let _style = {
+            let font = MonoFont {
+                image: ImageRaw::new_binary(&[1, 2, 3], 1),
+                character_size: Size::new(1, 2),
+                character_spacing: 0,
+                baseline: 0,
+                strikethrough: DecorationDimensions::default_strikethrough(2),
+                underline: DecorationDimensions::default_strikethrough(2),
```
jamwaffles, line 1167: Should this be?

```suggestion
                underline: DecorationDimensions::default_underline(2),
```

Doesn't matter much for a test but probably good for consistency/regression catching.

jamwaffles (changes requested):
Mostly small doc nits, otherwise looks really good! I'm glad we can support more languages now :).

I tested these changes by porting the [profont](https://crates.io/crates/profont) crate and it worked well there. Getting strikethrough/underline for "free" is a great addition. It took me a few moments to figure out the glyph mapping stuff but the docs helped so it wasn't too difficult. 

I like the hijacking of `\0` to separate glyph ranges, but what was the reasoning to use that character instead of the slice of ranges used previously? It feels a bit hacky to be "parsing" strings at runtime (plus a bit of performance impact?), but if there's no better solution I'm cool with it.

jamwaffles:
If you're curious, the WIP Profont upgrade is here: https://github.com/jamwaffles/profont/blob/4e41953f5a48a889451a73c2cec7d0278365cc86/src/lib.rs#L45-L58

There was originally [this character mapping function](https://github.com/jamwaffles/profont/blob/4e41953f5a48a889451a73c2cec7d0278365cc86/src/lib.rs#L60-L72) which works as is with only a change from `u32` -> `usize`, although I switched it out for a `&StrGlyphMapping::new("\0 ~\0\u{00A0}ÿ", '?' as usize - ' ' as usize)` which is a really nice oneline (modulo my comments about string parsing above).

rfuest:
> I like the hijacking of `\0` to separate glyph ranges, but what was the reasoning to use that character instead of the slice of ranges used previously? It feels a bit hacky to be "parsing" strings at runtime (plus a bit of performance impact?), but if there's no better solution I'm cool with it.

This is a trade off between size and performance. Some mappings, like ISO_8859_2, are quite large and don't contain many contiguous ranges of unicode code points.  By using a string we can reuse the space efficient UTF-8 encoding. The previous implementation used two `u32`s per range, which was up to 8 times more data for a single character.

I hope that the performance won't be a huge issue, because even in non english languages most characters will be in the ASCII range, which should limit the impact of the expensive linear search for glyphs.

If we later notice that there is a better way to encode this or we notice performance problems we can still change it by implementing a new `GlyphMapping`.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#587](https://github.com/embedded-graphics/embedded-graphics/pull/587) Added `From<&TextStyle>` impl for `TextStyleBuilder` and `From<&MonoFont>` for `MonoFontBuilder`.
 - [#589](https://github.com/embedded-graphics/embedded-graphics/pull/589) Implemented `From` trait to convert from RGB colors to grayscale colors, between different grayscale colors and from grayscale and RGB colors to `BinaryColor`.
 
+### Changed
+
+- **(breaking)** - [#596](https://github.com/embedded-graphics/embedded-graphics/pull/596) Added more glyph subsets for bundled `MonoFont`s. The `latin1` subset was renamed to `iso8859-1`.
```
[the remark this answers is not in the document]
rfuest: IMO your first suggestion is fine. More details can be added to the migration guide.

src/mono_font/mono_text_style.rs
rfuest: Good catch.

jamwaffles (approved):
Ah I see, that's a neat solution. We can also make the C programmers out there cry with `\0` everywhere ;)

The rest of the non-generated changes look good. Should we do another beta release seeing as this is a breaking change?

I also remember you mentioning some more text/font work. Is this PR the last of that?

bugadani:
>  Should we do another beta release seeing as this is a breaking change?

I'd appreciate it as it would give me at least the chance to try and catch up with e-t :)

rfuest:
> > Should we do another beta release seeing as this is a breaking change?
> 
> I'd appreciate it as it would give me at least the chance to try and catch up with e-t :)

Yes, definitely. And we also need the beta to update the examples.

src/mono_font/mapping.rs
rfuest: Sorry, I had missed this comment. Why do you think that `into_iter` would be better? For me `into_iter` should be used to convert a value into an iterator, but this function doesn't depend on a value.

jamwaffles:
I just released 0.7.0-beta.2 :+1:

src/mono_font/mapping.rs
bugadani: @rfuest I guess github doesn't like to notify about comments (as I missed your response as well).

Technically, Mapping _is_ a "value" (well, an enum, sure) and this method returns an iterator to this. Maybe I should change my preference to `iter()` as there is no actual conversion.

My main issue with `all()` is that `all` is an iterator method and if you ever have to write `Mapping::all().all([...])` it gets at least visually confusing.

rfuest: > My main issue with `all()` is that `all` is an iterator method and if you ever have to write `Mapping::all().all([...])` it gets at least visually confusing.

That does look confusing, but IMO it's very unlikely that anyone will ever need to use this construct.

I'm be OK with changing it to `iter`. While all `iter` methods in the standard library are methods and not associated functions, the `strum` crate does also use `iter` in the same way.

If the name bothers you enough feel free to open a PR.

PR #602, opened 2021-06-05
Implement Display for Point and Size

### LINKED ISSUE #601: Implementing Display for Point and Size
Should we implement the `core::fmt::Display` trait for `Point` and `Size`?

The use case I have in mind for this are error messages for assertions. Using the `Debug` output can get noisy and you need to manually print the components if you want a cleaner output:
```rust
// without `Display` impl
assert_eq!(
    point, other_point,
    "point ({}, {}) must be equal to other_point ({}, {})",
    point.x, point.y,
    other_point.x, other_point.y
);

// with `Display` impl
assert_eq!(point, other_point, "point ({}) must be equal to other_point ({})", point, other_point);
```

If we decide to implement this, how should the output look of the `Display` impls look?
`Point::new(1, 2)` -> `1, 2`
`Size::new(1, 2)` -> `1x2` or `1 x 2` or also `1, 2`?

rfuest:
This PR implements the `Display` trait for `Point` and `Size`, as discussed in #601.

I've slightly changed the output for `Size`s: `123 x 456` instead of `123x456`, because it makes the output more readable IMO.

Closes #601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- [#602](https://github.com/embedded-graphics/embedded-graphics/pull/602) Implemented `core::fmt::Display` for `Point` and `Size`.
+
 ### Changed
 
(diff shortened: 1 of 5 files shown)


jamwaffles (approved):
LGTM! I guess this doesn't need any new beta releases for core/e-g, right?

rfuest:
> I guess this doesn't need any new beta releases for core/e-g, right?

No it doesn't.

PR #603, opened 2021-06-05
Replace MockDisplay::eq by PartialEq impl

rfuest:
When we introduced `EG_FANCY_PANIC` I had decided to remove the `PartialEq` impl for `MockDisplay` to make sure the new assertions are used. I think this was useful for the transition, but it might not be the best permanent solution. This PR reverts this change, which simplifies the migration guide and makes clippy a little happier.


jamwaffles (approved):
LGTM :shipit:

PR #604, opened 2021-06-05
Remove incomplete JIS X 201 fonts

rfuest:
Not all of our source BDF fonts contain the required glyphs for the JIS X 201 subset. This was supposed to be detected by the font conversion tool, but I decided to just remove them manually now and fix the conversion tool later.


jamwaffles (approved):
Good catch, lgtm

PR #610, opened 2021-06-22
Update release docs with upstream notes

jamwaffles:

## PR description

I also made the release Just commands require at least one argument now. Not sure if this will be a hindrance, but it's a little bit more safety around requiring at least a version/level to release.

### DIFF
--- a/doc/releasing.md
+++ b/doc/releasing.md
@@ -27,13 +27,13 @@ Target audience: crate maintainers who wish to release `embedded-graphics` or `e
 - Double check the release level (major, minor, patch)
 - Release the crates in the following order:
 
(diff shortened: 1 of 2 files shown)


rfuest (approved):

jamwaffles:
Argh, the was-beta-now-stable docs build error is failing. I'll force merge this PR and fix that error later. Maybe time to bump MSRV?...

rfuest:
> Maybe time to bump MSRV?...

I'm not sure what the best way to raise the MSRV would be. A MSRV change would be a breaking change and should therefore be included in e-g 0.8. But as long as we might want to release more patch releases for 0.7 we cannot easily change the MSRV of the master branch.

Maybe we need to reconsider having a development branch and one for maintenance releases.

Issue #609, opened 2021-06-21
Iterating over pixels in `RawImage` and `Image`


jamwaffles:
Context: starting [from this message](https://app.element.io/#/room/#rust-embedded-graphics:matrix.org/$6nOEaC4HWJjrwKGiwDDNHhPgHDRgBCBJmlx5f3dbMdQ) by @jacobrosenthal:

> We have all the sexy new from impls for pixels colors, i was hoping to be able to say greyscale a rgb rawimage or image, either wholesale `rawimage.into()` or via a pixel iterator `image.into_iter().map(|p|Gray8::from(p))).collect()` but I cant seem to figure a way, am I missing anything?

Did we miss out adding methods to iterate over the pixels in both a `RawImage` an `Image`? I think these could be added pretty easily, and would be useful methods to have.

rfuest:
The need for providing a pixel iterator was deliberately removed from `ImageDrawable`, because it isn't possible to efficiently implement a pixel iterator for all image drawables. `ImageDrawable` could, for example, also be implemented for a vector graphics file.

We had discussed this earlier and decided to add additional traits if these features are required later. One trait could be used to provide an iterator over all pixels and another trait to access individual pixels: `fn pixel(Point) -> C`.

jacobrosenthal:
For context I'm basically trying to use eg as the image-rs library. Run
convolution over images etc.

rfuest:
I've created PR #612 to add a `pixel` getter trait and implement it for `ImageRaw`s.

> For context I'm basically trying to use(abuse) eg as the image-rs library. Run convolution over images etc.

Interesting. Do you plan to output the result to a `DrawTarget` or generate another image?

jacobrosenthal:
I might need an intermediate image to do more work, but that wouldnt be the
goal. The end result would be the DrawTarget

LLFourn:
I was looking for something like `Framebuffer::iter_pixels`. I was able to produce it with `RawDataSlice` manually but it was tricky to figure out. It seems like `ImageRaw` and `FrameBuffer` would benefit from helper methods.

rfuest:
`GetPixel` trait implementations for `Framebuffer` and `ImageRaw` have been added since this PR was created. It is now possible to iterate over the pixels in an image by using code like this:
```rust
let fb = Framebuffer::<...>::new();
let pixels_iter = fb.bounding_box().points().map(|p| (p, fb.pixel(p));
```
[...]

PR #612, opened 2021-06-28
Add ImagePixelGetter trait

rfuest:
This PR adds an `ImagePixelGetter` trait which can optionally be implemented by `ImageDrawable`s to allow access to the pixel data.


jamwaffles (approved):
Over a year old... apologies.

Not sure if this closes #609 as that issue mentions iterators.

rfuest:
> Over a year old... apologies.

Don't worry, I would have bugged you earlier when this would have been important.

> Not sure if this closes #609 as that issue mentions iterators.

Iterators are a separate issue IMO. The `pixel` getter can be used as a replacement, but wouldn't be as efficient if you need to iterate over all pixels.

PR #614, opened 2021-07-11
Fix link to migration guide from 0.6 to 0.7

sirhcel:
Just fixes the link to the migration guide from 0.6 to 0.7 in `README.md`. Checked the link target in the preview of `README.md` manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ calculation.
 ## Migrating from older versions
 
 * [Migration guide from 0.5 to 0.6](https://github.com/embedded-graphics/embedded-graphics/blob/master/MIGRATING-0.5-0.6.md).
-* [Migration guide from 0.6 to 0.7](https://github.com/embedded-graphics/embedded-graphics/blob/master/MIGRATING-0.5-0.6.md).
+* [Migration guide from 0.6 to 0.7](https://github.com/embedded-graphics/embedded-graphics/blob/master/MIGRATING-0.6-0.7.md).
 
 ## Implementing `embedded_graphics` support for a display driver
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -109,7 +109,7 @@
 //! # Migrating from older versions
 //!
 //! * [Migration guide from 0.5 to 0.6](https://github.com/embedded-graphics/embedded-graphics/blob/master/MIGRATING-0.5-0.6.md).
-//! * [Migration guide from 0.6 to 0.7](https://github.com/embedded-graphics/embedded-graphics/blob/master/MIGRATING-0.5-0.6.md).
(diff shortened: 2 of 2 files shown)


sirhcel:
Reading through the checklist and deciding that they don't apply to this change, i forgot something: Saying a big thank you for providing and maintaining this crate! And especially providing a migration guide (which brought me here)! :)

jamwaffles (changes requested):
Thanks for the fix! The readme is generated from `src/lib.rs`, so the [associated line 112 there](https://github.com/embedded-graphics/embedded-graphics/blob/master/src/lib.rs#L112) needs changing too. 

If you have `just` installed, running `just generate-readmes` will update everything, otherwise you can make the change manually in `src/lib.rs`.

jamwaffles:
>  a big thank you for providing and maintaining this crate! And especially providing a migration guide (which brought me here)! :)

Thank you from all of us :blush: I'm pretty proud of our docs, so it's great to hear they're useful!

jamwaffles (changes requested):
Please add the same change to `src/lib.rs` or run `just generate-readmes` and commit both files - the build should pass then.

sirhcel:
Thank you for pointing me to `just`! I managed to re-generate the readmes but missed `git push --force` before re-requesting the review. I'm sorry for bothering you with this request!

Things should be good to go now.

sirhcel:
> Things should be good to go now.

Well ... apparently not. When running `just build-without-fmt-check` locally it fails at link checks:
```
[...]
cargo doc --workspace --features "nalgebra_support fixed" --no-deps
 Documenting embedded-graphics-core v0.3.2 ([...]/embedded-graphics/core)
 Documenting embedded-graphics v0.7.1 ([...]/embedded-graphics)
    Finished dev [unoptimized + debuginfo] target(s) in 2.84s
cargo deadlinks --dir target/doc/embedded_graphics
Found invalid urls in mono_font/struct.MonoFont.html:
        Fragment #tymethod.from_subset at mono_font/struct.MonoFont.html does not exist!
        Fragment #tymethod.vzip at mono_font/struct.MonoFont.html does not exist!
        Fragment #tymethod.is_in_subset at mono_font/struct.MonoFont.html does not exist!
        Fragment #tymethod.to_subset_unchecked at mono_font/struct.MonoFont.html does not exist!
[...]
error: Recipe `check-links` failed on line 77 with exit code 1
```
It does not look like there is a dead link/reference in the documentation because I can't find `from_subset` in the sources using `git grep` at all. This looks a bit mysterious to me at the moment.

jamwaffles (approved):
Thanks for the docs fix! 

> Well ... apparently not. When running just build-without-fmt-check locally it fails at link checks:

Don't worry about that - we've been seeing odd false positives like that for a while now so it's nothing to do with this PR :sweat_smile:

PR #619, opened 2021-08-07
Fix doc link false positives on stable builds

jamwaffles:

## PR description

Fixes the odd errors when building the docs with newer versions of Rust. This should make the beta build pass again too, as well as `stable`. I think the issue was caused by intra doc links being stabilised and/or doc  rendering changing slightly.


jamwaffles:
After some fettling, the only way I could find to fix both 1.40 and `stable` is to stop checking fragments which isn't ideal. When we migrate to intra-doc-links, we can remove all the fragment links anyway. We could also drop `cargo-deadlinks` at that point too, or use it to check external links, etc.

I've also had to pin Criterion to a specific version as it bumped the minimum MSRV to 1.46 in [0.3.5](https://github.com/bheisler/criterion.rs/blob/master/CHANGELOG.md#035---2021-07-26).

PR #621, opened 2021-08-23
add rgb666 support

almindor:

## PR description

Adds `Rgb666` support for displays such as the [ILI9486](https://www.aliexpress.com/item/1005002632269797.html?spm=a2g0s.9042311.0.0.68974c4dmBHD93)

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- [#621](https://github.com/embedded-graphics/embedded-graphics/pull/621) Added `Rgb666` and `Bgr666` color type support.
+
 ## [0.3.2] - 2021-06-05
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- [#621](https://github.com/embedded-graphics/embedded-graphics/pull/621) Added `Rgb666` and `Bgr666` color type support.
+
 ## [0.7.1] - 2021-06-15
 
 ### Changed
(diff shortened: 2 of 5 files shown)


almindor:
@jamwaffles ping? Also question, does it make any sense to create something like RawU18 for this?

core/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- Added `Rgb666` type support.
```
jamwaffles, line 11:
```suggestion
- [#621](https://github.com/embedded-graphics/embedded-graphics/pull/621) Added `Rgb666` color type support.
```

Please also add the same line to the top-level `CHANGELOG.md` too - these types are re-exported by e-g itself so it's good to let direct users know.

jamwaffles (changes requested):
Thanks for the nudge!

> does it make any sense to create something like RawU18 for this?

Yes, I think it does. The other colour types represent their exact storage so doing so with `Rgb666` keeps things consistent.

jamwaffles:
p.s. please rebase from master - the CI should start passing after #619 was merged.

almindor:
> p.s. please rebase from master - the CI should start passing after #619 was merged.

Seems there one more issue still in the CI for MSRV?

almindor:
I'm also unclear how to handle `ToBytes`for `RawU18` since padding is necessary for any kind of endianness swap logic to work.

Right now I do LE MCU to BE display this way:

```rust
        let mut iter = colors.into_iter().flat_map(|c| {
            let s = c.into_storage(); // bit-packed 18bits
                                      // we need to un-pack and pad with 2 bits each into 3 bytes of 6bit info
            let red = ((s & 0x3F) as u8) << 2;
            let green = (((s >> 6) & 0x3F) as u8) << 2;
            let blue = (((s >> 12) & 0x3F) as u8) << 2;
            [red, green, blue]
        });
```

almindor:
This should be good to go, the RawU18 representation is a bit funky tho when it comes to the endianness conversions, but then again it never really makes sense.

bugadani:
Could you please update the PR with a display that uses rgb666? Right now this looks like "yet another random color format". (I read your use case on matrix but IMO it would make sense to include it in the OP.)

jamwaffles (approved):
LGTM, thanks for your patience while I got round to this. I agree the `ToBytes` impl is a bit odd for 18 bit colours - it certainly makes more sense for multiples of 8 bits - but it might be useful for somebody, and it's consistent with the other RGB colours.

jamwaffles:
> Right now I do LE MCU to BE display this way:

Because `Rgb666` implements `RgbColor` you should be able to call `c.r()`, `c.g()` and `c.b()` to clean that code up a bit, FWIW.

jamwaffles:
This is released in e-g-core 0.3.3. e-g itself should pick the changes up with a `cargo update`.

PR #636, opened 2021-12-29
Set merge=union git attribute for CHANGELOG.md

### LINKED ISSUE #634: Add merge=union for CHANGELOG.md
Inspired by rust-embedded/embedded-hal#335. Should hopefully reduce the amount of small conflicts in PRs.

peterstuart:

## PR description

Set merge=union git attribute for CHANGELOG.md

Closes #634

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG.md merge=union


jamwaffles:
Sweet, thanks for the quick fix :). I can't see any issues with adding this attribute but I've asked for a quick review from @rfuest in case I've missed anything.

rfuest (approved):
The only possible problem I could imagine would be the ordering of changelog entries. They are currently ordered by issue number and automatic merging might put them in a different order.

But that's a pretty minor problem which could also be resolved by ordering the list once before a release or by just ignoring the ordering from now on.

jamwaffles:
Good point, although a minor one I agree. We could add a line to the release guide to check/apply ordering to the changelog entries, but it's not a huge deal if something slips through unordered.

jamwaffles:
Not sure why the build isn't being picked up. CircleCI is configured to build PRs. I'll merge this change anyway as it's probably not going to break CI.

PR #637, opened 2021-12-29
Add rust-toolchain.toml

jamwaffles:
I've deliberately left extra stuff like different targets and non-standard-development stuff used in CI out of this config as the build image we use preinstalls a few things for itself anyway.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,4 @@
+[toolchain]
+channel = "1.40.0"
+components = [ "rustfmt" ]
+profile = "minimal"


rfuest:
Would this require any contributor to install the `1.40.0` toolchain? This would make sure that CI wouldn't fail because of newer features, but it's also another hurdle for new contributors.

jamwaffles:
It will get Rustup to download 1.40.0 automatically if it's not already available when building anything in the repository. If the user isn't using Rustup I don't think it will make any difference.

It looks like an override can be set too with `rustup override set <version>`  which takes priority over the contents of `rust-version.toml`.

I see your point with it being a hurdle. For example, 1.40.0 isn't available for M1 Macs (which I recently switched to) so I've just been using latest stable. Perhaps it's better to just specify the minimum version in `Cargo.toml` and close this PR?

rfuest:
> Perhaps it's better to just specify the minimum version in `Cargo.toml` and close this PR?

Adding the minimum version to `Cargo.toml` as part of the MSRV bump is a good idea, regardless of this PR.

I'm unsure about this PR. Perhaps we just wait until the MSRV is bumped and specify a newer, M1 compatible, version in the toolchain file.

jamwaffles:
> I'm unsure about this PR. Perhaps we just wait until the MSRV is bumped and specify a newer, M1 compatible, version in the toolchain file.

Sure, let's go for that. I'll change this PR to draft until that point.

PR #638, opened 2021-12-29
Upgrade to ~1.56.1~ 1.57

jamwaffles:
I'll try to keep this tightly scoped to:

- [x] Updating  all 1.40.0 references to ~1.56.1 (as well as adding edition 2021 in certain places)~ 1.57
- [x] Using ~1.56.1 in CI - see https://github.com/embedded-graphics/build-tools/pull/2~ 1.57 - see https://github.com/embedded-graphics/build-tools/pull/3
- [x] Running `cargo fix --edition` - no changes
- [x] Fixing any other warnings - `clap` had some compile errors from some renames, otherwise seems to be pretty clear


justfile
```diff
@@ -2,7 +2,7 @@ targets := "arm-unknown-linux-gnueabi armv7-unknown-linux-gnueabihf x86_64-unkno
 
 target_dir := "target"
 
-# list of all features except criterion
+# list of all features
 all_features := "nalgebra_support fixed"
```
jamwaffles, line 6: We could probably just use the `--all-features` flag here now I think?

justfile
```diff
@@ -14,20 +14,15 @@ doc_assets_dir := doc_dir + "/assets"
 
 build: check-formatting check-drawing-examples build-without-fmt-check
 
-build-without-fmt-check: test test-all check-readmes check-links
-
-# TODO: Remove this and move `build-benches` back into `build-without-fmt-check` when MSRV is
-# increased enough to make all the benchmark tooling compile again. Also remove
-# `tools/generate-drawing-examples/rust-toolchain`.
-build-without-fmt-check-stable: test test-all build-benches check-readmes check-links
+build-without-fmt-check: test test-all check-readmes check-links build-benches
```
jamwaffles, line 17: Fixed hack from #632

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ### Changed
 
+- **(breaking)** - [#638](https://github.com/embedded-graphics/embedded-graphics/pull/638) Bump all crate Minimum Supported Rust Versions (MSRV) to 1.56.1, aka edition 2021.
```
jamwaffles, line 17: While this PR doesn't set any hard rules on MSRV, it changes it in the docs and CI so I thought it appropriate to add the changelog entry for it instead of a future one with more significant changes.

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ jobs:
   # Check that everything (tests, benches, etc) builds using the MSRV
   precheck-msrv:
     docker: &docker
-      - image: jamwaffles/circleci-embedded-graphics:1.40.0-3
+      - image: jamwaffles/circleci-embedded-graphics:1.56.1
```
rfuest, line 25:
```suggestion
      - image: jamwaffles/circleci-embedded-graphics:1.56.1-0
```

CHANGELOG.md
rfuest: Agreed.

Cargo.toml
```diff
@@ -34,12 +34,9 @@ micromath = { version = "1.1.0", default-features = false }
 embedded-graphics-core = { path = "core", version = "0.3.2"}
 byteorder = { version = "1.3.4", default-features = false }
 
-# criterion is not listed as a dev-dependency to work around a problem with no_std compatibility
-# and because dev-dependencies can't be optional
-criterion = { version = "=0.3.4", optional = true }
-
```
rfuest, line 40: So nice we can finally get rid of this ugly hack.

justfile
rfuest: I think so.

tools/bdf-to-mono/src/main.rs
```diff
@@ -2,19 +2,19 @@ use std::path::PathBuf;
...
 use bdf_parser::BdfFont;
 use bdf_to_mono::{Encoding, MonoFontData};
-use clap::Clap;
+use clap::Parser;
 
-#[derive(Clap)]
+#[derive(clap::Parser)]
 struct Args {
-    #[clap(about = "BDF input")]
+    #[clap(help = "BDF input")]
```
rfuest, line 9: I haven't tried to run this, but the `Cargo.toml` still specifies `clap` version `3.0.0-beta.2` and I think these changes were implemented in a later `clap` version. I would be OK with just leaving this crate as it was, because someday I might actually finish the replacement tool in the `bdf` repo.

rfuest (changes requested):

Cargo.toml
jamwaffles: Yeah! Much cleaner now :)

tools/bdf-to-mono/src/main.rs
jamwaffles: Hmm, if I lock the version to `"=3.0.0-beta.2"` (confirmed with `cargo tree`) and revert to the old code, I see the same compile errors as before this PR, so I'm not sure what's going on there, potentially caused by `cargo update`? I can revert this change and dig into why these errors are popping up but if this tool will be replaced eventually anyway, should we spend much time on it?

rfuest: No, don't spend any time on this and commit whatever works.

rfuest:
Did you look into the CI failure? Seems to be some kind of difference in the fixed point math, but why wasn't this showing up before?

jamwaffles:
I had to go do other stuff so not yet, but I saw the same issue on my laptop. Good to know it's reproducible. I'll fix it in a bit.

jamwaffles:
I'm unsure _why_ it fixed it, but reverting the `--all-features` commit back to a specific list makes things work again. I wonder if there's a feature within `fixed` or `micromath` or a dependency somewhere that gets turned on by `--all-features`? It's a really weird behaviour. Any ideas?

peterstuart:
I think it may be worth upgrading all the way to 1.57, since that release stabilized panicking in `const` functions, which is necessary to "constify" some of the constructors while preserving existing behavior. I'll have a  PR posted with that work shortly.

rfuest:
> Any ideas?

The `all_features` variable was wrong. It should have been `fixed_point` instead of `fixed`. Because of this error we never actually tested the fixed point variant of our code.

jamwaffles:
Oh that was silly 🤦 at least we caught it at some point...

Considering the change between fixed/non-fixed is quite minor, we could change the test params so it passes in both fixed and non-fixed cases. I'm not sure how strict you want to be about rendering differences between the two features but I'm personally ok with a couple of pixels here and there (as long as it's not a marker for a larger difference).

jamwaffles:
Oh that was silly 🤦 at least we caught it at some point...

Considering the change between fixed/non-fixed is quite minor, we could change the test params so it passes in both fixed and non-fixed cases. I'm not sure how strict you want to be about rendering differences between the two features but I'm personally ok with a couple of pixels here and there (as long as it's not a marker for a larger difference).

rfuest:
Agreed, at some point there will be a difference between fixed point and floating point arithmetic. I would suggest to disable the test (at least for the fixed point version) now. But to make sure it's not a larger problem I would add an item to #484 to take another look at it later.

jamwaffles:
[The](https://matrix.to/#/!SfJCDXZbMHXkPovtKL:matrix.org/$0tw18ST9nLRfy9fSyWjpgMRp4eGpAmbHXvU5accy4H0?via=matrix.org&via=envs.net&via=eea.crans.org) [polls](https://twitter.com/jam_waffles/status/1476954886445142017) I posted range from "idc what the MSRV is" to "just use 1.57 already", so I've updated this PR to target 1.57. Image already pushed in https://github.com/embedded-graphics/build-tools/pull/3.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ### Changed
 
+- **(breaking)** - [#638](https://github.com/embedded-graphics/embedded-graphics/pull/638) Bump all crate Minimum Supported Rust Versions (MSRV) to 1.57.
```
peterstuart, line 17: I think this is mistakenly under 0.7.1 instead of Unreleased.

jamwaffles: Good catch, fixed that. It also reminded me that I need to add the same item to `core/CHANGELOG.md` which is also done.

rfuest (approved):

jamwaffles:
Thanks everyone!

PR #639, opened 2021-12-31
Use const fn for constructors

peterstuart:

## PR description

This makes the following functions `const`:

- `Point::component_mul`
- `Point::component_div`
- `Size::saturating_add`
- `Size::saturating_sub`
- `Size::component_mul`
- `Size::component_div`
- `BinaryColor::invert`
- `BinaryColor::is_on`
- `BinaryColor::is_off`
- `Rectangle::with_center`
- `Rectangle::is_zero_sized`
- `RawDataSlice::new`
- `MonoTextStyle::new`
- `Arc::with_center`
- `Arc::from_circle`
- `Arc::to_circle`
- `Circle::with_center`
- `Ellipse::with_center`
- `Sector::with_center`
- `Sector::from_circle`
- `Sector::to_circle`
- `Styled::new`

One of biggest obstacles to making functions `const` was that the arithmetic operators can't be used in a `const` context except on integers. `Rectangle::with_center` in particular couldn't be `const` because of that limitation, and is used by several other constructors, so I added `const` versions of the necessary `sub` and `div` functions to `Point` and `Size` as a workaround so that I could make it `const`. I also had to bump the MSRV up to 1.57 since it stabilized panicking in `const` functions because `Point`'s implementation of `Sub<Size>` panics. Do you think it's worth it?

## Related

https://github.com/embedded-graphics/embedded-graphics/issues/205
https://github.com/embedded-graphics/embedded-graphics/pull/638

### DIFF
--- a/core/src/geometry/point.rs
+++ b/core/src/geometry/point.rs
@@ -196,7 +196,7 @@ impl Point {
     ///
     /// assert_eq!(result, Point::new(-40, 90));
     /// ```
-    pub fn component_mul(self, other: Self) -> Self {
+    pub const fn component_mul(self, other: Self) -> Self {
         Self::new(self.x * other.x, self.y * other.y)
     }
 
@@ -213,7 +213,7 @@ impl Point {
     ///
     /// assert_eq!(result, Point::new(2, -10));
     /// ```
-    pub fn component_div(self, other: Self) -> Self {
+    pub const fn component_div(self, other: Self) -> Self {
         Self::new(self.x / other.x, self.y / other.y)
     }
 }
--- a/core/src/geometry/point.rs
+++ b/core/src/geometry/point.rs
@@ -157,6 +157,24 @@ impl Point {
         Point::new(self.x.abs(), self.y.abs())
     }
 
+    /// Offsets a point by subtracting a size.
+    ///
+    /// This method provides a workaround for the `Sub` trait not being usable in `const` contexts.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if `width` or `height` are too large to be represented as an `i32`
+    /// and debug assertions are enabled.
+    pub const fn sub_size(self, other: Size) -> Point {
+        let width = other.width as i32;
+        let height = other.height as i32;
+
+        debug_assert!(width >= 0, "width is too large");
+        debug_assert!(height >= 0, "height is too large");
+
+        Point::new(self.x - width, self.y - height)
+    }
+
     /// Returns the componentwise minimum of two `Point`s
     ///
     /// # Examples
@@ -290,13 +308,7 @@ impl Sub<Size> for Point {
     /// This function will panic if `width` or `height` are too large to be represented as an `i32`
     /// and debug assertions are enabled.
     fn sub(self, other: Size) -> Point {
-        let width = other.width as i32;
-        let height = other.height as i32;
-
-        debug_assert!(width >= 0, "width is too large");
-        debug_assert!(height >= 0, "height is too large");
-
-        Point::new(self.x - width, self.y - height)
+        self.sub_size(other)
     }
 }
 
--- a/core/src/geometry/point.rs
+++ b/core/src/geometry/point.rs
@@ -165,7 +165,7 @@ impl Point {
     ///
     /// This function will panic if `width` or `height` are too large to be represented as an `i32`
     /// and debug assertions are enabled.
-    pub const fn sub_size(self, other: Size) -> Point {
+    pub(crate) const fn sub_size(self, other: Size) -> Point {
         let width = other.width as i32;
         let height = other.height as i32;
 
--- a/core/src/pixelcolor/binary_color.rs
+++ b/core/src/pixelcolor/binary_color.rs
@@ -84,10 +84,7 @@ impl BinaryColor {
     /// ```
     #[inline]
     pub const fn is_on(self) -> bool {
-        match self {
-            BinaryColor::On => true,
-            BinaryColor::Off => false,
-        }
+        matches!(self, BinaryColor::On)
     }
 
     /// Returns `true` if this color is `Off`.
@@ -101,10 +98,7 @@ impl BinaryColor {
     /// ```
     #[inline]
     pub const fn is_off(self) -> bool {
-        match self {
-            BinaryColor::On => false,
-            BinaryColor::Off => true,
-        }
+        matches!(self, BinaryColor::Off)
     }
 
     /// Maps active and inactive colors to a different type.
--- a/core/src/pixelcolor/binary_color.rs
+++ b/core/src/pixelcolor/binary_color.rs
@@ -66,7 +66,7 @@ impl BinaryColor {
     /// assert_eq!(BinaryColor::On.invert(), BinaryColor::Off);
     /// ```
     #[inline]
-    pub fn invert(self) -> Self {
+    pub const fn invert(self) -> Self {
         match self {
             BinaryColor::On => BinaryColor::Off,
             BinaryColor::Off => BinaryColor::On,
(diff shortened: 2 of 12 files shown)


.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ jobs:
   # Check that everything (tests, benches, etc) builds using the MSRV
   precheck-msrv:
     docker: &docker
-      - image: jamwaffles/circleci-embedded-graphics:1.56.1-0
+      - image: jamwaffles/circleci-embedded-graphics:1.57.0-0
```
peterstuart, line 25: It looks like this would require something like https://github.com/embedded-graphics/build-tools/pull/2 for this to work?

core/src/pixelcolor/binary_color.rs
```diff
@@ -83,8 +83,11 @@ impl BinaryColor {
     /// assert!(BinaryColor::On.is_on());
     /// ```
     #[inline]
-    pub fn is_on(self) -> bool {
-        self == BinaryColor::On
+    pub const fn is_on(self) -> bool {
+        match self {
```
peterstuart, line 87: `==` can't be used in a `const` function on `BinaryColor`, so I switched this to a `match`.

src/iterator/raw.rs
```diff
@@ -49,7 +49,7 @@ pub struct RawDataSlice<'a, R, BO> {
 
 impl<'a, R, BO> RawDataSlice<'a, R, BO> {
     /// Creates a new raw data slice.
-    pub fn new(data: &'a [u8]) -> Self {
+    pub const fn new(data: &'a [u8]) -> Self {
```
peterstuart, line 52: I'm not really familiar with how this is used. Maybe there's no point to making this one `const`?

core/src/pixelcolor/binary_color.rs
rfuest: This could also be written as: `matches!(self, BinaryColor::On)`

core/src/pixelcolor/binary_color.rs
```diff
@@ -97,8 +100,11 @@ impl BinaryColor {
     /// assert!(BinaryColor::Off.is_off());
     /// ```
     #[inline]
-    pub fn is_off(self) -> bool {
-        self == BinaryColor::Off
+    pub const fn is_off(self) -> bool {
```
rfuest, line 103: See comment above.

core/src/geometry/point.rs
```diff
@@ -157,6 +157,24 @@ impl Point {
...
 
+    /// Offsets a point by subtracting a size.
+    ///
+    /// This method provides a workaround for the `Sub` trait not being usable in `const` contexts.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if `width` or `height` are too large to be represented as an `i32`
+    /// and debug assertions are enabled.
+    pub const fn sub_size(self, other: Size) -> Point {
```
rfuest, line 168: IMO we shouldn't make the workaround methods public. If a method is only used in `embedded-graphics` and not in `embedded-graphics-core` it could be moved to the `PointExt` trait.

src/mono_font/mono_text_style.rs
```diff
@@ -386,7 +386,7 @@ pub struct MonoTextStyleBuilder<'a, C> {
 
 impl<C> MonoTextStyleBuilder<'_, C> {
     /// Creates a new text style builder.
-    pub fn new() -> Self {
+    pub const fn new() -> Self {
```
rfuest, line 389: The builder can't be used in a const context, because none of the other methods are `const`. I would leave the constructor as is until it is possible to make the entire builder work in a const context.

src/primitives/arc/mod.rs
```diff
@@ -91,7 +91,7 @@ impl Arc {
     /// Creates an arc based on a circle.
     ///
     /// The resulting arc will match the `top_left` and `diameter` of the base circle.
-    pub fn from_circle(circle: Circle, angle_start: Angle, angle_sweep: Angle) -> Self {
+    pub const fn from_circle(circle: Circle, angle_start: Angle, angle_sweep: Angle) -> Self {
```
rfuest, line 94: `Sector` contains the same constructor, which you missed.

rfuest (changes requested):
I'll leave the MSRV decision to @jamwaffles. The rest of the PR looks good except for some minor nits.

core/src/pixelcolor/binary_color.rs
peterstuart: Ok, updated to use `matches!`.

core/src/geometry/point.rs
peterstuart: I don't think that'll work because trait functions can't be `const`, so I changed them to be `pub(crate)`.

src/mono_font/mono_text_style.rs
peterstuart: Ok, reverted.

src/primitives/arc/mod.rs
peterstuart: Thanks, added.

core/src/geometry/point.rs
rfuest: OK, if the method is only used in one crate that's the better solution anyway.

.circleci/config.yml
jamwaffles: It would indeed. I can prepare a PR for that once we settle on a version.

README.md
```diff
@@ -230,7 +230,7 @@ Additional examples can be found in the [examples](https://github.com/embedded-g
 
 ## Minimum supported Rust version
 
-The minimum supported Rust version for embedded-graphics is `1.56.1` (edition 2021) or greater.
+The minimum supported Rust version for embedded-graphics is `1.57` (edition 2021) or greater.
```
jamwaffles, line 233: The chosen version is no longer actually edition 2021. That said, should we still mention that this chosen version is "edition 2021 + a bit"?

```suggestion
The minimum supported Rust version for embedded-graphics is `1.57` or greater.
```

Don't forget the other 3 places too.

jamwaffles (changes requested):
> I'll leave the MSRV decision to @jamwaffles. 

I threw a [quick poll on Twitter](https://twitter.com/jam_waffles/status/1476954886445142017) as well as one on the e-g Matrix channel. I'll let those get a few responses before the final decision, but I'm sliding towards 1.57 at the moment.

> The rest of the PR looks good except for some minor nits.

Agreed :)

README.md
peterstuart: Ok, updated to remove mentions of the edition. IMO it'd be a little confusing to mention the edition since 1.57 still supports the 2021 edition, and it doesn't matter to consumers of the library anyway, since they can remain on a previous edition as long as they have 1.57.

.circleci/config.yml
jamwaffles: I've pushed a 1.57.0-0 image to Docker Hub for testing with. This PR should now at least run in CI.

jamwaffles:
Looks like we're good for 1.57 instead of 1.56.1. Hopefully not too much work to update this PR!

README.md
jamwaffles: That makes sense yeah :)

peterstuart:
👍🏻 I rebased it on top of the latest version of your PR branch.

rfuest (approved):

jamwaffles:
Oops, looks like Github autoclosed this PR and won't let me reopen because the original branch was deleted. Can you try rebasing on `master` now that #638 is merged? If not, feel free to open a new PR with the changes.

peterstuart:
Rebasing didn't give me the option to re-open it, so I made a new PR: https://github.com/embedded-graphics/embedded-graphics/pull/640.

PR #642, opened 2022-01-06
Convert doc links to intra-doc-links

jamwaffles:

## PR description


jamwaffles:
First pass done with `cargo intraconv`. Some doc links e.g. between core and e-g don't resolve properly (as do some of the trait method links) so I've left them as relative HTML paths at time of writing. It would be very good to remove them as we can then remove `cargo-deadlinks` in the build too.

rfuest:
I also had to clean up the `intraconv` result for the simulator repo.

Any idea how we could handle the deep links from the README files into the docs? I don't see an easy way of changing them to intra-doc links.

jamwaffles:
> I also had to clean up the intraconv result for the simulator repo.

Did you find a solution for any links that still needed relative paths? I had some errors where some trait methods weren't being found.

> Any idea how we could handle the deep links from the README files into the docs? I don't see an easy way of changing them to intra-doc links.

Hm that's frustrating. It might be possible to replace `::` with `/` and prefix the path with `https://docs.rs/embedded-graphics/` maybe?

I've got to do other stuff for a bit, but feel free to push to this branch if you want to continue - just post me a comment saying as such :)

rfuest:
> Did you find a solution for any links that still needed relative paths? I had some errors where some trait methods weren't being found.

I don't think I encountered such problems in the simulator repo.

> Hm that's frustrating. It might be possible to replace `::` with `/` and prefix the path with `https://docs.rs/embedded-graphics/` maybe?

The full path contains a prefix, e.g. `struct.Line.html`, but as long as all linked items are structs we might get away with some `sed` magic. Modules should be distinguishable from struct, because they don't contain any upper case letters.

 > I've got to do other stuff for a bit, but feel free to push to this branch if you want to continue - just post me a comment saying as such :)

I'm done for today, but I can take a look at this tomorrow, if you haven't finished the PR by then.

rfuest:
I've removed some of the remaining links to html files.

The remaining links to `docs.rs` from the core crate to e-g will be harder to remove. There is an issue about this problem: https://github.com/rust-lang/rust/issues/74481

One solution suggested in the issue would be to document the re-exported type instead of the core type. IMO the e-g documentation is more important than the core documentation, so this is something we should consider to use until a better solution is available.

jamwaffles:
> I've removed some of the remaining links to html files.

Nice! I saw the `super` failing when I tried it but maybe that was something else.

> IMO the e-g documentation is more important than the core documentation, so this is something we should consider to use until a better solution is available.

Agreed. It's a bit of a pain, but core can stay using HTML links for now. I'd really like to get rid of cargo-deadlinks but admit that doing so would stop those HTML links being checked. What are your thoughts on that?

rfuest:
> Agreed. It's a bit of a pain, but core can stay using HTML links for now.

The problem with this is that the core docs are imported into the e-g docs. Because of this some links, mainly in the  `drawable` module docs, are linking to docs.rs instead of the local version. This could cause confusion because the user might end up viewing the docs for a different e-g version without noticing.

> I'd really like to get rid of cargo-deadlinks but admit that doing so would stop those HTML links being checked. What are your thoughts on that?

There shouldn't be many HTML links left in the docs after this PR and IMO we can manage them without using `cargo-deadlinks`. I would am more concerned about the links in the README.md files. Is there a good link checker for markdown files?

jamwaffles:
I've just pushed two commits which:

1. Strip out intra-doc links from the READMEs. I'm no sed expert but they seem to work (on macOS' `sed` even)
2. Add [`lychee`](https://github.com/lycheeverse/lychee), a link checker, to the Just commands to check the links in the READMEs. It won't work in CI yet as I haven't added it to the build image in case it's not the right way to go. I'll push a new build image and update this PR to use it if you think `lychee` is decent enough. The only issue I see with it is the low monthly download count, but it seems reliable so 🤷.

I'm still not sure what to do about the intra-crate links.

jamwaffles:
There's also [`cargo-rdme`](https://github.com/orium/cargo-rdme) (not a typo) which does doc link conversion to `docs.rs` as well. On second thoughts, that sounds like a better solution so I'll try it now.

jamwaffles:
`cargo-rdme` is quite flaky and requires the use of `crate::...` syntax which could be easy to forget. It also can't resolve re-exports, so the list of features at the top of the e-g readme is missing links for items reexported from `core`. I decided to go back to just removing the intra-doc links with `sed` as it gives a more consistent output.

rfuest:
`lychee` seems to be a good way to check the links in README.md.

I would really like to keep the links in the first paragraph of the README and not simply strip them out. What is you opinion on replacing the `sed` based mess with a simple Rust program in the `tools` directory? IMO that would be more flexible and easier to maintain for non sed experts.

README.md
```diff
@@ -137,7 +137,7 @@ rendered by the display. See the [`DrawTarget`] documentation for implementation
 
 [![A grid of screenshots showing primitives, text and other items that can be drawn using embedded-graphics.](https://raw.githubusercontent.com/embedded-graphics/embedded-graphics/4c680418cc4c37b8f7ed92d8b26edfde880d4c96/doc/assets/all_drawing_ops.png)](https://docs.rs/embedded-graphics/latest/embedded_graphics/examples/index.html)
 
-Example usage of drawing primitives, text and images with embedded-graphics can be found [here](https://docs.rs/embedded-graphics/latest/embedded_graphics/examples/index.html).
+Example usage of drawing primitives, text and images with embedded-graphics can be found here.
```
rfuest, line 140: This link shouldn't be stripped out.

jamwaffles:
> I would really like to keep the links in the first paragraph of the README and not simply strip them out. What is you opinion on replacing the sed based mess with a simple Rust program in the tools directory? IMO that would be more flexible and easier to maintain for non sed experts.

That would certainly be cleaner, although I think it would be quite a lot of code to get this right. We could look at forking/fixing [carg-rdme](https://github.com/orium/cargo-rdme) which has already done a lot of the heavy lifting, it just has issues around requiring `crate::` on everything, and not supporting re-exports or aliases very well.

rfuest:
I'm really starting to dislike README generation, it's a much harder problem than it should be.

Here is my suggestion how we could fix the links until some better solution is available:
- Convert all intra doc links in `lib.rs` to the reference form
   e.g. use `[Lines]` in the text and define `[Lines]: primitives::line::Line` at the bottom
- Insert a commented out block of links to docs.rs before the intra doc links
  ```
  Some text that talks about [Lines].

  <!-- README-LINKS
  [Lines]: https://docs.rs/embedded-graphics/...
  README_LINKS -->

  [Lines]: primitives::line::Line
  ```
  These will be ignored by `cargo doc`.
- Generate the README with `cargo-readme`
- Remove all lines containing the `README-LINKS` marker from the generated README.md
  The earlier definitions of link references will override later ones. This behavior is guaranteed by the CommonMark and GitHub markdown specs (see https://github.github.com/gfm/#example-173 and https://spec.commonmark.org/0.30/#example-204)
- Run `lychee`
  This will report a broken link if we forgot to include a reference in the `README-LINKS` block, because it will still use the intra doc link

One additional check for missing link references would be needed. `lychee` doesn't check for missing references and at the moment `README.md` doesn't define a reference for ``[`Point`]`` without causing an error. I didn't find a markdown linter that checks for these cases and doesn't use javascript/npm, which I would like to avoid. But it's trivial to implement one using the `pulldown-cmark` crate:
```rust
fn main() {
    let file = std::env::args().nth(1).unwrap();
    let readme = std::fs::read_to_string(&file).unwrap();

    let broken_link_callback = &mut |broken_link: BrokenLink| {
        panic!("broken link: {}", broken_link.reference);
    };

    let parser =
        Parser::new_with_broken_link_callback(&readme, Options::all(), Some(broken_link_callback));

    for _event in parser {}
}
```

jamwaffles:
Yeah this has turned into a bit of a nightmare 😬 

That's a hacky yet clever way around the link generation. I like it 😈. I've pushed a change implementing your suggestion and it seems to work reasonably well. Let me know what you think of the impl and the output.

filter-readme-embedded-graphics-core.sed
```diff
@@ -4,3 +4,8 @@
 s#\[(.+)\]\(\.\/(.*)\)#[\1](https://docs.rs/embedded-graphics-core/latest/embedded_graphics_core/\2)#g
 s#\[(.+)\]\: +\.\/(.*)#[\1]: https://docs.rs/embedded-graphics-core/latest/embedded_graphics_core/\2#g
 
+# Delete lines containing reference links like `[Foo]: bar::Foo`
+/\[([^]]+)\]\: ([A-Za-z_]+::)*[A-Za-z_]+$/d
```
jamwaffles, line 8: This means that:

a. We don't have duplicate links in the README source and
b. The broken reference checker actually sees a broken reference if there's a typo between the original link and the reference. This is required as `lychee` happily ignores references to crate paths.

rfuest: I like it, that's cleaner than keeping the unused links in the README.

rfuest:
I did take a quick look at the generated docs and it looks good. One small change I would suggest is to use a intra-doc link for `embedded-graphics-core` and `DrawTarget`.

rfuest:
How will we handle adding links to new types in the README in the future? The new types will exist in the README before they are available on `docs.rs´. Or is this a problem we ignore until we encounter it?

rfuest:
It's a bit outside the scope of this PR, but should we move the `Minimum supported Rust version` and `Development setup` sections from the README template into `lib.rs`? Or is there a good reason not to include this in the docs?

filter-readme-embedded-graphics.sed
```diff
@@ -4,3 +4,8 @@
 s#\[(.+)\]\(\.\/(.*)\)#[\1](https://docs.rs/embedded-graphics/latest/embedded_graphics/\2)#g
 s#\[(.+)\]\: +\.\/(.*)#[\1]: https://docs.rs/embedded-graphics/latest/embedded_graphics/\2#g
```
rfuest, line 5: These replacements shouldn't be necessary anymore, or are they still used somewhere?

jamwaffles:
> Or is this a problem we ignore until we encounter it?

`lychee` should error with a 404 in the case it's not present on docs.rs, at which point we can add an `--exclude` argument to the `lychee` command until we release, at which point we can PR a removal of the `--exclude`. It's manual and prone to errors/forgetfulness but I don't think we can get perfect here.

> It's a bit outside the scope of this PR, but should we move the Minimum supported Rust version and Development setup sections from the README template into lib.rs? Or is there a good reason not to include this in the docs?

MSRV can certainly go into `lib.rs` as it's relevant to doc.rs viewers, but I don't think the same applies for the dev setup - that seems very geared towards the _repo_ not the _docs_.

rfuest:
Did you notice this comment: https://github.com/embedded-graphics/embedded-graphics/pull/642#discussion_r834444520

jamwaffles:
> Did you notice this comment:

Apologies, was in a rush and hit review before reading through your comments. I've pushed a commit that addresses the comment. It also revealed a bug - the examples image link wasn't linking to the right place. Because no crate-specific stuff is in the sed rules, we can also now just use a single sed config which is nice - I've forgotten to update one or the other in the past.

> OK. If we add manual excludes we should add the removal of these excludes to the release process in doc/releasing.md. > Do you want to add this now or wait?

Later, if you don't mind, but this is the approach I had in mind too.

tools/check-md-refs/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "check-md-refs"
+version = "0.1.0"
+edition = "2021"
+
```
rfuest, line 5:
```suggestion
publish = false
```

rfuest (changes requested):
One small nit left.

rfuest (approved):
LGTM! :tada:

jamwaffles:
Thanks for the review! Good to be getting a bit of momentum back on this crate. 

I'll update the CI image in a future PR - currently having issues building the Linux CI image on my Mac.

rfuest:
> Thanks for the review! Good to be getting a bit of momentum back on this crate.

Yeah, that feels really good.

> I'll update the CI image in a future PR - currently having issues building the Linux CI image on my Mac.

OK. Can you open an issue to make sure we don't forget this?

jamwaffles:
> OK. Can you open an issue to make sure we don't forget this?

Working on it as we speak, PR inbound soon now that I've dug my old Linux laptop out :)

PR #648, opened 2022-02-02
Fix invalid MonoStyle text dimensions calculation

derchr:
The method len() on &str does return the number of bytes of a character.
Using characters such as '°' (which are two bytes long) caused an
incorrect calculation of the text bounding box.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
   - `Sector::to_circle`
   - `Styled::new`
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


rfuest:
Good catch, thanks for fixing this. There is another use of `text.len()` in the `draw_string` method, which also needs to be fixed.

Could you also add a unit test for the invalid bounding box? You can simply copy `transparent_text_dimensions_one_line` and modify it to use a string with multi byte characters and e.g. a `latin1` font.

derchr:
I'll look into it this weekend!

derchr:
I also changed the calculation in the `draw_string`method and added a test case. Are my changes sufficient? :)

jamwaffles (approved):
LGTM, thanks for the fix! I'll let @rfuest merge this in case there's anything else he'd like to see.

rfuest:
Merged. Thanks again.

Issue #650, opened 2022-02-18
Faster color conversions


rfuest:
The current color conversions, implemented by the `From` trait, are focused on accuracy and not conversion speed. This makes drawing of `Rgb888` images on a `Rgb565` target slow.

I did compare our current implementation to a lookup table based conversion with this test code:
```
let bb = display.bounding_box();
display.color_converted().fill_contiguous(&bb, core::iter::repeat(Rgb888::MAGENTA))?;
```
Note that the test uses `fill_contiguous` instead of `fill_solid` to force one color conversion per pixel which simulates drawing of an image without the image decoding overhead.

Using a STM32F0 with an 320x240 ILI9341 display connected via SPI at 10 MHz i got the following results:
* Current implementation: 1103 ms
* Lookup table: 421 ms

I believe the performance impact of our current implementation will be far less on a microcontroller with hardware divider, but I didn't have time to confirm this.

The decision whether to use a LUT or arithmetic is a tradeoff between flash usage and performance. The LUTs for the `Rgb888` to ` Rgb565`, for example, consume 512 bytes of flash. This is a pretty large overhead if the color conversion performance isn't important, because the conversion isn't used to draw an entire image and is only needed to convert a few color values.

Should we support both implementations or decide to use one of them?

rfuest:
I'm closing this now that the performance improvements in #651 and #652 are merged.

PR #651, opened 2022-02-19
Faster color conversions

rfuest:
This PR changes the implementation of the color conversion to be more efficient when no hardware divider is available. The division is replaced by a cheap right shift. It is as fast as the LUT based implementation mentioned in #650, but doesn't need to store any tables in flash.

### DIFF
--- a/core/src/pixelcolor/conversion.rs
+++ b/core/src/pixelcolor/conversion.rs
@@ -8,12 +8,8 @@ const fn convert_channel<const FROM_MAX: u8, const TO_MAX: u8>(value: u8) -> u8 
     const SHIFT: usize = 24;
     const CONST_0_5: u32 = 1 << (SHIFT - 1);
 
-    /// Returns `from_max / to_max` scaled by `1 << SHIFT`.
-    const fn factor<const FROM_MAX: u8, const TO_MAX: u8>() -> u32 {
-        ((TO_MAX as u32) << SHIFT) / (FROM_MAX as u32)
-    }
-
-    let result = value as u32 * factor::<FROM_MAX, TO_MAX>();
+    // `value * from_max / to_max` scaled by `1 << SHIFT`.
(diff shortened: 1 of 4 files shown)


core/src/pixelcolor/conversion.rs
```diff
@@ -1,8 +1,18 @@
 use crate::pixelcolor::{binary_color::*, gray_color::*, rgb_color::*};
 
 /// Convert color channel values from one bit depth to another.
-const fn convert_channel(value: u8, from_max: u8, to_max: u8) -> u8 {
-    ((value as u16 * to_max as u16 + from_max as u16 / 2) / from_max as u16) as u8
+///
+/// Fixed point implementation of the conversion formula:
+/// out = round(in * from_max / to_max)
```
jamwaffles, line 6: Minor readability improvement?:

```suggestion
/// `out = round(in * from_max / to_max)`
```

jamwaffles (approved):
Tiny nit, otherwise looks good. Nice use of all that `const` magic!

I went to run a benchmark for this and realised we don't have any for the colour conversion stuff. It shouldn't block this PR, but perhaps worth considering now or later? At any rate, how does it compare to the timings in #650?

rfuest:
> I went to run a benchmark for this and realised we don't have any for the colour conversion stuff. It shouldn't block this PR, but perhaps worth considering now or later?

IMO this would be something that needs to run on real target hardware. A beefy desktop CPU won't have a big problem dealing with some additional divisions.

> At any rate, how does it compare to the timings in #650?

On my Cortex M0 test setup the benchmark result is within a few milliseconds of the LUT based implementation. But the LUT based approach might have advantages on platforms with less than 32 bit word width, no hardware multiplier and/or no barrel shifter. At the moment I think the color conversions are now fast enough until someone reports otherwise.

jamwaffles:
True,  although it would have been interesting to see any deltas if they were somewhat representative of other HW. At any rate, 50% is a pretty good improvement!

PR #652, opened 2022-02-20
Faster color conversions (Part 2)

rfuest:
This PR is a small addition to #651. Color conversions between color types that share some channel bit depths, like `Rgb565` and `Rgb555`, will now only calculate the conversion for the channels with different bit depths. The effect of this change was minimal to non existing on my desktop, but I think it can't hurt to add it. The compiler seems to already optimize away some redundant calculations now that `convert_channel` uses const generics instead of parameters. 

I've also added benches for the color conversions. Here is a comparison from before #651 and after this PR:
grafik
Note that `Rgb888` to `Gray8` doesn't use any of the code that was modified in the recent PRs and is only included for completeness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
   - `Sector::from_circle`
   - `Sector::to_circle`
   - `Styled::new`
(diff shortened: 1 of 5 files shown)


jamwaffles (approved):
Looks good. I ran the benches on my rPi 4 with similar results 👍

rfuest:
> Looks good. I ran the benches on my rPi 4 with similar results +1

Great, thanks for verifying the result. I hadn't expected such a big improvement.

PR #653, opened 2022-02-23
Better Support Trailing Newline in Text draw

kevinastone:

- [X] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) if your changes affect the **public API**

## PR description

The previous implementation of `Text::draw` ignored a trailing newline so if you chained Text draw calls (with the returned position), the line-spacing was ignored.

There's an example failing test in the first commit of this PR and the second commit is the fix.

For context, I ran into this issue trying to use `core::writeln!` on a wrapper of the display which calls `Text::draw` on each string fragment including a bare newline for a sample string such as:
```
writeln!("Value1: {}", value1);
writeln!("Value 2: {}", value2);
```

### DIFF
--- a/src/text/text.rs
+++ b/src/text/text.rs
@@ -105,10 +105,17 @@ impl<S: Clone> Transform for Text<'_, S> {
 }
 
 impl<S: TextRenderer> Text<'_, S> {
+    fn line_height(&self) -> i32 {
+        self.text_style
+            .line_height
+            .to_absolute(self.character_style.line_height())
+            .saturating_as::<i32>()
+    }
+
     fn lines(&self) -> impl Iterator<Item = (&str, Point)> {
         let mut position = self.position;
 
-        self.text.lines().map(move |line| {
+        self.text.split('\n').map(move |line| {
             let p = match self.text_style.alignment {
                 Alignment::Left => position,
                 Alignment::Right => {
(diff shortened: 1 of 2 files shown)


src/text/text.rs
```diff
@@ -107,10 +107,25 @@ impl<S: Clone> Transform for Text<'_, S> {
...
 
-        self.text.lines().map(move |line| {
+        self.text.split_inclusive('\n').flat_map(move |mut line| {
+            let is_terminated = match line.strip_suffix('\n') {
+                None => false,
+                Some(s) => {
+                    line = s;
+                    true
+                }
+            };
```
rfuest, line 127: Simply using `self.text.split('\n')` should also work without the additional `is_terminated`and `extra_termination` code. The iterator will return an empty string for the trailing newline.

rfuest (changes requested):
Thanks for the PR, this way of handling trailing newlines is better.

One issue with this change is that `lines()` did also support windows style line endings `\r\n`, which the implementation in this PR doesn't. But it wouldn't be hard to replicate this behavior: https://doc.rust-lang.org/src/core/str/mod.rs.html#2503-2505

src/text/text.rs
kevinastone: > Thanks for the PR, this way of handling trailing newlines is better.
> 
> One issue with this change is that `lines()` did also support windows style line endings `\r\n`, which the implementation in this PR doesn't. But it wouldn't be hard to replicate this behavior: https://doc.rust-lang.org/src/core/str/mod.rs.html#2503-2505

Yeah, you could do the same `strip_suffix('\r')` => `line = s` as removing the terminator.

kevinastone: > Simply using `self.text.split('\n')` should also work without the additional `is_terminated`and `extra_termination` code. The iterator will return an empty string for the trailing newline.

That sounds correct.  Originally I was attempting a different solution without using empty string iterations, so you had to do some math depending on the trailing terminator.

rfuest: > Yeah, you could do the same `strip_suffix('\r')` => `line = s` as removing the terminator.

You could, but the version from the core library will probably have slightly better performance, because it is byte based and not char based.

jamwaffles (approved):
Looks good, thanks for the test too!

rfuest (approved):
Thanks @kevinastone for this PR and sorry for the long delay to merge it.

PR #662, opened 2022-03-25
Apply MSRV comments

rfuest:
This PR contains the changes we had marked with MSRV comments. Except the two remaining comments, which rely on not yet stabilized features.


jamwaffles (approved):
Looks good. Nice to see some cleaner code possible now :)

PR #664, opened 2022-03-27
Inline some usages of self in closures

### LINKED ISSUE #629: Review closure usages
Edition 2021 no longer captures entire structs, so there might be an opportunity to clean up some code.

https://twitter.com/m_ou_se/status/1451259982385754114?s=12

jamwaffles:

## PR description

This PR is a result of #629, although I don't really like the changes made here as I think they make the code slightly _less_ readable. Unless I've missed any egregious uses of a temporary binding of `self.<var>` somewhere, I'd be inclined to close both this PR and #629.

Closes #629

### DIFF
--- a/src/draw_target/clipped.rs
+++ b/src/draw_target/clipped.rs
@@ -40,11 +40,9 @@ where
     where
         I: IntoIterator<Item = Pixel<Self::Color>>,
     {
-        let clip_area = self.clip_area;
-
         let pixels = pixels
             .into_iter()
-            .filter(|Pixel(p, _)| clip_area.contains(*p));
+            .filter(|Pixel(p, _)| self.clip_area.contains(*p));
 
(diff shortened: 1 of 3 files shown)


rfuest:
IMO they are very similar in how readable they are. Just use whatever variant you prefer.

Or perhaps do a quick benchmark run to see if there is any difference. I don't see why there would be, but sometimes the compiler does give surprising results.

jamwaffles:
`just bench arc` gave no perf difference between `master` and this branch so that's good. I think I'll merge the changes here since I've already made them. For beaurocratic reasons, could you approve a review so we can keep the Germans happy? ;)

rfuest (approved):
> For beaurocratic reasons, could you approve a review so we can keep the Germans happy? ;)

Sure, Germans love bureaucracy.

PR #666, opened 2022-03-27
Add eg-seven-segment to crate list

rfuest:
After almost 1.5 years I finally managed to publish the first version of `eg-seven-segment`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,6 +36,7 @@
 //! * [Simple layout/alignment functions - `embedded-layout`](https://crates.io/crates/embedded-layout)
 //! * [TextBox with text alignment options - `embedded-text`](https://crates.io/crates/embedded-text)
 //! * [Heapless plotting library for small embedded targets - `embedded-plots`](https://crates.io/crates/embedded-plots)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -36,6 +36,7 @@
 //! * [Simple layout/alignment functions - `embedded-layout`](https://crates.io/crates/embedded-layout)
 //! * [TextBox with text alignment options - `embedded-text`](https://crates.io/crates/embedded-text)
 //! * [Heapless plotting library for small embedded targets - `embedded-plots`](https://crates.io/crates/embedded-plots)
+//! * [Seven-segment displays - `eg-seven-segment`](https://crates.io/crates/eg-seven-segment)
```
jamwaffles, line 39: I admit I initially thought "e-g on seven segment displays? How?!" when I read this description. Should we maybe add some wording around "virtual" or "emulated" SS displays?

rfuest: That makes sense, I've added "virtual".

jamwaffles (approved):
Looks good, and congrats 🎉

Issue #608, opened 2021-06-16
In-memory display driver?


kpreid:
Is there a `DrawTarget` implementation (display driver) which supports writing to an in-memory (or memory-mapped) framebuffer (given `&[u8]`, a bounding rectangle, and perhaps a pixel format)? There is [`MockDisplay`](https://docs.rs/embedded-graphics/0.7.0/embedded_graphics/mock_display/struct.MockDisplay.html), of course, but that's an implementation designed for testing rather than performance. I took a look at the dependent crates for drivers but they all seemed to be hardware-specific.

If not, would such an implementation be a reasonable addition to the `embedded-graphics` crate?

Besides being reusable code (index computation, bounds checking, and implementations of the optional acceleration methods to avoid redundant index computation) for drawing to displays which want this format (if there are any such), it might be useful for applications which would like to draw into images (and then perhaps draw those images to other `DrawTarget`s).

jamwaffles:
Hey, thanks for opening the discussion on this. I think embedded-graphics should remain target agnostic and including a framebuffer with e-g itself would go against that. It could certainly be a separate crate to cater for more generic target requirements, however some or all of what you're looking for might already be covered by the [embedded-graphics-simulator](https://crates.io/crates/embedded-graphics-simulator) crate. 

The simulator can be used to generate images if that's what you're looking to do, but it would be great to know what concrete requirements you have that e-g isn't currently fulfilling.

kpreid:
> I think embedded-graphics should remain target agnostic and including a framebuffer with e-g itself would go against that.

I was imagining an implementation that is both a small amount of code and sufficiently generic that, in addition to being usable for images or buffered drawing, would be useful to multiple concrete targets that use addressable framebuffers, either because the final destination is in-memory or because they use a local buffer and a flush operation. For examples in existing e-g drivers, [epd_waveshare](https://docs.rs/epd-waveshare/0.4.0/src/epd_waveshare/graphics.rs.html), [st7920](https://github.com/wjakobczyk/st7920/blob/master/src/lib.rs), and [ls010b7dh01](https://github.com/byronwasti/ls010b7dh01/blob/master/src/lib.rs) all draw pixels to in-memory arrays, which could be supported by a common implementation which knows how to do the range checks and index arithmetic.

[...]

PR #673, opened 2022-04-24
Add framebuffer

rfuest:
This draft PR adds the framebuffer implementation discussed in #608. In contrast to the existing implementation in the [`embedded-graphics-framebuf`](https://crates.io/crates/embedded-graphics-framebuf) crate this implementation packs multiple pixels into one byte for color types with < 8BPP.

The internal storage format is the same as the format used by `ImageRaw`. With the `Framebuffer::as_image` getter the content of the framebuffer can be used as an image to display it on another draw target.

A large portion of this code was written last year and I've only updated it to use const generics. I haven't taken a closer the look at the old code and it will need some more polish before it can be merged. The `framebuffer` macro to calculate the `N` parameter isn't great either, but expressions with const generics aren't supported in stable rust.

TODO / open questions:
- [x] Add getters and setters for individual pixels, similar to #612
- [x] Add `data_mut`?
- [ ] Add `SliceFramebuffer` variant, which would use a mutable slice instead of an array?
- [ ] Add benches & improve performance
- [ ] Documentation
- [ ] Examples

### DIFF
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -61,13 +61,13 @@ where
     C: PixelColor,
 {
     /// Static assertion that N is correct.
-    /// MSRV: remove N when constant generic expressions are stabilized
+    // MSRV: remove N when constant generic expressions are stabilized
     const CHECK_N: () = if N != buffer_size::<C>(WIDTH, HEIGHT) {
         panic!("Invalid N: see Framebuffer documentation for more information");
     };
 
     /// Creates a new framebuffer.
-    pub fn new() -> Self {
+    pub const fn new() -> Self {
         #[allow(path_statements)]
         {
             // Make sure CHECK_N isn't optimized out.
@@ -83,7 +83,7 @@ where
     }
 
     /// Returns a reference to the raw framebuffer data.
-    pub fn data(&self) -> &[u8; N] {
+    pub const fn data(&self) -> &[u8; N] {
         &self.data
     }
 
@@ -101,13 +101,14 @@ where
     for<'a> RawDataSlice<'a, C::Raw, BO>: IntoIterator<Item = C::Raw>,
 {
     /// Returns an image based on the framebuffer content.
-    pub fn as_image(&self) -> ImageRaw<'_, C, BO> {
+    pub const fn as_image(&self) -> ImageRaw<'_, C, BO> {
         ImageRaw::new(&self.data, WIDTH as u32)
     }
 
     /// Get pixel color.
     ///
     /// Returns `None` is `p` is outside the framebuffer bounding box.
+    // TODO: Optimise implementation by moving away from `ImageRaw::pixel` and directly calculating skip value.
     pub fn pixel(&self, p: Point) -> Option<C> {
         self.as_image().pixel(p)
     }
@@ -174,7 +175,7 @@ where
 {
     /// Sets the color of a pixel.
     ///
-    /// Trying to set a pixel outside the framebuffer is a noop.
+    /// Setting a pixel outside the framebuffer's bounding box will be a noop.
     pub fn set_pixel(&mut self, p: Point, c: C) {
         if let (Ok(x), Ok(y)) = (usize::try_from(p.x), usize::try_from(p.y)) {
             if x < WIDTH && y < HEIGHT {
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -22,7 +22,15 @@ use crate::{
 /// This function is a workaround for current limitations in Rust const generics.
 /// It can be used to calculate the `N` parameter based on the size and color type of the framebuffer.
 pub const fn buffer_size<C: PixelColor>(width: usize, height: usize) -> usize {
-    (width * C::Raw::BITS_PER_PIXEL + 7) / 8 * height
+    buffer_size_bpp(width, height, C::Raw::BITS_PER_PIXEL)
+}
+
+/// Calculates the required buffer size.
+///
+/// This function is a workaround for current limitations in Rust const generics.
+/// It can be used to calculate the `N` parameter based on the size and bit depth of the framebuffer.
+pub const fn buffer_size_bpp(width: usize, height: usize, bpp: usize) -> usize {
+    (width * bpp + 7) / 8 * height
 }
 
 /// A framebuffer.
@@ -45,7 +53,7 @@ pub const fn buffer_size<C: PixelColor>(width: usize, height: usize) -> usize {
 ///     .draw(&mut fb)
 ///     .unwrap();
 /// ```
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct Framebuffer<C, R, BO, const WIDTH: usize, const HEIGHT: usize, const N: usize> {
     data: [u8; N],
     color_type: PhantomData<C>,
@@ -58,9 +66,11 @@ impl<C, BO, const WIDTH: usize, const HEIGHT: usize, const N: usize>
 where
     C: PixelColor,
 {
+    const BUFFER_SIZE: usize = buffer_size::<C>(WIDTH, HEIGHT);
+
     /// Static assertion that N is correct.
     // MSRV: remove N when constant generic expressions are stabilized
-    const CHECK_N: () = if N != buffer_size::<C>(WIDTH, HEIGHT) {
+    const CHECK_N: () = if N < Self::BUFFER_SIZE {
         panic!("Invalid N: see Framebuffer documentation for more information");
     };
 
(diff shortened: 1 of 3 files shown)


jamwaffles:
I'm sorry I can't give a proper review this week as I'm travelling for work, but have you seen https://github.com/embedded-graphics/packed-display-buffer? It focusses on 1BPP but I'm moderately happy with the optimised `fill_solid` impl over there. The `fill_contiguous` impl is still WIP, but perhaps that code could serve as inspiration for some speedups in this PR?

rfuest:
The framebuffer implementation still needs some work, like documentation and testing/improving the performance. But I think it should be finished enough to merge this PR and finish it in future PRs.

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+///     framebuffer,
+///     framebuffer::{Framebuffer, buffer_size},
+///     pixelcolor::{Rgb565, raw::LittleEndian},
+///     prelude::*,
+///     primitives::PrimitiveStyle,
+/// };
+///
+/// let mut fb = <framebuffer!(Rgb565, 320, 240)>::new();
+/// // or:
+/// let mut fb = Framebuffer::<Rgb565, _, LittleEndian, 320, 240, {buffer_size::<Rgb565>(320, 240)}>::new();
```
rfuest, line 43: Having to use a macro to determine the type isn't great, but IMO better than requiring the user to calculate 'N` manually.

rfuest: The non macro version is already much better than it used to be. Before we increased the MSRV the `buffer_size` function couldn't use a type parameter.

jamwaffles:
> The framebuffer macro to calculate the N parameter isn't great either, but expressions with const generics aren't supported in stable rust.

I feel your pain. I'm aching for `generic_const_exprs` in some of my other projects. It can't come soon enough. That said, I think the macro is a pretty clever solution - I might have to employ it elsewhere.

> Add SliceFramebuffer variant, which would use a mutable slice instead of an array?

`Framebuffer::new` can be made `const` which I think would fix some/most issues around storing a reusable framebuffer in a `static`, e.g. for some RTIC situations perhaps. Plus, it's better to start with a small interface and add to it if needed.

I'll post some review comments shortly.

I'd also like to find a way to reduce the macro codegen if possible, as well as porting my optimised `fill_solid` impl from https://github.com/embedded-graphics/packed-display-buffer but I won't block this PR on those changes.

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+    byte_order: PhantomData<BO>,
+}
+
+impl<C, BO, const WIDTH: usize, const HEIGHT: usize, const N: usize>
+    Framebuffer<C, C::Raw, BO, WIDTH, HEIGHT, N>
+where
+    C: PixelColor,
+{
+    /// Static assertion that N is correct.
+    /// MSRV: remove N when constant generic expressions are stabilized
```
jamwaffles, line 64: Nit

```suggestion
    // MSRV: remove N when constant generic expressions are stabilized
```

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+    const CHECK_N: () = if N != buffer_size::<C>(WIDTH, HEIGHT) {
+        panic!("Invalid N: see Framebuffer documentation for more information");
+    };
+
+    /// Creates a new framebuffer.
+    pub fn new() -> Self {
```
jamwaffles, line 70:
```suggestion
    pub const fn new() -> Self {
```

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+        Self {
+            data: [0; N],
+            color_type: PhantomData,
+            raw_type: PhantomData,
+            byte_order: PhantomData,
+        }
+    }
+
+    /// Returns a reference to the raw framebuffer data.
+    pub fn data(&self) -> &[u8; N] {
```
jamwaffles, line 86:
```suggestion
    pub const fn data(&self) -> &[u8; N] {
```

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+
+impl<C, BO, const WIDTH: usize, const HEIGHT: usize, const N: usize>
+    Framebuffer<C, C::Raw, BO, WIDTH, HEIGHT, N>
+where
+    C: PixelColor + From<C::Raw>,
+    BO: ByteOrder,
+    for<'a> RawDataSlice<'a, C::Raw, BO>: IntoIterator<Item = C::Raw>,
+{
+    /// Returns an image based on the framebuffer content.
+    pub fn as_image(&self) -> ImageRaw<'_, C, BO> {
```
jamwaffles, line 104:
```suggestion
    pub const fn as_image(&self) -> ImageRaw<'_, C, BO> {
```

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+        ImageRaw::new(&self.data, WIDTH as u32)
+    }
+
+    /// Get pixel color.
+    ///
+    /// Returns `None` is `p` is outside the framebuffer bounding box.
```
jamwaffles, line 110: Mostly a note to self for a future PR unless you beat me to it. I think we could get some pretty good performance gains here.

```suggestion
    /// Returns `None` is `p` is outside the framebuffer bounding box.
    // TODO: Optimise implementation by moving away from `ImageRaw::pixel` and directly calculating skip value.
```

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+impl_bit!(RawU4);
+
+impl<C, BO, const WIDTH: usize, const HEIGHT: usize, const N: usize>
+    Framebuffer<C, RawU8, BO, WIDTH, HEIGHT, N>
+where
+    C: PixelColor + Into<RawU8>,
+{
+    /// Sets the color of a pixel.
+    ///
+    /// Trying to set a pixel outside the framebuffer is a noop.
```
jamwaffles, line 177: Slightly better?

```suggestion
    /// Setting a pixel outside the framebuffer's bounding box will be a noop.
```

Or maybe something like `framebuffer's drawing area`?

src/framebuffer.rs
```diff
@@ -0,0 +1,707 @@
...
+    for Framebuffer<C, R, BO, WIDTH, HEIGHT, N>
+{
+    fn size(&self) -> Size {
+        Size::new(WIDTH as u32, HEIGHT as u32)
+    }
+}
+
+/// Calculate the framebuffer generic constants.
+#[macro_export]
+macro_rules! framebuffer {
```
jamwaffles, line 278: Do we need these macros beyond some convenience for the user? This is another API item we'd end up removing in the future once more const features are stablised. I think I'd prefer to export `buffer_size` and rely on our docs to help users write the correct type definitions.

I'm also enjoying not having any exported macros in e-g 😁

jamwaffles (changes requested):

jamwaffles:
Some initial comments. Lots of room for improvement but let's fix forward instead of trying for one huge, perfect PR

rfuest:
> > Add SliceFramebuffer variant, which would use a mutable slice instead of an array?
> 
> `Framebuffer::new` can be made `const` which I think would fix some/most issues around storing a reusable framebuffer in a `static`, e.g. for some RTIC situations perhaps. Plus, it's better to start with a small interface and add to it if needed.

Agreed that we don't need to add this now. My main reason for adding a `SliceFramebuffer` or something similar is to also support dynamically allocated buffers. But at that point it might be better to add an optional `alloc` feature.

src/framebuffer.rs
rfuest: Agreed. I wanted to add a similar comment, but forgot to add it in the end. But this is also something that should be benchmarked before changing it IMO.

src/framebuffer.rs
rfuest: It is better. I wasn't happy with that sentence, but couldn't come up with something better.

src/framebuffer.rs
rfuest: > Do we need these macros beyond some convenience for the user?

No.

> I'm also enjoying not having any exported macros in e-g grin

OK. I've made the macro private inside the tests module.

rfuest:
I've noticed another issue caused by limitations in Rusts const generics. Trying to use `Framebuffer` inside `MockDisplay` didn't work:
```rust
pub struct MockDisplay<C>
where
    C: PixelColor,
{
    pixels: Framebuffer<C, C::Raw, LittleEndian, SIZE, SIZE, {buffer_size::<C>(SIZE, SIZE)}>,
    ....
}
```
The compiler doesn't allow `C` to be used in a const expression. I don't think that it is possible to circumvent this limitation, even with a macro.

I've changed the buffer size check to also allow oversized buffers. While this isn't a great solution for memory constrained systems it should work as a workaround for `MockDisplay` and similar applications.

rfuest:
I'll be away for the next two weeks. If you want to try to integrate the performance improvements from `packed-display-buffer` you can merge this PR.

jamwaffles (approved):

jamwaffles:
Thanks for making a good start on this impl. Have a good break!

PR #679, opened 2022-06-16
Add `embedded-fps` crate

elpiel:
## PR description

This is a small crate that gives the ability to count the Frames per second on embedded devices - `no_std` & no `alloc`.

I've provided an example with `embedded_graphics` in the docs as well.


jamwaffles (approved):
Looks good, thanks!

PR #682, opened 2022-08-13
Add st7565 driver

Finomnis:
## PR description
Adds the st7565 driver to the list of existing drivers


jamwaffles (approved):
Nice, thanks!

PR #688, opened 2022-09-01
Require `GlyphMapping` implementors to also be `Sync`.

### LINKED ISSUE #687: `MonoFont: Send + Sync`
- Version of embedded-graphics in use (if applicable): 0.7.1

## Description of the feature request

[The `MonoFont` struct](https://docs.rs/embedded-graphics/0.7.1/embedded_graphics/mono_font/struct.MonoFont.html) contains a field

```rust
    pub glyph_mapping: &'a dyn GlyphMapping,
```

which causes the whole `MonoFont` to be `!Send + !Sync`, meaning that it cannot be stored or referenced in a `static` variable, or sent between threads. Is this an intended restriction? If not, I'd like to request that it be made `Send + Sync`, either by changing the field type to 

```rust
    pub glyph_mapping: &'a dyn GlyphMapping + Send + Sync,
```

or by changing the trait definition,

```rust
pub trait GlyphMapping: Send + Sync {
```

whichever seems more appropriate. This should not be onerous since the mapping should generally be a stateless function. (But it is a semver breaking change that could affect any code that constructs a `MonoFont` or implements `GlyphMapping`.)

I would be happy to provide a PR to change this, but there's at least one design choice consideration, so I'm presenting it as an issue first.

## Test case

```rust
use embedded_graphics::mono_font::{
    ascii::{FONT_9X15, FONT_9X15_BOLD},
    MonoFont,
};

static FONT_LIST: [MonoFont<'_>; 2] = [FONT_9X15, FONT_9X15_BOLD];
```

```
error[E0277]: `(dyn GlyphMapping + 'static)` cannot be shared between threads safely
 --> src/lib.rs:6:19
  |
6 | static FONT_LIST: [MonoFont<'_>; 2] = [FONT_9X15, FONT_9X15_BOLD];
  |                   ^^^^^^^^^^^^^^^^^ `(dyn GlyphMapping + 'static)` cannot be shared between threads safely
  |
  = help: within `[MonoFont<'static>; 2]`, the trait `Sync` is not implemented for `(dyn GlyphMapping + 'static)`
  = note: required because it appears within the type `&'static (dyn GlyphMapping + 'static)`
  = note: required because it appears within the type `MonoFont<'static>`
  = note: required because it appears within the type `[MonoFont<'static>; 2]`
  = note: shared static variables must have a type that implements `Sync`
```

kpreid:
- [X] Run `just build`

## PR description

Added `Sync` as a supertrait to `GlyphMapping`. This means that the `&dyn GlyphMapping` in `MonoFont` no longer causes `MonoFont` to be `!Send + !Sync`. Now, it is `Send + Sync`.

This is a semver breaking change because it imposes a new requirement on implementors of `GlyphMapping`. This should not be onerous since the mapping should generally be a stateless function.

Closes #687.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
   - `Styled::new`
 - [#651](https://github.com/embedded-graphics/embedded-graphics/pull/651), [#652](https://github.com/embedded-graphics/embedded-graphics/pull/652) Improved performance of color conversions.
 - [#662](https://github.com/embedded-graphics/embedded-graphics/pull/662) `ImageRaw::new` no longer panics if `width == 0`.
+- **(breaking)** [#688](https://github.com/embedded-graphics/embedded-graphics/pull/688) `MonoFont` is now `Send + Sync`; implementations of `GlyphMapping` must be `Sync`.
 - **(breaking)** [#663](https://github.com/embedded-graphics/embedded-graphics/pull/663) Upgraded Cargo dependencies to their latest versions.
 - **(breaking)** [#689](https://github.com/embedded-graphics/embedded-graphics/pull/689) Bump Minimum Supported Rust Version (MSRV) to 1.61.
 
--- a/src/mono_font/mapping.rs
+++ b/src/mono_font/mapping.rs
@@ -46,7 +46,7 @@
 use core::ops::RangeInclusive;
 
 /// Mapping from characters to glyph indices.
-pub trait GlyphMapping {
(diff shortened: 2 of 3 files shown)


jamwaffles (approved):
This looks good to me, although to be extremely nitpicky, could you move the changelog item to the bottom of the list? 

As for a breaking change, there are other breaking changes already in `master` so I'm not worried about that :). @rfuest any additional comments?

rfuest (approved):
LGTM

jamwaffles:
Thanks!

PR #691, opened 2022-09-02
Don't run tests in release mode

rfuest:
This PR changes the `justfile` to not run the tests in release mode. Running the tests in release mode disables the integer overflow checks and the debug assertions in our code, which could mask some errors.


jamwaffles (approved):
Makes sense. The CI build isn't appreciably slower either which is nice.

PR #693, opened 2022-09-07
Update u8g2-fonts description

Finomnis:
u8g2-fonts just released `v0.2.0`, which integrates with `embedded_graphics::text`.

So it's no longer just an external renderer, it can also be used as a character style for the internal renderer.


jamwaffles (approved):
Congrats on the release! Thanks for the update PR :)

PR #694, opened 2022-09-20
Framebuffer optimisations

jamwaffles:

### DIFF
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -131,6 +131,14 @@ where
     }
 }
 
+fn check<C>() -> usize
+where
+    C: PixelColor,
+    <<C as PixelColor>::Raw as RawData>::Storage: Into<u8>,
+{
+    C::Raw::BITS_PER_PIXEL
+}
+
 macro_rules! impl_bit {
     ($raw_type:ident) => {
         impl<C, BO, const WIDTH: usize, const HEIGHT: usize, const N: usize>
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -126,7 +126,6 @@ where
     /// Get pixel color.
     ///
     /// Returns `None` if `p` is outside the framebuffer bounding box.
-    // TODO: Optimise implementation by moving away from `ImageRaw::pixel` and directly calculating skip value.
     fn pixel(&self, p: Point) -> Option<C> {
         self.as_image().pixel(p)
     }
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -131,14 +131,6 @@ where
     }
 }
 
-fn check<C>() -> usize
-where
-    C: PixelColor,
-    <<C as PixelColor>::Raw as RawData>::Storage: Into<u8>,
-{
-    C::Raw::BITS_PER_PIXEL
-}
-
 macro_rules! impl_bit {
     ($raw_type:ident) => {
         impl<C, BO, const WIDTH: usize, const HEIGHT: usize, const N: usize>
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -123,9 +123,6 @@ where
 {
     type Color = C;
 
-    /// Get pixel color.
-    ///
-    /// Returns `None` if `p` is outside the framebuffer bounding box.
     fn pixel(&self, p: Point) -> Option<C> {
         self.as_image().pixel(p)
     }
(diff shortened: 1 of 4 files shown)


src/framebuffer.rs
```diff
@@ -141,7 +140,7 @@ macro_rules! impl_bit {
                         let byte_index = bytes_per_row * y + (x / pixels_per_bit);
                         let bit_index = 8 - (x % pixels_per_bit + 1) * C::Raw::BITS_PER_PIXEL;
 
-                        let mask = !($raw_type::new(0xFF).into_inner() << bit_index);
+                        let mask = !((2u8.pow(C::Raw::BITS_PER_PIXEL as u32) - 1) << bit_index);
```
rfuest, line 143: Nice performance improvement for such a small change!

jamwaffles: Yeah! I was surprised that it would help at all, moreso how much.

rfuest: My guess would be that the compiler doesn't recognize that the value is constant because `into_inner` isn't a const fn.

But I'm currently working on the `tinytga` PR and did some benchmarking/optimizing myself and I was surprised many times how much/little some changes impacted the performance. Maybe compilers are magic and we shouldn't even try to understand them.

jamwaffles: > My guess would be that the compiler doesn't recognize that the value is constant because into_inner isn't a const fn.

That could be it, but it's hard to say. I tried to get some output from godbolt.org with the two approaches yesterday but didn't get anywhere; it either produced nothing (presumably optimised out) or way too much debug cruft so I couldn't see what the line was compiled to.

rfuest:
Regarding the performance of the `pixel` getter: Should we just accept the limitation that `Framebuffer::pixel` is much slower than `set_pixel`? In some use cases it should be possible to get better performance by converting the framebuffer into an image with `as_image` and then read the pixels via `ImagePixelGetter::pixel`. But that wouldn't work for something like an image filter, which needs to get and set pixels.

`ImagePixelGetter::pixel` can be optimized a bit, but `Framebuffer::pixel` still has a lot over overhead:
```rust
    fn pixel(&self, p: Point) -> Option<Self::Color> {
        if p.x < 0 || p.y < 0 || p.x >= self.size.width as i32 || p.y >= self.size.height as i32 {
            return None;
        }

        RawDataSlice::new(self.data)
            .into_iter()
            .nth(p.x as usize + p.y as usize * self.data_width() as usize)
            .map(|r| r.into())
    }
```
(I can open a separate PR for this change if you prefer)

jamwaffles:
There's still more that could be done, particularly around `fill_solid` with <= 8bpp displays, but I'm drawing a line in the sand with the current improvements so we can continue to (slowly) roll the ball.

With the following commands:

```bash
# Or whatever the hash of the commit titled "Add framebuffer benchmarks" is at time of writing
git checkout 565c10011aa70be59643b51e66ab57a6fca0a6e8
cargo bench --bench framebuffer -- --save-baseline base
git checkout -
cargo bench --bench framebuffer -- --baseline base
```

I got these results on an AMD Ryzen 5950X, 32GB RAM, Windows 11 (I know and I'm sorry):

| bench name                   | % time   |
| ---------------------------- | -------- |
| framebuffer set pixel 1bpp   | -24.430% |
| framebuffer get pixel 1bpp   | -30.703% |
| framebuffer set pixel rgb565 | -3.8469% |
| framebuffer get pixel rgb565 | -30.686% |
| framebuffer 1bpp draw iter   | -16.023% |

Not mindblowing, but still decent. Thank you also for your suggestion in https://github.com/embedded-graphics/embedded-graphics/pull/694#issuecomment-1252512517. I don't recall the exact improvement, but it helped by double-digit percent which is great to see.

src/framebuffer.rs
```diff
@@ -123,7 +123,9 @@ where
 {
     type Color = C;
 
-    // TODO: Optimise implementation by moving away from `ImageRaw::pixel` and directly calculating skip value.
+    /// Get pixel color.
```
rfuest, line 126: What is the benefit of adding a doc comment for this? Wouldn't it be enough to let rustdoc reuse the trait's doc comment?

If you want to keep the comment it should start with "Gets" or "Returns".

jamwaffles: Hm. I think this was from a botched rebase, but either way I've removed it as there is indeed no benefit.

rfuest (approved):
LGTM. It's a nice start and the benches will help with future optimizations.

PR #698, opened 2022-10-11
[WIP] Bounding boxes for Arc and Sector

### LINKED ISSUE #405: Bounding boxes for Arc and Sector don't take the angles into account

cedeber:

## PR description

Fix #405 Bounding boxes for Arc and Sector don't take the angles into account.

### DIFF
--- a/src/primitives/arc/mod.rs
+++ b/src/primitives/arc/mod.rs
@@ -9,6 +9,7 @@ use crate::{
 mod points;
 mod styled;
 
+use crate::geometry::Trigonometry;
 pub use points::Points;
 pub use styled::StyledPixelsIterator;
 
@@ -124,7 +125,29 @@ impl PointsIter for Arc {
 
 impl Dimensions for Arc {
     fn bounding_box(&self) -> Rectangle {
-        Rectangle::new(self.top_left, Size::new(self.diameter, self.diameter))
+        // Center
+        let bb = Rectangle::new(self.top_left, Size::new(self.diameter, self.diameter));
+        let cx: f32 = bb.center().x as f32;
+        let cy: f32 = bb.center().y as f32;
+
+        // Starting point position
+        let start_cos: f32 = self.angle_start.cos().into();
+        let start_sin: f32 = self.angle_start.sin().into();
+
+        let start_x = cx + ((self.diameter as f32) / 2.0) * start_cos;
+        let start_y = cy - ((self.diameter as f32) / 2.0) * start_sin;
+
+        // End point position
+        let end_cos: f32 = (self.angle_start + self.angle_sweep).cos().into();
+        let end_sin: f32 = (self.angle_start + self.angle_sweep).sin().into();
+
+        let end_x = cx + ((self.diameter as f32) / 2.0) * end_cos;
+        let end_y = cy - ((self.diameter as f32) / 2.0) * end_sin;
+
+        Rectangle::with_corners(
+            Point::new(start_x as i32, start_y as i32),
+            Point::new(end_x as i32, end_y as i32),
+        )
     }
 }
 


jamwaffles (commented):
I'm by no means an expert on the arc/sector code, nor the algorithms behind it, so I've asked @rfuest to give it a check over. Superficially, to me it looks quite clean but my gut tells me there are some optimisations to be made somewhere.

It also looks like there are some regressions in the tests - please check the CI output or `cargo test` locally if you can't get into CircleCI.

src/primitives/arc/mod.rs
```diff
@@ -106,9 +107,51 @@ impl Arc {
...
+
+    /// Returns the end angle of the arc.
+    pub fn angle_end(&self) -> Angle {
+        self.angle_start + self.angle_sweep
+    }
+
+    /// Returns the Point from a certain angle.
+    pub fn point_from_angle(&self, angle: &Angle) -> Point {
+        let center = self.center();
+        let radius = self.radius();
```
rfuest, line 127: To get accurate results you'll need to use `Circle::center_2x` instead of `center` and scale all coordinates by 2. `center` does round the center point to the nearest integer, which isn't accurate for circles/arcs with even diameters. The radius in the scaled up coordinate system is then equal to `diameter`.

src/primitives/arc/mod.rs
```diff
@@ -122,9 +165,35 @@ impl PointsIter for Arc {
...
 }
 
+// TODO Check rect.styled_bounding_box()
 impl Dimensions for Arc {
     fn bounding_box(&self) -> Rectangle {
-        Rectangle::new(self.top_left, Size::new(self.diameter, self.diameter))
+        // Create a rectangle between the start and the end angle points.
+        let rect = Rectangle::with_corners(self.point_start(), self.point_end());
+        let mut p1 = rect.top_left;
+        let mut p2 = p1 + rect.size.x_axis() + rect.size.y_axis();
```
rfuest, line 174: `p2` isn't correct, because it would be outside the rect. This should work:
```suggestion
        let start = self.point_start();
        let end = self.point_start();        
        let mut p1 = start.component_min(end);
        let mut p2 = start.component_max(end);
```

rfuest (changes requested):

rfuest:
I only did take a quick look at the code, but the general concept looks good. If you haven't already noticed it I would also like to point you to the `debug-tools` repo which has tools to test the bounding box of `Arc` and `Sector`: https://github.com/embedded-graphics/debug-tools/blob/master/debug-tools/examples/arc.rs

src/primitives/arc/mod.rs
cedeber: I feel like dividing the returned `Point` by 2 will have the same effect, isn't it?

rfuest: Which `Point` do you mean? The result of `point_from_angle`?

cedeber: Yes. I feel I could avoid to work with the radius on the `bounding_box` function and simply reuse `top_left + diameter`, etc.

rfuest: I'm still not sure I understand what you mean. But you could get rid of using `Real` for the `radius` function by scaling the coordinate system up by a factor of 2.

cedeber:
```rust
let arc = Arc::new(Point::new(10, 10), 5, 0.0.deg(), 90.0.deg());
let center = self.to_circle().center_2x();
```

center is `Point { x: 24, y: 24 }`, and it should be `Point { x: 25, y: 25 }`

rfuest: `(24, 24)` or `(12, 12)` (unscaled) is the correct center. Here is a diagram that shows a circle with top left at `(10, 10)` and a diameter of 5:

grafik

Integer coordinates in e-g are centered inside the corresponding pixel (blue dots in the diagram). But the outside of filled shapes are defined to be aligned with the boundaries between pixels (grey grid lines in the diagram).

cedeber: Thanks a lot for the help and explanation. Not easy 😅
I pushed the changes.

cedeber:
> It also looks like there are some regressions in the tests - please check the CI output or `cargo test` locally if you can't get into CircleCI.

It's because of the bounding box usage for the dimension test for instance. I fixed them by simply using the circle conversion and add other tests for the bounding box.
I still need to do the code for the styling bounding box.

src/primitives/arc/mod.rs
rfuest: > Not easy sweat_smile

No it's not, the half pixel offset that creeps into many calculations can get really confusing.

The BB isn't correct for some negative start angles:
grafik

rfuest: Negative sweep angles are also broken:
grafik

cedeber: Oh, completely forgot to test the negatives angles 😅
But I've seen that I forgot to normalize the angles during the refactoring, which is probably why it's broken. (hopefully)

cedeber: Ok, I fixed it by using the PlaneSector that @jamwaffles introduced on the other PR. Although it's not always pixel perfect.

With the debug tools, the stroke size pushes the bounding box in a weird way. Shouldn't it stay the same, disregarding the stroke width?

cedeber:
I have the feeling that the stroke for Sector is wrong at the center. It looks twice as large as what I think it should be.
In the screenshot bellow the stroke from the center is 30px, not 15px. Isn't it wrong?
Capture d’écran 2022-11-16 à 17 51 15

rfuest:
I think it is correct, or at least the way it was designed to work. The two line segments should be drawn the same way as a `Polyline` with the two arc endpoints and the center point of the arc. You could verify that it behaves correctly by drawing a `Polyline`, but keep in mind that stroke alingment isn't supported for polylines.

rfuest:
But something definitely isn't right with some edge cases. I'll add some of my findings to #484.

cedeber:
I quickly checked how Sketch or Affinity Designer behave with polygones and large strokes. Well, it's bad too with narrow angles :-/ Definitely not an easy topic.

Now I am wondering if it wouldn't be better to offer only rounded corners instead, because it has the advantage to be consistant with every angle values. (For Sector and polygones at least)

Not sure it's easier to implement though ^_^

Issue #696, opened 2022-10-01
Should `ImagePixelGetter` be renamed?


Nerixyz:
As discussed in embedded-graphics/tinybmp#35:

I think `ImagePixelGetter` (added in #612) should be named something like `GetPixel` which is more Rust-like (when comparing to traits from std: `FromStr`, `AddAssign`, `IntoFuture`).

rfuest:
@jamwaffles I would like to hear your opinion on this. Do we have any other traits with similar naming conventions?

jamwaffles:
I wouldn't be opposed to a rename. A quick peruse shows we already have stuff like `Drawable`, `DrawTarget`, `AngleUnit`, `OriginDimensions`, `WebColors`, etc. I can't think of a better name than `GetPixel`, but I feel there should be something in it that relates to the source data being a 2D array of pixels, as opposed to e.g. a primitive that generates its pixels dynamically. `GetImagePixel`? Still sounds weird...

rfuest:
I think that `GetPixel` would be my favorite name if the trait wouldn't depend on the super trait `ImageDrawable`. We could change the trait to include an associated color type instead of reusing `ImageDrawable::Color`:
```rust
pub trait GetPixel {
    type Color: PixelColor;
    fn pixel(&self, p: Point) -> Option<Self::Color>;
}
```
This would make the trait more flexible and we could also use it for `Framebuffer::pixel` (`Framebuffer` doesn't implement `ImageDrawable`)

jamwaffles:
Sounds good to me!

PR #699, opened 2022-10-16
Rename ImagePixelGetter to GetPixel

rfuest:
This PR renames the pixel getter as discussed in #696.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 ### Added
 
 - [#592](https://github.com/embedded-graphics/embedded-graphics/pull/592) Add support for defmt behind a feature (`defmt`).
-- [#612](https://github.com/embedded-graphics/embedded-graphics/pull/612) Added `GetPixel` trait and an implementations for `ImageRaw` and `Framebuffer`.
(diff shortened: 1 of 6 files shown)


rfuest:
Is there a better place to put this trait than the `image` module, now that it isn't used exclusively for images?

jamwaffles (approved):
lgtm

> Is there a better place to put this trait than the image module, now that it isn't used exclusively for images?

That is an excellent question. It doesn't quite fit with `image` any more but I can't think of a better place to put it. Maybe we could generalise the image/framebuffer/pixel getter stuff into a newly named module? I can't think of a name for that either... `SquareStuffWithPixels` doesn't exactly roll off the toungue.

rfuest:
> `SquareStuffWithPixels` doesn't exactly roll off the toungue.

That would be the perfect name, but unfortunately the trait could also be used for non square stuff :wink: 

The only other place that would fit IMO is the crate root, but the trait might not be important enough to deserve the honor of being there.

jamwaffles:
> unfortunately the trait could also be used for non square stuff 

I spent a long time and all of my sponsorship money on a focus group for that name. What a waste :(

> The only other place that would fit IMO is the crate root, but the trait might not be important enough to deserve the honor of being there.

I agree. It seems the best place for `GetPixel` is in `image`, for now at least.

PR #709, opened 2022-11-15
Add Image::with_center

rfuest:
This PR adds `Image::with_center` to position images centered around a given point.


jamwaffles (approved):

jamwaffles:
Post-approval thought: is it a good idea to add a test for odd-sized images, or is just relying on the rectangle code, which is already tested, enough do you think?

rfuest:
> Post-approval thought: is it a good idea to add a test for odd-sized images, or is just relying on the rectangle code, which is already tested, enough do you think?

IMO it is enough and I think we already rely on `Rectangle::with_center` for other constructors, like `Circle::with_center`. But I can add an addition test if you prefer that.

jamwaffles:
No I agree the current tests are fine :)

Issue #703, opened 2022-11-07
Add one dimensional variants of Rectangle::resized


rfuest:
Using `Rectangle::resized` to only change either the width or height of a rectangle is a fairly common operation in my code. But the resulting code is very noisy and this could be improved by adding two new methods:
```rust
// Current code:
let bounding_box = target.bounding_box(); // Note that a temporary variable is necessary
let right = bounding_box.resized(Size::new(1, bounding_box.size.height), AnchorPoint::TopRight);

// With the new method:
let right = target.bounding_box().resized_x(1, AnchorPoint::TopRight);
```

Remaining questions:
- How should the methods be named?
  `resized_x`/`resized_y`, `resized_horizontal`/`resized_vertical`, `change_width`/`change_height`, ...
- Should we reuse `AnchorPoint` or create two new enums?
  `resized_x(width, AnchorPoint::TopLeft)`, `resized_x(width, AnchorPoint::CenterLeft)`, and `resized_x(width, AnchorPoint::BottomLeft)` would, for example, produce the same output.

  The new types could look like this and provide some methods to convert from/to `AnchorPoint`:

  ```rust
  pub enum AnchorPointX { Left, Center, Right }
  pub enum AnchorPointY { Top, Center, Bottom }
  
  impl AnchorPoint {
      pub fn from_xy(x: AnchorPointX, y: AnchorPointY) -> Self { ... }
      pub fn x(self) -> AnchorPointX { ... }
      pub fn y(self) -> AnchorPointY { ... }
  }
  ```

jamwaffles:
> How should the methods be named?

How about the 4th, secret option: `resized_width`/`resized_height`? I feel like that fits with `Size` which also uses the "width" and "height" wording.

> Should we reuse AnchorPoint or create two new enums?

Separate X/Y types would encode the intent in the type system which is nice. Reusing `AnchorPoint` for individual axes would make "half" of the given variant meaningless, e.g. resizing on X with `AnchorPoint::TopRight`, the `Top` component is ignored.

As we're going for a bit of API sugar here too, we could call the two new enums `AnchorX` and `AnchorY` to reduce chars typed slightly.

rfuest:
> How about the 4th, secret option: `resized_width`/`resized_height`? I feel like that fits with `Size` which also uses the "width" and "height" wording.

SGTM.

[...]

PR #710, opened 2022-11-15
Add 1D rectangle resizing

rfuest:
This PR adds `Rectangle::resized_x` and `Rectangle::resized_y` to resize rectangles in one dimension. Some support types and methods are also added.

### DIFF
--- a/core/src/geometry/mod.rs
+++ b/core/src/geometry/mod.rs
@@ -116,7 +116,7 @@ pub enum AnchorPoint {
 }
 
 impl AnchorPoint {
-    /// Creates an anchor point from a X and Y component.
+    /// Creates an anchor point from an X and Y component.
     pub fn from_xy(x: AnchorX, y: AnchorY) -> Self {
         match (y, x) {
             (AnchorY::Top, AnchorX::Left) => AnchorPoint::TopLeft,
--- a/core/src/geometry/mod.rs
+++ b/core/src/geometry/mod.rs
@@ -114,3 +114,100 @@ pub enum AnchorPoint {
     /// Bottom right.
     BottomRight,
 }
+
+impl AnchorPoint {
+    /// Creates an anchor point from a X and Y component.
+    pub fn from_xy(x: AnchorX, y: AnchorY) -> Self {
+        match (y, x) {
+            (AnchorY::Top, AnchorX::Left) => AnchorPoint::TopLeft,
+            (AnchorY::Top, AnchorX::Center) => AnchorPoint::TopCenter,
+            (AnchorY::Top, AnchorX::Right) => AnchorPoint::TopRight,
+            (AnchorY::Center, AnchorX::Left) => AnchorPoint::CenterLeft,
+            (AnchorY::Center, AnchorX::Center) => AnchorPoint::Center,
+            (AnchorY::Center, AnchorX::Right) => AnchorPoint::CenterRight,
+            (AnchorY::Bottom, AnchorX::Left) => AnchorPoint::BottomLeft,
+            (AnchorY::Bottom, AnchorX::Center) => AnchorPoint::BottomCenter,
+            (AnchorY::Bottom, AnchorX::Right) => AnchorPoint::BottomRight,
+        }
+    }
+
+    /// Returns the X axis component.
+    pub fn x(self) -> AnchorX {
+        match self {
+            AnchorPoint::TopLeft | AnchorPoint::CenterLeft | AnchorPoint::BottomLeft => {
+                AnchorX::Left
+            }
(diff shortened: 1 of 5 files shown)


rfuest:
Oops, I just noticed I didn't used the names we had decided to use in #703.

core/src/geometry/mod.rs
```diff
@@ -114,3 +114,100 @@ pub enum AnchorPoint {
     /// Bottom right.
     BottomRight,
 }
+
+impl AnchorPoint {
+    /// Creates an anchor point from a X and Y component.
```
jamwaffles, line 119:
```suggestion
    /// Creates an anchor point from an X and Y component.
```

core/src/primitives/rectangle/mod.rs
```diff
@@ -268,24 +268,93 @@ impl Rectangle {
...
+        let mut resized = self.clone();
+        resized.resize_width_mut(size.width, anchor_point.x());
+        resized.resize_height_mut(size.height, anchor_point.y());
+
+        resized
+    }
+
+    /// Returns a resized rectangle with the given width.
+    ///
+    /// The rectangle is resized relative to the given anchor edge.
```
jamwaffles, line 280: I'm not convinced this is better, but it's nice to have the full behaviour described in one line. Might be helpful for some IDEs that only show the first line too. If you think the current wording is better, feel free to leave it as is.

```suggestion
    /// Returns a new rectangle with the given width, resized relative to the given anchor edge.
```

core/src/primitives/rectangle/mod.rs
```diff
@@ -268,24 +268,93 @@ impl Rectangle {
...
+    pub fn resized_width(&self, width: u32, anchor_x: AnchorX) -> Self {
+        let mut resized = self.clone();
+        resized.resize_width_mut(width, anchor_x);
+
+        resized
+    }
+
+    /// Returns a resized rectangle with the given height.
+    ///
+    /// The rectangle is resized relative to the given anchor edge.
```
jamwaffles, line 308: Same as above

```suggestion
    /// Returns a new rectangle with the given height, resized relative to the given anchor edge.
```

core/src/primitives/rectangle/mod.rs
```diff
@@ -321,21 +390,63 @@ impl Rectangle {
...
     pub fn anchor_point(&self, anchor_point: AnchorPoint) -> Point {
+        Point::new(
+            self.anchor_x(anchor_point.x()),
+            self.anchor_y(anchor_point.y()),
+        )
+    }
+
+    /// Returns the X coordinate of an anchor edge.
+    ///
+    /// # Examples
```
jamwaffles, line 401:
```suggestion
    /// # Examples
    ///
```

core/src/primitives/rectangle/mod.rs
```diff
@@ -321,21 +390,63 @@ impl Rectangle {
     /// );
     /// ```
     pub fn anchor_point(&self, anchor_point: AnchorPoint) -> Point {
+        Point::new(
+            self.anchor_x(anchor_point.x()),
+            self.anchor_y(anchor_point.y()),
+        )
+    }
+
+    /// Returns the X coordinate of an anchor edge.
```
jamwaffles, line 399: Maybe this would work better:

```suggestion
    /// Returns the X coordinate of a given anchor edge of the rectangle.
```

I found the original a bit too general, although it _is_ within the context of `Rectangle` so it doesn't matter that much.

core/src/primitives/rectangle/mod.rs
```diff
@@ -321,21 +390,63 @@ impl Rectangle {
...
+
+        self.top_left.x
+            + match anchor_x {
+                AnchorX::Left => 0,
+                AnchorX::Center => delta / 2,
+                AnchorX::Right => delta,
+            }
+    }
+
+    /// Returns the Y coordinate of an anchor edge.
```
jamwaffles, line 426:
```suggestion
    /// Returns the Y coordinate of a given anchor edge of the rectangle.
```

core/src/primitives/rectangle/mod.rs
```diff
@@ -321,21 +390,63 @@ impl Rectangle {
...
+    ///
+    /// # Examples
```
jamwaffles, line 428:
```suggestion
    /// # Examples
    ///
```

jamwaffles (changes requested):
Looks good, just some minor nits and comment suggestions.

rfuest:
Thanks for the suggestions to improve the docs, I've applied all of them.

jamwaffles (approved):
Good to go!

Issue #705, opened 2022-11-13
Drawing an invalid MonoFont can panic


rfuest:
`MonoFont`s are constructed using struct literals and not by builder, which could check that the values in the `MonoFont` constant make sense. If `character_size.width == 0` or `image.size().width < character_size` an divide by 0 error occurs in the `MonoFont::glyph`. While there wouldn't be any usable output anyway, there shouldn't be divide by zero error IMO. For debug builds the error condition could explicitly checked with a debug assertion and an appropriate error message.

PR #713, opened 2022-11-28
Fix possible panic in MonoFont

rfuest:
This PR fixes a possible panic if an invalid `MonoFont` struct is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 ### Fixed
 
 - [#648](https://github.com/embedded-graphics/embedded-graphics/pull/648) Fixed incorrect text bounding box calculation when multi-byte characters are used.
+- [#713](https://github.com/embedded-graphics/embedded-graphics/pull/713) Fixed possible panic when drawing an invalid `MonoFont`.
 
 ## [0.7.1] - 2021-06-15
 
--- a/src/mono_font/mod.rs
+++ b/src/mono_font/mod.rs
@@ -98,6 +98,10 @@ pub struct MonoFont<'a> {
 impl MonoFont<'_> {
     /// Returns a subimage for a glyph.
     pub(crate) fn glyph(&self, c: char) -> SubImage<'_, ImageRaw<BinaryColor>> {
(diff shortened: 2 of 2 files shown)


jamwaffles:
#705 mentions triggering a debug assertion if the config is invalid. Wouldn't it be better to do that than allow the creation of a `MonoFont` which would silently never output anything?

rfuest:
> #705 mentions triggering a debug assertion if the config is invalid. Wouldn't it be better to do that than allow the creation of a `MonoFont` which would silently never output anything?

Yes, it would. I guess I should have read my own comment before working on this :thinking:

jamwaffles:
Hm good point. I guess we could `cargo test` and `cargo test --release` in CI with some `cfg` options, but that doesn't feel clean. Perhaps the code is simple enough to assume release mode is ok? I'm not sure which way I'd lean on that.

rfuest:
That wouldn't work either. If we don't add `cfg` attributes, but test for panics caused by debug assertions, it would no longer be possible to run the tests successfully in release mode.

Perhaps running the tests in debug and release mode in CI is the safest option? I'm not sure I like the solution and it will probably impact the CI runtime, but I don't have a better idea at the moment.

jamwaffles:
No I don't like it either :(. Perhaps it's easier to add something to the docs to mention that nothing will be drawn if the bitmap is too small. At any rate, fixing the panic is the focus of this PR which is already achieved :)

rfuest:
> At any rate, fixing the panic is the focus of this PR which is already achieved :)

OK, let's just ignore the problem :stuck_out_tongue_winking_eye: I think it's reasonable to assume that noting will be drawn if the image is smaller than the character size.

In the long term we could add a `MonoFontBuilder` to check that the data in the `MonoFont` object will be valid. But this will require a const version of `unwrap`, which isn't stabilized yet.

jamwaffles (approved):
> OK, let's just ignore the problem 😜 I think it's reasonable to assume that noting will be drawn if the image is smaller than the character size.

Agreed

> In the long term we could add a MonoFontBuilder to check that the data in the MonoFont object will be valid. But this will require a const version of unwrap, which isn't stabilized yet.

There's a lot of `const` stuff I'm sorely missing and couldn't come soon enough to be honest, so this is going on that list.

Issue #655, opened 2022-03-19
Point e-g to e-g-core 0.3.3


jamwaffles:
The repo `Cargo.toml` asks for `0.3.2` whereas the latest is `0.3.3`. Not sure why `cargo-release` didn't pick this up, but it's been so long since a release I can't remember if it's manual or not.

This is a minor issue, but may help fix some strange dependency issues a user on Matrix is seeing.

rursprung:
if/when #717 gets merged dependabot should take care of these things (by creating PRs for updates if they can be applied automatically).

PR #717, opened 2022-12-28
enable GitHub Dependabot

rursprung:
this ensures that the dependencies are kept up to date. see [the docs][] for further information.

note: changelog intentionally left away as this only touches the build infrastructure and has no direct impact on consumers of the library.

[the docs]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "."
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "/tools"
+    schedule:
+      interval: "weekly"


rursprung:
to see how this has been added to rtic-examples see https://github.com/rtic-rs/rtic-examples/pull/49 and https://github.com/rtic-rs/rtic-examples/pull/79

jamwaffles:
Hey [rursprung](https://github.com/rursprung), I appreciate the time taken to make this contribution but I just... hate Dependabot. I think it adds more noise than anything else useful, so I'd very much prefer to not include it in embedded-graphics. #655 is an isolated case (IIRC anyway) and can be fixed manually. Unless @rfuest has some very compelling reasons to include Dependabot, I'm afraid it's a decline from my side.

rfuest:
I don't have any experience with maintaining a project which uses Dependabot. But I'll have to agree witch @jamwaffles that it can add a lot of noise.

Intra repository dependencies, like in #655, should be checked when the crates are released. A weekly check for out of date dependencies wouldn't help much to recognize these issues.

rursprung:
> I think it adds more noise than anything else useful

i have the contrary experience. yes, it adds PRs (which you can consider noise?), but these automate away things which you'd usually have to do manually and thus would ignore. see also my comment here where i go into a bit more details: https://github.com/rust-embedded/cortex-m/pull/457#issuecomment-1367150610

jamwaffles:
From my maintainer's perspective, I don't see an overall benefit to using Dependabot so I'm going to close this PR. We have few dependencies, and don't currently have major issues with keeping them up to date. For those reasons, I would prefer not to put up with the noise of Dependabot for no tangible benefit. Thanks again for raising the discussion - I really do appreciate it - but Dependabot is not something I want to use in https://github.com/embedded-graphics at this time.

PR #723, opened 2023-05-16
Cleanup

bugadani:

## PR description

This PR fixes a few typos and clippy warnings. This PR also excludes defmt from the MSRV guarantees because the crate itself doesn't guarantee anything beyond "latest stable".

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,8 @@ calculation.
 
 * `defmt` - provide implementations of `defmt::Format` for all types where possible. [`defmt`]
   is a library for logging that moves as much work as possible over to a separate logging
-  machine, making it especially suited to low-resource MCUs.
+  machine, making it especially suited to low-resource MCUs. Note that `defmt` might not work with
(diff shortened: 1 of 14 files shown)


bugadani:
how did that check fail here but not on previous PRs?

bugadani:
CI is failing because of defmt-parse. How should we approach this?

rfuest:
> CI is failing because of defmt-parser which has a "latest stable" MSRV. How should we approach this?

The only solution I can think of is to remove the `defmt` feature from all MSRV builds. I've created a PR to remove the feature: https://github.com/embedded-graphics/embedded-graphics/pull/724
At the momement CI still fails because the new `tinytga` version hasn't been published yet.

bugadani:
I tried doing that but I couldn't figure out how the justfile works 😅

rfuest:
> I tried doing that but I couldn't figure out how the justfile works sweat_smile

The justfile could use some cleanup, but I think the CI docker file still uses an ancient version of just, which would need to be updated first to use some of the newer just features.

PR #727, opened 2023-07-21
Add gc9a01 driver to list

IniterWorker:
## PR description

Adds the gc9a01 driver to the list of existing drivers


jamwaffles (changes requested):
Hey, thanks for the PR! Could you please make the same change to `src/lib.rs` - the CI is failing because we check that the README (which is generated from `src/lib.rs` is in sync).

If you have `just` installed, you can run `just generate-readmes`, otherwise manually is fine :)

IniterWorker:
The last CI's errors are not related.
```bash
error: package `half v2.3.1` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.61.0
error: Recipe `build-target` failed on line 46 with exit code 101
error: Recipe `build-targets` failed with exit code 101
```

jamwaffles (approved):
Looks good, thanks! And yeah, don't worry about the CI.

PR #730, opened 2023-08-09
Make `MonoTextStyleBuilder` fns const

bugadani:

## PR description

cc #205

### DIFF
--- a/src/mono_font/mono_text_style.rs
+++ b/src/mono_font/mono_text_style.rs
@@ -48,7 +48,10 @@ pub struct MonoTextStyle<'a, C> {
     pub font: &'a MonoFont<'a>,
 }
 
-impl<'a, C: PixelColor> MonoTextStyle<'a, C> {
+impl<'a, C> MonoTextStyle<'a, C>
+where
+    C: PixelColor,
+{
     /// Creates a text style with transparent background.
     pub const fn new(font: &'a MonoFont<'a>, text_color: C) -> Self {
         MonoTextStyleBuilder::new()
@@ -181,7 +184,10 @@ impl<'a, C: PixelColor> MonoTextStyle<'a, C> {
     }
 }
 
-impl<C: PixelColor> TextRenderer for MonoTextStyle<'_, C> {
+impl<C> TextRenderer for MonoTextStyle<'_, C>
+where
+    C: PixelColor,
+{
     type Color = C;
 
     fn draw_string<D>(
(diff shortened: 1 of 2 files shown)


bugadani:
I guess technically this is a breaking change as the PR doesn't allow nonsense, half-finished builders that have a non-PixelColor type parameter. So technically, the PR can break code that wasn't ever meant to be working anyway, IMO. Any throughts on this?

luisroel91:
Not so sure I can comment on the correctness of this...but it would be great, so we're able to define styles as consts and use them throughout our program instead of having to rebuild them on every call to a display.

src/mono_font/mono_text_style.rs
```diff
@@ -530,6 +524,11 @@ mod tests {
         ..FONT_6X9
     };
 
+    const _CAN_BUILD_CONST: MonoTextStyle<'static, BinaryColor> = MonoTextStyleBuilder::new()
```
spookyvision, line 527: I'd probably move that to a test fn

bugadani: It would be the same thing, just with more lines, and it would also create a test that asserts nothing, and does nothing except print that it passes. Would it really be better?

spookyvision: personal preference: either test fn or remove entirely - it's just a showcase, right? Either way no strong feelings

jamwaffles (approved):
Looks good, thank you! I'm ok with the technically breaking change with the new trait bound - it's a good addition to make.

src/mono_font/mono_text_style.rs
```diff
@@ -385,7 +385,7 @@ pub struct MonoTextStyleBuilder<'a, C> {
     style: MonoTextStyle<'a, C>,
 }
 
-impl<C> MonoTextStyleBuilder<'_, C> {
+impl<'a, C: PixelColor> MonoTextStyleBuilder<'a, C> {
```
jamwaffles, line 388: Nit: could you please change this to:

```suggestion
impl<'a, C> MonoTextStyleBuilder<'a, C> where C: PixelColor {
```

to be consistent with the rest of the codebase.

jamwaffles (changes requested):
Ah sorry, I was reading through the rest of e-g again and remembered we chose the more verbose style. Needs a `cargo fmt` if you apply my suggestion too. Don't worry about the CI failure - we're just on a too-old Rust version.

bugadani:
Sorry for not going with your suggestion directly, but there were other bounds to move, too.

jamwaffles (approved):
All good! Thanks for the changes.

PR #734, opened 2023-11-07
Fix crash in README example

BroderickCarlin:

## PR description

Addresses https://github.com/embedded-graphics/embedded-graphics/issues/729

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,14 +142,15 @@
 //!
 //! ## Shapes and text
 //!
-//! The following example draws some shapes and text to a [`MockDisplay`] in place of target
-//! hardware. The [simulator](https://docs.rs/embedded-graphics-simulator/) can also be used for
+//! The following example draws some shapes and text to a [`FrameBuffer`](crate::framebuffer::Framebuffer)
+//! in place of target hardware. The [simulator](https://docs.rs/embedded-graphics-simulator/) can also be used for
 //! debugging, development or if hardware is not available.
 //!
 //! ```rust,no_run
 //! use embedded_graphics::{
+//!     framebuffer::{buffer_size, Framebuffer},
 //!     mono_font::{ascii::FONT_6X10, MonoTextStyle},
-//!     pixelcolor::BinaryColor,
+//!     pixelcolor::{raw::LittleEndian, BinaryColor},
 //!     prelude::*,
 //!     primitives::{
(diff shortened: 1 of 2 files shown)


rfuest:
IMO it would be better to replace `MockDisplay` with the new `Framebuffer`, that was added in e-g 0.8. `MockDispay` was only used in the example because no other `DrawTarget` was included in e-g at the time the example was written.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed 
+
+- [#734](https://github.com/embedded-graphics/embedded-graphics/pull/734) Fix panic if the README example is ran
+
```
rfuest, line 12: I don't think this change needs a changelog entry. Our policy had been to only include important API changes in the changelog to make it easier for users to find the relevant changes.

BroderickCarlin:
Removed the changelog entry and updated to use the `FrameBuffer` type

src/lib.rs
```diff
@@ -159,9 +160,15 @@
 //! };
 //!
 //! fn main() -> Result<(), std::convert::Infallible> {
-//!     // Create a new mock display
-//!     let mut display: MockDisplay<BinaryColor> = MockDisplay::new();
-//! #   display.set_allow_overdraw(true);
+//!     // Create a new mock framebuffer
```
rfuest, line 163:
```suggestion
//!     // Create a new framebuffer
```

rfuest (changes requested):

rfuest:
> My initial thought is that this made the example more verbose and I'm not sure it is a great README level example given the type complexity of the `FrameBuffer`.

Using `MockDisplay` also doesn't make a lot of sense IMO, because it's only intended for writing tests and won't be used by most users. They are more likely to use `Framebuffer` in real applications, but I agree that the complex type isn't great for an example.

In the past we had used the unmodified simulator example (https://github.com/embedded-graphics/examples/blob/main/eg-next/examples/hello-world.rs), which would be the more useful, but we did run into CI issues with cyclic dependencies.

rfuest:
The paragraph above the example code still contains a reference to `MockDisplay`.

BroderickCarlin:
Good catch on the lingering `MockDisplay` - fixed

> They are more likely to use `Framebuffer` in real applications

I think this is a fair argument and points to e-g providing more than just graphical primitives. I'm fine leaving it as-is

rfuest (approved):
LGTM

PR #744, opened 2024-02-12
Adjust performance expectations in the readme

Kezii:
In my testing, using esp32 and GC9A01A (240x240), writing to the display with the DrawTarget interface is ~1000% slower for a simple framebuffer, up to >5000% slower for elaborate geometry, this is compared to using a (tiled) framebuffer and copying it with DMA, following the conventional industry approach.

This does not change with -s or -O3 optimization, so the description of the project should be updated to better reflect the scope of the project.

Thank you


rfuest:
Some operations are certainly not as fast as they could be and the documentation is lacking performance optimization tips. But the performance does depend on many factors and is are also depended on external factors, like the display driver and the `display-interface` crate.

If you want to investigate the performance issues further feel free to open an issue with your findings, but I will close this PR for now.

PR #756, opened 2024-07-24
Implemented PartialEq and Eq for Image and SubImage

za-songguo:

## PR description

Implemented `PartialEq` and `Eq` for Image and `SubImage`.

### DIFF
--- a/benches/primitives.rs
+++ b/benches/primitives.rs
@@ -58,7 +58,7 @@ fn sector_150(c: &mut Criterion) {
         Sector::with_center(
             BOUNDING_BOX.center(),
             BOUNDING_BOX.size.width,
-            -30.0.deg(),
+            -(30.0.deg()),
             150.0.deg(),
         )
     });
@@ -69,7 +69,7 @@ fn sector_360(c: &mut Criterion) {
         Sector::with_center(
             BOUNDING_BOX.center(),
             BOUNDING_BOX.size.width,
-            -30.0.deg(),
+            -(30.0.deg()),
             150.0.deg(),
         )
     });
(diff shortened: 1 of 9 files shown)


benches/primitives.rs
```diff
@@ -58,7 +58,7 @@ fn sector_150(c: &mut Criterion) {
         Sector::with_center(
             BOUNDING_BOX.center(),
             BOUNDING_BOX.size.width,
-            -30.0.deg(),
+            -(30.0.deg()),
```
rfuest, line 61: Is this a clippy fix or why did you include this change in this PR?

src/image/sub_image.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
 ///
 /// [`sub_image`]: trait.ImageDrawableExt.html#tymethod.sub_image
 /// [module-level documentation]: super#sub-images
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
```
rfuest, line 19: I think we should also add `Clone` and `Copy` to the derives.

rfuest (changes requested):
Looks good, but you should add a changelog entry for the added `Default` impls and for making it possible to compare `Image`s and `SubImage`s.

benches/primitives.rs
za-songguo: > Is this a clippy fix or why did you include this change in this PR?

I ran `cargo clippy --fix` in the repo.

za-songguo: OK, I will add them to `CHANGELOG.md`.

src/image/sub_image.rs
za-songguo: I have just committed the new changes to the repository.

rfuest (approved):
LGTM. Thanks for the PR.

PR #763, opened 2024-09-06
Add swap_xy for Point and Size

rfuest:
This PR adds `Point::swap_xy` and `Size::swap_xy` to swap the coordinates. The main advantage of having these functions instead of using `Point::new(p.y, p.x)` is that they can be used as part of a longer expression without having to us a temporary variable.

@jamwaffles: Would you like to review this PR or should I just merge it? This was a part of #711, which I want to slowly merge in smaller PRs.

### DIFF
--- a/core/src/geometry/point.rs
+++ b/core/src/geometry/point.rs
@@ -21,7 +21,7 @@ use crate::geometry::Size;
 ///
 /// ## Create a `Point` from two integers
 ///
-/// ```rust
+/// ```
 /// use embedded_graphics::geometry::Point;
 ///
 /// // Create a coord using the `new` constructor method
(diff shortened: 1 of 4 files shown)


core/src/geometry/point.rs
```diff
@@ -234,6 +238,21 @@ impl Point {
     pub const fn component_div(self, other: Self) -> Self {
         Self::new(self.x / other.x, self.y / other.y)
     }
+
+    /// Returns a point with swapped x and y coordinates.
```
jamwaffles, line 242: Nit: we use uppercase in a couple of other places at least

```suggestion
    /// Returns a point with swapped X and Y coordinates.
```

jamwaffles (approved):
Thanks for asking :). Just a tiny nit, otherwise this looks good. It's good to see #711 being driven slowly forward!

rfuest:
Thanks for the review.

PR #764, opened 2024-09-07
Return error from ImageRaw::new

rfuest:
This PR changes the signature of `ImageRaw` to take a `Size` argument and to return a `Result`. If I remember correctly the previous method of only specifying the `width` of the image was introduced to work around some const limitation at the time. Returning an error instead of silently truncating the height of the image is a much better approach in my opinion.

The changes in this PR affected the auto generated font code which makes it noisy and appear much more extensive than it is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 ### Changed
 
+- **(breaking)** [#764](https://github.com/embedded-graphics/embedded-graphics/pull/764) Changed `ImageRaw::new` to return an error instead of truncating the image height.
 - [#732](https://github.com/embedded-graphics/embedded-graphics/pull/732) Added `Rgb444` to support 12bit RGB displays. Note that this type is currently stored as a 16 bit value in `ImageRaw`s and `Framebuffer`s.
 
 ### Added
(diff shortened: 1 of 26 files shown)


CHANGELOG.md
```diff
@@ -17,6 +18,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#756](https://github.com/embedded-graphics/embedded-graphics/pull/756) Added `PartialEq, Eq` derives to `Image` and `SubImage`.
 - [#756](https://github.com/embedded-graphics/embedded-graphics/pull/756) Added `Clone, Copy` derives to `SubImage` and `Default` impls to `Framebuffer`, `MonoTextStyleBuilder` and `TextStyleBuilder`.
 - [#763](https://github.com/embedded-graphics/embedded-graphics/pull/763) Added `swap_xy` method to `Point` and `Size`.
+- [#764](https://github.com/embedded-graphics/embedded-graphics/pull/764) Added `ImageRaw::new_const` as a panicing alternative to the new `ImageRaw::new` for ease of use in const contexts.
```
jamwaffles, line 21:
```suggestion
- [#764](https://github.com/embedded-graphics/embedded-graphics/pull/764) Added `ImageRaw::new_const` as a panicking alternative to the new `ImageRaw::new` for ease of use in const contexts.
```

jamwaffles (approved):
This is a much nicer API. The magic single number from before never felt quite right. A small spelling nit in the changelog, otherwise this LGTM

PR #766, opened 2024-09-16
Make conversion from and to RawUx mandatory

rfuest:
This PR makes `From<RawUx>` and `Into<RawUx>` mandatory for all `PixelColor`s. This is different from #711, where an additional trait `StorablePixelColor` was used.

At the moment this change doesn't have many advantages, except for making some trait bounds a bit simpler. But I think that requiring all `PixelColor`s to have a binary representation will have advantages if we want to add DMA transfers.

### DIFF
--- a/core/src/pixelcolor/mod.rs
+++ b/core/src/pixelcolor/mod.rs
@@ -24,8 +24,8 @@
 //!     Red,
 //! }
 //!
-//! /// Implement the `PixelColor` trait to mark this type as a embedded-graphics
-//! /// color and to specify the binary storage format.
+//! /// Implement the `PixelColor` trait to mark this type as an embedded-graphics
+//! /// color and associated binary storage format.
 //! impl PixelColor for EpdColor {
 //!     // 2 bits per pixel are required to store the 3 colors.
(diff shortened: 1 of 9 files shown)


core/src/pixelcolor/mod.rs
```diff
@@ -23,10 +24,33 @@
 //!     Red,
 //! }
 //!
-//! /// The `Raw` can be is set to `()` because `EpdColor` doesn't need to be
-//! /// converted to raw data for the display and isn't stored in images.
+//! /// Implement the `PixelColor` trait to mark this type as a embedded-graphics
+//! /// color and to specify the binary storage format.
```
jamwaffles, line 28: I had a go at wording this a little better:

```suggestion
//! /// Implement the `PixelColor` trait to mark this type as an embedded-graphics
//! /// color and associated binary storage format.
```

jamwaffles (approved):

core/src/pixelcolor/mod.rs
rfuest: Thanks, that's much better.

PR #767, opened 2024-09-22
Add load and store methods to RawData trait

rfuest:
This PR adds the `load` and `store` methods from #711 to the `RawData` trait. These methods allow random access to individual pixels in a byte slice and are easier to use than the previous iterator based approach of reading pixels. In a later PR the new methods will be used in more places and will, for example, make it possible to remove all bit depth dependent code from `Framebuffer`, but I didn't want to let this PR get too large.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 
 - **(breaking)** [#764](https://github.com/embedded-graphics/embedded-graphics/pull/764) Changed `ImageRaw::new` to return an error instead of truncating the image height.
 - **(breaking)** [#765](https://github.com/embedded-graphics/embedded-graphics/pull/765) Made conversion to and from `RawUx` types mandatory for all `PixelColor` implementations.
+- **(breaking)** [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Renamed `ByteOrder`, `LittleEndian`, and `BigEndian` to `DataOrder`, `LittleEndianMsb0`, and `BigEndianLsb0`.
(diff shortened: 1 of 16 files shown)


rfuest:
`check-drawing-examples` does currently fail because of a renamed type, which made `tinytga` 0.5 incompatible. I'm not sure how we have handled problems like this in the past, but I would suggest to ignore this failed CI step for the moment and later temporarily change the `tinytga` dependency for the examples to a GIT dependency to a draft PR in the `tinytga` repo.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - **(breaking)** [#764](https://github.com/embedded-graphics/embedded-graphics/pull/764) Changed `ImageRaw::new` to return an error instead of truncating the image height.
 - **(breaking)** [#765](https://github.com/embedded-graphics/embedded-graphics/pull/765) Made conversion to and from `RawUx` types mandatory for all `PixelColor` implementations.
 - [#732](https://github.com/embedded-graphics/embedded-graphics/pull/732) Added `Rgb444` to support 12bit RGB displays. Note that this type is currently stored as a 16 bit value in `ImageRaw`s and `Framebuffer`s.
+- **(breaking)** [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Renamed `ByteOrder`, `LittleEndian`, and `BigEndian` to `DataOrder`, `LittleEndianMsb0`, and `BigEndianLsb0`.
+- **(breaking)** [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Changed default data order for `ImageRaw` to `LittleEndianMsb0`.
```
jamwaffles, line 15: Can you add the old data order type name into this sentence? I think that would be useful for users migrating.

jamwaffles (approved):
Pretty cursory review but nothing caught my attention

PR #768, opened 2024-10-17
Add 8bit RGB332 support

edwardgeorge:

## PR description

Adds `Rgb332` support which I need for a project. Mostly follows the changes from the RGB444 PR.
I needed to allow `trivial_numeric_casts` in two locations in the macro due to the storage type `u8` being the same as the `PixelColour::MAX_[RGB]` associated constant type. That seemed the simplest approach but I'm open to alternatives.


rfuest (approved):
LGTM, thanks! I agree that allowing `trivial_numeric_casts` inside the macro is the best solution.

Issue #771, opened 2025-02-15
Feature request: Dashed stroke/fills


LaineZ:
Currently, the embedded graphics library seems to be not support dashed or dotted strokes and fills, similar to the `border-style: dashed` and `border-style: dotted` CSS properties. This feature would provide additional flexibility in styling elements, especially on `BinaryColor` displays.

![Image](https://github.com/user-attachments/assets/4393e3de-2210-4050-9079-b3e3c840511e)

inesvar:
Hello,

I'm interested in trying to implement this feature.
This would be my first contribution and I am currently discovering the code, I'll make an update here if I'm stuck.

rfuest:
This would be a great addition. Implementing this for all primitives could get complicated, e.g. for lines with > 1px stroke width and line joins, but we could always start with a subset of supported primitives and expand support to additional types later.

@inesvar: Thanks for trying to implement this, let us know if we can help in any way.

inesvar:
Hello again, I need some advice on how to test my future commits.

I get a compilation error when trying to run an example from the embedded-graphics/examples repo using the embedded-graphics source code instead of the crates.io package.

`eg-0.8/Cargo.toml` was changed from
```
embedded-graphics = "0.8.1"
```
to
```
# embedded-graphics = "0.8.1"
# embedded-graphics = { path = "../../embedded-graphics", version = "0.8.1"} # this doesn't work either
# embedded-graphics = { git = "https://github.com/embedded-graphics/embedded-graphics.git", tag = "embedded-graphics-v0.8.0" } # same
embedded-graphics = { git = "https://github.com/embedded-graphics/embedded-graphics.git", tag = "embedded-graphics-v0.8.1" }
```

However while the example "primitives-stroke" was compiling in the first case, I now get a compilation error:

![Image](https://github.com/user-attachments/assets/9fa9a828-a106-4655-89e0-9f259fc410bf)

```
cargo run --example primitives-stroke
....
error[E0599]: the function or associated item `new` exists for struct `SimulatorDisplay<Rgb888>`, but its trait bounds were not satisfied
   --> examples/primitives-stroke.rs:58:51
    |
58  |     let mut display = SimulatorDisplay::<Rgb888>::new(Size::new(512, 512));
    |                                                   ^^^ function or associated item cannot be called on `SimulatorDisplay<Rgb888>` due to unsatisfied trait bounds
    |
[...]

PR #772, opened 2025-02-22
Add feature dotted border for Rectangle

inesvar:

## PR description

Support for dotted/dashed borders and fills was requested in #771. This pull request is only the beginning of the implementation, starting with dotted borders for the `Rectangle` primitive.

### Developer tests

I've tested the existing commits by comparing screenshots of the regular and dotted border (with all stroke alignment styles), see [screenshots](https://github.com/inesvar/test-embedded-graphics/blob/ec6cfb16cfd1ab0ef81f9b06369c83fe1ebb0baf/README.md) (the testing code is in the linked repo).

### Testing issue

I'm not sure what kind of tests should be added for this feature, any advice?
Checking that the dotted border fits the regular border seemed relevant, are there existing tests doing this kind of stuff using screenshots ?

### Scope of the PR

I plan on continuing the implementation of #771 with dotted borders for `Line` and `Polyline`. If you think it would be best to group this in one PR, just let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 - [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Added `load` and `store` methods to `RawData` trait.
 - [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Added `MASK` constant to `RawData` trait.
 - [#768](https://github.com/embedded-graphics/embedded-graphics/pull/768) Added 8bit `Rgb332` support.
+- [#772](https://github.com/embedded-graphics/embedded-graphics/pull/772) Added `PrimitiveStyle::stroke_style` property to draw dotted borders (currently only supported for `Rectangle`).
 
 ## [0.8.1] - 2023-08-10
 
--- a/src/primitives/primitive_style.rs
+++ b/src/primitives/primitive_style.rs
@@ -42,6 +42,14 @@ where
     /// This property only applies to closed shapes (rectangle, circle, ...) and is
     /// ignored for open shapes (line, ...).
     pub stroke_alignment: StrokeAlignment,
+
+    /// Stroke style.
+    ///
+    /// The stroke style sets the border style (default is [`StrokeStyle::Solid`]).
+    ///
+    /// [`StrokeStyle::Dotted`] is only implemented for the [`Rectangle`](crate::primitives::Rectangle)
+    /// primitive, the default will be used for all other primitives.
+    pub stroke_style: StrokeStyle,
 }
 
 impl<C> PrimitiveStyle<C>
@@ -131,6 +139,7 @@ where
             stroke_color: None,
             stroke_width: 0,
             stroke_alignment: StrokeAlignment::Center,
+            stroke_style: StrokeStyle::const_default(),
         }
     }
 }
@@ -254,6 +263,16 @@ where
         self
     }
 
+    /// Sets the stroke style.
+    ///
+    /// This feature is not supported for all primitives, see [PrimitiveStyle::stroke_style]
+    /// for the complete list.
+    pub const fn stroke_style(mut self, stroke_style: StrokeStyle) -> Self {
+        self.style.stroke_style = stroke_style;
+
+        self
+    }
+
     /// Builds the primitive style.
     pub const fn build(self) -> PrimitiveStyle<C> {
         self.style
@@ -287,6 +306,24 @@ impl Default for StrokeAlignment {
     }
 }
 
+/// Stroke style.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
+#[cfg_attr(feature = "defmt", derive(::defmt::Format))]
+#[non_exhaustive]
+pub enum StrokeStyle {
+    /// Solid.
+    #[default]
+    Solid,
+    /// Dotted.
+    Dotted,
+}
+
+impl StrokeStyle {
+    const fn const_default() -> Self {
+        Self::Solid
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -301,6 +338,7 @@ mod tests {
                 stroke_color: None,
                 stroke_width: 0,
                 stroke_alignment: StrokeAlignment::Center,
+                stroke_style: StrokeStyle::Solid,
             }
         );
 
--- a/src/primitives/primitive_style.rs
+++ b/src/primitives/primitive_style.rs
@@ -126,7 +126,13 @@ where
     /// Returns the fill area.
     pub(in crate::primitives) fn fill_area<P: OffsetOutline>(&self, primitive: &P) -> P {
         // saturate offset at i32::min_value() if stroke width is to large
-        let offset = -self.inside_stroke_width().saturating_as::<i32>();
+        let offset = if self.stroke_style == StrokeStyle::Solid {
+            -self.inside_stroke_width().saturating_as::<i32>()
+        } else {
+            // do not shrink the fill area for non solid borders, because the entire fill
+            // area is visible through the gaps in the border
+            0
+        };
 
         primitive.offset(offset)
     }
--- a/src/primitives/rectangle/styled.rs
+++ b/src/primitives/rectangle/styled.rs
@@ -545,4 +545,102 @@ mod tests {
 
         assert_eq!(transparent_rect.bounding_box(), filled_rect.bounding_box(),);
     }
+
+    #[test]
+    fn dotted_border_is_inside_the_regular_border() {
+        let base = PrimitiveStyleBuilder::from(&PrimitiveStyle::with_stroke(BinaryColor::On, 5));
+
+        let rectangles = [
+            Rectangle::new(Point::new(5, 6), Size::new(40, 3)),
+            Rectangle::new(Point::new(4, 5), Size::new(40, 39)),
+        ];
+
+        for rect in &rectangles {
+            let mut regular = MockDisplay::new();
+            let mut dotted = MockDisplay::new();
+
+            rect.into_styled(base.build()).draw(&mut regular).unwrap();
+            rect.into_styled(base.stroke_style(StrokeStyle::Dotted).build())
+                .draw(&mut dotted)
+                .unwrap();
+
+            for p in regular.bounding_box().points() {
+                if dotted.get_pixel(p) == Some(BinaryColor::On) {
+                    assert_eq!(regular.get_pixel(p), Some(BinaryColor::On));
+                }
+            }
+        }
+    }
+
+    #[test]
+    fn dotted_border_dots_have_correct_shape_and_size() {
+        // The diameter of the dots should be the border width, except in the case
+        // where this causes corner dots to overlap.
+        // Rectangles are used for dots with size <= 3.
+        let base = PrimitiveStyleBuilder::new()
+            .stroke_color(BinaryColor::On)
+            .stroke_width(5)
+            .stroke_alignment(StrokeAlignment::Inside);
+
+        let rectangle1 = Rectangle::new(Point::new(5, 6), Size::new(40, 7));
+        let rectangle2 = Rectangle::new(Point::new(4, 5), Size::new(40, 39));
+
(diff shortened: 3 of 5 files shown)


rfuest:
Thanks for the PR. I didn't have time to look at the code yet, but I can at least give you some initial feedback.

---

> ### Developer tests
> 
> I've tested the existing commits by comparing screenshots of the regular and dotted border (with all stroke alignment styles), see [screenshots](https://github.com/inesvar/test-embedded-graphics/blob/771-dotted-border-for-rectangle/README.md) (the testing code is in the linked repo).

Have you seen the [`debug-tools`](https://github.com/embedded-graphics/debug-tools) repo? It provides an easy way to write interactive debug tools, which can help to find edge cases that still need work. But the repo might not have been updated to the latest version of e-g yet.

> ### Testing issue
> 
> I'm not sure what kind of tests should be added for this feature, any advice? Checking that the dotted border fits the regular border seemed relevant, are there existing tests doing this kind of stuff using screenshots ?

First I would suggest to add a few tests using [`MockDisplay`](https://docs.rs/embedded-graphics/latest/embedded_graphics/mock_display/index.html), that compare against an expected pattern inside the source code. This should work well for small(ish) border sizes.

Checking against a regular border is a good idea, if you think that 64x64 pixel is enough space you could also use `MockDisplay` for this, otherwise you would need to use `Framebuffer`. For `MockDisplay` the test could look like this:
```rust
let mut regular = MockDisplay::new();
draw_regular_border(&mut regular).unwrap();

let mut dotted = MockDisplay::new();
draw_dotted_border(&mut dotted).unwrap();

for p in regular.bounding_box().points() {
   // checks that the same pixels were drawn in both cases,
   // but ignores the color
   assert_eq!(regular.get_pixel(p).is_some(), dotted.get_pixel(p).is_some());
}
```
`Framebuffer` doesn't track which pixels have been drawn and you would need to pick distinct colors to distinguish between border, dots, fill, and background, but otherwise the test should work pretty much the same.

> ### Scope of the PR
> 
> I plan on continuing the implementation of #771 with dotted borders for `Line` and `Polyline`. If you think it would be best to group this in one PR, just let me know.

Let's keep the PRs focused on one primitive at a time to make reviewing easier.

---

I would like to change the way the dots are a aligned on the border, at least for small border width:
image

The gaps in the center are, in my opinion, more distracting than having some of the corner dots missing:
image

inesvar:
Thanks for your detailed answer.
I will get back to you in the coming week.

Any commits named "temp" are WIP and not meant to be reviewed.

inesvar:
I changed the implementation so that missing corner dots are allowed for border widths of 1px and 2px.
Here are screenshots of various rectangles with the [previous code](https://raw.githubusercontent.com/inesvar/test-embedded-graphics/f386b050925a9e19bb92c875c4abdec744dc7caa/screenshots/various_rectangles_v1.png) and the [new one](https://raw.githubusercontent.com/inesvar/test-embedded-graphics/f386b050925a9e19bb92c875c4abdec744dc7caa/screenshots/various_rectangles_v2.png) .

Let me know what you think, I'll start working on tests in the next week.

rfuest:
I've played with the code a bit and it works well in most cases. But these drawing routines unfortunately have lots of details that need to be considered. Here are some nitpicky comments: 

### Large border sizes with small rectangle sizes

If the length of a vertical or horizontal border is small enough that 3 dots don't fit anymore there is a very large gap between the border dots (right image). If only one pixel more is added the display looks much better (left image). I think we should allow the gaps to shrink a bit to make the jump between 2 and 3 dots less apparent. Having the allowed values for the gap size centered around the ideal value, with some wiggle room in the positive and negative direction, might also improve the spacing in general and not only this case.

image

### Shape of 3 pixel wide borders

Here is a comparison of border width from 1 px to 5 px:

image

In my opinion the 3 pixel wide border doesn't look good an I would never use is. The dots seem to be visually smaller than the chunky 2 pixel ones.

What do you think about using rectangles for border width <= 3 px? Like this:

image

Using rectangles instead of circles for these small sizes could also improve the drawing performance a bit.

### Stroke alignment

At the moment the stroke alignment option doesn't influence how the dots are drawn in relation to the fill. Here is how it should work (from left to right: outside, center, inside):

image

src/primitives/primitive_style.rs
```diff
@@ -42,6 +42,13 @@ where
     /// This property only applies to closed shapes (rectangle, circle, ...) and is
     /// ignored for open shapes (line, ...).
     pub stroke_alignment: StrokeAlignment,
+
+    /// Stroke style.
+    ///
+    /// The `stroke_style` sets a particular style, for instance the dotted style.
+    /// This feature is only implemented for the [`Rectangle`](crate::primitives::Rectangle)
+    /// primitive, and `stroke_style` will be ignored for all other primitives.
+    pub stroke_style: Option<StrokeStyle>,
```
rfuest, line 51: There is no need to use an `Option` here. Just add a `Solid` stroke style and make that the default.

inesvar:
Hello !

I have applied the suggested changes, I do agree that the result is more aesthetic.
That is, except for the stroke alignment, because the code already worked as described.

I didn't have time to update the PR yet, but I did the development and I drew the results in this commit :
https://github.com/inesvar/test-embedded-graphics/commit/5333f34492b06b03b02ccb0050054f720772d615.

screenshots_v5 show the final results, and screenshots_v6 show that the dotted border is correctly aligned with the fill.

You can play with it yourself as the Cargo.toml has a dependency on my git dev branch.

rfuest:
> (That is, except for the stroke alignment, because the code already worked as described).
> ...
> screenshots_v5 show the final results, and screenshots_v6 show that the dotted border is correctly aligned with the fill.

I didn't have time to look at the changes in your latest revision yet, but I think I have an idea why our stroke alignment results were different. Your test draws the fill and stroke separately using two different `PrimitiveStyle`s. But in my test I was using a single `PrimitiveStyle` with both stroke and fill enabled.

inesvar:
Oh I see how it works now. I will update `fill_area` accordingly.

On another subject, I realized that it was a bad idea to use central symmetry in `draw_dotted_rectangle_border_with_dotted_corners` as it will clash with the dotted fill. I will revert to axial symmetry.

I'll be waiting for your appraisal of the code on the dev branch https://github.com/inesvar/embedded-graphics/tree/develop before updating the PR, as I expect you will do further testing before the review. The dev branch should be updated this week-end.

src/primitives/primitive_style.rs
```diff
@@ -118,7 +126,11 @@ where
     /// Returns the fill area.
     pub(in crate::primitives) fn fill_area<P: OffsetOutline>(&self, primitive: &P) -> P {
         // saturate offset at i32::min_value() if stroke width is to large
-        let offset = -self.inside_stroke_width().saturating_as::<i32>();
+        let offset = if self.stroke_style == StrokeStyle::Solid {
+            -self.inside_stroke_width().saturating_as::<i32>()
+        } else {
+            0
```
rfuest, line 132:
```suggestion
            // do not shrink the fill area for non solid borders, because the entire fill
            // area is visible through the gaps in the border            
            0
```

src/primitives/primitive_style.rs
```diff
@@ -287,6 +310,23 @@ impl Default for StrokeAlignment {
...
+/// Stroke style.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[cfg_attr(feature = "defmt", derive(::defmt::Format))]
+#[non_exhaustive]
+pub enum StrokeStyle {
+    /// Solid.
+    Solid,
+    /// Dotted.
+    Dotted,
+}
```
rfuest, line 322:
```suggestion
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
#[cfg_attr(feature = "defmt", derive(::defmt::Format))]
#[non_exhaustive]
pub enum StrokeStyle {
    /// Solid.
    #[default]
    Solid,
    /// Dotted.
    Dotted,
}
```

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Added `load` and `store` methods to `RawData` trait.
 - [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Added `MASK` constant to `RawData` trait.
 - [#768](https://github.com/embedded-graphics/embedded-graphics/pull/768) Added 8bit `Rgb332` support.
+- [#772](https://github.com/embedded-graphics/embedded-graphics/pull/772) Added feature dotted border (through the `stroke_style` helper), implemented for `Rectangle`.
```
rfuest, line 28: Why do you call `stroke_style` a "helper"?
```suggestion
- [#772](https://github.com/embedded-graphics/embedded-graphics/pull/772) Added `PrimitiveStyle::stroke_style` property to draw dotted borders (currently only supported for `Rectangle`).
```

core/src/geometry/point.rs
```diff
@@ -399,6 +399,16 @@ impl Index<usize> for Point {
...
+impl IndexMut<usize> for Point {
+    fn index_mut(&mut self, idx: usize) -> &mut i32 {
+        match idx {
+            0 => &mut self.x,
+            1 => &mut self.y,
+            _ => panic!("index out of bounds: the len is 2 but the index is {}", idx),
+        }
+    }
+}
+
```
rfuest, line 411: If we don't need this (see other comment) for this PR I would like to remove it. It might be a useful feature, but if it is implemented for `Point` it should also be implemented for `Size` and I would like to keep this PR focused to the dotted border.

rfuest (changes requested):
The stroke alignment works great now! I couldn't find any more major issues and the PR should almost be ready to be merged.

src/primitives/rectangle/styled.rs
```diff
@@ -69,6 +70,95 @@ impl<C: PixelColor> StyledPixels<PrimitiveStyle<C>> for Rectangle {
...
+            translation[idx] = Into::<i32>::into((dot_offset * Real::from(dot_idx)).round());
+            draw_dot_translated_by(top_left_dot, &translation, style, target)?;
+
+            translation[1 - idx] = border_size[1 - idx] as i32;
+            draw_dot_translated_by(top_left_dot, &translation, style, target)?;
+        }
+    }
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
inesvar: Your suggestion is great, it's applied now (I'm pushing soon).

I also made the following changes that you might have an opinion on :
- adding a second helper function for the second dotted border drawing function
- passed `top_left` by reference in both functions (the Circle/Rectangle dot is now created inside the drawing function)
- updated the second function comment to mention it's using Rectangle

core/src/geometry/point.rs
inesvar: It's removed.

inesvar:
`just build` and the CI currently fail, namely because this link https://gitlab.freedesktop.org/xorg/font/misc-misc/-/tree/master returns 503. 
Otherwise the doc, tests and fmt are fine.

rfuest:
> `just build` and the CI currently fail, namely because this link https://gitlab.freedesktop.org/xorg/font/misc-misc/-/tree/master returns 503. Otherwise the doc, tests and fmt are fine.

Don't worry about the broken link and the CI test that fails because of a MSRV issue. I'll look into these errors after this PR is merged.

But this one is a genuine problem with a divide by zero: https://github.com/embedded-graphics/embedded-graphics/actions/runs/13887811895/job/38859234599?pr=772#step:7:586 It seems to only happen when the `fixed_point` feature is enabled:
```shell
# works
cargo test -p embedded-graphics

# broken
cargo test -p embedded-graphics --features=fixed_point
```

inesvar:
It's fixed.

On another note, I noticed the `round` function in geometry/real.rs has a #[allow(unused)] flag. I'm using this function, should I remove the flag?

rfuest (approved):
> On another note, I noticed the round function in geometry/real.rs has a #[allow(unused)] flag. I'm using this function, should I remove the flag?

That's a good idea. The rest of the PR looks good now and I'll merge it after you've made that change.

inesvar:
It's done :+1:

Next up, if you're amenable I would like to make a PR to lower the number of clippy warnings (there are 28 currently).

That way I can take a small break before implementing the dashed border for the Rectangle, which seems like the natural next step. Or did you have another idea in mind?

Issue #492, opened 2020-11-19
Investigate inlining images in docs


jamwaffles:
An interesting solution was presented in [this tweet](https://twitter.com/m_ou_se/status/1329192390536597504):

![image](https://user-images.githubusercontent.com/800791/99647881-4647f700-2a4a-11eb-9b29-895b50872078.png)

I wonder if it would be possible to generate, or at least read images out of a folder, and inline them into the docs.

This would help for examples and would fix the stale link issues going forward. It would also be very helpful to describe some of the algorithms in use.

None of this may be feasible, but I thought I'd raise it as an investigation topic.

rfuest:
> I wonder if it would be possible to generate, or at least read images out of a folder, and inline them into the docs.

Reading the images out of a folder shouldn't be to hard. We could convert the png files to base64 and use a simple `include_str` in the docs to include them. This wouldn't require any additional build dependencies, if we add the base64 encoded versions to the repo.

Automatically generating images from code in the doc comments would be really nice to have, but I guess it would be a lot harder to implement.

I've found this post which describes a way to use a build script to include images, which might also be worth investigating: https://mrh.io/cargo-build-images-in-rust-docs/

> It would also be very helpful to describe some of the algorithms in use.

Do you mean that in general or by adding images?

jamwaffles:
> We could convert the png files to base64 and use a simple include_str in the docs to include them.

That's what I was thinking too.

> I've found this post which describes a way to use a build script to include images, which might also be worth investigating

I wonder if docs.rs supports that though. I guess it should, if the assets are copied to the `target/doc/` folder.

> Do you mean that in general or by adding images?

By adding images :)

rfuest:
> I wonder if docs.rs supports that though. I guess it should, if the assets are copied to the `target/doc/` folder.

I did some a quick research before posting the idea and I think it should. The docs.rs provides some information how to test doc builds locally, which would be useful for trying this out.

> By adding images :)

I assume this wouldn't be screenshots in most cases, but something like in https://github.com/embedded-graphics/embedded-graphics/issues/182 or https://github.com/embedded-graphics/embedded-graphics/issues/225. An inline SVG file might also be a possible solution for these kind of graphics.

jamwaffles:
> The docs.rs provides some information how to test doc builds locally, which would be useful for trying this out.

It's surprisingly simple, so props to the docs.rs team for making it easy!

[...]

PR #776, opened 2025-04-13
Add support for `StrokeStyle::Dotted` for `Line`

inesvar:

## PR description

This PR implements support for dotted lines following the issue #771. It uses 2 different strategies based on the width of the line :
- for thin lines, the regular line iterator `ThickPoints` is used but part of the pixels are skipped
- for thicker lines, dots are drawn regularly along the line using primitive `Circle` and new iterator `DottedLinePoints`

Thin lines have width below 3, I tried to apply the thin line strategy for width 3 but I didn't manage (more details in the comment from the 2nd commit).

The new iterator `DottedLinePoints` computes relevant points (from a bresenham line) to draw a dotted line with the desired number of points. In the previous PR for the rectangle dotted border, I used floats (`Real` type) to determine the dot positions. I realized that I could use a variant of the bresenham algorithm to do this without using floats. The result should be exactly the same (it's tested in `dotted_line_dots_are_correct`). 

Eventually, the dotted rectangle border could also rely on `DottedLinePoints` instead of using floats.

## Testing the code

I used the debug-tools repo, attached is a patch that updates the Cargo.toml (the extension must be changed to .patch).

[debug-tools-dotted-line.patch.txt](https://github.com/user-attachments/files/19727098/debug-tools-dotted-line.patch.txt)

## Changelog ?

I didn't update it (only a public comment was changed in `PrimitiveStyle`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display 
 - [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Added `load` and `store` methods to `RawData` trait.
 - [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Added `MASK` constant to `RawData` trait.
 - [#768](https://github.com/embedded-graphics/embedded-graphics/pull/768) Added 8bit `Rgb332` support.
-- [#772](https://github.com/embedded-graphics/embedded-graphics/pull/772) Added `PrimitiveStyle::stroke_style` property to draw dotted borders (currently only supported for `Rectangle`).
+- [#772](https://github.com/embedded-graphics/embedded-graphics/pull/772), [#776](https://github.com/embedded-graphics/embedded-graphics/pull/776) Added `PrimitiveStyle::stroke_style` property to draw dotted borders (currently only supported for `Rectangle` and `Line`).
 
 ## [0.8.1] - 2023-08-10
 
--- a/src/primitives/line/dotted_bresenham.rs
+++ b/src/primitives/line/dotted_bresenham.rs
@@ -4,18 +4,18 @@
 use super::{bresenham::MajorMinor, Line, Points};
 use crate::geometry::Point;
 
-/// Dotted bresenham parameters.
+/// Bresenham algorithm for dotted lines.
 ///
 /// [`super::bresenham::BresenhamParameters`] describes a major and
 /// a minor step, classically vectors along opposing axes.
 ///
-/// [`DottedBresenhamParameters`] describes a scalar major step and a
+/// [`DottedBresenham`] describes a scalar major step and a
 /// scalar minor step. It is used to draw a dotted line with the
-/// desired number of dots by selecting the appropriate points in a
+/// desired number of dots by selecting the appropriate indices in a
 /// bresenham line.
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 #[cfg_attr(feature = "defmt", derive(::defmt::Format))]
-struct DottedBresenhamParameters {
+struct DottedBresenham {
     /// Error threshold.
     ///
     /// If the accumulated error exceeds the threshold a minor move is made.
@@ -26,10 +26,18 @@ struct DottedBresenhamParameters {
 
     /// Change in index for major and minor steps.
     index_step: MajorMinor<usize>,
+
+    /// Current index increment.
+    ///
+    /// It is used to retrieve the next dot from a [`Points`] iterator.
+    index_nth: Option<usize>,
+
+    /// Error accumulator.
+    error: i32,
 }
 
-impl DottedBresenhamParameters {
-    /// Create a new bresenham parameters object.
+impl DottedBresenham {
+    /// Create a new bresenham object.
     fn new(line: &Line, nb_dots_desired: i32) -> Self {
         let delta = line.delta().abs();
         let line_max_index = delta.x.max(delta.y);
@@ -49,6 +57,8 @@ impl DottedBresenhamParameters {
             error_threshold,
             error_step: MajorMinor::new(remainder * 2, error_threshold * 2),
             index_step,
+            index_nth: Some(0),
+            error: 0,
         }
     }
 
@@ -56,10 +66,10 @@ impl DottedBresenhamParameters {
     ///
     /// If the error threshold is reached the error is reduced by a minor step and
     /// `true` is returned.
-    fn increase_error(&self, error: &mut i32) -> bool {
-        *error += self.error_step.major;
-        if *error >= self.error_threshold {
-            *error -= self.error_step.minor;
+    fn increase_error(&mut self) -> bool {
+        self.error += self.error_step.major;
+        if self.error >= self.error_threshold {
+            self.error -= self.error_step.minor;
 
             true
         } else {
@@ -68,38 +78,19 @@ impl DottedBresenhamParameters {
     }
 }
 
-/// Bresenham algorithm for dotted lines.
-#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
-#[cfg_attr(feature = "defmt", derive(::defmt::Format))]
-struct DottedBresenham {
-    /// Current index increment.
-    ///
-    /// It is used to retrieve the next dot from a [`Points`] iterator.
-    index_nth: Option<usize>,
-
-    /// Error accumulator.
-    error: i32,
-}
-
-impl DottedBresenham {
-    /// Create a new bresenham object.
-    const fn new() -> Self {
-        Self {
-            index_nth: Some(0),
-            error: 0,
-        }
-    }
+impl Iterator for DottedBresenham {
+    type Item = usize;
 
     /// Return the increment to the next point on the line.
     /// This iterator is infinite, except if `parameters.index_step.major = 0`.
-    fn next(&mut self, parameters: &DottedBresenhamParameters) -> Option<usize> {
+    fn next(&mut self) -> Option<usize> {
         let ret = self.index_nth;
 
         if self.index_nth.is_some() {
-            let mut increment = parameters.index_step.major;
+            let mut increment = self.index_step.major;
 
-            if parameters.increase_error(&mut self.error) {
-                increment += parameters.index_step.minor;
+            if self.increase_error() {
+                increment += self.index_step.minor;
             }
 
             self.index_nth = if increment > 0 {
@@ -118,7 +109,6 @@ impl DottedBresenham {
 #[cfg_attr(feature = "defmt", derive(::defmt::Format))]
 pub(super) struct DottedLinePoints {
     points: Points,
-    index_parameters: DottedBresenhamParameters,
     index_bresenham: DottedBresenham,
 }
 
@@ -127,12 +117,10 @@ impl DottedLinePoints {
     /// taking into account the desired number of dots.
     pub(super) fn new(line: &Line, nb_dots_desired: i32) -> Self {
         let points = Points::new(line);
-        let index_parameters = DottedBresenhamParameters::new(line, nb_dots_desired);
-        let index_bresenham = DottedBresenham::new();
+        let index_bresenham = DottedBresenham::new(line, nb_dots_desired);
 
         Self {
             points,
-            index_parameters,
             index_bresenham,
         }
     }
@@ -142,7 +130,6 @@ impl Iterator for DottedLinePoints {
     type Item = Point;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.points
-            .nth(self.index_bresenham.next(&self.index_parameters)?)
+        self.points.nth(self.index_bresenham.next()?)
     }
 }
--- a/src/primitives/line/dotted_bresenham.rs
+++ b/src/primitives/line/dotted_bresenham.rs
@@ -1,0 +1,148 @@
+//! A variant of the bresenham algorithm that is used to select
+//! points from a bresenham line in order to draw a dotted line.
+
+use super::{bresenham::MajorMinor, Line, Points};
+use crate::geometry::Point;
+
+/// Dotted bresenham parameters.
+///
+/// [`super::bresenham::BresenhamParameters`] describes a major and
+/// a minor step, classically vectors along opposing axes.
+///
+/// [`DottedBresenhamParameters`] describes a scalar major step and a
+/// scalar minor step. It is used to draw a dotted line with the
+/// desired number of dots by selecting the appropriate points in a
+/// bresenham line.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[cfg_attr(feature = "defmt", derive(::defmt::Format))]
+struct DottedBresenhamParameters {
+    /// Error threshold.
+    ///
+    /// If the accumulated error exceeds the threshold a minor move is made.
+    error_threshold: i32,
+
+    /// Change in error for major and minor steps.
+    error_step: MajorMinor<i32>,
+
+    /// Change in index for major and minor steps.
+    index_step: MajorMinor<usize>,
+}
+
+impl DottedBresenhamParameters {
+    /// Create a new bresenham parameters object.
+    fn new(line: &Line, nb_dots_desired: i32) -> Self {
+        let delta = line.delta().abs();
+        let line_max_index = delta.x.max(delta.y);
+        let nb_pixels_in_line = line_max_index + 1;
+
+        // Enforce the use of at least 2 dots to prevent division by 0 when the line is reduced to a point.
+        // `clamp` can't be used here (`nb_pixels_in_line` is less than 2 when the line is reduced to a point).
+        let nb_dots = nb_dots_desired.min(nb_pixels_in_line).max(2);
+
+        let integer_quotient = line_max_index / (nb_dots - 1);
+        let remainder = line_max_index - integer_quotient * (nb_dots - 1);
+
+        let index_step = MajorMinor::new(integer_quotient as usize, 1);
+        let error_threshold = nb_dots - 1;
+
+        Self {
+            error_threshold,
+            error_step: MajorMinor::new(remainder * 2, error_threshold * 2),
+            index_step,
+        }
+    }
+
+    /// Increases the error by a major step.
+    ///
+    /// If the error threshold is reached the error is reduced by a minor step and
+    /// `true` is returned.
+    fn increase_error(&self, error: &mut i32) -> bool {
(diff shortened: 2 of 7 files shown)


rfuest:
Thanks for sticking with this feature. Most of the lines already look very good. I'll take a closer look at the code when I have some more time.

The pixel skipping approach does work very well for 1px lines, but I don't think the result for 2 pixels wide lines looks as good as simple `Rectangle` dots:
image

I would suggest to change the drawing code to this:
| width | drawing operation |
|---|---|
| 1px | Pixel skipping |
| 2px - 3px | `Rectangle` dots using `DottedLinePoints` |
| > 3px | `Circle` dots using `DottedLinePoints` |

inesvar:
Hello ! 

I'm planning on working on this feature/project over the next year :+1:
My free time is limited though, so progress will come in bursts.

I will push a new version of the code in the coming week. Applying your suggestions will require a bit of tweaking (2x2 dots can be in contact in the current code, cf image below).

2x2_dots_in_contact

Also I just realized I missed an issue. Opposite lines (with swapped endpoints) are actually different, and I didn't take this into account when drawing dots (see below).

well_aligned_dots
badly_aligned_dots

rfuest:
> I'm planning on working on this feature/project over the next year 👍
> My free time is limited though, so progress will come in bursts.

Sounds great.

> Also I just realized I missed an issue. Opposite lines (with swapped endpoints) are actually different, and I didn't take this into account when drawing dots (see below).

This is an unfortunate side effect of having to fit everything into the pixel grid. This ancient issue explains some of my reasons for implementing it this way if you are interested: https://github.com/embedded-graphics/embedded-graphics/issues/182

The TLDR is that lines with even stroke widths will always have one more pixel on the left side if you are looking from the start point towards the end point.

inesvar:
It was a very interesting read thank you !

It would be nice to include these images in the documentation, like requested in issue #492. I will try to do this in the coming months.

rfuest:
Sorry for the late reply, I wasn't able to look at the code or try it yet. Is the PR ready for a final review, because it's still marked as a draft?

> I see the check-docs task failed. I don't know what the issue is yet, when I run the same command on my machine there's no problem.

Are you using the latest Rust release? I was able to reproduce the error locally with Rust 1.87.0.

rfuest:
> Fixing the problematic links wouldn't take a lot of time either but I think it makes more sense not to keep unused code.

You were right, it didn't take long. I had to take a look anyway and removed them in #780. I'll look into the other CI problem next to finally have a working CI system again.

inesvar:
Great ! I rebased and undrafted both PRs.

inesvar:
Hi,

I will have a look at the problem with `isqrt` (the CI fail) in the coming days, and also slightly change the 1px wide dotted line implementation.

After yesterday's changes to the dotted rectangle implementation (support for the case when the dotted area is flattened to a 1px line), I think it would be a good idea to slightly change the implementation of the 1px wide dotted line so that the first pixel is always drawn (instead of the last pixel).

This way, the flattened dotted rectangle will be compatible with the 2 first dotted lines in clockwise order.

inesvar:
The CI failed in the previous version because `isqrt` is unstable in Rust 1.81.

I added the dependency `integer-sqrt` to avoid this issue, another solution could be to re-implement integer sqrt in e-g.

src/primitives/line/dotted_bresenham.rs
```diff
@@ -0,0 +1,246 @@
...
+impl DottedBresenhamParameters {
+    /// Create a new bresenham parameters object.
+    pub fn new(line: &Line, nb_dots_desired: i32) -> Self {
+        let delta = line.delta().abs();
+        let line_max_index = delta.x.max(delta.y);
+
+        // Enforce the use of at most `line_max_index + 1` dots.
+        let mut nb_dots = min(nb_dots_desired, line_max_index + 1);
+        // Then enforce the use of at least 2 dots to prevent division by 0 when the line is reduced to a point.
+        nb_dots = max(2, nb_dots);
```
rfuest, line 41: This could use `clamp` instead of separate `min` and `max` calls:
```suggestion
        // Enforce the use of at least 2 dots to prevent division by 0 when the line is reduced to a point and enforce the use of at most `line_max_index + 1` dots.
        let mut nb_dots = nb_dots_desired.clamp(2, line_max_index + 1);
```

src/primitives/line/dotted_bresenham.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub fn increase_error(&self, error: &mut i32) -> bool {
+        *error += self.error_step.major;
+        if *error >= self.error_threshold {
+            *error -= self.error_step.minor;
+
+            true
+        } else {
+            false
+        }
+    }
```
rfuest, line 69: That's some not so great code I wrote 5 years ago and using that confusing `&mut i32` parameter isn't necessary here.

I don't think there is a reason to keep the split between `DottedBresenhamParameters` and `DottedBresenham`. I would suggest to merge them, or, if you prefer, even merge everything into `DottedLinePoints`.

src/primitives/line/dotted_bresenham.rs
```diff
@@ -0,0 +1,246 @@
...
+        Self::with_initial_error(initial_index, 0)
+    }
+
+    /// Create a new extended bresenham object with the initial error.
+    pub const fn with_initial_error(initial_index: usize, initial_error: i32) -> Self {
+        Self {
+            index_nth: Some(initial_index),
+            error: initial_error,
+        }
+    }
```
rfuest, line 97: There's no need to keep the `with_initial_error` constructor if it isn't used:

```suggestion
    /// Create a new bresenham object.
    pub const fn new(initial_index: usize) -> Self {
        Self {
            index_nth: Some(initial_index),
            error: 0,
        }
    }
```

src/primitives/line/styled.rs
```diff
@@ -52,6 +59,95 @@ impl<C: PixelColor> StyledPixels<PrimitiveStyle<C>> for Line {
...
+    dot: &TS,
+    dotted_line_points: &DottedLinePoints,
+    dot_style: &PrimitiveStyle<C>,
+    target: &mut D,
+) -> Result<(), D::Error>
+where
+    TS: Transform + StyledDrawable<PrimitiveStyle<C>, Output = ()>,
+    C: PixelColor,
+    D: DrawTarget<Color = TS::Color>,
+{
```
rfuest, line 72: Not very important, but this could have one less generic parameter:

```suggestion
fn draw_dotted_line<TS, D>(
    dot: &TS,
    dotted_line_points: &DottedLinePoints,
    dot_style: &PrimitiveStyle<D::Color>,
    target: &mut D,
) -> Result<(), D::Error>
where
    TS: Transform + StyledDrawable<PrimitiveStyle<D::Color>, Color = D::Color, Output = ()>,
    D: DrawTarget,
{
```

src/primitives/line/styled.rs
```diff
@@ -64,7 +160,61 @@ impl<C: PixelColor> StyledDrawable<PrimitiveStyle<C>> for Line {
     where
         D: DrawTarget<Color = C>,
     {
-        target.draw_iter(StyledPixelsIterator::new(self, style))
+        let dot_size = style.stroke_width as i32;
+        if dot_size == 0 {
```
rfuest, line 164: The cast from `u32` to `i32` could theoretically generate a negative value for `dot_size`, so this would be a bit safer:
```suggestion
        if dot_size <= 0 {
```

src/primitives/line/thick_points.rs
```diff
@@ -221,6 +221,11 @@ impl ThickPoints {
             iter: ParallelsIterator::new(line, thickness, StrokeOffset::None),
         }
     }
+
+    pub fn has_more_lines_than_expected(line: &Line, thickness: i32) -> bool {
```
rfuest, line 225: This method needs a comment.

rfuest (changes requested):
The output looks good and I wasn't able to break it in my test.

>I added the dependency integer-sqrt to avoid this issue, another solution could be to re-implement integer sqrt in e-g.

I'm OK with adding a new dependency for this, but we should open an issue to make sure to remember to migrate to the builtin function when the MSRV is updated to a version that includes it.

> Changelog? I didn't update it (only a public comment was changed in PrimitiveStyle).

The changelog entry does also mention that only `Rectangle` is supported and needs to be updated like the doc comment.

inesvar:
Hello :)
The requested changes were applied.
Additional changes: restricted visibilities in line::dotted_bresenham, corrected description of unit test `dotted_line_by_skipping_pixels` in line::styled.

rfuest:
> /!\ I was just reading the doc on the `Dimensions` trait and I realized that the bounding box implementation of a styled (dotted) line needs to be updated.
> 
> Unlike the dotted rectangle border that is included in the solid rectangle border, the dotted line has a bigger bounding box than a solid line (because of the endpoint dots). I will look into that this week.

Another good catch!

I'm thinking about relaxing the `Dimensions` trait a bit in the future, because for some styled primitives it's hard to determine a pixel accurate bounding box without actually drawing it. The best example for this are`Arc` and `Sector` which have two different draft PRs to fix the bounding box, but neither got close to being finished. In my opinion a good approximation that is computationally inexpensive and is guaranteed to never be too small is a better way to handle the bounding boxes of styles primitives.

For the thick line an appropriate approximation could be `Rectangle::with_corners(line.start, line.end).offset(stroke_width.div_ceil(2) as i32)`. I haven't tested it and it might be off a bit.

inesvar:
Hello :)

`Rectangle::with_corners(min, max).offset((style.stroke_width / 2) as i32)` is a little more restrictive and works well*. This bounding box is even included in `display.affected_area().offset(1)` :)

*That is, EXCEPT for the 2px line... The ending dot isn't included in the bounding box.


Yellow dots: start (100,70) & end(126,120)

I was let down by this, but this might be an opportunity to improve the function that translates the dots so that they fit the line. It turns out my reasons for choosing the current implementation over the alternative weren't that good.

<details>
<summary>Initial idea</summary>

I initially didn't take into account the number of bresenham lines. Just separating the lines into nearly horizontal/vertical and the others (depending on the distance between the first and the second pixel of the line). This simplifies the code (and makes the bounding box work perfectly).

However, there was the following issue: in the middle picture below, the positioning of the red dots on the orange line doesn't look ideal. 

4_bresenham_lines_to_top
5_bresenham_lines_dots_to_top
5_bresenham_lines_dots_to_top_left

Yellow dots: start (100,70) & end(124,120) / end(125,120) / end(126,120)

The number of bresenham lines is 5, but the line is still considered "nearly vertical" so the dot positioning is the same as in the left picture (offset to the top), while the dot positioning in the right picture would have been a better fit (offset to the top and to the left).

So there are (few) line orientations where the number of bresenham lines increased, but the line is still nearly horizontal/vertical. In these cases the dot positioning is off.

</details>

<details>
<summary>Current implementation</summary>

The current implementation computes the number of bresenham lines to fix the previous issue:

4_bresenham_lines
5_bresenham_lines
5_bresenham_lines_second

Yellow dots: start (100,70) & end(124,120) / end(125,120) / end(126,120)
 
 While the position of the last dot moves rather unpredictably, the dots are aligned well.
 However, the 2px line is weird. This is not really surprising given that the number of bresenham lines is always 2, so a 45° line would have the same dot translation as the closest horizontal/vertical line.
 



50OK

Yellow dots: start (100,70) & end(125,120) / end(126,120) / end(149,120) / end(150,120)
 
 This issue wasn't present in the initial implementation.
 
 25OK
26OK
49OK
50OK

Yellow dots: start (100,70) & end(125,120) / end(126,120) / end(149,120) / end(150,120)
 
 </details>
 
 ### Summing up

To sum up, I thought the current implementation looked better overall, but actually it has a blind spot with the 2px line. It has the disadvantage of being more complex codewise and is more unpredictable.

To get the best look overall, the current implementation could be used for thick lines and the initial idea for thin lines. Alternatively, if you think the alignment looks alright, the initial idea could be used for all the lines.

I will push quickly two commits that use the initial idea for all the lines and implement the bounding box. The new debug-tools patch is here:
[dotted_line_debug_tools.patch.txt](https://github.com/user-attachments/files/21330315/dotted_line_debug_tools.patch.txt)

rfuest:
> Alternatively, if you think the alignment looks alright, the initial idea could be used for all the lines (I like this option better because it's simpler).

In my opinion, horizontal and vertical lines will be the most commonly used dotted lines and we should focus on get them look as good as possible. Without having subpixel precision and antialiasing there will always be some jankiness for arbitrary angles. I'm OK with the alignment and keeping the code simpler sounds like a good idea for future maintenance.

inesvar:
Great ! Then you can take a look at the last 2 commits, they are ready for review :)

src/primitives/line/dotted_bresenham.rs
```diff
@@ -0,0 +1,223 @@
...
+
+            self.index_nth = if increment > 0 {
+                Some(increment - 1)
+            } else {
+                None
+            };
+        }
+
+        ret
+    }
```
rfuest, line 100: The comment says it's an iterator and the signature also matches `Iterator::next`, so why doesn't `DottedBresenham` impl `Iterator`?

src/primitives/line/styled.rs
```diff
@@ -52,6 +59,104 @@ impl<C: PixelColor> StyledPixels<PrimitiveStyle<C>> for Line {
...
+fn extend_line_by_one_unit(line: &Line) -> Line {
+    let mut points = Points::new(line);
+
+    if let Some(second) = points.nth(1) {
+        Line::new(line.start, line.end + second - line.start)
+    } else {
+        // If there is only one point in the iterator, the line is expected to be reduced to a point.
+        // In that case we don't extend the line.
+        *line
+    }
```
rfuest, line 157: It took me a moment to realize that the if statement is the other way around than in the function above. I would suggest to use the same structure for both functions:

```suggestion
    let Some(second) = points.nth(1) else {
        // If there is only one point in the iterator, the line is expected to be reduced to a point.
        // In that case we don't extend the line.
        return *line;
    };
    
    Line::new(line.start, line.end + second - line.start)
```

src/primitives/line/styled.rs
```diff
@@ -64,26 +169,87 @@ impl<C: PixelColor> StyledDrawable<PrimitiveStyle<C>> for Line {
...
+            let line = if dot_size % 2 == 0 {
+                // When drawing a dotted rectangle border, the distance between the endpoint dots
+                // is longer by one pixel when the dot size is even.
+                // So that the dotted rectangle border matches 4 lines drawn in clockwise order,
+                // the line is extended by one pixel when the dot size is even.
+                extend_line_by_one_unit(self)
+            } else {
+                *self
+            };
+            // Draw circles along the line.
```
rfuest, line 191:
```suggestion

            // Draw dots along the line.
```

src/primitives/line/styled.rs
```diff
@@ -64,26 +169,87 @@ impl<C: PixelColor> StyledDrawable<PrimitiveStyle<C>> for Line {
...
+            let mut length = line.delta().length_squared().integer_sqrt();
+            // The gaps between dots ideally have the same size as the dots
+            // If `dot_size <= 3`, only positive error is allowed,
+            // otherwise both positive and negative error are allowed.
+            if dot_size > 3 {
+                length += dot_size;
+            }
+            // The 2 endpoint dots take half the space of a regular dot.
+            let nb_dots_desired = length / (2 * dot_size) + 1;
```
rfuest, line 200: Wouldn't it be better to move this into `DottedLinePoints::new` to have all calculations in one place?

src/primitives/line/styled.rs
```diff
@@ -64,26 +169,87 @@ impl<C: PixelColor> StyledDrawable<PrimitiveStyle<C>> for Line {
...
+            let dotted_line_points =
+                DottedLinePoints::new(&line.translate(-line.start), nb_dots_desired);
+            let dot_style = PrimitiveStyle::with_fill(stroke_color);
+            // Improve the positioning of the dots.
+            let start = start_dot_offset(self, dot_size);
```
rfuest, line 205: If the code above is moved into `DottedLinePoints::new`, this should also be moved.

src/primitives/line/styled.rs
```diff
@@ -64,26 +169,87 @@ impl<C: PixelColor> StyledDrawable<PrimitiveStyle<C>> for Line {
...
+        }
     }
 }
 
 impl<C: PixelColor> StyledDimensions<PrimitiveStyle<C>> for Line {
     fn styled_bounding_box(&self, style: &PrimitiveStyle<C>) -> Rectangle {
-        let (l, r) = self.extents(style.stroke_width, StrokeOffset::None);
+        if style.stroke_style == StrokeStyle::Dotted {
+            let min = self.start.component_min(self.end);
+            let max = self.start.component_max(self.end);
```
rfuest, line 234: This is redundant, because `Rectangle::with_corners` also determines the min and max coordinates.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 - [#767](https://github.com/embedded-graphics/embedded-graphics/pull/767) Added `MASK` constant to `RawData` trait.
 - [#768](https://github.com/embedded-graphics/embedded-graphics/pull/768) Added 8bit `Rgb332` support.
 - [#772](https://github.com/embedded-graphics/embedded-graphics/pull/772) Added `PrimitiveStyle::stroke_style` property to draw dotted borders (currently only supported for `Rectangle`).
+- [#776](https://github.com/embedded-graphics/embedded-graphics/pull/776) Added dotted border support for primitive `Line`.
```
rfuest, line 30: Let's merge these entries.
```suggestion
- [#772](https://github.com/embedded-graphics/embedded-graphics/pull/772), [#776](https://github.com/embedded-graphics/embedded-graphics/pull/776) Added `PrimitiveStyle::stroke_style` property to draw dotted borders (currently only supported for `Rectangle` and `Line`).
```

rfuest (changes requested):
A few minor nits, but the majority of the code looks good.

inesvar:
Hello,

Thanks for your thoughtful review !
I applied most of the comments. However I'm still working on moving the computation of the number of dots into `DottedLinePoints::new`. I think it's a good idea, but it requires updating some unit tests to use `DottedBresenham` instead of `DottedLinePoints`.

src/primitives/line/styled.rs
inesvar: This was moved to `DottedLinePoints::with_dot_size`

src/primitives/line/styled.rs
inesvar: I didn't move the `start_dot_offset` call to the dotted_bresenham module because I think the "making dots fit in the line" and "improving compatibility with the dotted rectangle" should be grouped.

Let me know what you think of the most recent code, an alternative is to move both into the dotted_bresenham module.

PR #777, opened 2025-04-26
Correct dot spacing for thin rectangle borders

inesvar:

- [X] Add a `CHANGELOG.md` entry in the **Unreleased** section under the appropriate heading (**Added**, **Fixed**, etc) if your changes affect the **public API**

## PR description

This fixes a dotted rectangle border issue introduced in #772.

The `dot_positions_in_clockwise_order` function was originally intended to compute both dot and gap positions. However, it actually computed only dot positions, using `length / (2 * dot_size)` instead of `length / dot_size`, causing excessive spacing between dots for thin rectangle borders.

This probably happened because I made a copy-paste of the similar function `dot_positions_with_dotted_corners` during the last stage of the PR. Unfortunately I didn't notice the functionality change before.

This commit:
- fixes the computation
- renames the problematic function to `unit_positions_in_clockwise_order`
- moves the 2 functions closer together for comparison
- adds a unit test using `MockDisplay`

## Other remarks

After this PR, `dot_positions_with_dotted_corners`  and `dot_positions_in_clockwise_order` become similar with respect to the division by 0 risk. It's not possible with the input arguments given by the current code (because `dot_size` is chosen to be smaller than `length` in `draw_styled`), but it could happen with different input arguments.

Thus I think both functions should handle the eventuality in the same way, which is not the case here.
The better option seems to be adding a check in `*dotted_corners` and returning an empty iterator in the problematic case.

Should I add a commit for this in this PR?

### DIFF
--- a/src/primitives/rectangle/styled.rs
+++ b/src/primitives/rectangle/styled.rs
@@ -185,7 +185,18 @@ where
     let mut corner_dot = Rectangle::new(*top_left, Size::new_equal(dot_size));
     let mut unit_is_dot = true;
 
-    for (side_idx, side) in border_sides.iter().enumerate() {
+    let nb_sides_to_draw = match (
+        border_sides[0] == Point::zero(),
+        border_sides[1] == Point::zero(),
+    ) {
+        (false, false) => 4,
+        (true, true) => 0,
+        // In case one pair of opposite borders overlap,
+        // only draw the first 2 sides in clockwise order to avoid overwriting in the buffer.
+        (_, _) => 2,
+    };
+
+    for (side_idx, side) in border_sides[0..nb_sides_to_draw].iter().enumerate() {
         let length = side[side_idx % 2].unsigned_abs();
 
         for offset in unit_positions_in_clockwise_order(length, dot_size) {
@@ -194,13 +205,17 @@ where
                 corner_dot
                     .translate(translation)
                     .draw_styled(style, target)?;
-            };
+            }
             unit_is_dot = !unit_is_dot; // alternating dots and gaps
         }
         corner_dot.translate_mut(*side);
     }
 
-    Ok(())
+    if nb_sides_to_draw < 4 && unit_is_dot {
+        corner_dot.draw_styled(style, target)
+    } else {
+        Ok(())
+    }
 }
 
 impl<C: PixelColor> StyledDrawable<PrimitiveStyle<C>> for Rectangle {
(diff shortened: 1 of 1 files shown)


rfuest:
> Should I add a commit for this in this PR?

Sounds like a good idea to me. Even if the method should never be called with parameters that can cause a div/0 error, it's better to be safe and consistent across both methods.

inesvar:
Hello !

I've read through PR #772 to check for other issues and add more unit tests.
There are several issues I'd like your input on before pushing the code.

### Edge cases (huge rectangle, overflow rectangle)

I'm unsure if/how the problems occurring in some unlikely scenarios should be handled. 
These situations are quite impractical and I don't know how to test them easily yet.

1. the rectangle has a `size` bigger than i32::MAX

I think this case is handled correctly when drawing a regular border (I didn't actually check though).
Handling this would require some corrections in the dotted border code.
(I think the following should be done:
- dot positions computing: converting Real to u32 instead of i32 to avoid lossy clamping
- thin borders (drawn in clockwise order): using an array of `Size` (instead of `Point`) to represent the rectangle sides, computing the direction of the side (left to right or right to left?) using the index of the side)

2. the rectangle has a `top_left` and `size` such that their sum is bigger than i32::MAX

I’m not sure if the preexisting code clamps the drawing to the i32 range in both directions or if it just panics.
I would need a display that uses the i32 range to check - I guess.

### Opposing dotted borders

There are two ways to draw opposing dotted borders in a rectangle.

(1) Opposing borders could be identical (for instance both horizontal borders could be drawn left to right) or (2) one could be a 180° rotation of the other.
Another way to look at this is (1) axial vs (2) central symmetry.

As mentioned in #182, the rectangle border is expected to be the same thing as as the 4 sides drawn in clockwise order. 
This matches approach (2).

The PR implementing dotted borders (#772) uses (2) for thin borders and (1) for thicker borders => Should we apply a correction to use only (2) ?

Both have pros and cons, another one I can think of is that dotted fills could clash with (2).

rfuest:
> Remark: I didn't change the implementation so that huge dotted rectangles (with size bigger than i32::MAX) would render correctly. Unless you say it's otherwise, I'm assuming that it's not needed.

Some of the code in e-g does some clamping to prevent overflows, but I'm sure there are plenty of other drawing operations that wouldn't work correctly. It would be hard to make sure there are no possible overflows anywhere in the code and IMO not worth the effort. As long as overflows (when panicing on overlow is disabled) can't cause an infinite loop and only cause visual issues it's OK to not care about these edge cases.

> * changing the symmetry of the thick dotted rectangular border (from axial to central), this was done so that the rectangle is equivalent to 4 lines drawn in clockwise order (more detail in the last comment) - let me know if you thing this is a bad idea

I haven't tried it yet, but it sounds like a good idea to keep things more consistent.

rfuest:
I finally found some time to look at this. Most of it looks really good now. The only minor issue I found is that the border disappears completely for a 1 px wide or high rectangle if the stroke width is one (or gets reduced to one by the stroke alignment).

CI is fixed in the master branch and should work after you rebased the PR.

inesvar:
You're right, I had forgotten about this issue. It's fixed :)

rfuest (approved):
LGTM! Thanks for your patience with the PR review.

Issue #789, opened 2025-08-27
StrGlyphMapping performance concerns for non-ascii lookups


junkdog:
I've been playing around with [mousefood](https://github.com/j-g00da/mousefood) which enables running [ratatui](https://github.com/ratatui/ratatui) apps on embedded. Ratatui uses a lot of non-ascii characters for rendering, this requires fonts with a broader range of glyphs (~500-1500). 

Per default setup, `StrGlyphMapping` performs a linear search through all available glyphs. This incurs a pretty heavy cost when glyphs fall outside of the ASCII range. To mitigate this, I modified the `bdf-to-mono` tool to generate a monofont with deterministic glyph lookup + a `FontAtlasData : GlyphMapping`  to go along with it.

In my benchmarks, the atlas implementation provides ~~15-300x~~ 15-370x faster glyph lookup:

<img width="403" height="368" alt="Image" src="https://github.com/user-attachments/assets/746e4536-33a4-4605-a3bf-62903a704ff4" />

Measurements are in ms, 100k iterations over `GlyphMapping::index`.

The approach is quite simple (https://github.com/junkdog/embedded-graphics/tree/font-atlas):

```rust
pub struct FontAtlasData {
    blocks: Vec<UnicodeBlock>,
    other_symbols: Vec<char>, // sorted
}

pub struct UnicodeBlock {
    base_offset: u32,
    range: ops::RangeInclusive<char>,
}
```

Each `UnicodeBlock` maps a contiguous unicode range (like ASCII `0x20-0x7F` or Box Drawing `0x2500-0x257F`) to a base glyph index. When looking up a character with `GlyphMapping::index`:

1. ASCII fast path: `char as usize - 0x20`
2. Block lookup: Find which block contains the character, then `base_offset + (char - block_start)`
3. Other symbols: Binary search through the sorted additional characters array

This eliminates linear scanning through 1000+ glyphs.

Is this something that would be of interest as merge request in some form, as an alternative to `StrGlyphMapping` or extension to the `bnf-to-mono tool`?

rfuest:
Did you use the range feature built into `StrGlyphMapping`? As long as you include whole Unicode blocks with contiguous code points I wouldn't expect the performance to be so much worse.

> Is this something that would be of interest as merge request in some form, as an alternative to `StrGlyphMapping` or extension to the `bnf-to-mono tool`?

[...]

PR #790, opened 2025-08-27
Use ranges instead of chars iter

junkdog:
This is a tiny PR which updates `StrGlyphMapping` to use ranges instead of enumerating individual characters. It removes most of the overhead of looking up glyphs compared to the old approach.

Relates to #789


rfuest (approved):
LGTM

PR #792, opened 2025-09-04
Add the TransparentColor trait

peckpeck:

## PR description

As discussed in https://github.com/embedded-graphics/embedded-graphics/issues/791 this adds a TransparentColor trait.

The goal is to make sure all people producing and accepting transparent colors agree on the same API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Note that some drivers may not support the latest version of embedded-graphics.
 * [hub75](https://crates.io/crates/hub75): A rust driver for hub75 rgb matrix displays
 * [ili9341](https://crates.io/crates/ili9341): A platform agnostic driver to interface with the ILI9341 (and ILI9340C) TFT LCD display
 * [ls010b7dh01](https://crates.io/crates/ls010b7dh01): A platform agnostic driver for the LS010B7DH01 memory LCD display
+* [oled-async](https://crates.io/crates/oled_async): Extensible I2C/SPI async driver for OLED displays (SH1107, SH1108 & SSD1309...)
 * [push2_display](https://crates.io/crates/push2_display): Ableton Push2 embedded-graphics display driver
 * [retro-display](https://crates.io/crates/retro-display): Display drivers for retro-computers incl. Commodore 64
 * [sh1106](https://crates.io/crates/sh1106): I2C driver for the SH1106 OLED display
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,6 +76,7 @@
 //! * [hub75](https://crates.io/crates/hub75): A rust driver for hub75 rgb matrix displays
 //! * [ili9341](https://crates.io/crates/ili9341): A platform agnostic driver to interface with the ILI9341 (and ILI9340C) TFT LCD display
 //! * [ls010b7dh01](https://crates.io/crates/ls010b7dh01): A platform agnostic driver for the LS010B7DH01 memory LCD display
+//! * [oled-async](https://crates.io/crates/oled_async): Extensible I2C/SPI async driver for OLED displays (SH1107, SH1108 & SSD1309...)
 //! * [push2_display](https://crates.io/crates/push2_display): Ableton Push2 embedded-graphics display driver
 //! * [retro-display](https://crates.io/crates/retro-display): Display drivers for retro-computers incl. Commodore 64
 //! * [sh1106](https://crates.io/crates/sh1106): I2C driver for the SH1106 OLED display


rfuest:
I don't have time to review the PR at the moment or look into the alpha channel support more broadly. I'll get back to you in the end of September.

peckpeck:
Understood.

peckpeck:
Is it a better momet now ?

core/src/pixelcolor/transparent_color.rs
```diff
@@ -0,0 +1,140 @@
...
+/// ```
+pub trait TransparentColor<C: PixelColor=Self>: PixelColor {
+    /// Blend this color over another underlying color
+    fn blend_over(&self, other: C) -> C;
+}
+
+/// Trait for transparent colors that expose their alpha channel
+pub trait AlphaColor: TransparentColor<Self::OpaqueColor> {
+    /// Alpha channel type
+    type Alpha;
```
rfuest: Why did you add this associated type? Is it to allow a `bool` alpha channel or a bit depth > 8?

I don't think that an > 8 bits alpha channel is necessary if `RgbColor` is limited to 8 bits per channel max. While a `bool` alpha channel could be nice for on-off transparency, it might make the implementation of generic conversions between color types more difficult. And the `MAX_A` constant doesn't really make sense for a bool.

peckpeck: The `Alpha` associated type is because some people use float as alpha channel, but it might be useless in embedded-graphics.

The `MAX_A` is to mimic the behavior of `RgbColor` colors and allow colors like `Argb6666`, but see that it's not compatible with an associated Alpha type.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -0,0 +1,155 @@
...
+/// They can blend with other colors, transparent or nor.
+///
+/// Transparent colors are a special kind of [`PixelColor`].
+pub trait TransparentColor<C: PixelColor=Self>: PixelColor {
+    /// Blend this color over another underlying color
+    fn blend_over(self, other: C) -> C;
+}
+
+/// Trait for transparent colors that expose their alpha channel
+pub trait AlphaColor {
```
peckpeck: Maybe this should be `: TransparentColor<Self>`

core/src/pixelcolor/transparent_color.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// The maximum value of the alpha channel.
+    const MAX_A: u8;
+}
+
+/// Trait used to find a transparent color that hold a pixelcolor with a compatible alpha channel
+/// Example:
+/// ```
+/// use embedded_graphics_core::pixelcolor::{Rgb888,HasAlphaColor,AlphaColor};
+/// let color = Rgb888::new(0x0,0x80,0xFF);
+/// let transparent = color.into_alpha_color(0);
```
rfuest: I don't really like the name `into_alpha_color`. What do you think about `with_alpha`? While the `with_` prefix is more common in constructors, there are some examples in the `std` library that use it for other purposes.
```suggestion
/// let transparent = color.with_alpha(0);
```
And it's common to use the same name for the trait as for the single method it contains: `WithAlpha`.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -0,0 +1,155 @@
...
+/// image.draw(&mut display.color_converted()).unwrap();
+/// ```
+
+/// Transparent color trait.
+///
+/// This trait is used to represent color that handle transparency level.
+/// They can blend with other colors, transparent or nor.
+///
+/// Transparent colors are a special kind of [`PixelColor`].
+pub trait TransparentColor<C: PixelColor=Self>: PixelColor {
```
rfuest: We should find another name for this trait. To make it easier to implement code that works with transparent and opaque colors this trait can also be implemented for opaque colors:
```rust
impl TransparentColor for Rgb888 {
    fn blend_over(self, other: Rgb888) -> Rgb888 {
        self
    }
}
```
But the current name is a bit confusing, because `Rgb888` isn't transparent.

In other languages I would probably call this trait `BlendableColor`, but I think that using `able` is discouraged in some Rust API naming guideline.

core/src/pixelcolor/transparent_color.rs
rfuest: > The `Alpha` associated type is because some people use float as alpha channel, but it might be useless in embedded-graphics.

Interesting, I hadn't thought about floating point values at all. Without also supporting floats for the RGB channels this might not be of much use, but perhaps a conversion function or special constructor could be a useful addition later. Specifying alpha as a float from 0 to 1 is much easier that having to think about how many alpha channel bits the current color type has and scaling the value manually.

core/src/pixelcolor/transparent_color.rs
peckpeck: yes

core/src/pixelcolor/transparent_color.rs
peckpeck: I don't know of any such recomendation

core/src/pixelcolor/transparent_color.rs
peckpeck: It is easier ton understand, but if the color is based on integer, it needs a lot of conversions for blending.

PR #794, opened 2025-10-17
Make several geometric primitive operations computable in const-eval

REALERvolker1:

## PR description

This PR makes it possible to do several common-sense geometry functions in const-eval. It does not use any nightly features, instead I added some polyfills that could easily be removed if const traits were ever to theoretically be stabilized. I tried to keep it simple, and since it is going from non-const to const in the same MSRV, it is most likely not a breaking change. However, I also decided to make the `Size::from_bounding_box` function public, because I have some code in other projects that can be greatly simplified with that, and it made sense to do. I am most likely not alone in that situation, as many small displays use caset/paset commands that use values like those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Note that some drivers may not support the latest version of embedded-graphics.
 * [hub75](https://crates.io/crates/hub75): A rust driver for hub75 rgb matrix displays
 * [ili9341](https://crates.io/crates/ili9341): A platform agnostic driver to interface with the ILI9341 (and ILI9340C) TFT LCD display
 * [ls010b7dh01](https://crates.io/crates/ls010b7dh01): A platform agnostic driver for the LS010B7DH01 memory LCD display
+* [oled-async](https://crates.io/crates/oled_async): Extensible I2C/SPI async driver for OLED displays (SH1107, SH1108 & SSD1309...)
 * [push2_display](https://crates.io/crates/push2_display): Ableton Push2 embedded-graphics display driver
 * [retro-display](https://crates.io/crates/retro-display): Display drivers for retro-computers incl. Commodore 64
 * [sh1106](https://crates.io/crates/sh1106): I2C driver for the SH1106 OLED display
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,31 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
...
+  - `Rectangle::contains`
+  - `Rectangle::intersection`
+  - `Rectangle::envelope`
+  - `Rectangle::resized`
+  - `Rectangle::resized_width`
+  - `Rectangle::resized_height`
+  - `Rectangle::offset`
+  - `Rectangle::anchor_point`
+  - `Rectangle::anchor_x`
+  - `Rectangle::anchor_y`
```
rfuest: Move these into the `Changed` section and add a link to the PR.

CHANGELOG.md
```diff
@@ -6,6 +6,31 @@ Embedded Graphics is a `no_std` library for adding graphics features to display
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+- Made `Size::from_bounding_box` public
```
rfuest: The CHANGELOG should reflect API changes and from a user perspective it doesn't matter that this function has been available internally before:

```suggestion
- [#794](https://github.com/embedded-graphics/embedded-graphics/pull/794) Added `from_bounding_box` constructor to `Size`.
```

core/src/geometry/size.rs
```diff
@@ -182,7 +182,17 @@ impl Size {
...
+    /// ```
+    /// use embedded_graphics::geometry::{Size, Point};
+    ///
+    /// let top_left = Point::new(5, 5);
+    /// let bottom_right = Point::new(314, 234);
+    /// let bounded = Size::from_bounding_box(top_left, bottom_right);
+    ///
+    /// assert_eq!(bounded, Size::new(310, 230));
+    /// ```
+    pub const fn from_bounding_box(corner_1: Point, corner_2: Point) -> Self {
```
rfuest: There has been some discussion about the name of this method when it was added, but at that time we didn't really care because it was internal only. I don't like the current name, because I wouldn't expect a function that is called `from_BOUNDING_BOX` take points as parameters and not a `Rectangle`. But I cannot really think of anything better at the moment and perhaps I'm overthinking this. 

Another option would be to not make this method public and instead to make `Rectangle::with_corners` const. In my opinion `Rectangle::with_corners(top_left, bottom_right).size` is more readable than `Size::from_bounding_box(top_left, bottom_right)`. And in case it is used in a const context the possible overhead from also calculating `Rectangle::top_left` doesn't matter.

PR #798, opened 2025-11-01
Alpha color implementations

peckpeck:

## PR description

Implementation of some alpha colors, starting with RGB colors

### DIFF
--- a/core/src/pixelcolor/conversion.rs
+++ b/core/src/pixelcolor/conversion.rs
@@ -55,16 +55,17 @@ macro_rules! impl_rgb_conversion {
     };
 }
 
-impl_rgb_conversion!(Rgb332 => Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
-impl_rgb_conversion!(Rgb444 => Rgb332, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
-impl_rgb_conversion!(Rgb555 => Rgb332, Rgb444, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
-impl_rgb_conversion!(Bgr555 => Rgb332, Rgb444, Rgb555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
-impl_rgb_conversion!(Rgb565 => Rgb332, Rgb444, Rgb555, Bgr555, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
-impl_rgb_conversion!(Bgr565 => Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr666, Rgb888, Bgr888);
-impl_rgb_conversion!(Rgb666 => Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr666, Bgr565, Bgr888, Rgb888);
-impl_rgb_conversion!(Bgr666 => Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr565, Bgr888, Rgb888);
-impl_rgb_conversion!(Rgb888 => Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr666, Bgr565, Bgr888);
-impl_rgb_conversion!(Bgr888 => Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr666, Bgr565, Rgb888);
+impl_rgb_conversion!(Rgb332 => Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_conversion!(Rgb444 => Rgb332, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_conversion!(Bgr444 => Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_conversion!(Rgb555 => Rgb332, Rgb444, Bgr444, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_conversion!(Bgr555 => Rgb332, Rgb444, Bgr444, Rgb555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_conversion!(Rgb565 => Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_conversion!(Bgr565 => Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_conversion!(Rgb666 => Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr666, Bgr565, Bgr888, Rgb888);
+impl_rgb_conversion!(Bgr666 => Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr565, Bgr888, Rgb888);
+impl_rgb_conversion!(Rgb888 => Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr666, Bgr565, Bgr888);
+impl_rgb_conversion!(Bgr888 => Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Rgb666, Bgr666, Bgr565, Rgb888);
 
 /// Macro to implement conversion between grayscale color types.
 macro_rules! impl_gray_conversion {
@@ -108,7 +109,7 @@ macro_rules! impl_rgb_to_and_from_gray {
     }
 }
 
-impl_rgb_to_and_from_gray!(Gray2, Gray4, Gray8 => Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_to_and_from_gray!(Gray2, Gray4, Gray8 => Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
 
 /// Macro to implement conversion from `BinaryColor` to RGB and grayscale types.
 macro_rules! impl_from_binary {
@@ -122,7 +123,7 @@ macro_rules! impl_from_binary {
 }
 
 impl_from_binary!(
-    Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888, Gray2, Gray4,
+    Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888, Gray2, Gray4,
     Gray8
 );
 
@@ -150,7 +151,7 @@ macro_rules! impl_rgb_to_binary {
     };
 }
 
-impl_rgb_to_binary!(Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+impl_rgb_to_binary!(Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
 
 #[cfg(test)]
 mod tests {
@@ -275,7 +276,7 @@ mod tests {
             assert_eq!(ToC::from(FromC::WHITE), ToC::WHITE);
         }
 
-        type_matrix!(test_rgb_to_rgb; Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
+        type_matrix!(test_rgb_to_rgb; Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888);
     }
 
     #[test]
@@ -285,7 +286,7 @@ mod tests {
             assert_eq!(ToC::from(FromC::WHITE), ToC::WHITE);
         }
 
-        type_matrix!(test_rgb_to_gray; Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888 => Gray2, Gray4, Gray8);
+        type_matrix!(test_rgb_to_gray; Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888 => Gray2, Gray4, Gray8);
     }
 
     #[test]
@@ -298,7 +299,7 @@ mod tests {
             assert_eq!(BinaryColor::from(FromC::WHITE), BinaryColor::On);
         }
 
-        type_matrix!(test_rgb_to_binary; Rgb332, Rgb444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888 => BinaryColor);
+        type_matrix!(test_rgb_to_binary; Rgb332, Rgb444, Bgr444, Rgb555, Bgr555, Rgb565, Bgr565, Rgb666, Bgr666, Rgb888, Bgr888 => BinaryColor);
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
 pub trait HasAlphaColor: PixelColor {
     /// Associated AlphaColor
     type AlphaColor: AlphaColor + ColorBlend<Self>;
+
+    /// Create transparent color from opaque color
+    fn with_alpha(self, alpha: u8) -> Self::AlphaColor;
```
peckpeck: I don't know why this was removed from the prvious PR

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
...
+    }
+}
+
+argb_color!(Argb4444, Rgb444, RawU16, u16, Argb = (4, 4, 4, 4));
+argb_color!(Argb6666, Rgb666, RawU24, u32, Argb = (6, 6, 6, 6));
+argb_color!(Bgra6666, Bgr666, RawU24, u32, Bgra = (6, 6, 6, 6));
+argb_color!(Argb8888, Rgb888, RawU32, u32, Argb = (8, 8, 8, 8));
+argb_color!(Bgra8888, Bgr888, RawU32, u32, Bgra = (8, 8, 8, 8));
+
+// No obvious impl
```
peckpeck: I don't know if we should try to implement these

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
...
+        impl From<$base_type> for $type {
+            fn from(value: $base_type) -> Self {
+                // set alpha to max value
+                $type(value.into_storage() | Self::A_MASK)
+            }
+        }
+    }
+}
+
+argb_color!(Argb4444, Rgb444, RawU16, u16, Argb = (4, 4, 4, 4));
```
rfuest: We should also add the missing `Bgr444` type to make it possible to have `Bgra444`. If you don't want to add it as part of this PR just add a comment in the code and the tracking PR for the alpha support.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
...
+                let b_shifted = (b & Self::MAX_B) as $storage_type << $b_pos;
+                Self(a_shifted | r_shifted | g_shifted | b_shifted)
+            }
+        }
+
+        impl PixelColor for $type {
+            type Raw = $data_type;
+        }
+
+        impl ColorBlend<$base_type> for $type {
```
rfuest: Do you want to add `impl ColorBlend<$type> for $type` in this PR or leave it for a future PR?

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
...
+
+                Self(data & Self::ARGB_MASK)
+            }
+        }
+
+        impl From<$type> for $data_type {
+            fn from(color: $type) -> Self {
+                Self::new(color.0)
+            }
+        }
```
rfuest: The other color types only implement conversion from and to the `RawUx` types, which these types should also do.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
...
+
+            const BLACK: Self = Self::new(0, 0, 0, Self::MAX_A);
+            const RED: Self = Self::new(Self::MAX_R, 0, 0, Self::MAX_A);
+            const GREEN: Self = Self::new(0, Self::MAX_G, 0, Self::MAX_A);
+            const BLUE: Self = Self::new(0, 0, Self::MAX_B, Self::MAX_A);
+            const YELLOW: Self = Self::new(Self::MAX_R, Self::MAX_G, 0, Self::MAX_A);
+            const MAGENTA: Self = Self::new(Self::MAX_R, 0, Self::MAX_B, Self::MAX_A);
+            const CYAN: Self = Self::new(0, Self::MAX_G, Self::MAX_B, Self::MAX_A);
+            const WHITE: Self = Self::new(Self::MAX_R, Self::MAX_G, Self::MAX_B, Self::MAX_A);
+        }
```
rfuest: If this is the exact same code as in the `impl_rgb_color` macro it would be a good idea to extract that part into a new macro and use it in both cases.

rfuest (changes requested):
Thanks for sticking with this. Here is my first review pass.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
...
+        $type:ident,
+        $base_type:ty,
+        $data_type:ty,
+        $storage_type:ty,
+        ($a_bits:expr, $r_bits:expr, $g_bits:expr, $b_bits:expr),
+        ($a_pos:expr, $r_pos:expr, $g_pos:expr, $b_pos:expr)
+    ) => {
+        #[doc = "TODO"]
+        #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default, Debug)]
+        pub struct $type($storage_type);
```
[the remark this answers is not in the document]
peckpeck: indeed

core/src/pixelcolor/transparent_color.rs
peckpeck: indeed

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +156,183 @@ pub trait AlphaColor: PixelColor {
...
+                Self(a_shifted | r_shifted | g_shifted | b_shifted)
+            }
+        }
+
+        impl PixelColor for $type {
+            type Raw = $data_type;
+        }
+
+        impl ColorBlend<$base_type> for $type {
+            fn blend_over(self, other: $base_type) -> $base_type {
```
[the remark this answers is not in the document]
peckpeck: on a second thought it cannot overflow, but the rounding trick makes it overflow

rfuest (changes requested):
I wasn't sure if you are still working on this or are waiting for a review. I've taken a quick look at the code and found an issue with the blending of two alpha colors. It would be a good idea to add a few tests to make sure blending works as expected.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +158,190 @@ pub trait AlphaColor: PixelColor {
...
+                <$base_type>::new(r, g, b)
+            }
+        }
+
+        impl ColorBlend<$type> for $type {
+            fn blend_over(self, other: $type) -> $type {
+                let a1 = self.alpha() as u16 ;
+                let a2 = r_div(a1 * (Self::MAX_A as u16 - a1), Self::MAX_A) as u16;
+
+                let r0 = self.r() as u16 * a1 + other.r() as u16 * a2;
```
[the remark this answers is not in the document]
rfuest: I agree that a call with 4 integer parameters isn't great, but I think that the per channel calculations are easier to follow if they are extracted into a function. Any duplicate calculations should be detected by the compiler, especially when the function is marked with an inline attribute.

But if you prefer to leave it as is I'm also OK with that.

nn00xx:
I wanted to ask whether work on the feature is still ongoing. I would really like to use this crate in our project but missing Alpha color is currently holding us back from using it.

rfuest:
If @peckpeck doesn't want to finish the PR I can take over, but I would like to her from them first.

nn00xx:
@rfuest Thank you. If you can take over I would also be free to help you out with review/testing since we would use the feature in productive system and I can test it on real hardware.

peckpeck:
I'm back, sorry to have been silent for so long. I tend to work in bursts and this subject fell of my attention.
I'm going to complete this PR now, but next time, don't hesitate to take over my work if needed, i won't be offended.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -155,4 +158,190 @@ pub trait AlphaColor: PixelColor {
...
+
+                let r = r_div(r0, Self::MAX_A);
+                let g = r_div(g0, Self::MAX_A);
+                let b = r_div(b0, Self::MAX_A);
+                <$base_type>::new(r, g, b)
+            }
+        }
+
+        impl ColorBlend<$type> for $type {
+            fn blend_over(self, other: $type) -> $type {
```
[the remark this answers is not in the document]
peckpeck: Indeed

core/src/pixelcolor/transparent_color.rs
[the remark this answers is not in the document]
peckpeck: We could, but if we do this, we must choose which alpha color is the "natural" one since `Rgb888` (for example) can have only one implementation of `HasAlphaColor`.

core/src/pixelcolor/transparent_color.rs
peckpeck: `Argb2222` could be implemented but with dedicated code because current `blend_over` on base type assumes that the base type has the same channel encoding as the alpha type. And `From` conversion would be lossy

core/src/pixelcolor/transparent_color.rs
```diff
@@ -275,35 +334,28 @@ macro_rules! impl_argb_color {
...
-                let r = r_div(r0, a0);
-                let g = r_div(g0, a0);
-                let b = r_div(b0, a0);
-                <$type>::new(r, g, b, a0)
+                let other_alpha = r_div(u16::from(other.alpha()) * (u16::from(Self::MAX_A) - u16::from(self.alpha())), Self::MAX_A);
+
+                let r = blend_over_transparent(self.r(), self.alpha(), other.r(), other_alpha);
+                let g = blend_over_transparent(self.g(), self.alpha(), other.g(), other_alpha);
+                let b = blend_over_transparent(self.b(), self.alpha(), other.b(), other_alpha);
+                let a = self.alpha() + other_alpha;
```
rfuest: I haven't checked the rest yet, but this cannot be right because it would, for example, overflow for two opaque colors.

core/src/pixelcolor/transparent_color.rs
rfuest: I don't remember exactly what my plan for `Rgb332` was back in when I wrote the comment. But I don't think that adding an alpha variant for `Rbg332` is necessary.

Instead we could add `Argb2222` and the corresponding `Rgb222`/`Bgr222` types, which fit in one byte per pixel without lossy conversion between opaque and alpha types.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -146,13 +149,263 @@ pub trait AlphaColor: PixelColor {
...
+                let b = blend_over_opaque(self.b(), self.alpha(), other.b(), Self::MAX_A);
+                <$base_type>::new(r, g, b)
+            }
+        }
+
+        impl ColorBlend<$type> for $type {
+            /// For simplicity this implementation ignores gamma correction
+            /// This might give visually surprising results when blending between dark and bright colors
+            fn blend_over(self, other: $type) -> $type {
+                let other_alpha = r_div(u16::from(other.alpha()) * (u16::from(Self::MAX_A) - u16::from(self.alpha())), Self::MAX_A);
```
peckpeck: I should have renamed other_alpha properly

rfuest: I missed that line.

core/src/pixelcolor/transparent_color.rs
peckpeck: * a2 * (MAX-a1) is <= MAX^2
* a2 * (MAX-a1)/MAX  is <= MAX (in floating point term, <= 1.0)
** equiv to a2*(1-a1) in fp version
* so other_alpha = a1 + a2*(1-a1) = a1 + a2 - a1*a2

which as long as a1 and a2 are <= MAX (<=1.0)  is always <=MAX

This might be confusing because i used a lookalike name for `other.alpha()` and `other_alpha`

peckpeck:
Ping

rfuest:
> Ping

I wanted to wait with a review until I've had time to think about whether #813 should be included in the next e-g release.

core/src/pixelcolor/conversion.rs
```diff
@@ -55,16 +55,87 @@ macro_rules! impl_rgb_conversion {
...
+    ($from_type:ident <=> $($to_type:ident),+) => {
+        $(impl From<$from_type> for $to_type {
+            fn from(other: $from_type) -> Self {
+                Self::new(
+                    convert_channel::<{$from_type::MAX_R}, {$to_type::MAX_R}>(other.r()),
+                    convert_channel::<{$from_type::MAX_G}, {$to_type::MAX_G}>(other.g()),
+                    convert_channel::<{$from_type::MAX_B}, {$to_type::MAX_B}>(other.b()),
+                )
+            }
+        })*
```
rfuest, line 116: I don't think we should include this conversion. Just dropping the alpha channel is only one option to convert RGBA to RGB. A user might be surprised if `Rgba8888::new(255, 255, 255, 0)` gets converted into `Rgb888::WHITE` and not `Rgb888::BLACK`. It could also lead to users trying to draw something (semi-)transparent onto a framebuffer with `fb.color_converted().draw(&image)?`, which would draw the image without alpha blending.

Do you have some use case in mind where having this conversion would improve usability? In that case we would at least need to document the implemented behavior and decide whether ignoring the alpha channel or blending the color with black is the better way to convert RGBA to RGB.

core/src/pixelcolor/transparent_color.rs
```diff
@@ -146,13 +149,386 @@ pub trait AlphaColor: PixelColor {
...
+        }
+    }
+}
+
+// 4 variations for each major rgb types
+
+argb_color!(Argb4444, Rgb444, RawU16, u16, Argb = (4, 4, 4, 4));
+argb_color!(Bgra4444, Bgr444, RawU16, u16, Bgra = (4, 4, 4, 4));
+//argb_color!(Rgba4444, Rgb444, RawU16, u16, Rgba = (4, 4, 4, 4));
+//argb_color!(Abgr4444, Bgr444, RawU16, u16, Abgr = (4, 4, 4, 4));
```
rfuest, line 405: Are these intentionally commented out?

core/src/pixelcolor/transparent_color.rs
```diff
@@ -146,13 +149,386 @@ pub trait AlphaColor: PixelColor {
...
+argb_color!(Bgra6666, Bgr666, RawU24, u32, Bgra = (6, 6, 6, 6));
+//argb_color!(Rgba6666, Rgb666, RawU24, u32, Rgba = (6, 6, 6, 6));
+//argb_color!(Abgr6666, Bgr666, RawU24, u32, Abgr = (6, 6, 6, 6));
+
+argb_color!(Argb8888, Rgb888, RawU32, u32, Argb = (8, 8, 8, 8));
+argb_color!(Bgra8888, Bgr888, RawU32, u32, Bgra = (8, 8, 8, 8));
+//argb_color!(Rgba8888, Rgb888, RawU32, u32, Rgba = (8, 8, 8, 8));
+//argb_color!(Abgr8888, Bgr888, RawU32, u32, Abgr = (8, 8, 8, 8));
+
+// supported bySTM32 microcontrollers
```
rfuest, line 417:
```suggestion
```

core/src/pixelcolor/transparent_color.rs
```diff
@@ -146,13 +149,386 @@ pub trait AlphaColor: PixelColor {
...
+argb_color!(Argb1555, Rgb555, RawU16, u16, Argb = (1, 5, 5, 5));
+
+// This probably need a specific implementation since it will be lossy
+//argb_color!(???, Rgb332, RawU8, u8, Rgb = (3, 3, 2));
+
```
rfuest, line 422: Let's remove this for now and perhaps add `Rgb222` and `Rgba2222` (plus permutations) in a later PR.
```suggestion
```

core/src/pixelcolor/conversion.rs
peckpeck: Indeed, I could just drop it

core/src/pixelcolor/transparent_color.rs
peckpeck: Yes because of a comment I made earlier : we must choose which alpha color is the "natural" one since Rgb888 (for example) can have only one implementation of HasAlphaColor.

This default conversion conflicts with the existence of both `BgraXXXX` and  `AbgrXXXX`.

rfuest: Right, I forgot about that. What do you think about this as an alternative to the `HasAlphaColor` trait, which removes the need for "natural" color combinations:
```rust
/// Trait for transparent colors that expose their alpha channel
pub trait AlphaColor: PixelColor {
    ///Returns the alpha channel value.
    fn alpha(&self) -> u8;

    fn with_alpha(self, alpha: u8) -> Self;

    /// The maximum value of the alpha channel.
    const MAX_A: u8;
}
```
By adding `with_alpha` to the `AlphaColor` trait a user can modify the alpha channel of any alpha color. This would make the example from `HasAlphaColor` look like this:
```rust
let color = Rgb888::new(0x0,0x80,0xFF);
let transparent = Argb8888::from(color).with_alpha(0);
// or
let transparent = Rgba8888::from(color).with_alpha(0);
```
Being able to set the value of individual color channels could be useful in general and also be extended to `RgbColor` with `with_r`, `with_g`, and `with_b` methods.

peckpeck: We could.
Isn't converting to a different color type, then modifying a channel less efficient than directly converting ?

rfuest: > Isn't converting to a different color type, then modifying a channel less efficient than directly converting ?

Maybe in some cases, but most of the time the compiler should be able to optimize the extra steps away. Here is a Rust Playground link, if you want to look at the generated assembly for the different variants: https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=0fdab016e3610bd0853123b982ebf555 I haven't had time to play more with this, but it shows that at least for constant colors the compiler just inserts a constant integer into the assembly in both cases.

peckpeck: it is easy to optimise const calculation but, harder for classical variables.
What about a Generic parameter `AlphaColor` for `HasAlphaColor` instead of an associated type, this way, there is no default type, but `with_alpha` would convert directly from opaque to transparent.

Direct modification of single channels could be implemented separately.

rfuest: > it is easy to optimise const calculation but, harder for classical variables.

Sure, but a bit of bit shifting and masking shouldn't be hard to optimize in any case.

> What about a Generic parameter `AlphaColor` for `HasAlphaColor` instead of an associated type, this way, there is no default type, but `with_alpha` would convert directly from opaque to transparent.

That sounds like a good solution.

PR #803, opened 2026-01-10
README: Add embedded-mono-img and embedded-graphics-colorcast

kpcyrd:
## PR description

This adds two projects I developed for use with embedded-graphics to the README list of related projects. :)

Thanks for making this library, I use it extensively. :sparkles:


rfuest (changes requested):
I'm glad that you enjoy working with e-g and thanks for expanding the e-g ecosystem further.

There are two small issues that need to be fixed before the PR can be merged:
1. `README.md` is generated from `src/lib.rs` and both files need to be in sync to make CI happy. I would recommend to manually copy the changes over, so that you don't have to install the required tooling first.
2. There is a formatting inconsistency with one of the entries (see comment below).

PR #806, opened 2026-02-14
Fix build on MSRV by using tighter dependencies

rfuest:
This PR fixes the build of embedded-graphics 0.8 on the MSRV version of Rust by using tighter dependencies and by adding a Cargo.lock file for the development dependencies.


msrd0:
I am unhappy to see this change. It seems that there were no changes to the embedded-graphics code itself to fix compiling the MSRV, just restricting some dependencies. This, IMHO, was unnecessary and bad. Building on MSRV should've been possible with a suitable `Cargo.lock` file even before this change, which [newer versions of cargo can already build for you](https://blog.rust-lang.org/2025/01/09/Rust-1.84.0/). Restricting dependencies from newer, semver-compatible versions means that if I wanted to use embedded-graphics and, say, az 1.3 in my project, I just couldn't, because cargo doesn't allow two different yet semver-compatible versions of the same crate in your dependencies. Not every user of embedded-graphics uses the minimum Rust version, and IMHO you shouldn't stop those people from using latest versions of dependencies.

rfuest:
I'm sorry that this caused issue's for your project and I agree that the fix I chose was far from perfect, but I wouldn't call it unnecessary. When we specify e-g's MSRV to be a certain version it should build on that version without any manual `Cargo.lock` changes. The issue came up when another crate that depends on e-g could no longer be built using its MSRV. 

Thanks for the link you provided, I heard about the feature when it was introduced by never looked into it. It seems like a much more robust solution and I wouldn't mind raising the MSRV for e-g 0.9 to 1.84.0 especially for that feature.

PR #732, opened 2023-08-19
adds Rgb444

catherinejones:
This PR adds Rgb444 and RawU12 to support 12bit RGB on displays like the ST7789.

This fixes 1/2 of https://github.com/embedded-graphics/embedded-graphics/issues/726 (missing BGR444).

PR #815, opened 2026-06-14
Add RawU12

catherinejones:
## Checklist

## PR description

Adds a RawU12 type and supporting code for storing 12-bit pixels in u8 slices.

### DIFF
--- a/core/src/pixelcolor/raw/mod.rs
+++ b/core/src/pixelcolor/raw/mod.rs
@@ -261,6 +261,7 @@ impl_raw_data!(RawU1: u8, 1, "1 bit");
 impl_raw_data!(RawU2: u8, 2, "2 bits");
 impl_raw_data!(RawU4: u8, 4, "4 bits");
 impl_raw_data!(RawU8: u8, 8, "8 bits");
+impl_raw_data!(RawU12: u16, 12, "12 bits");
 impl_raw_data!(RawU16: u16, 16, "16 bits");
 impl_raw_data!(RawU24: u32, 24, "24 bits");
 impl_raw_data!(RawU32: u32, 32, "32 bits");
@@ -316,6 +317,7 @@ mod tests {
         assert_eq!(RawU1::new(u8::max_value()).0, 0x1);
         assert_eq!(RawU2::new(u8::max_value()).0, 0x3);
         assert_eq!(RawU4::new(u8::max_value()).0, 0xF);
+        assert_eq!(RawU12::new(u16::max_value()).0, 0xFFF);
         assert_eq!(RawU24::new(u32::max_value()).0, 0xFFFFFF);
     }
 }
--- a/core/src/pixelcolor/raw/mod.rs
+++ b/core/src/pixelcolor/raw/mod.rs
@@ -120,6 +120,7 @@
 
 mod load_store;
 mod to_bytes;
+mod u12;
 
 pub use to_bytes::ToBytes;
 
--- a/core/src/pixelcolor/raw/u12.rs
+++ b/core/src/pixelcolor/raw/u12.rs
@@ -1,0 +1,343 @@
+//! Raw u12 support.
+//!
+//! This module contains helper functions to support reading and writing 12-bit values into and out
+//! of byte slices. This is necessary as every pixel value will have 4 bits in a separate byte from
+//! the other 8 bits, which means one out of every three bytes will contain 4 bits from two different
+//! pixels.
+
+use super::OutOfBoundsError;
+
+/// pixel_index_to_byte_and_nibble takes a pixel index and returns the starting byte index and the
+/// nibble index for that pixel when stored.
+#[inline]
+fn pixel_index_to_byte_and_nibble(index: usize) -> (usize, usize) {
+    let byte_index = (index * 12) / 8;
+    let nibble_index = index % 2;
+    (byte_index, nibble_index)
+}
+
+/// store_u12_be replaces the pixel (found in data, which is a mutable slice) at the passed index
+/// with the bytes in the u16. This is the big endian version of this
+/// function.
+#[inline]
+pub fn store_u12_be(data: &mut [u8], index: usize, value: u16) -> Result<(), OutOfBoundsError> {
+    let (byte_index, nibble_index) = pixel_index_to_byte_and_nibble(index);
+
+    // the value is in the next two bytes, so we need to validate that the byte
+    // index does not go beyond the size of the data.
+    if (byte_index + 1) > data.len() {
+        return Err(OutOfBoundsError);
+    }
+
+    // Throught this function we will use an example 12 bit value of 0x123.
+    // This could be a RGB444 pixel, with a red value of 1, green 2, and blue 3.
+    //
+    // As big endian bytes, this would then be:
+    //
+    // [0x01, 0x23]
+
+    if nibble_index == 0 {
+        // To get the start of the pixel, we bitshift the entire u16 value by 4 to the right
+        // and cast it to a u8 (dropping the most significant 8 bits from the u16).
+        //
+        // = (0x0123 >> 4) as u8
+        // = 0x0012 as u8
+        // = 0x12
+        //
+        // This is replaced in data at byte_index.
+        //
+        // [ 0x12, 0x?A, 0xAA ]
+        //
+        data[byte_index] = (value >> 4) as u8;
+
+        // Next, the subsequent byte in the data is masked and the remaining
+        // 4 bits in u16 are OR'd into it:
+        //
+        //   0x?A & 0x0F | ((0x0123 & 0x000F) << 4) as u8
+        // = 0x0A        | (0x0003 << 4) as u8
+        // = 0x0A        | 0x0030 as u8
+        // = 0x0A        | 0x30
+        // = 0x3A
+        //
+        // As a result, the pixel is now stored in the slice of bytes.
+        //
+        // [ 0x12, 0x3A, 0xAA ]
+
+        data[byte_index + 1] = data[byte_index + 1] & 0x0F | ((value & 0x000F) << 4) as u8;
+    } else {
+        // Since the start of the pixel is 4 bits, we bitshift the entire u16 right by 8
+        // and OR the result into into the first byte at the index in the data.
+        //
+        //   0xA? & 0xF0 | (0x0123 >> 8) as u8
+        // = 0xA0        | 0x0001 as u8
+        // = 0xA0        | 0x01
+        // = 0xA1
+        //
+        data[byte_index] = data[byte_index] & 0xF0 | (value >> 8) as u8;
+
+        // The least significant 8 bits of the u16 can be put directly into the
+        // subsequent byte in the data.
+        //
+        // This results in the byte at byte_index+1 in the data being:
+        //
+        // 0x23
+        data[byte_index + 1] = (value & 0x0FF) as u8;
+    }
+
+    Ok(())
+}
+
+/// load_u12_be extracts a u12 value from a slice of u8s at the passed index
+/// returning Some containing a u16 or None if the index goes out of bounds.
+/// This is the big endian version of this function.
+#[inline]
+pub fn load_u12_be(data: &[u8], index: usize) -> Option<u16> {
+    let (byte_index, nibble_index) = pixel_index_to_byte_and_nibble(index);
+
+    // the value is in the next two bytes, so we need to validate that the byte
+    // index does not go beyond the size of the data.
+    if (byte_index + 1) > data.len() {
+        return None;
+    }
+
+    // Throught this function we will use an example 12 bit value of 0x123.
+    // This could be a RGB444 pixel, with a red value of 1, green 2, and blue 3.
+    //
+    // As bytes in a byte slice of big endian u12 values, this would then be:
+    //
+    // [0x12, 0x30]
+    //
+    // or
+    //
+    // [0x01, 0x23]
+
+    let value: u16 = if nibble_index == 0 {
+        // The first eight bits are contained in the first byte, so we can shift that
+        // left 4 bits.
+        // The last four bits are in the second byte, which we bitshift right 4 bits
+        // so it can be ORd with the now shifted first 8.
+        //
+        // = 0x0012 << 4 | 0x0030 >> 4
+        // = 0x0120 | 0x0003
+        // = 0x0123
+        //
+        (u16::from(data[byte_index]) << 4) | u16::from(data[byte_index + 1] >> 4)
+    } else {
+        // Only the first four bits are at the end of the first byte, so we need to bitshift
+        // that to the left by eight bits.
+        // The second byte can then be ORd directly against the first value.
+        //
+        //   0x0001 << 8 | 0x0023
+        // = 0x0100      | 0x0023
+        // = 0x0123
+        //
+        (u16::from(data[byte_index] & 0x0F) << 8) | u16::from(data[byte_index + 1])
+    };
+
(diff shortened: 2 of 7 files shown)


core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+/// with the bytes in the u16. This is the big endian version of this
+/// function.
+#[inline]
+pub fn replace_u12_be(data: &mut [u8], index: usize, value: u16) -> Result<(), OutOfBoundsError> {
+    let byte_index = (index * 12) / 8;
+    if (byte_index + 1) > data.len() {
+        return Err(OutOfBoundsError);
+    }
+
+    let raw_values = value.to_be_bytes();
```
catherinejones: This should either be switched to use a u16 value or the other version of this function should be switched to use bytes for consistency

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+    //
+    // GGGGRRRR, ????BBBB
+    // ^      ^      ^  ^
+    // LSB   MSB   LSB  MSB
+    //
+    // or
+    //
+    // RRRR????, BBBBGGGG
+    // ^         ^      ^
+    // LSB      LSB    MSB
```
rfuest: This looks right, but I've checked the ST7789 datasheet and couldn't find this format being mentioned. Could you send me some reference for this?

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+//! pixels.
+
+use super::OutOfBoundsError;
+
+/// replace_u12_be replaces the pixel (found in data, which is a mutable slice) at the passed index
+/// with the bytes in the u16. This is the big endian version of this
+/// function.
+#[inline]
+pub fn replace_u12_be(data: &mut [u8], index: usize, value: u16) -> Result<(), OutOfBoundsError> {
+    let byte_index = (index * 12) / 8;
```
rfuest: See comment below.
```suggestion
    let byte_index = (index * 12) / 8;
    let nibble_index = index % 2;
```

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+    let raw_values = value.to_be_bytes();
+
+    // Throught this function we will use an example 12 bit value of 0x123.
+    // This could be a RGB444 pixel, with a red value of 1, green 2, and blue 3.
+    //
+    // As big endian bytes, this would then be:
+    //
+    // [0x01, 0x23]
+
+    if byte_index % 3 == 0 {
```
rfuest: In my opinion it's easier to understand to calculate a separate `nibble_index` (see above). Checking for divisibility by 2 should also be a little faster.
```suggestion
    if nibble_index == 0 {
```

core/src/pixelcolor/raw/u12.rs
rfuest: The index calculation is common to all functions in this file and should be extracted into a common helper function.

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+        // = 0x3A
+        //
+        // As a result, the pixel is now stored in the slice of bytes.
+        //
+        // [ 0x12, 0x3A, 0xAA ]
+
+        data[byte_index + 1] = data[byte_index + 1] & 0x0F | (raw_values[1] << 4);
+        return Ok(());
+    }
+    // Since the start of the pixel is 4 bits, we can OR the start of the pixel directly
```
rfuest: Please put this part in an else block to get rid of the return statement. This also applies to the other functions.

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+//! the other 8 bits, which means one out of every three bytes will contain 4 bits from two different
+//! pixels.
+
+use super::OutOfBoundsError;
+
+/// replace_u12_be replaces the pixel (found in data, which is a mutable slice) at the passed index
+/// with the bytes in the u16. This is the big endian version of this
+/// function.
+#[inline]
+pub fn replace_u12_be(data: &mut [u8], index: usize, value: u16) -> Result<(), OutOfBoundsError> {
```
rfuest: I would prefer if the functions in this module followed the naming scheme used in the `LoadStore` trait:
```suggestion
pub fn store_u12_be(data: &mut [u8], index: usize, value: u16) -> Result<(), OutOfBoundsError> {
```

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+        // The first eight bits are contained in the first byte, so we can shift that
+        // left 4 bits.
+        // The last four bits are in the second byte, which we bitshift right 4 bits
+        // so it can be ORd with the now shifted first 8.
+        //
+        // = 0x0012 << 4 | 0x0030 >> 4
+        // = 0x0120 | 0x0003
+        // = 0x0123
+        //
+        let value: u16 = ((data[byte_index] as u16) << 4) | ((data[byte_index + 1] >> 4) as u16);
```
rfuest: For safe casts from a smaller to a larger integer type the e-g code uses `from` instead of `as`:
```suggestion
        let value = u16::from(data[byte_index]) << 4 | u16::from(data[byte_index + 1]) >> 4;
```

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+
+    // Only the first four bits are at the end of the first byte, so we need to bitshift
+    // that to the left by eight bits.
+    // The second byte can then be ORd directly against the first value.
+    //
+    //   0x0001 << 8 | 0x0023
+    // = 0x0100      | 0x0023
+    // = 0x0123
+    //
+    let value: u16 = (((data[byte_index] & 0x0F) as u16) << 8) | (data[byte_index + 1] as u16);
```
rfuest:
```suggestion
    let value = u16::from(data[byte_index] & 0x0F) << 8 | u16::from(data[byte_index + 1]);
```

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+    data[byte_index + 1] = raw_values[1];
+
+    Ok(())
+}
+
+/// get_u12_be extracts a u12 value from a slice of u8s at the passed index
+/// returning a Result containing a u16 or an OutOfBoundsError. This is the
+/// big endian version of this function.
+#[inline]
+pub fn get_u12_be(data: &[u8], index: usize) -> Result<u16, OutOfBoundsError> {
```
rfuest:
```suggestion
pub fn load_u12_be(data: &[u8], index: usize) -> Result<u16, OutOfBoundsError> {
```

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+    let value: u16 = (((data[byte_index] & 0x0F) as u16) << 8) | (data[byte_index + 1] as u16);
+
+    Ok(value)
+}
+
+/// replace_u12_le replaces the pixel (found in data, which is a mutable slice containing
+/// exactly two bytes) with the bytes in the u16. This is the little endian version of this
+/// function.
+#[inline]
+pub fn replace_u12_le(data: &mut [u8], index: usize, value: u16) -> Result<(), OutOfBoundsError> {
```
rfuest:
```suggestion
pub fn store_u12_le(data: &mut [u8], index: usize, value: u16) -> Result<(), OutOfBoundsError> {
```

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,334 @@
...
+    data[byte_index + 1] = (((raw_value & 0x0F) as u8) << 4) | (((raw_value & 0xF0) as u8) >> 4);
+
+    Ok(())
+}
+
+/// get_u12_le extracts a u12 value from a slice of u8s at the passed index
+/// returning a Result containing a u16 or an OutOfBoundsError. This is the
+/// little endian version of this function.
+#[inline]
+pub fn get_u12_le(data: &[u8], index: usize) -> Result<u16, OutOfBoundsError> {
```
rfuest:
```suggestion
pub fn load_u12_le(data: &[u8], index: usize) -> Result<u16, OutOfBoundsError> {
```

rfuest (changes requested):
Thanks for sticking with this feature. Looks good except for a few stylistic nitpicks.

core/src/pixelcolor/raw/u12.rs
rfuest: I would use a u16 in all functions. In my opinion this function would get easier to understand by using a u16 instead of the byte array.

core/src/pixelcolor/raw/u12.rs
catherinejones: I think I might be mistaken here, the only mention of little endianness in the ST7789 datasheet indicates that it's for 65k (RGB565) in 8bit and 9bit modes only (page 258 of the ST7789 datasheet, detailing the "RAMCTRL" command). I haven't tested the little endian code with a real display either so I'm happy to remove it!

core/src/pixelcolor/raw/u12.rs
catherinejones: Done!

catherinejones:
@rfuest I've renamed all of the functions from "replace_u12*" to "store_u12*" and from "get_u12" to "load_u12*"! I'm going to resolve your comments about those renames so I can focus on the other changes, I hope that's ok!

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,349 @@
...
+    //
+    // This would mean that this function should return either:
+    //
+    // [0x21, 0x03]
+    //
+    // or
+    //
+    // [0x10, 0x32]
+
+    if nibble_index % 3 == 0 {
```
catherinejones: Oops, fixing this now!

core/src/pixelcolor/raw/u12.rs
rfuest: I'm not sure if anyone will ever use it, but we need to have some little endian encoding because all other `RawUx` types support an alternative layout.

My suggestion would be to change the format a bit:
For even start nibble indices ` [ 0bGGGGBBBB, 0b????RRRR ]` and for odd start nibble indices `[ 0bBBBB????, 0bRRRRGGG ]`.

The layout currently used in the PR would still be supported by using `Bgr444` instead of `Rgb444`.

core/src/pixelcolor/raw/u12.rs
catherinejones: Done! Should I also do the same for u16 to u8s?

core/src/pixelcolor/raw/u12.rs
```diff
@@ -0,0 +1,344 @@
...
+//! of byte slices. This is necessary as every pixel value will have 4 bits in a separate byte from
+//! the other 8 bits, which means one out of every three bytes will contain 4 bits from two different
+//! pixels.
+
+use super::OutOfBoundsError;
+
+/// pixel_index_to_byte_and_nibble takes a pixel index and returns the starting byte index and the
+/// nibble index for that pixel when stored.
+#[inline]
+fn pixel_index_to_byte_and_nibble(index: usize) -> (usize, usize) {
```
rfuest, line 13: You could also include the common bounds check in this function.
```suggestion
fn pixel_index_to_byte_and_nibble(data: &[u8], index: usize) -> Result<(usize, usize), OutOfBoundsError> {
```

core/src/pixelcolor/raw/u12.rs
rfuest: `from` isn't implemented for `u16` to `u8`, because it might change the integer value by throwing away the upper 8 bits. The usage of `as` instead of `from` is used to document cases in the e-g code where this behavior is expected.

catherinejones: kk! So `as u8` is a shorthand to indicate that dropping all but the least significant 8-bits is intended?

core/src/pixelcolor/raw/u12.rs
catherinejones: Would it be OK if I left them in the individual functions? From my perspective it keeps "pixel_index_to_byte_and_nibble" cleaner and I'm a fan of keeping boundary checks in the functions that need them.

I think in the case of the `load_u12` I could also switch the boundary check to return None and the default value to return Some(value), which would be also be more inline with the LoadStore trait!

core/src/pixelcolor/raw/u12.rs
rfuest: Yes. And because `from` isn't implemented all other alternatives would be quite a bit longer and awkward to use (e.g. `u8::try_from(value & 0xFF).unwrap()` or `value.to_be_bytes()[0]`).

core/src/pixelcolor/raw/u12.rs
rfuest: > Would it be OK if I left them in the individual functions? From my perspective it keeps "pixel_index_to_byte_and_nibble" cleaner and I'm a fan of keeping boundary checks in the functions that need them.

OK, just make sure to have the comment about the `+1` in all copies.

> I think in the case of the `load_u12` I could also switch the boundary check to return None and the default value to return Some(value), which would be also be more inline with the LoadStore trait!

Sounds good!

core/src/pixelcolor/raw/u12.rs
catherinejones: Done! Though I have a question, is there any risk that this code would break in a big endian system due to me not using either `to_be` or `to_le` to convert the u16 before manipulating the bits?

catherinejones:
@rfuest I believe I've now refactored everything based on your requests! No rush to re-review this, thank you for your quick responses today! :)

core/src/pixelcolor/raw/u12.rs
rfuest: Good question, but in this case the code doesn't depend on the endianness of the system. All operations on a `u16` variable, like shifting and anding, work the same on both kinds. The internal representation inside the CPU might be different, but the actual integer values stored in the variable are the same. The only part of the code where memory layout might be important is where the `data` array is accessed, but because it is a `u8` array and the code manually writes individual bytes in the right order this part also works independent of the endianness.

catherinejones: OK! This is what I assumed but I wanted to get a second opinion since I don't have a setup that allows me to test big-endian systems.

core/src/pixelcolor/raw/mod.rs
```diff
@@ -120,6 +120,7 @@
 
 mod load_store;
 mod to_bytes;
+pub mod u12;
```
rfuest: The `u12` helper module shouldn't be public.
```suggestion
mod u12;
```

rfuest (changes requested):
One last nit, the rest looks good.

Please add a copy of the changelog entry to the main e-g crate changelog. The note about the 16 bit storage in the changelog entry for #732 can also be removed.

core/src/pixelcolor/raw/mod.rs
catherinejones: Done!

rfuest (approved):
LGTM, thanks again for finishing this feature!

catherinejones:
> LGTM, thanks again for finishing this feature!

No worries! Thanks for taking the time to review it! 😄

PR #823, opened 2026-08-08
Update README with ST7789 EOL notice and mipidsi driver

nigelsim:
Added a note indicating the end of life for the ST7789 driver library and included a new driver library for MIPI displays.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,8 @@ Note that some drivers may not support the latest version of embedded-graphics.
 * [st7567_rs](https://crates.io/crates/st7567_rs): SPI driver for ST7567 displays
 * [st7567s](https://crates.io/crates/st7567s): Driver for the ST7567S LCD controller
 * [st7735-lcd](https://crates.io/crates/st7735-lcd): Rust library for displays using the ST7735 driver
-* [st7789](https://crates.io/crates/st7789): A Rust driver library for ST7789 displays
+* [st7789](https://crates.io/crates/st7789): A Rust driver library for ST7789 displays (EOL)
+* [mipidsi](https://github.com/almindor/mipidsi): A generic driver library for SPI and parallel TFT displays that implement the MIPI Display Command Set (including ST7789)
 * [st7920](https://crates.io/crates/st7920): ST7920 LCD driver in Rust
 * [gc9a01-rs](https://crates.io/crates/gc9a01-rs): SPI 4-wire driver for the Gc9a01 display driver
 * [rp-cvideo](https://crates.io/crates/rp-cvideo): Driver to generate a binary monochrome composite video signal on the RP2040 and RP235x MCU's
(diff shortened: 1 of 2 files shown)


rfuest:
Thanks for the PR. I was surprised that `mipidsi` wasn't mentioned in the list yet.

For users it would be beneficial to have a list of all controllers supported by the `mipidsi` crate, but that list would need to be updated each time `mipidsi` gains support for a new controller.

@almindor: Please take a look at the description of your crates. Do you think we should still include the old `st7789` crate, or remove it completely? Maybe a more prominent deprecation warning in the README of that project would be a good idea (at least for visitors of the GH project).

nesk:
> For users it would be beneficial to have a list of all controllers supported by the mipidsi crate

Currently the list goes the other way: we list displays, and for each of them we link to a crate. My suggestion is to keep the list that way, replace all non-maintained crates (TBD) by mipidsi if it supports them, and add all the missing displays supported by mipidsi to the list.

almindor:
> @almindor: Please take a look at the description of your crates. Do you think we should still include the old `st7789` crate, or remove it completely? Maybe a more prominent deprecation warning in the README of that project would be a good idea (at least for visitors of the GH project).

I'll add a better deprecation notice, definitely use mipidsi. I don't even look at the old st7789 anymore.

almindor:
> Thanks for the PR. I was surprised that `mipidsi` wasn't mentioned in the list yet.
> 
> For users it would be beneficial to have a list of all controllers supported by the `mipidsi` crate, but that list would need to be updated each time `mipidsi` gains support for a new controller.

We could perhaps link into a pure list file of mipidsi directly, would that work?
> 
> @almindor: Please take a look at the description of your crates. Do you think we should still include the old `st7789` crate, or remove it completely? Maybe a more prominent deprecation warning in the README of that project would be a good idea (at least for visitors of the GH project).

Archived st7789 and updated readme to make it clear mipidsi is the crate to use.

rfuest:
> Currently the list goes the other way: we list displays, and for each of them we link to a crate. My suggestion is to keep the list that way, replace all non-maintained crates (TBD) by mipidsi if it supports them, and add all the missing displays supported by mipidsi to the list.

While the current list also lists crates and not controller types, this is effectively the same because most crates only support a single controller type. But I agree that this is the most useful format for users. I had thought about reformatting the list as a table, which would also allow listing of multiple driver crates for a single controller type (in case there is more than one actively maintained crate for the same controller type).

Something like:

| Controller | Description | Crate |
|------------|----|---|
| `ili9431` | 240RGBx320 LCD Controller | [`mipidsi`](#) |
| `ssd1306` | 128x64 OLED Controller | [`ssd1306`](#) |
| `st7789` |  240RGBx320 LCD Controller | [`mipidsi`](#), [`st7789-lcd`](#) |
| ... | ... |

> We could perhaps link into a pure list file of mipidsi directly, would that work?

I don't see the benefit of a list file over linking to the supported controllers list in the `mipidsi` docs. This also has the disadvantage that it wouldn't be searchable, like the table suggested above.

nesk:
> Something like:
> 
> | Controller | Description | Crate |
> | --- | --- | --- |
> | `ili9431`| 240RGBx320 LCD Controller | [`mipidsi`](#) |
> | `ssd1306` | 128x64 OLED Controller | [`ssd1306`](#) |
> | `st7789` | 240RGBx320 LCD Controller | [`mipidsi`](#), [`st7789-lcd`](#) |
> ... 	... 	

Seems good to me.

Issue #719, opened 2023-03-28
draw triangle panics 'attempt to multiply with overflow' (only when not --release)


uliano:
- Version of embedded-graphics in use (if applicable):"0.7.
- Target: real hardware
- If using real hardware: mipidsi = "0.6.0"

[(https://crates.io/crates/mipidsi)]

## Description of the problem/feature request/other
I can draw the triangle when in --release mode, both the triangle and the rectangle by omitting --release

## Test case (if applicable)

```rust
        let triangle = Triangle::new(
            Point::new(10, 10),
            Point::new(10, 229),
            Point::new(309, 229),
        )
        .into_styled(blue_with_red_outline);

        let rectangle = Rectangle::new(
            Point::new(10, 10),
            Size::new(300, 220)
        )
        .into_styled(blue_with_red_outline);
        
        rectangle.draw(lcd).unwrap();
        rprintln!("all is good up to here");
        
        triangle.draw(lcd).unwrap();
        rprintln!("can't reach");
```

results in

```11:15:40.299 all is good up to here
11:15:40.299 panicked at 'attempt to multiply with overflow', /rustc/8460ca823e8367a30dda430efda790588b8c84d3/library/core/src/num/mod.rs:251:5```

PR #824, opened 2026-08-08
Fix integer overflow when calculating thick line joints

### LINKED ISSUE #817: i32 overflow panic in IntersectionParams::nearly_colinear_has_error when drawing thick-stroked Triangle   with large coordinates
I'm working on some demo apps for the Lily Go T5 E-Paper S3 Pro.  When I tried to draw a triangle with a thick line it crashed with an overflow error.  Once I figured out the root issue I had AI help me reduce it to minimal test case.  This is in *version: embedded-graphics 0.8.2*

Edited Claude output below:

  ---
  Description                                                   

  Drawing a Triangle with a stroke width ≥ 1 panics at runtime with attempt to multiply with overflow when the
  triangle's edge vectors produce a cross-product denominator whose magnitude exceeds √i32::MAX ≈ 46,340.

  The panic occurs in nearly_colinear_has_error in `src/primitives/line/intersection_params.rs`:

``` rust
  pub fn nearly_colinear_has_error(&self) -> bool {
      self.denominator.pow(2) < self.line1.delta().dot_product(self.line2.delta()).abs()
      //  ^^^^^^^^^^^^^^^^^^^^ panics if |denominator| > 46,340
  }
```

  denominator is the determinant of the two edge normal vectors
  (le1.normal_vector.determinant(le2.normal_vector)). For the triangle below, the joint between the diagonal and
  horizontal edges produces normal vectors (-210, -120) and (0, 240), giving determinant −50,400. Squaring that:
  50,400² = 2,540,160,000 > i32::MAX (2,147,483,647).

  The panic fires inside `Styled::draw → Triangle::styled_bounding_box → ClosedThickSegmentIter::new →
  intersections → nearly_colinear_has_error`, so it triggers for any stroke width ≥ 1 on an affected triangle,
  even before a single pixel is rendered.

  ---
  Minimal reproducer
                                                                
  Cargo.toml:
```toml
  [dependencies]
  embedded-graphics = "0.8.2"
  ```

  src/main.rs:                                                  
```rust
  use embedded_graphics::{
      draw_target::DrawTarget,
      geometry::{OriginDimensions, Size},
      pixelcolor::BinaryColor,
      prelude::*,
      primitives::{PrimitiveStyle, Triangle},                   
  };  
  
  struct NullDisplay;
  
  impl DrawTarget for NullDisplay {
      type Color = BinaryColor;
      
      fn draw_iter<I>(&mut self, pixels: I) -> Result<(), Self::Error>
      where
          I: IntoIterator<Item = Pixel<Self::Color>>,
      {   
          for _ in pixels {}
          Ok(())
      }   
  }   
  
  impl OriginDimensions for NullDisplay {
      fn size(&self) -> Size {
          Size::new(1000, 1000)
      }   
  }   
  
  fn main() {
      let mut display = NullDisplay;
      
      // Any triangle whose edge normals produce |det| > 46340 (= sqrt(i32::MAX))
      // will panic. Here: edges (-120,210) and (240,0) yield det = -50400.
      Triangle::new(
          Point::new(480, 310),
          Point::new(360, 520),
          Point::new(600, 520),
      )   
      .into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 4))
      .draw(&mut display)                                       
      .unwrap();
  }
```

**Expected**: triangle draws without panic.

 **Actual**: attempt to multiply with overflow at intersection_params.rs:73.

---
  Root cause
                                                                
  denominator is typed i32. The squaring in denominator.pow(2) should either use i64/u64 (widen before squaring),
  use checked_mul, or be compared differently — for example:

```rust
  // Safe equivalent:
  (self.denominator as i64).pow(2)
      < self.line1.delta().dot_product(self.line2.delta()).abs() as i64
```

---
  Workaround (until fixed)
                                                                
  Draw the triangle as three separate Line primitives. Line's thick-stroke iterator (ThickPoints) does not use
  ClosedThickSegmentIter or IntersectionParams, so it is unaffected.

```rust
  let a = Point::new(480, 310);
  let b = Point::new(360, 520);
  let c = Point::new(600, 520);
  let style = PrimitiveStyle::with_stroke(BinaryColor::On, 4);
  Line::new(a, b).into_styled(style).draw(&mut display).unwrap();
  Line::new(b, c).into_styled(style).draw(&mut display).unwrap();
  Line::new(c, a).into_styled(style).draw(&mut display).unwrap();
```

teddytennant:
Fixes #817. This is very likely the same root cause as the older report in #719.

## The bug

`IntersectionParams::nearly_colinear_has_error` squares the determinant of the two normal vectors using `i32` arithmetic:

```rust
self.denominator.pow(2) < self.line1.delta().dot_product(self.line2.delta()).abs()
```

`denominator` is `le1.normal_vector.determinant(le2.normal_vector)`, which is equal to the cross product of the two line deltas, i.e. roughly twice the area of the triangle they span. As soon as its magnitude exceeds `sqrt(i32::MAX)` (46340), the squaring overflows.

That is reachable with ordinary display coordinates: a triangle with a 240 x 210 pixel corner already yields a determinant of -50400. Because `Triangle::styled_bounding_box` goes through `ClosedThickSegmentIter`, this triggers for any stroke width >= 1, before a single pixel is drawn.

Reproducer from #817:

```rust
Triangle::new(
    Point::new(480, 310),
    Point::new(360, 520),
    Point::new(600, 520),
)
.into_styled(PrimitiveStyle::with_stroke(BinaryColor::On, 4))
.draw(&mut display)
.unwrap();
```

```
thread 'main' panicked at library/core/src/num/mod.rs:394:5:
attempt to multiply with overflow
```

`Polyline` is affected in the same way, since it shares `line_join.rs`. `Polyline::new(&[(480, 310), (360, 520), (600, 520)])` with a stroke width of 4 panics on the current `master` and returns a bounding box of `Rectangle { top_left: (358, 309), size: 243 x 213 }` with this change.

The release build is arguably worse than the panic: `50400i32.pow(2)` wraps to `-1754807296`, which is less than the dot product, so `nearly_colinear_has_error` returns `true` instead of `false` and the joint is treated as degenerate. For the triangle above that produces a bounding box of `Rectangle { top_left: (360, 309), size: 242 x 214 }` instead of the correct `Rectangle { top_left: (357, 307), size: 247 x 216 }`, so part of the stroke ends up outside the reported bounding box.

## The fix

Square the determinant with `i64` arithmetic and widen the dot product to match:

```rust
i64::from(self.denominator).pow(2)
    < i64::from(self.line1.delta().dot_product(self.line2.delta()).abs())
```

This is the smallest change that keeps the predicate's meaning intact. Both operands already fit in `i32`, so widening cannot change the comparison result for any input that didn't overflow before — only the intermediate square is affected. I deliberately did not use `checked_pow(...).is_none()` mapped to a fixed `true`/`false`, because that changes the boolean for near-threshold inputs unless you reason it through; widening is unconditionally faithful. If you would rather avoid 64 bit arithmetic on 32 bit targets, `denominator.checked_pow(2).is_none_or(...)`-style handling with `false` on overflow is equivalent (a squared determinant that doesn't fit in an `i32` can never be smaller than an `i32` dot product), happy to switch.

## Tests

`nearly_colinear_large_determinant` in `intersection_params.rs` covers boundary values on both sides of the overflow threshold:

- determinant 46340 (the largest value that squares without overflow) — this already worked, and the test pins that the result is unchanged
- determinant 46341 — the first value that used to overflow
- determinant -50400 — the actual joint from #817

`nearly_colinear` pins the exact `<` comparison (determinant 100 with dot products of 10000 and 10100, either side of the boundary) so the predicate can't drift.

`large_determinant` in `triangle/styled.rs` is the end-to-end regression test and asserts the styled bounding box, so it also fails on the silent wraparound rather than only on the debug-build panic.

Verification, with the source change reverted but the tests kept:

```
running 3 tests
test primitives::line::intersection_params::tests::nearly_colinear ... ok
test primitives::line::intersection_params::tests::nearly_colinear_large_determinant ... FAILED
test primitives::triangle::styled::tests::large_determinant ... FAILED

---- primitives::line::intersection_params::tests::nearly_colinear_large_determinant stdout ----
panicked at library/core/src/num/mod.rs:394:5:
attempt to multiply with overflow

test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 442 filtered out
```

Same thing with `-C overflow-checks=off`, which is what a release build does:

```
---- primitives::line::intersection_params::tests::nearly_colinear_large_determinant stdout ----
assertion failed: !nearly_colinear_has_error(Point::new(46341, 1), Point::new(0, 1))

---- primitives::triangle::styled::tests::large_determinant stdout ----
assertion `left == right` failed
  left: Rectangle { top_left: Point { x: 360, y: 309 }, size: Size { width: 242, height: 214 } }
 right: Rectangle { top_left: Point { x: 357, y: 307 }, size: Size { width: 247, height: 216 } }

test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 442 filtered out
```

With the fix applied, both configurations pass. Full run:

- `cargo test --workspace`: 442 + 140 + 82 + 54 + 7 passed, 0 failed (442 lib tests, up from 439)
- `cargo fmt --all -- --check`: clean
- `cargo clippy --all-targets -- --deny=warnings` and `--all-features`: clean
- `just check-drawing-examples`: no diff in `doc/` or `src/examples.rs`, so no documented example rendering changed

## Scope

Only the squaring in `nearly_colinear_has_error` is changed. `PointExt::dot_product` and `PointExt::determinant` can also overflow in principle, but only for deltas around 46000 pixels, which is a much more remote case than this one and would be a wider change to the numeric model.


teddytennant:
Anything you need from me here before a review?

